Issue #545, opened 2021-09-06
safekeeper test: `--sync-safekeepers` with lagging safekeepers


kelvich:
Write utility to put specially crafted records in safekeepers. Write tests to check `postgres --sync-safekeepers` in various setups: different amounts of records; maximal generation lsn < prev gen lsn; etc.

PR #647, opened 2021-09-22
Support for `--sync-safekeepers` in tests

petuhovskiy:
New command has been added to append specially crafted records in safekeeper WAL. This command takes json for append, encodes LogicalMessage based on json fields, and processes new AppendRequest to append and commit WAL in safekeeper.

Python test starts up walkeepers and creates config for walproposer, then appends WAL and checks `--sync-safekeepers` works without errors. This test is simplest one, more useful test cases (like in #545) for different setups will be added soon.


walkeeper/src/safekeeper.rs
```diff
@@ -154,24 +154,24 @@ pub struct VoteResponse {
...
+    pub end_lsn: Lsn,
     /// LSN committed by quorum of safekeepers
-    commit_lsn: Lsn,
+    pub commit_lsn: Lsn,
     /// minimal LSN which may be needed by proposer to perform recovery of some safekeeper
-    truncate_lsn: Lsn,
+    pub truncate_lsn: Lsn,
     // only for logging/debugging
-    proposer_uuid: PgUuid,
+    pub proposer_uuid: PgUuid,
```
SomeoneToIgnore, line 174: > `pub`

Not sure how important here, but feels quite odd to make a few of those fields publicly available.
Might make changing the code later a bit harder when this all leaks being used as a public api.

I notice that you use those only to create a struct, so maybe instead of making every field public, we can have another `pub create_for_test(..) -> Self` method?

SomeoneToIgnore (commented):
👋 I'm trying to get onboard other components' logic and reviewing the PRs and help with the small things I note myself.

So far, nothing clever to say about the main logic by me, but spotted a few nits that might make sense to look at.

walkeeper/src/json_ctrl.rs
```diff
@@ -0,0 +1,220 @@
...
+}
+
+pub fn handle_json_ctrl(
+    swh: &mut SendWalHandler,
+    pgb: &mut PostgresBackend,
+    cmd: &Bytes,
+) -> Result<()> {
+    let cmd = cmd.strip_prefix(b"JSON_CTRL");
+    if cmd.is_none() {
+        return Err(anyhow!("invalid prefix"));
```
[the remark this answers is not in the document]
petuhovskiy: I just started writing in Rust, so I think that's useful for me to learn such little nice things. Feel free to bring more of this, thanks!

walkeeper/src/safekeeper.rs
petuhovskiy: It also feels quite odd to me. These messages are part of walproposer <-> safekeeper protocol, and some of them are already declared public. I think there may be some sense to make all these structures public, because they are already used in walproposer as identical C definitions, but not really sure.

test_runner/batch_others/test_wal_proposer.py
```diff
@@ -0,0 +1,102 @@
...
+    pg = ProposerPostgres(pgdata_dir, pg_bin, timeline_id, tenant_id)
+    pg.create_dir_config(wa_factory.get_connstrs())
+
+    # run sync to init safekeepers with ProposerGreeting
+    initial_lsn = pg.sync_safekeepers()
+    # should be 0/0 for empty safekeepers
+    print(f"initial lsn = {initial_lsn}")
+
+    # valid lsn, which is not in the segment start, nor in zero segment
+    epoch_start_lsn = 23826824  # 0/16B9188
```
kelvich, line 75: Could be just `0x16B9188`

kelvich (approved):
Nice, thanks! I played a bit with it and to my surprise `--sync-safekeepers` went well on the several log entry patterns that I've tried. `pg_waldump` does not decode resulting WAL, bu I've used `hexdump -C` along with different strings in logical messages to check what is in the resulting file.

Would be good to have a WAL read API too (not in this PR though) to check WAL contents in more automated fashoin.

petuhovskiy:
> Would be good to have a WAL read API too (not in this PR though) to check WAL contents in more automated fashoin.

Yeah, I actually had this for debug while implementing this feature, but decided to leave this for later. Currently WAL reading is handled only by `ReplicationConn`, it needs some refactoring to have API for reads.

walkeeper/src/json_ctrl.rs
```diff
@@ -0,0 +1,218 @@
...
+//! This module implements JSON_CTRL protocol, which allows exchange
+//! JSON messages over psql for testing purposes.
+//!
+//! Currently supports AppendLogicalMessage, which is used for WAL
+//! modifications in tests.
+//!
+
+use anyhow::{anyhow, Result};
+use bytes::{BufMut, Bytes, BytesMut};
+use crc32c::*;
```
LizardWizzard, line 11: looks like we use only crc32c_append from this package, so it can be imported directly without wildcard use

walkeeper/src/json_ctrl.rs
```diff
@@ -0,0 +1,218 @@
...
+    };
+
+    let mainrdata = logical_message.encode();
+    let mainrdata_len: usize = mainrdata.len() + prefix_bytes.len() + message_bytes.len();
+    // only short mainrdata is supported for now
+    assert!(mainrdata_len <= 255);
+    let mainrdata_len: u8 = mainrdata_len as u8;
+
+    let mut data: Vec<u8> = vec![pg_constants::XLR_BLOCK_ID_DATA_SHORT, mainrdata_len];
+    data.extend_from_slice(&mainrdata);
```
LizardWizzard, line 175: just as an idea, as you know needed capacity in advance you can use Vec::with_capacity

walkeeper/src/json_ctrl.rs
```diff
@@ -0,0 +1,218 @@
...
+    let crc = crc32c_append(0, &data);
+    let crc = crc32c_append(crc, &header_bytes[0..xlog_utils::XLOG_RECORD_CRC_OFFS]);
+    header.xl_crc = crc;
+
+    let mut wal: Vec<u8> = Vec::new();
+    wal.extend_from_slice(&header.encode());
+    wal.extend_from_slice(&data);
+
+    const PADDING: usize = 8;
+    let padding_rem = wal.len() % PADDING;
```
LizardWizzard, line 201: could you please add a comment describing padding requirement

LizardWizzard (commented):
Looks good to me, left a few nits

walkeeper/src/json_ctrl.rs
```diff
@@ -0,0 +1,218 @@
...
+        end_lsn,
+        append_response,
+    })
+}
+
+#[repr(C)]
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+struct XlLogicalMessage {
+    db_id: Oid,
+    transactional: uint32, // 4 bytes bool?
```
arssher, line 139: hm, why? in Postgres this is single byte

petuhovskiy: maybe because of data structure fields alignment, current version seems to work like in postgres

arssher: ah, indeed, I see

arssher (approved):

test_runner/batch_others/test_wal_proposer.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def create_dir_config(self, wal_acceptors: str):
+        """ Create dir and config for running --sync-safekeepers """
+
+        mkdir_if_needed(self.pg_data_dir_path())
+        with open(self.config_file_path(), "w") as f:
+            f.write("zenith.zenith_timeline = '{}'\n".format(self.timeline_id))
+            f.write("zenith.zenith_tenant = '{}'\n".format(self.tenant_id))
+            f.write("synchronous_standby_names = '{}'\n".format("walproposer"))
+            f.write("wal_acceptors = '{}'\n".format(wal_acceptors))
```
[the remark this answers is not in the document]
petuhovskiy: Seems that `Postgres.config` only write lines to file, not too much to reuse

test_runner/batch_others/test_wal_proposer.py
```diff
@@ -0,0 +1,102 @@
...
+    tenant_id = uuid.uuid4().hex
+
+    # write config for proposer
+    pgdata_dir = os.path.join(repo_dir, "proposer_pgdata")
+    pg = ProposerPostgres(pgdata_dir, pg_bin, timeline_id, tenant_id)
+    pg.create_dir_config(wa_factory.get_connstrs())
+
+    # run sync to init safekeepers with ProposerGreeting
+    initial_lsn = pg.sync_safekeepers()
+    # should be 0/0 for empty safekeepers
```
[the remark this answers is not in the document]
petuhovskiy: sure, I'll add it

walkeeper/src/json_ctrl.rs
petuhovskiy: yeah, but seems overkill for tests :)

PR #650, opened 2021-09-22
bump vendor/postgres: Catch walkeeper ErrorResponse

sharnoff:
Per postgres commit message:

> PQgetCopyData can sometimes indicate that the copy is done if the
> backend returns an error response. So while we still expect that the
> walkeeper never sends CopyDone, we can't expect it to never produce
> errors.


arssher (approved):
+1, but let's put postgres commit into `main`.

Fixes first point from https://github.com/zenithdb/zenith/issues/635

kelvich (approved):

PR #652, opened 2021-09-22
walredo - don't use RefCell on stdin/stdout

patins:
Remove some runtime checks


hlinnaka (approved):
Nice! I remember I fought with the borrow checker for quite a while on this :-).

PR #655, opened 2021-09-23
Allow `LeSer`/`BeSer` impls missing either `Serialize` or `Deserialize`

sharnoff:
This is a little bit of a niche change, but I think it improves quality of life a bit.

Currently, using any of the methods from `BeSer` or `LeSer` requires that the type implement *both* `Serialize` and `DeserializeOwned`. In some places, we unnecessarily implement `Serialize` or `Deserialize` because of this -- i.e. we're implementing an unused trait in order to get access to methods.

The first commit here moves the type bounds into the methods themselves, which allows types that don't implement one of the traits to still access the other methods.

The second commit cleans up all of the places where there's unused derived `Serialize` or `Deserialize` implementations.

### DIFF
--- a/pageserver/src/walredo.rs
+++ b/pageserver/src/walredo.rs
@@ -22,7 +22,7 @@ use byteorder::{ByteOrder, LittleEndian};
 use bytes::{Buf, BufMut, Bytes, BytesMut};
 use lazy_static::lazy_static;
 use log::*;
-use serde::{Deserialize, Serialize};
+use serde::Serialize;
 use std::fs;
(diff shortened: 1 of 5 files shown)


hlinnaka (approved):

LizardWizzard (approved):
Nice!

petuhovskiy (approved):

sharnoff:
Ran into a build issue because `impl Deserialize for AppendResponse` started being used. Rebased, should be ok now. Waiting for CI to finish before merging

PR #663, opened 2021-09-24
Pageserver defaults cleanup

yeputons:
I've noted and fixed some discrepancies in pageserver defaults: some specify port 5430 instead of 64000, as well as an incorrect copy-paste of the HTTP default.

I took the liberty of bringing in [const_format](https://crates.io/crates/const_format). If it's better to remove it, I will remove the first commit and rewrite the second via `format!`


pageserver/src/bin/pageserver.rs
```diff
@@ -228,14 +231,14 @@ fn main() -> Result<()> {
                 .long("listen-pg")
                 .alias("listen") // keep some compatibility
                 .takes_value(true)
-                .help("listen for incoming page requests on ip:port (default: 127.0.0.1:5430)"),
+                .help(formatcp!("listen for incoming page requests on ip:port (default: {DEFAULT_PG_LISTEN_ADDR})")),
```
yeputons, line 234: Note that standard `format!` does not have [implicit arguments](https://github.com/rust-lang/rfcs/blob/master/text/2795-format-args-implicit-identifiers.md) yet — suggested in `const_format`'s docs.

hlinnaka (approved):
> I took the liberty of bringing in const_format. If it's better to remove it, I will remove the first commit and rewrite the second via format!

Seems OK to me. I'm wary of adding dependencies to big libraries that bring in cascade of other dependencies, but little helpers like this are not a big deal.

SomeoneToIgnore:
I've run `cargo +nightly-2021-09-08 build --all --all-targets -Z timings` twice on current `main` and this branch cleaning `target` on my not-so-beefy machine the and got the time stats it took to build the new deps:
* const_format_proc_macros v0.2.21 | 3.4s
* const_format v0.2.21 | 1.1s

Not that big for a nice way to tie clean up all the messages IMO.

LizardWizzard (approved):

PR #666, opened 2021-09-27
Proxy for existing users

kelvich:


kelvich:
I want this to be in main to have docker images pushed. Let's go with post-review cc @funbringer @patins

proxy/src/mgmt.rs
```diff
@@ -78,34 +81,47 @@ impl postgres_backend::Handler for MgmtHandler {
...
-            .get(&resp.session_id)
-            .ok_or_else(|| anyhow::Error::msg("psql_session_id is not found"))?;
+    let waiters = mgmt.state.waiters.lock().unwrap();
 
-        match resp.result {
-            PsqlSessionResult::Success(db_info) => {
-                sender.send(Ok(db_info))?;
+    let sender = waiters
+        .get(&resp.session_id)
+        .ok_or_else(|| anyhow::Error::msg("psql_session_id is not found"))?;
```
patins, line 109: I think we should drop `waiters` before doing IO

Issue #354, opened 2021-07-21
Implement drop subcommand for tenant, branch and pg


lubennikovaav:

shanyp:
@lubennikovaav is this still relevant ?

jcsp:
Stale.

PR #669, opened 2021-09-27
PR to discuss thread desing

lubennikovaav:
Let's review our multithreading design, keeping in mind following goals:

- shutdown pageserver gracefully; #477 
- delete branches/timelines/tenants; #354 
- offload inactive tenants; (do we need it?)
- relocate tenants from one pageserver to another (do we need it?)

### DIFF
--- a/docs/thread_design_rfc.md
+++ b/docs/thread_design_rfc.md
@@ -1,0 +1,69 @@
+## Current state
+
+NOTE: I used current names of threads, which are quite chaotic.
+It would be nice to use one uniform style.
+
+#### pageserver
+Main entry point for the pageserver executable.
+At start pageserver sets up logging, auth, metrics collection and starts following threads:
+
+- http_endpoint_thread
+- Page Service thread
+
+#### http_endpoint_thread
+Thread that serves incoming service requests, such as create tenant/branch, list tenants/branches.
+
+#### Page Service thread
+Listens for connections, and launches a new `conn_handler` handler thread for each.
+
+#### conn_handler() // these threads are not named
+Per connection handler.
+Communicates with one PostgresBackend using libpq protocol. Serves queries.
+
+#### Threads that run per tenant (repository)
+-------------
+
+#### wal_redo_manager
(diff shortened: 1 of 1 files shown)


docs/thread_design_rfc.md
```diff
@@ -0,0 +1,69 @@
...
+
+#### Checkpointer
+Flushes in-memory layers to disk.
+Currently it is launched at pageserver start only for pre-existing tenants.
+
+#### GC
+Removes on-disk layers that are not visible anymore.
+Currently it is launched at pageserver start only for pre-existing tenants.
+
+#### relish_uploader
```
SomeoneToIgnore, line 41: Just a note: current code placement for this kind of threads was selected naively.
A follow-up set of changes in the [branch](https://github.com/zenithdb/zenith/compare/kb/download_storage?expand=1) runs `relish_uploader` threads the same way as `http_endpoint_thread` is run.

docs/thread_design_rfc.md
```diff
@@ -0,0 +1,69 @@
...
+
+3. Should we use per tenant `page_service` threads?
+
+4. What actions are needed for graceful pageserver shutdown?
+    - Stop receiving new incoming connections.
+    - Cancel existing connections.
+    - Stop receiveing new WAL. Does WAL service need special actions on pageserver shutdown?
+    - Shutdown checkpointer and GC to avoid interference with final checkpoint.
+    - Checkpoint all timelines for each tenant.
+    - Wait till everything is uploaded to s3.
```
SomeoneToIgnore, line 62: I would wait for current uploads to be finished, to avoid truncated files, otherwise the rest of the uploads can wait 'till the server comes back: the branch mentioned above is capable of restarting itself for non-uploaded files only.

SomeoneToIgnore:
I've got the relish_storage update that impacts the document here.

https://github.com/zenithdb/zenith/pull/689

@lubennikovaav please have a look and let me know if the updates are ok/not ok from the document's standpoint.

PR #674, opened 2021-09-28
Use logging in python tests

petuhovskiy:
Originally proposed [here](https://github.com/zenithdb/zenith/pull/637#issuecomment-923885623). This PR replaces `print` calls to `log.info`, to see logs in CI when running tests in parallel (with pytest-xdist)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -251,7 +251,7 @@ jobs:
             # multiple tests in one file)
             # -rA prints summary in the end
             # -n4 uses four processes to run tests via pytest-xdist
-            pipenv run pytest --junitxml=$TEST_OUTPUT/junit.xml --tb=short -s --verbose -rA $TEST_SELECTION $EXTRA_PARAMS
+            pipenv run pytest --junitxml=$TEST_OUTPUT/junit.xml --tb=short --verbose -rA $TEST_SELECTION $EXTRA_PARAMS
       - run:
           # CircleCI artifacts are preserved one file at a time, so skipping
           # this step isn't a good idea. If you want to extract the
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -245,12 +245,12 @@ jobs:
             #
             # The junit.xml file allows CircleCI to display more fine-grained test information
             # in its "Tests" tab in the results page.
-            # -s prevents pytest from capturing output, which helps to see
-            # what's going on if the test hangs
             # --verbose prints name of each test (helpful when there are
             # multiple tests in one file)
             # -rA prints summary in the end
             # -n4 uses four processes to run tests via pytest-xdist
+            # -s is not used to prevent pytest from capturing output, because tests are running
+            # in parallel and logs are mixed between different tests
             pipenv run pytest --junitxml=$TEST_OUTPUT/junit.xml --tb=short --verbose -rA $TEST_SELECTION $EXTRA_PARAMS
       - run:
           # CircleCI artifacts are preserved one file at a time, so skipping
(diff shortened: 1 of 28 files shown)


.circleci/config.yml
```diff
@@ -245,7 +245,7 @@ jobs:
             # multiple tests in one file)
             # -rA prints summary in the end
             # -n4 uses four processes to run tests via pytest-xdist
-            pipenv run pytest --junitxml=$TEST_OUTPUT/junit.xml --tb=short -s --verbose -rA $TEST_SELECTION $EXTRA_PARAMS
+            pipenv run pytest --junitxml=$TEST_OUTPUT/junit.xml --tb=short --verbose -rA $TEST_SELECTION $EXTRA_PARAMS
```
petuhovskiy, line 248: It disables instant test output display, as test outputs are mixed with each other when running in parallel. Logs in python tests are captured and always displayed after all tests are completed.

test_runner/batch_others/test_branch_behind.py
```diff
@@ -1,6 +1,9 @@
 import subprocess
 from fixtures.zenith_fixtures import PostgresFactory, ZenithPageserver
 
+import logging
+import fixtures.log_helper  # configures loggers
```
funbringer, line 5: There are no fixtures in `log_helper`, so I doubt that it really belongs to `fixtures`.

test_runner/batch_others/test_branch_behind.py
```diff
@@ -1,6 +1,9 @@
...
+log = logging.getLogger('root')
```
funbringer, line 6: This logging setup sequence seems too complicated. Ideally, I'd like to just import `logging` or whatever and call `info`, `warning` etc. Could we do all the required setup in `log_helper`?

test_runner/fixtures/zenith_fixtures.py
```diff
@@ -1099,9 +1104,9 @@ def check_restored_datadir_content(zenith_cli: ZenithCli, test_output_dir: str,
                                                 restored_dir_path,
                                                 pgdata_files,
                                                 shallow=False)
-    print('filecmp result mismatch and error lists:')
-    print(mismatch)
-    print(error)
+    log.info('filecmp result mismatch and error lists:')
+    log.info(mismatch)
+    log.info(error)
```
funbringer, line 1109: I'd print this as a single message.

petuhovskiy: I fixed this and looked at other prints, but couldn't find similar multiline messages anywhere else. There may be other places to optimize the logs, but it's not as obvious to improve them.

Anyway, all logs are captured and displayed in the summary (when running `pytest -rA`). In this summary logs from different tests are not mixed with each other, so concurrent tests don't really affect anything log-related in CI.

.circleci/config.yml
petuhovskiy: Yep, added some info why we don't need to use `-s` in CI

test_runner/batch_others/test_branch_behind.py
petuhovskiy: Couldn't find a better place for log_helper, any suggestions? Even though there is no real pytest fixture in log_helper, I think it has purpose similar to utils and zenith_fixtures in the same directory.

arssher: I can propose rename `fixtures` to `lib` or `utils`, as there is already utils.py which is used outside fixtures; or do something like
```
test_runner/
├── fixtures/
│   ├── zenith_fixtures.py
├── lib/
│   ├── utils.py
│   └── log_helper.py
```
but it seems like more hassle than needed.

arssher:
One thing I noticed is still duplicated captured output in CI logs -- it looks like pytest intercepts logging calls and issues them on its own as well. Apart from that I like it.

@funbringer @LizardWizzard do you want to review this?

PR #681, opened 2021-09-29
Make `pageserver_` prefix in global metric names configurable

yeputons:
Some metrics were extracted from pageserver to a shared code, but retained `pageserver_` prefix on them, which makes no sense.


LizardWizzard:
IMO it might be useful to distinguish between same metrics in different services when querying in tools like grafana.
What do you think about using label with a service name?

LizardWizzard:
I think you are right, AFAIK it won’t be able to distinguish different instances. I’ve seen usage of labels with host, or container id for kubernetes. So I suggest to start with prefixes for simplicity and check/fix multi instance behaviour separately

yeputons:
Updated. As common metrics like disk I/O are defined in `zenith_metrics`/`zenith_utils` crates, it's impossible to prepend different names in compile-time because they're compiled separately. Hence, some run-time mechanics is required, I really liked semantics of `OnceBox` from [`once_cell`](https://docs.rs/once_cell/1.8.0/once_cell/). Looks like it was already our transitive dependency, `Cargo.lock` did not update much.

kelvich:
> IMO it might be useful to distinguish between same metrics in different services when querying in tools like grafana.
What do you think about using label with a service name?

Service itself may completely ignore such namespacing. That is usually done by a local collector which mixes in node metrics. Along with the other things it adds custom tag with the service name and host ip address. That is how metrics exposed from host itself look like (we use telegraf as a local collector):
```
# HELP pageserver_wal_redo_wait_time Telegraf collected metric
# TYPE pageserver_wal_redo_wait_time histogram
pageserver_wal_redo_wait_time_bucket{service="pageserver",host="ip-172-32-24-61",url="http://localhost:9898/metrics",le="0.005"} 112
pageserver_wal_redo_wait_time_bucket{service="pageserver",host="ip-172-32-24-61",url="http://localhost:9898/metrics",le="0.01"} 112
pageserver_wal_redo_wait_time_bucket{service="pageserver",host="ip-172-32-24-61",url="http://localhost:9898/metrics",le="0.025"} 112
pageserver_wal_redo_wait_time_bucket{service="pageserver",host="ip-172-32-24-61",url="http://localhost:9898/metrics",le="0.05"} 112
pageserver_wal_redo_wait_time_bucket{service="pageserver",host="ip-172-32-24-61",url="http://localhost:9898/metrics",le="0.1"} 112
pageserver_wal_redo_wait_time_bucket{service="pageserver",host="ip-172-32-24-61",url="http://localhost:9898/metrics",le="0.25"} 112
pageserver_wal_redo_wait_time_bucket{service="pageserver",host="ip-172-32-24-61",url="http://localhost:9898/metrics",le="0.5"} 112
pageserver_wal_redo_wait_time_bucket{service="pageserver",host="ip-172-32-24-61",url="http://localhost:9898/metrics",le="1"} 112
pageserver_wal_redo_wait_time_bucket{service="pageserver",host="ip-172-32-24-61",url="http://localhost:9898/metrics",le="2.5"} 112
pageserver_wal_redo_wait_time_bucket{service="pageserver",host="ip-172-32-24-61",url="http://localhost:9898/metrics",le="5"} 112
pageserver_wal_redo_wait_time_bucket{service="pageserver",host="ip-172-32-24-61",url="http://localhost:9898/metrics",le="10"} 112
pageserver_wal_redo_wait_time_bucket{service="pageserver",host="ip-172-32-24-61",url="http://localhost:9898/metrics",le="+Inf"} 112
pageserver_wal_redo_wait_time_sum{service="pageserver",host="ip-172-32-24-61",url="http://localhost:9898/metrics"} 0.000133281
pageserver_wal_redo_wait_time_count{service="pageserver",host="ip-172-32-24-61",url="http://localhost:9898/metrics"} 112
```

kelvich:
So sorry for weighting in lately -- that discussion slipped from attention. But I would rather have unprefixed metrics, since it is actually makes querying easier. cc @cicdteam

kelvich:
@cicdteam made an argument that prefixing metrics with a service name is also a common practice along with relabeling. So, yeah, let's merge this PR in it's current form.

kelvich (approved):

cicdteam:
`process_` more related to processes metrics exactly (about resources - cpu, mem, io) but `zenith_` looks good. Currently metrics came from pageserver mostly but "tomorrow" we can collect metrics from other binaries, so `zenith_` is good variant.

PR #691, opened 2021-09-30
More nightly clippy fixes (2021-09-29)

sharnoff:
Most of the changes are for the new `if-then-panic` lint added in https://github.com/rust-lang/rust-clippy/pull/7669. The error messages look like:
```
error: only a `panic!` in `if`-then statement
   --> zenith_utils/src/lsn.rs:195:9
    |
195 | /         if prev.checked_add(val).is_none() {
196 | |             panic!("AtomicLsn overflow");
197 | |         }
    | |_________^ help: try: `assert!(!prev.checked_add(val).is_none(), "AtomicLsn overflow");`
    |
    = note: `-D clippy::if-then-panic` implied by `-D warnings`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#if_then_panic
```
At first I thought it generally made things less readable, but it actually surprised me - being able to see the `assert!` at the start of the line makes it abundantly clear what the purpose is.

### DIFF
--- a/pageserver/src/layered_repository/interval_tree.rs
+++ b/pageserver/src/layered_repository/interval_tree.rs
@@ -283,6 +283,7 @@ mod tests {
             write!(f, "{}", self.val)
         }
     }
+    #[rustfmt::skip]
     fn assert_search(
         tree: &IntervalTree<MockItem>,
         key: u32,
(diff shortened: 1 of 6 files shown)


pageserver/src/layered_repository/interval_tree.rs
```diff
@@ -291,24 +292,20 @@ mod tests {
         if let Some(v) = tree.search(key) {
             let vstr = v.to_string();
 
-            if expected.is_empty() {
-                panic!("search with {} returned {}, expected None", key, v);
-            }
+            assert!(!expected.is_empty(), "search with {} returned {}, expected None", key, v);
```
SomeoneToIgnore, line 295: The `!expected.is_empty()` check in conjunction with `expected None` is quite confusing, but I guess I might not know something 🙂

SomeoneToIgnore (approved):

pageserver/src/layered_repository/interval_tree.rs
sharnoff: It took me a moment to figure out as well. From a higher level, what the function is doing is really just `assert!(result.is_none() == expected.is_empty())` -- it should have returned `None` iff we weren't expecting anything. This is just the case if we already know the result is `Some(_)`.

https://github.com/zenithdb/zenith/blob/d92412554fe81f22c5d3f52173de3a2e82d9f1d1/pageserver/src/layered_repository/interval_tree.rs#L292-L295

PR #697, opened 2021-10-01
pageserver - repository test cleanup

patins:


zenith_utils/src/zid.rs
```diff
@@ -78,6 +78,10 @@ macro_rules! zid_newtype {
             pub fn generate() -> Self {
                 $t(ZId::generate())
             }
+
+            pub const fn from_array(b: [u8; 16]) -> Self {
+                $t(ZId(b))
+            }
```
patins, line 84: I'd like to use the existing `From` but it doesn't take `const`?

SomeoneToIgnore (approved):

zenith_utils/src/zid.rs
sharnoff: Possibly worth making this `#[cfg(test)]`, but maybe there's use cases outside of that.

sharnoff (approved):
Looks good!

PR #698, opened 2021-10-03
More failsafe relish_storage impls

SomeoneToIgnore:
Follow up of https://github.com/zenithdb/zenith/pull/639#pullrequestreview-769259701

### DIFF
--- a/pageserver/src/relish_storage.rs
+++ b/pageserver/src/relish_storage.rs
@@ -19,7 +19,7 @@ use anyhow::Context;
 
 use self::local_fs::LocalFs;
 pub use self::synced_storage::schedule_timeline_upload;
-use crate::relish_storage::rust_s3::RustS3;
+use crate::relish_storage::rust_s3::S3;
 use crate::{PageServerConf, RelishStorageKind};
 
 pub fn run_storage_sync_thread(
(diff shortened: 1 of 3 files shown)


patins:
Is local storage strictly for dev/testing? Does it need these semantics?

SomeoneToIgnore:
Yep, the local storage is a thing for local testing only currently.

Not sure what do you mean about semantics, better error messages?

patins:
Semantics as in does it need to be crash safe?

SomeoneToIgnore:
Oh, got it, thanks.

Yes, local storage at its current state duplicates the workdir manipulations pageserver does when dealing with layers, and we care the s3 impl only in production now.

So consistency feels pretty redundant at this point, I've tried to apply it to verify I'm doing things right and to bring some level of parity to future IT tests for the feature.

SomeoneToIgnore:
So looks like we will be changing the S3 file format eventually, according to the discussion around

https://discord.com/channels/869525774699462656/869526396878323733/895643938554253312 

Also, we slow down things when syncing per-file too if I remember, so is local storage sync when downloading necessary now in your opinion, @patins ?

SomeoneToIgnore:
I'll close this one for now then, to avoid confusion.

PR #719, opened 2021-10-08
pageserver - use VecMap for in-memory PageVersions

patins:
In the future I plan to use VecMap in

* InMemoryLayer segsizes
* DeltaLayer metadata
* DeltaLayer segsizes

### DIFF
--- a/pageserver/src/layered_repository/inmemory_layer.rs
+++ b/pageserver/src/layered_repository/inmemory_layer.rs
@@ -24,6 +24,8 @@ use std::sync::{Arc, RwLock};
 use zenith_utils::accum::Accum;
 use zenith_utils::lsn::Lsn;
 
+use super::page_versions::PageVersions;
+
 pub struct InMemoryLayer {
     conf: &'static PageServerConf,
     tenantid: ZTenantId,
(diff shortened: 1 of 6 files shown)


pageserver/src/layered_repository/inmemory_layer.rs
```diff
@@ -722,11 +722,7 @@ impl InMemoryLayer {
                 before_segsizes.insert(*lsn, *size);
             }
         }
-        let mut before_page_versions = inner.page_versions.iter().filter(|tup| {
-            let ((_blknum, lsn), _pv) = tup;
-
-            *lsn < end_lsn
-        });
+        let mut before_page_versions = inner.page_versions.ordered_page_version_iter(Some(end_lsn));
```
patins, line 725: The order going into DeltaLayer doesn't matter. At the moment it will all go into a BTreeMap, which will take care of the order for us. I'm just using this pre-sorted setup now to avoid changing `DeltaLayer` code much. In the future, I will come up with a more customized routine for efficiently going from `PageVersions` to `VecMap<(u32, Lsn), PageVersion>` (once we use VecMap in DeltaLayer)

PR #740, opened 2021-10-12
Rename `wal_acceptor` binary to `safekeeper`

### LINKED ISSUE #704: Rename `wal_acceptor` binary to `safekeeper`

yeputons:
Fixes #704 (partially).

This is first stage of renaming: old name is kept in Dockerfiles until our private scripts are updated. The next PR will remove aliases from Dockerfiles.

Lots of other renaming lives separately in #741. This PR interleaves with our private repository, so it's separate.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,7 +38,7 @@ RUN apt-get update && apt-get -yq install libreadline-dev libseccomp-dev openssl
 
 COPY --from=build /zenith/target/release/pageserver /usr/local/bin
 COPY --from=build /zenith/target/release/safekeeper /usr/local/bin
-# Temporary alias for compatibility, see https://github.com/zenithdb/zenith/pull/740
+# TODO: temporary alias for compatibility, see https://github.com/zenithdb/zenith/pull/740
 RUN ln -s /usr/local/bin/safekeeper /usr/local/bin/wal_acceptor
 COPY --from=build /zenith/target/release/proxy /usr/local/bin
 COPY --from=pg-build /zenith/tmp_install postgres_install
--- a/Dockerfile.alpine
+++ b/Dockerfile.alpine
@@ -82,7 +82,7 @@ RUN apk add --update openssl build-base libseccomp-dev
 RUN apk --no-cache --update --repository https://dl-cdn.alpinelinux.org/alpine/edge/testing add rocksdb
 COPY --from=build /zenith/target/release/pageserver /usr/local/bin
 COPY --from=build /zenith/target/release/safekeeper /usr/local/bin
-# Temporary alias for compatibility, see https://github.com/zenithdb/zenith/pull/740
+# TODO: temporary alias for compatibility, see https://github.com/zenithdb/zenith/pull/740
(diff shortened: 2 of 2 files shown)


test_runner/fixtures/zenith_fixtures.py
```diff
@@ -64,15 +64,15 @@ def pytest_configure(config):
          raise Exception('Too many workers configured. Cannot distrubute ports for services.')
 
     # does not use -c as it is not supported on macOS
-    cmd = ['pgrep', 'pageserver|postgres|wal_acceptor']
+    cmd = ['pgrep', 'pageserver|postgres|safekeeper']
```
yeputons, line 67: Although there may be some old versions called `wal_acceptor` running locally or in prod, I assume they will be dropped in full very quickly and should not cause any disturbance after few weeks.

yeputons (commented):
~~Not sure what is the capitalization for `Safekeeper`/`SafeKeeper` is. There already was `SafeKeeperState` in the code base.~~

yeputons:
There was a spurious test failure (`test_pageserver_restart`), disappeared after re-running. Weird. I don't think renaming would increase flakiness of tests, so I don't think this PR is the cause.

End-to-end tests, however, fail, because some deployment scripts call `wal_acceptor` directly. Possible work around: keep both versions in Docker for a while, update deployment scripts, remove the `wal_acceptor` symlink.

I think it's non-trivial enough to warrant removing non-binary-related renames from this PR.

Dockerfile
```diff
@@ -37,7 +37,9 @@ RUN apt-get update && apt-get -yq install libreadline-dev libseccomp-dev openssl
     mkdir zenith_install
 
 COPY --from=build /zenith/target/release/pageserver /usr/local/bin
-COPY --from=build /zenith/target/release/wal_acceptor /usr/local/bin
+COPY --from=build /zenith/target/release/safekeeper /usr/local/bin
+# Temporary alias for compatibility, see https://github.com/zenithdb/zenith/pull/740
```
funbringer, line 41: Could you mark this as FIXME or TODO? I'm quite used to searching for those.

Dockerfile
```diff
@@ -37,7 +37,9 @@ RUN apt-get update && apt-get -yq install libreadline-dev libseccomp-dev openssl
...
+RUN ln -s /usr/local/bin/safekeeper /usr/local/bin/wal_acceptor
```
funbringer, line 42: I think this is fine for the time being. During final transition there probably will be a moment of time when E2E tests might be broken, but I don't think that's a problem as long as we finish it quickly.

test_runner/fixtures/zenith_fixtures.py
```diff
@@ -895,7 +895,7 @@ def append_logical_message(self, tenant_id: str, timeline_id: str, request: Dict
 class WalAcceptorFactory:
```
funbringer, line 895: Consider renaming this class as well.

yeputons:
Got flaky test on CircleCI: `test_isolation - batch_pg_regress.test_isolation`. Last flaky on Oct 8, as it reports. Will try re-running.

PR #758, opened 2021-10-14
Bump MSRV to 1.55

patins:


SomeoneToIgnore (commented):
LGTM

@hlinnaka should approve that though to be sure we don't break his workflow too much 🙂

PR #762, opened 2021-10-15
Update postgres to use walproposer-as-an-extension

MMeent:
Companion merge request to https://github.com/zenithdb/postgres/pull/96, to verify that the changes work and pass the tests.


MMeent:
This is not going to get merged as-is; and needs some serious time that I don't have right now; so I'm closing this.

PR #763, opened 2021-10-15
Track state of the repository for a tenant using RepositoryState

lubennikovaav:


pageserver/src/repository.rs
```diff
@@ -9,10 +9,29 @@ use std::time::Duration;
...
+    // Repository working threads periodically check the state
+    // and stop their job when shutdown is requested.
+    ShutdownRequested,
+}
+
 ///
 /// A repository corresponds to one .zenith directory. One repository holds multiple
 /// timelines, forked off from the same initial call to 'initdb'.
 pub trait Repository: Send + Sync {
+    fn set_state(&self, state: RepositoryState);
```
SomeoneToIgnore, line 31: I've got an impression, that `state` could be kept outside of the repository's public api, similar to the proposal in https://github.com/zenithdb/zenith/pull/689#issuecomment-931216193

I see a few reasons for that:

* feels a bit odd to plug in the "download layer on demand" case into the picture: when we trigger a download, there's no timeline created yet and no files to create the timeline from, so it's hard to set any "Downloading" status to nothing :slightly_smiling_face: 
* seems like all `set_state` calls are located in `tenant_mgr`, where we have to lock to get a repository anyway
* all `get_state`/`wait_state_update` calls are done in `impl LayeredRepository` and all check for the shutdown flag only

To put it all together, feels like `tenant_mgr` could hold the state indeed in the repo map under the mutex and update it when needed?
The repo can get an `AtomitBoolean` flag for shutdown and leave the state management and waiting for the managing code

pageserver/src/tenant_mgr.rs
```diff
@@ -89,18 +122,26 @@ pub fn register_relish_download(
...
     }
+
+    // init repo updates Tenant state
+    init_repo(conf, tenant_id);
+    let new_repo = get_repository_for_tenant(tenant_id).unwrap();
+    init_timeline(&new_repo, timeline_id);
 }
 
-fn init_timeline(repo: &dyn Repository, timeline_id: ZTimelineId) {
+fn init_timeline(repo: &Arc<dyn Repository>, timeline_id: ZTimelineId) {
```
SomeoneToIgnore, line 145: NIT: `&Arc<dyn Repository>` looks odd, I think it should be the previous one still, `repo: &dyn Repository` and use https://doc.rust-lang.org/std/sync/struct.Arc.html#method.as_ref to properly pass it.

pageserver/src/tenant_mgr.rs
```diff
@@ -171,7 +228,7 @@ pub fn get_timeline_for_tenant(
 }
 
 fn list_tenants() -> Result<Vec<ZTenantId>> {
-    let o = &mut REPOSITORY.lock().unwrap();
+    let o = &mut TENANTS.lock().unwrap();
```
SomeoneToIgnore, line 231: NIT: Looks like another `access_tenants()` call

pageserver/src/tenant_mgr.rs
```diff
@@ -140,25 +182,40 @@ pub fn create_repository_for_tenant(
...
 }
 
 pub fn get_repository_for_tenant(tenantid: ZTenantId) -> Result<Arc<dyn Repository>> {
-    access_repository()
+    let m = access_tenants();
+    let tenant = m
         .get(&tenantid)
-        .map(Arc::clone)
-        .ok_or_else(|| anyhow!("repository not found for tenant name {}", tenantid))
+        .ok_or_else(|| anyhow!("Tenant not found for tenant name {}", tenantid));
```
SomeoneToIgnore, line 213: NIT: 

> .ok_or_else(|| anyhow!("Tenant not found for tenant name {}", tenantid))

* in the log message, it's id, not `name`, right?

* I see that we have a lot of (at least 6?) `access_tenants() -> get/get_mut via tenant_id -> .unwrap()` calls in various forms, here we even add some anyhow context, but still unwrap it.
Might make sense to extract that into the method (but might not work good, since requires dealing with the mutex somehow)

SomeoneToIgnore (approved):
Found a few style nits, but otherwise good, thanks for adding some handy enum varialnts for uploading already 🙂

PR #766, opened 2021-10-19
Change 'zenith tenant list' API to return tenant state added in 0dc7a3fc

lubennikovaav:
current code only uses two tenant states: CloudOnly (short moment while the tenant is bootstrapping) and Active

### DIFF
--- a/pageserver/src/tenant_mgr.rs
+++ b/pageserver/src/tenant_mgr.rs
@@ -9,6 +9,7 @@ use crate::PageServerConf;
 use anyhow::{anyhow, bail, Context, Result};
 use lazy_static::lazy_static;
 use log::{debug, info};
+use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::fmt;
 use std::fs;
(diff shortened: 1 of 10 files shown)


yeputons (commented):
My concerns are mostly with Python in tests:

* `openapi_spec.yml` is not updated. Not sure if it's actually used, but if it is, maybe a code which validates the schema would be nice. In a separate unrelated issue/PR, of course.
* Direct JSON usage in tests => no type checking. I would suggest a `@dataclass`, and there is no nesting here, you can instantiate it with `**res.json()` (dictionary unpack).
* Manual parsing of `tenant list`'s output in tests. I think looking for `.split()[0]` when the format is changed once again may get nasty. I'd suggest adding a separate method to `zenith_cli` which does all the parsing and returns a `@dataclass`.
  * Alternative is to avoid parsing everywhere except `test_tenant_list_psql` and use HTTP in `start()`

I'm also not sure if it's ok to list tenants from some in-memory struct instead of listing directory entries, but the former _definitely_ looks cleaner.

pageserver/src/tenant_mgr.rs
```diff
@@ -247,3 +248,22 @@ fn list_tenantids() -> Result<Vec<ZTenantId>> {
...
+    pub state: String,
+}
+
+pub fn list_tenants() -> Result<Vec<TenantInfo>> {
+    let m = access_tenants();
+    m.iter()
+        .map(|v| {
+            let (id, tenant) = v;
+            Ok(TenantInfo {
+                id: id.to_string(),
```
SomeoneToIgnore, line 264: Not strongly necessary, but might look cleaner: `ZTenantId` derives `Serialize`, making the `TenantState` serializable is also trivial, so you can avoid string creations entirely here, letting `serde` to do its job instead.

But up to you.

PR #771, opened 2021-10-19
use cached-property package because python<3.8 doesnt have cached_property in functools

LizardWizzard:
we claim in readme that we support python > 3.6 but there is a feature we use that appeared only in python3.8. This pr replaces usage of new stdlib feature with dedicated package as people used it before cached_property become part of stdlib

### DIFF
--- a/test_runner/Pipfile
+++ b/test_runner/Pipfile
@@ -11,6 +11,7 @@ pyjwt = {extras = ["crypto"], version = "*"}
 requests = "*"
 pytest-xdist = "*"
 asyncpg = "*"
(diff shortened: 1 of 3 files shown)


MMeent:
I'm not on a working branch right now, but I did need to also change 
```diff
-        self.pidfile.unlink(missing_ok=True)
+        try:
+            self.pidfile.unlink()
+        except FileNotFoundError as ex:
+            pass
+        except Error as ex:
+            print(ex)
```
as `unlink()`'s `missing_ok`  is also new in 3.8.

PR #773, opened 2021-10-19
[proxy] Prevent TLS stream from hanging

funbringer:
This change causes writer halves of a TLS stream to always flush after a
portion of bytes has been written by `std::io::copy`. Furthermore, some
cosmetic and minor functional changes are made to facilitate debug.

### DIFF
--- a/proxy/src/proxy.rs
+++ b/proxy/src/proxy.rs
@@ -6,7 +6,6 @@ use anyhow::bail;
 use tokio_postgres::NoTls;
 
 use rand::Rng;
-use std::io::Write;
 use std::{io, sync::mpsc::channel, thread};
 use zenith_utils::postgres_backend::Stream;
 use zenith_utils::postgres_backend::{PostgresBackend, ProtoState};
@@ -28,11 +27,13 @@ pub fn thread_main(
         println!("accepted connection from {}", peer_addr);
         socket.set_nodelay(true).unwrap();
 
-        thread::spawn(move || {
-            if let Err(err) = proxy_conn_main(state, socket) {
-                println!("error: {}", err);
-            }
-        });
+        thread::Builder::new()
+            .name("Proxy thread".into())
+            .spawn(move || {
+                if let Err(err) = proxy_conn_main(state, socket) {
+                    println!("error: {}", err);
+                }
+            })?;
     }
 }
 
@@ -158,6 +159,7 @@ impl ProxyConnection {
     fn handle_existing_user(&mut self) -> anyhow::Result<DatabaseInfo> {
         // ask password
         rand::thread_rng().fill(&mut self.md5_salt);
+
         self.pgb
             .write_message(&BeMessage::AuthenticationMD5Password(&self.md5_salt))?;
         self.pgb.state = ProtoState::Authentication; // XXX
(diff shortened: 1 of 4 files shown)


funbringer:
I see no trace of tests for proxy, and we need those to keep it from decaying. Where should we put new ones?

proxy/src/proxy.rs
```diff
@@ -250,51 +252,64 @@ databases without opening the browser.
...
-        writer.flush()?;
-        writer.shutdown(std::net::Shutdown::Both)
+    fn do_proxy(mut reader: impl io::Read, mut writer: WriteStream) -> io::Result<u64> {
+        /// FlushWriter will make sure that every message is sent as soon as possible
+        struct FlushWriter<W>(W);
+
+        impl<W: io::Write> io::Write for FlushWriter<W> {
+            fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+                let res = self.0.write(buf);
+                let _ = self.0.flush(); // we should always return `res`
```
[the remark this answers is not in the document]
funbringer: Not only it is ok, it's also necessary for correctness. By method's contract, you should always report the amount of bytes written.

zenith_utils/src/sock_split.rs
```diff
@@ -107,21 +107,12 @@ impl io::Write for WriteStream {
...
-    stream: BufStream,
-    session: rustls::ServerSession,
-}
-
-impl TlsBoxed {
-    fn rustls_stream(&mut self) -> rustls::Stream<rustls::ServerSession, BufStream> {
-        rustls::Stream::new(&mut self.session, &mut self.stream)
-    }
-}
+type TlsStream<T> = rustls::StreamOwned<rustls::ServerSession, T>;
```
patins, line 110: Nice cleanup

proxy/src/proxy.rs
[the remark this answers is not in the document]
funbringer: I changed my mind. Although this may not always be a good idea, in this particular case propagating flush errors is better, because any error in read or write will cause `std::io::copy` to exit, per https://doc.rust-lang.org/stable/std/io/fn.copy.html#errors. This wouldn't be possible if we just silently cast the error away.

proxy/src/proxy.rs
```diff
@@ -250,51 +252,66 @@ databases without opening the browser.
...
+    fn do_proxy(mut reader: impl io::Read, mut writer: WriteStream) -> io::Result<u64> {
+        /// FlushWriter will make sure that every message is sent as soon as possible
+        struct FlushWriter<W>(W);
+
+        impl<W: io::Write> io::Write for FlushWriter<W> {
+            fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+                // `std::io::copy` is guaranteed to exit if we return an error,
+                // so we can afford to lose `res` in case `flush` fails
+                let res = self.0.write(buf);
+                self.0.flush()?;
```
patins, line 274: Should we bother flushing if there's a write error?

LizardWizzard (approved):

proxy/src/proxy.rs
funbringer: Second time is a charm :smiley:

PR #774, opened 2021-10-19
Fix mypy errors and force it on CI

yeputons:


yeputons:
Just updated from the `main` branch.

test_runner/Pipfile
```diff
@@ -15,10 +15,13 @@ cached-property = "*"
 
 [dev-packages]
 flake8 = "*"
-mypy = "*"
 # Behavior may change slightly between versions. These are run continuously,
 # so we pin exact versions to avoid suprising breaks. Update if comfortable.
 yapf = "==0.31.0"
+mypy = "==0.910"
+# Non-pinned packages follow.
```
MMeent, line 22: flake8 is not pinned to a version either, could you make that more consistent?

kelvich (approved):

test_runner/fixtures/zenith_fixtures.py
```diff
@@ -693,9 +705,9 @@ def stop_and_destroy(self) -> 'Postgres':
         """
 
         assert self.node_name is not None
-        assert self.tenant_id is not None
```
MMeent, line 696: Why remove the assertion if that should always be the case here?

MMeent (changes requested):
Overall OK with the changes, some small changes comments added

test_runner/Pipfile
yeputons: However, if someone is to run `flake8` and check their code for warnings, I think running the most recent version is better than an arbitrary pinned one.

PR #777, opened 2021-10-19
Remove `stop_point` argument from Safekeeper's `START_REPLICATION` handler, do not handle `start_point == 0`

yeputons:
It was there from the very first Rust version: https://github.com/zenithdb/zenith/blob/02ca245081a7f2cbed02b0f79426d9a4c8c4cdd9/src/wal_service.rs

However, the `START_REPLICATION` command only provides a single LSN. So I'm not sure what the parsing was for.

I've checked that both branches of the `stop_pos` initialization are covered:

* `test_normal_work` from `batch_others/test_wal_acceptor.py` [fails](https://app.circleci.com/pipelines/github/zenithdb/zenith/2739/workflows/a798efe0-f77c-473e-be36-f866c4d04584/jobs/20990/tests#failed-test-4) if there is a `panic!` in the non-`wal_proposer_recovery` case: 
* Following three tests [hang](https://app.circleci.com/pipelines/github/zenithdb/zenith/2740/workflows/210e51a1-0fa1-40dd-8fdc-06b6b57035f4/jobs/20997) (and only they) if there is a `panic!` in the `wal_proposer_recovery` case:
  * `test_restart_compute[with_wal_acceptors=True]` from `batch_others/test_restart_compute.py` (should complete in less than a minute)
  * `test_sync_safekeepers` from `batch_others/test_wal_acceptor.py` (should be instant)
  * `test_timeline_status` from `batch_others/test_wal_acceptor.py` (should complete in about ten seconds or less)

However, replacing `stop_pos` with `Lsn(0)` [does not fail any](https://app.circleci.com/pipelines/github/zenithdb/zenith/2743/workflows/9b3c9b10-e38b-48a9-b1f7-aaf35d01fe05) tests (modulo flaky `test_restarts_under_load`). Is `stop_pos` just a performance optimization?


arssher:
Yes, clearly currently it is pointless and should be removed, let's do that. 

Looking here, there is a small bug in WalProposerRecovery: proposer should check that he is provided with enough wal till its epochStartLsn, not bail out on some error earlier. It is separate, of course.

There is also a small risk of overproviding wal proposer with too much WAL in the presence of concurrent compute, but 1) this can be checked on proposer side as well 2) it is eventually harmless as terms will settle the situation, though better to notice this early.

yeputons:
I've also ensured that the `start_pos == Lsn(0)` check is never triggered in our tests: see 3be358411ce8b77a1e3987b76c5733fc992bde64

I'm not sure if removing it altogether is reasonable, but PostgreSQL's docs don't mention a special value of zero for `START_REPLICATION`. I think having it mean "give me the whole WAL" is more meaningful than "give me some random suffix of WAL". I don't know any good application for the latter except for some live monitoring.

lubennikovaav (approved):
This cleanup looks good to me.

I guess later we will need to introduce something like `stop_point` for safekeeper recovery, but it's another issue.

PR #784, opened 2021-10-21
Use more robust way to persist safekeeper control file.

### LINKED ISSUE #678: WAL Safekeeper's FileStorage::persist may theoretically leave control file corrupt
I suppose [this `persist` function](https://github.com/zenithdb/zenith/blob/6d993410c98a67043eaad2f1be2adccf1b7808e9/walkeeper/src/timeline.rs#L331-L338) is assumed to be atomic. However, it overrides the existing file. If something fails during the write, the file may be left corrupted because some changes have reached the filesystem and some are not.

Possible fix is (at least according [to the discussion here](https://users.rust-lang.org/t/how-to-write-replace-files-atomically/42821/13)) to write to a temporary file on the same filesystem (i.e. not just `/tmp`), sync it, atomically rename it, sync the folder.

LizardWizzard:
Now safekeeper control file updated in a following way:
1. Write data to temp file
2. Fsync the temporary file (if sync option is specified)
3. Rename temporary file to actual control file
4. Fsync containing directory (if sync option is specified)

Also because of the rename machinery switch to use dedicated lock file
to prevent running several safekeepers concurrently on the same data

I've also added a few `.with_context` calls for better error visibility

Resolves #678

### DIFF
--- a/walkeeper/src/timeline.rs
+++ b/walkeeper/src/timeline.rs
@@ -1,7 +1,7 @@
 //! This module contains timeline id -> safekeeper state map with file-backed
 //! persistence and support for interaction between sending and receiving wal.
 
-use anyhow::{bail, Result};
+use anyhow::{anyhow, bail, ensure, Context, Result};
 use fs2::FileExt;
 use lazy_static::lazy_static;
 use log::*;
@@ -9,7 +9,8 @@ use postgres_ffi::xlog_utils::find_end_of_wal;
 use std::cmp::{max, min};
 use std::collections::HashMap;
 use std::fs::{self, File, OpenOptions};
-use std::io::{Seek, SeekFrom, Write};
+use std::io::{Read, Seek, SeekFrom, Write};
+use std::path::PathBuf;
 use std::sync::{Arc, Condvar, Mutex};
 use std::time::Duration;
 use zenith_metrics::{register_histogram_vec, Histogram, HistogramVec, DISK_WRITE_SECONDS_BUCKETS};
@@ -24,9 +25,16 @@ use crate::safekeeper::{
 };
 use crate::SafeKeeperConf;
 use postgres_ffi::xlog_utils::{XLogFileName, XLOG_BLCKSZ};
+use std::convert::TryInto;
 
+// contains persistent metadata for safekeeper
 const CONTROL_FILE_NAME: &str = "safekeeper.control";
+// needed to atomically update the state using `rename`
+const CONTROL_FILE_NAME_PARTIAL: &str = "safekeeper.control.partial";
+// dedicated lockfile to prevent running several safekeepers on the same data
+const LOCK_FILE_NAME: &str = "safekeeper.lock";
 const POLL_STATE_TIMEOUT: Duration = Duration::from_secs(1);
+const CHECKSUM_SIZE: usize = std::mem::size_of::<u32>();
 
 /// Replica status: host standby feedback + disk consistent lsn
 #[derive(Debug, Clone, Copy)]
@@ -126,18 +134,10 @@ impl SharedState {
         timelineid: ZTimelineId,
         create: CreateControlFile,
     ) -> Result<Self> {
-        let (cf, state) = SharedState::load_control_file(conf, timelineid, create)?;
-        let timelineid_str = format!("{}", timelineid);
-        let storage = FileStorage {
-            control_file: cf,
-            conf: conf.clone(),
-            persist_sync_control_file_seconds: PERSIST_SYNC_CONTROL_FILE_SECONDS
-                .with_label_values(&[&timelineid_str]),
-            persist_nosync_control_file_seconds: PERSIST_NOSYNC_CONTROL_FILE_SECONDS
-                .with_label_values(&[&timelineid_str]),
-        };
+        let (file_storage, state) = SharedState::load_from_control_file(conf, timelineid, create)
+            .with_context(|| "failed to load from control file")?;
         let (flush_lsn, tli) = if state.server.wal_seg_size != 0 {
-            let wal_dir = conf.workdir.join(format!("{}", timelineid));
+            let wal_dir = conf.timeline_dir(&timelineid);
             find_end_of_wal(
                 &wal_dir,
                 state.server.wal_seg_size as usize,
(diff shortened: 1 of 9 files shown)


yeputons (changes requested):
* It looks like syncing the parent directory is Linux-specific, isn't it? The [`atomicwrites` crate](https://github.com/untitaker/rust-atomicwrites/blob/c65e456a4ee7dcb999fab3718d15e322acf7a57b/src/lib.rs#L229) does something different on Windows. If so, maybe documenting this is a good idea.
* If someone moves the `timeline_dir` while we're syncing the small control file (very unlikely, I guess), there is a race between all these file operations, and especially the last directory sync. Not sure if we can do anything about it, though.
*

walkeeper/src/timeline.rs
```diff
@@ -134,44 +143,56 @@ impl SharedState {
...
             control_file_path.display(),
-            create
+            create,
+            lock_file_path.display(),
         );
+
+        let lock_file = OpenOptions::new()
+            .create(true)
+            .write(true)
+            .open(&lock_file_path)
```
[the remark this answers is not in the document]
LizardWizzard: Fixed

yeputons:
I'm paranoid in this regard as well and would prefer "Out of space" error to thinking about partially overwritten files. Especially considering I don't know how to reason about FS/OS/HDD guarantees on the latter.

LizardWizzard:
@yeputons I've talked to @kelvich and he convinced me that this approach is ok, because when we rewrite file there is no extra space used. But I agree, here we rely on practical assumptions rather than some well-defined standard. 512 bytes is indeed a common block size on hdd. By the way, AFAIK only nvme protocol added capability to ask the disk about blocks of what size can be written atomically, but I don't know is it supported by vendors or not.

I've implemented this approach and tried to make common abstraction for that. Thanks to @funbringer for a hint about better api (IsSafeControlFile macro) Please take a look. 

FYI @patins @funbringer

arssher:
Hang on, what was wrong with atomic rename? I added term history to safekeeper control file (in my branch for now), which now potentially makes it larger than 512 bytes. The only advantage of single sector atomic write approach is simplicity, and more general solution was already implemented...

zenith_utils/src/control_file.rs
```diff
@@ -0,0 +1,167 @@
...
+    path::Path,
+};
+
+// Taken from PG_CONTROL_MAX_SAFE_SIZE
+pub const MAX_CONTROL_SAFE_SIZE: usize = 512;
+const CHECKSUM_SIZE: usize = std::mem::size_of::<u32>();
+pub const MAX_CONTROL_DATA_SIZE: usize = MAX_CONTROL_SAFE_SIZE - CHECKSUM_SIZE;
+
+// this trait serves as a proof that size of the implementing struct is less than MAX_CONTROL_DATA_SIZE
+// see IsSafeControlFile macro at the end of the file
```
funbringer, line 17: Could you make this a doc comment? Then you'd be able to reference `[IsSafeControlFile]` etc.

zenith_utils/src/control_file.rs
```diff
@@ -0,0 +1,167 @@
...
+pub unsafe trait SafeControlFile {}
+
+// This is a useful utility that allows atomic writes of a structures that
+// serialize themselves to a byte sequence smaller than MAX_CONTROL_DATA_SIZE.
+// Bincode docs mention that the size of the encoded object will be the same
+// or smaller than the size that the object takes up in memory in a running Rust program.
+// To use this utility wrap desired struct in a IsSafeControlFile.
+// So because `IsSafeControlFile` adds a compile time check for structure size
+// we can avoid runtime checks for the size of the structure.
```
funbringer, line 26: Would benefit from being a doc as well.

LizardWizzard:
@arssher There is nothing wrong with atomic rename. When applicable rewrite in place should perform better. Alongside I’ve added checksum calculation/validation. If in your PR control file content no longer fits 512 bytes I can roll it back and add checksum validation to atomic rename approach. Also should I leave atomic rewrite in utils? (I can split it out of this PR)

patins:
> hint about better api (IsSafeControlFile macro)

I think the `std::mem::size_of` doesn't need to equal the serialized size (consider container types), so it's nice but not perfect.

If we revert to the atomic rename, I think it's ok to just leave the control file stuff in pageserver and not pull it out to utils

LizardWizzard:
> doesn't need to equal the serialized size (consider container types)

Yes, this popped in my mind when I was going to bed yesterday :)

> I think it's ok to just leave the control file stuff in pageserver and not pull it out to utils
👍🏻

zenith_utils/src/control_file.rs
LizardWizzard: missclicked on resolve, sorry

LizardWizzard:
I've rolled this back to atomic rename and added checksum verification. ControlFile abstraction was imperfect but IMO still improved some aspects. @patins what do you think about using it in timeline metadata in pageserver?

walkeeper/src/timeline.rs
```diff
@@ -150,80 +150,114 @@ impl SharedState {
...
             }
-            Err(e) => {
+            if state.format_version != SK_FORMAT_VERSION {
                 bail!(
-                    "failed to open control file {:?}: {}",
-                    &control_file_path,
-                    e
+                    "incompatible format version: {} vs. {}",
+                    state.format_version,
+                    SK_FORMAT_VERSION
```
yeputons, line 241: Nit while we are here: `got {}, expected {}`

walkeeper/src/timeline.rs
```diff
@@ -150,80 +150,114 @@ impl SharedState {
...
+        // Empty file is legit on 'create', don't try to deser from it.
+        let state = if control_file.metadata().unwrap().len() == 0 {
+            if let CreateControlFile::False = create {
+                bail!("control file is empty");
+            }
+            SafeKeeperState::new()
+        } else {
+            let mut buf = Vec::new();
+            control_file
+                .read_to_end(&mut buf)
```
yeputons, line 209: Nit: I'd `ensure!` that file is not too big before reading. Say, ten megabytes. Maybe log an error if it's greater than 5 MB. Should never happen, but who knows.

walkeeper/src/timeline.rs
```diff
@@ -397,24 +432,70 @@ impl GlobalTimelines {
...
+
+        let len = buf.len(); // overcome borrowing error
+        buf[len - CHECKSUM_SIZE..].copy_from_slice(&checksum.to_le_bytes());
+
+        control_partial.write_all(&buf).with_context(|| {
+            format!(
+                "failed to write safekeeper state into control file at: {}",
+                control_partial_path.display()
+            )
+        })?;
```
yeputons, line 477: I'd just call `write_all` twice or use `extend_from_slice`.

walkeeper/src/timeline.rs
```diff
@@ -515,3 +591,79 @@ impl Storage for FileStorage {
...
+
+    fn file_storage(
+        conf: &SafeKeeperConf,
+        timeline_id: ZTimelineId,
+    ) -> Result<(FileStorage, SafeKeeperState)> {
+        fs::create_dir_all(&conf.timeline_dir(&timeline_id))
+            .expect("failed to create timeline dir");
+        let (storage, state) =
+            SharedState::load_from_control_file(conf, timeline_id, CreateControlFile::True)?;
+        Ok((storage, state))
```
yeputons, line 623: I think it's possible to just return the `Result` from `load_from_control_file`

walkeeper/src/timeline.rs
```diff
@@ -515,3 +591,79 @@ impl Storage for FileStorage {
...
+            let (mut storage, mut state) =
+                file_storage(&conf, timeline_id).expect("failed to read state");
+            // change something
+            state.wal_start_lsn = Lsn(42);
+            storage
+                .persist(&state, true)
+                .expect("failed to persist state");
+        }
+
+        let (_, state) = file_storage(&conf, timeline_id).expect("failed to read state");
```
yeputons, line 640: I'd call `SharedState::load_from_control_file` with `CreateControlFile::False` here instead

walkeeper/src/timeline.rs
```diff
@@ -515,3 +591,79 @@ impl Storage for FileStorage {
...
+
+    fn stub_conf() -> SafeKeeperConf {
+        let workdir = tempfile::tempdir().unwrap().into_path();
+        SafeKeeperConf {
+            workdir,
+            ..Default::default()
+        }
+    }
+
+    fn file_storage(
```
yeputons, line 615: `load_create_control_file`? It's basically a wrapper around `load_from_control_file`, and it's used both to get the `FileStorage` and load `SafeKeeperState` for checking.

yeputons (changes requested):

walkeeper/src/timeline.rs
LizardWizzard: I think it's ok to just issue a warning after the read if its too large. If we care about it's size we should check it on write too. @arssher thoughts?

yeputons: I'm mostly worried about accidentally **ram**ming 5 tb of data. Unconstrained reads scare me a little.

LizardWizzard: If this happens it is a consequence, not the root cause. If we intent to guard against accidently large safekeeper state we should prevent it from growing that big. Otherwise you see normally working safekeeper which fails if you restart it.

walkeeper/src/timeline.rs
LizardWizzard: write_all returns a result, but.resize and copy_from_slice are infallible (sort of), so I decided to have one place to handle the error. Do not see much of a difference

walkeeper/src/timeline.rs
yeputons: True, but this file is out of our control while it lies on the disk. Random `cp some-big-data-file control-file` would be bad. Or if FS is corrupted and thinks the file is much bigger than it is.

Of course, should never happen, and I don't have experience with such scenarios, but still.

walkeeper/src/timeline.rs
yeputons: Good point. `extend_from_slice` (if it does what I think it does) still should be easier to read than three lines of `resize`/`copy_from_slice`

walkeeper/src/timeline.rs
LizardWizzard: Also note that we rewrite it on every `persist` call

yeputons (approved):
LGTM.

Your call on whether to add extra checks before reading control file.

LizardWizzard:
We do the same for other files that are expected to be small like configs, public/private keys etc. So I do not see a problem with that. Feel free to open separate issue to discuss that

walkeeper/src/timeline.rs
```diff
@@ -397,24 +432,66 @@ impl GlobalTimelines {
...
+        // rename should be atomic
+        fs::rename(
+            control_partial_path,
+            self.timeline_dir.join(CONTROL_FILE_NAME),
+        )?;
+        if sync {
+            // fsync the directory (linux specific)
+            File::open(&self.timeline_dir)
+                .and_then(|f| f.sync_all())
+                .with_context(|| "failed to sync control file directory")?;
```
arssher, line 494: From pedantic department: note that Postgres does fsync on the file twice, before and after renaming, see e.g. `durable_rename`. I couldn't quickly find standard or something confirming this is required, but let's add it.

arssher (changes requested):

walkeeper/src/timeline.rs
arssher: I'm not too worried about this, the probability is small and the consequences are not fatal. However, note that if you put checksum in the beginning of the file, you can feed fd directly to `des_from` which will bail out immediately once file structure doesn't match.

arssher (approved):

PR #810, opened 2021-10-28
Tenant threads idle state

lubennikovaav:
Main idea of the PR is to launch and shutdown tenant threads together with walreceiver.
At first tenant exists in an `Idle` state. 
Then `callmemaybe` request comes and pageserver launches walreceiver for the timeline. TenantState changes to `Active` and we spawn tenant threads.
When all walreceivers are gone, the tenant becomes `Idle`, and tenant threads exit after they complete current iteration.

The problem is that now walreceiver in setup with safekeepers never disconnects:
1. In case of connection failure, walreceiver endlessly retries.
I've fixed this in PR with retry_count.

2. Safekeeper doesn't handle failed connection
There is a warning in log, but message loop continues after this message
`WARN query handler for b"START_WAL_PUSH\0" failed: connection closed unexpectedly` 

3. Safekeeper runs `request_callback` in an infinite loop, where constantly attempts to establish replication channel.

I've attempted to fix 2, by breaking the loop and 3 by disabling recall, but walreceiver still hangs in `physical_stream.next()?`
What am I missing?

### DIFF
--- a/pageserver/src/tenant_threads.rs
+++ b/pageserver/src/tenant_threads.rs
@@ -35,33 +35,35 @@ lazy_static! {
     .expect("failed to define a metric");
 }
 
+// Launch checkpointer and GC for the tenant.
+// It's possible that the threads are running already,
+// if so, just don't spawn new ones.
 pub fn start_tenant_threads(conf: &'static PageServerConf, tenantid: ZTenantId) {
-    //ensure that old threads are stopeed
-    wait_for_tenant_threads_to_stop(tenantid);
-
-    let checkpointer_handle = std::thread::Builder::new()
-        .name("Checkpointer thread".into())
-        .spawn(move || {
-            checkpoint_loop(tenantid, conf).expect("Checkpointer thread died");
-        })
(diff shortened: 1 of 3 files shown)


pageserver/src/tenant_threads.rs
```diff
@@ -69,6 +81,12 @@ pub fn wait_for_tenant_threads_to_stop(tenantid: ZTenantId) {
 /// Checkpointer thread's main loop
 ///
 fn checkpoint_loop(tenantid: ZTenantId, conf: &'static PageServerConf) -> Result<()> {
+    let gauge = TENANT_THREADS_COUNT.with_label_values(&["checkpointer"]);
```
SomeoneToIgnore, line 84: NIT: Feels like we could move both of those gauge fiddlings one level above, into `spawn(move || {` block, since in the loop functions not much of a thread as a context left.

SomeoneToIgnore (approved):
I would use different names for `TenantState` variants, but that can always happen later.

Issue #791, opened 2021-10-22
Improve typing for Python tests


yeputons:
Currently the following checks are disabled for `mypy`. I think it's worth enabling them back, fixing issues as we go:

* [x] `check_untyped_defs` (#3142)
* [ ] `disallow_incomplete_defs`
* [ ] `disallow_untyped_calls`
* [ ] `disallow_untyped_decorators`
* [ ] `disallow_untyped_defs`

Of course, we still have the issue of no type hints for `asyncpg` (although [there is some work-in-progress](https://github.com/MagicStack/asyncpg/pull/577)), `cached_property` and `pytest`. I'm wondering how well `disallow_untyped_decorators` and friends work with `ignore_missing_imports` for those libraries.

Also:

* [ ] Change most `Any`s to something more specific
* [ ] Get rid of most `cast`s

funbringer:
Personally, I doubt that we should treat those settings as undisputed setbacks on our road to typechecking everything. There's a concept of gradual typing, which allows one to refine types as necessary. Originally, I wanted to enable go-to-definition etc for our fixture code, preserving the ability to omit types in tests themselves. That's why I've relaxed some of the mypy's settings.

One of the reasons we don't write tests in rust is that it's not as forgiving and unsophisticated as python.

PR #813, opened 2021-10-28
Check python for the whole repository

yeputons:
* Reformat `pre-commit.py` accordingly
* Exclude `vendor/` directory from mypy
* Add `.yapfignore` from `.gitignore`: yapf tries to detect Python files by
  reading them in full. If one generates 10G of binaries and test data locally,
  it may take a while. Excluding typically big directories explicitly.


yeputons:
Good questions. I will think what to do.

Currently the situation is:

1. Both `yapf` and `mypy` read from `setup.cfg`. However, yapf needs list of ignored files exactly in `.yapfignored`, see https://github.com/google/yapf/issues/357 and https://github.com/google/yapf/pull/543 . Unexpected, and PR is probably not that easy, as they only load "style" from `setup.cfg`, not whole settings.
2. `yapf` correctly finds `setup.cfg` a parent directory, but `mypy` does not and it's unexpectedly by design: https://github.com/python/mypy/issues/3377#issuecomment-302308046 . Another argument being that you cannot check arbitrary sources as they probably `import` each another and that may or may not (I'm not sure) depend or your initial current directory.
3. Yes, confusing indeed.
4. Updating `pre-commit.py` is a good idea, maybe in a separate PR or here.
5. They use slightly different executors (for no great reason, though). I'll update name in this PR.

My intermediate thoughts:

* `.yapfignored` is here mostly for the `vendor/` directory and as some optimization. So it's ok if it's not read.
* `setup.cfg` for `mypy` is a different beast. Currently all knobs are actually defaults, but I hope to change it in #791. It's bad if some stricter checks are disabled unless you run `mypy` exactly right.
* Does not seem like there is an easy way to alter mypy. Probably need to learn a bit about Python imports to see if it even makes sense to run `mypy` from non-root directories (I think it worked for me..?)

funbringer (approved):
The more checks the better!

> Also I find it confusing that we now have Pipfile in the project dir (symlink), some config files in test_runner dir (setup.cfg, pytest.ini, original Pipfile) but again yapfignore in the root dir.

True, I'm also concerned. And, since we have to do something about this, I think it's high time we started moving towards poetry + black, which we already use in https://github.com/zenithdb/console.

Adding those `make {test,fmt,mypy,whatever}` rules would make such transitions easier for the team.

yeputons:
# My new plan for this PR

1. [ ] Put all Python-related configuration files in the repository's root only:
    * `Pipfile` and `Pipfile.lock`: dependencies and dev dependencies
    * `setup.cfg`: linter settings, mypy settings
    * `.yapfignore`: list of top-level directories for linter to ignore (e.g. `.git/`, `test_output/`)
    * `pytest.ini`: log settings for pytest
2. [ ] Update `pre-commit.py` accordingly (`make fmt` just calls it). TODO: whether to run `pipenv install --dev`?
3. [ ] Rename `check-codestyle` to `check-codestyle-rust`.
4. [ ] Move Python-related docs to a separate repository-wide doc. In particular, ask the user to run `mypy` from the root directory only (or provide `--config` option)

That way this PR is more like a clean-up for the status quo.

All of the following are out of scope at the moment: adding `tox` (if necessary?), migrating `yapf` to `black`, `pipenv` to `poetry`, `mypy` to some other typechecker (`pytype`, `pyright`, `pyre`). I don't their either of these changes will significantly simplify this PR. Migration in the future may be warranted, though. E.g. to avoid configuration quirks of `mypy` or `yapf`, but they're relatively painless at the moment, I think.

# Reasoning
## Our current Python code
We have independent Python "projects": `test_runner/` and two individual scripts: `pre-commit.py`, `scripts/generate_perf_report_page.py` from #631. A "project" here is a bunch of Python files which may `import` each other directly or indirectly.

I expect there will be more scripts and maybe one or two directories akin to `test_runner/`. I don't expect these scripts to be run except when building from source.

## `pipenv`
Dependencies of the forementioned "projects" may differ, but we don't have lots of them. A single virtual environment is simpler to manage, so I propose we use the same `pipenv` for all "projects". Running `pipenv` from a child directory will look for `Pipfile` in parent directories just fine.

So the proposal is to "unify" all these projects. Later I check how different tools behave when run from a subdirectory of the repository.

I don't see any landmines here.

## `pytest`
`pytest` looks for `pytest.ini` in parent directories just fine. The one where it's located becomes the `rootdir`. It's ok to have repository as pytest's `rootdir`, as all imports are still [resolved independently of it](https://docs.pytest.org/en/6.2.x/pythonpath.html#pythonpath) by default for compatibility.

Basically, all packages become available for everyone irrlevent of their location: `import fixtures` from anywhere will always find `test_runner/fixtures/`. That plan to break it in the future by switching `--import-mode prepend` to `--import-mode importlib`, but we can just run `python3 -m pytest` from the `test_runner/` directory then so the `test_runner/` directory is prepended to module search path.

I don't see any landmines here.

## `mypy`
`mypy` [resolves imports](https://mypy.readthedocs.io/en/stable/running_mypy.html#finding-imports) ([permalink](https://github.com/python/mypy/blob/f5fc579cf07f2078c9312044f6bcb132f891d746/docs/source/running_mypy.rst#following-imports)) lenientlier than Python. Similar to `pytest`, actually.

* In this regard, running `mypy` from (almost) any parent directory is ok unless there are packages with the same name in different parts of the filesystem. So it's safe to run both from repository's root and from `test_runner/`. However, when run from `test_runner/batch_others/`, it fails to find the `fixtures/` package.
* Little more details: it looks for all directories which contain an `__init__.py` and do not have `__init__.py` above them. All such directories are considered to be separate "root" modules. E.g. the `test_runner/fixtures/` directory provides `fixtures` module.

`mypy` only looks for the configuration file in the current directory (as well as some user-specific directories) or at the `--config` option, no way around it. We need the configuration for two things: a) suppress errors about missing type hints; b) provide stricter flags in the future. Unfortunately, both are important.

**Landmine**: it's necessary to run `mypy` from the repository's root in the CI and require users to do the same.

## `yapf`
It only cares about syntax and does not try to resolve any imports. It reads `setup.cfg` from parent directories, so it's safe to call it from anywhere.

`.yapfignore` is only needed to speed up Python source search by excluding some big directories in the repository's root. So if one is running `yapf` in a subdirectory, it's ok to ignore it.

**Landmine**: we are not able to consistently exclude non-top-level directories/files from `yapf` checks if the user runs it from non-root directory.

## `flake8`, `pycodestyle`
Both of these successfully load `setup.cfg` from parent directories. I haven't used them much, and I'm not aware of any gotchas.

test_runner/README.md
```diff
@@ -3,21 +3,13 @@
 This directory contains integration tests.
 
 Prerequisites:
-- Python 3.7 or later
-    - Development headers may also be needed to build `psycopg2` from source.
```
yeputons, line 7: I've removed this line in c5b8fd71703e90ff1bed8cfad803515e74ac84ff because we use `psycopg2-binary` instead of `psycopg2` since #631.

Although I'm a little concerned it says "The binary package is a practical choice for development and testing but in production it is advised to use the package built from sources." [here](https://pypi.org/project/psycopg2-binary/).

yeputons:
Resolved merge conflicts, ready for review.

The following files should have no changes at all compared to the current `main`: `Pipfile`, `Pipfile.lock`, `pytest.ini`. I've checked it by running `git diff origin/main:test_runner/pytest.ini check-python-toplevel:pytest.ini`

test_runner/README.md
LizardWizzard: We use it only for testing purposes, so I don't see any problems with that

LizardWizzard (approved):
> In particular, ask the user to run mypy from the root directory only (or provide --config option)

I think this is not ideal behaviour. IMO there should be one `make fmt` that just does the right thing. Otherwise if we change something in our linters/checkers flow it would confuse people and require them to learn various quirks of invoking different tools.

If our current pre-commit.py does the right thing (as I сan see this is the case) we can directly point to it (or make fmt) in the docs.

> We have independent Python "projects"

Currently scripts are independent (may not be necessary the case in the future, but IMO it is ok to assume that). 
As an idea we can move pre-commit to scripts directory.

>  it looks for all directories which contain an __init__.py

I had this problem when working on performance tests. And mypy forced me to add `test_runner/performance/__init__.py`. 

Looks good to me. This is definitely an improvement. In the future it will be good to investigate applicability of tools that use more modern approach with single pyproject.toml. But a do not think that this has high priority

PR #814, opened 2021-10-29
Enable async deserialization of FeMessage

funbringer:
Now it's possible to call Fe{Startup,}Message in both sync and async contexts, which is good for proxy.

### DIFF
--- a/zenith_utils/src/pq_proto.rs
+++ b/zenith_utils/src/pq_proto.rs
@@ -2,17 +2,17 @@
 //! <https://www.postgresql.org/docs/devel/protocol-message-formats.html>
 //! on message formats.
 
+use crate::sync::{AsyncishRead, SyncFuture};
 use anyhow::{bail, ensure, Context, Result};
-use byteorder::{BigEndian, ByteOrder};
-use byteorder::{ReadBytesExt, BE};
 use bytes::{Buf, BufMut, Bytes, BytesMut};
 use postgres_protocol::PG_EPOCH;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
-use std::io::Read;
+use std::future::Future;
 use std::io::{self, Cursor};
 use std::str;
 use std::time::{Duration, SystemTime};
+use tokio::io::AsyncReadExt;
 use tracing::info;
 
 pub type Oid = u32;
@@ -96,14 +96,14 @@ impl FeMessage {
     /// One way to handle this properly:
     ///
     /// ```
-    /// # use std::io::Read;
+    /// # use std::io;
     /// # use zenith_utils::pq_proto::FeMessage;
     /// #
     /// # fn process_message(msg: FeMessage) -> anyhow::Result<()> {
     /// #     Ok(())
     /// # };
     /// #
-    /// fn do_the_job(stream: &mut impl Read) -> anyhow::Result<()> {
+    /// fn do_the_job(stream: &mut (impl io::Read + Unpin)) -> anyhow::Result<()> {
     ///     while let Some(msg) = FeMessage::read(stream)? {
     ///         process_message(msg)?;
     ///     }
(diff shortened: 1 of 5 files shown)


zenith_utils/src/sync.rs
```diff
@@ -0,0 +1,133 @@
...
+impl<T: io::Read + Unpin> tokio::io::AsyncRead for AsyncishRead<T> {
+    #[inline(always)]
+    fn poll_read(
+        mut self: Pin<&mut Self>,
+        _cx: &mut task::Context<'_>,
+        buf: &mut tokio::io::ReadBuf<'_>,
+    ) -> task::Poll<io::Result<()>> {
+        task::Poll::Ready(
+            // This will block, meaning we don't need a real event loop!
+            self.0
```
LizardWizzard, line 84: Could you elaborate? Why this would block? Will it block when called from async context?

LizardWizzard (commented):
IMO this is a really clever implementation, but it comes with some non-obvious parts, and some unsafe code.

If I'm understanding it correctly we make roughly two read calls, for "header" with message type and body length, and the second for body itself. So what do you think about having the buffer reading separately sync/async and the parsing moved to common sync part? Is it possible?

zenith_utils/src/sync.rs
funbringer: It will block regardless of the context because we call `Read::read`, which is inherently synchronous.

LizardWizzard: Then it seems that there is no benefit in async execution, right?

funbringer:
> So what do you think about having the buffer reading separately sync/async and the parsing moved to common sync part? Is it possible?

Frankly speaking, I'm not interested in your proposal for several reasons:

* This allows us to reuse the code without any macro magic, which I don't like.
* As written, this function only consumes as many bytes as it needs to perform the conversion. The buffering may happen in the channel implementation, which means they will be loosely coupled.
* I'm the author, so obviously I'm biased towards picking my own solution. I don't see any problem in that as long as it meets the requirements.

zenith_utils/src/sync.rs
funbringer: > Then it seems that there is no benefit in async execution, right?

Not quite. You can still use `read_fut`, which is fully asynchronous, as it doesn't use `AsyncishRead`.

funbringer: To elaborate: `SyncFuture` is a future which we may synchronously wait for **under certain circumstances**. AsyncishRead serves as a proof that we may do that. In async context, we may always just `.await` it.

LizardWizzard:
> This allows us to reuse the code without any macro magic, which I don't like.

Didn't get it, what macro magic you are referring to?

> this function only consumes as many bytes as it needs to perform the conversion

Maybe we are not on the same page, there is already a buffer for every message body. If we add a tag to it, then we can split reading and processing. In that case reading should support both, sync/async and processing always stays sync. In that case we actually utilize async machinery to have more concurrency than with the sync version.

In general I do not see any benefits in having special magic to use blocking calls in async, in that case it mimics (sort of) spawn_blocking from tokio isn't it? And transition of proxy to async is intended to bring async benefits to connection handling, but if we use blocking calls there I see almost no difference. Please correct me if I'm wrong

> but they don't oblige us to
This is correct, but IMO almost always pointless

zenith_utils/src/sync.rs
```diff
@@ -0,0 +1,133 @@
...
+        nested: T,
+        _marker: std::marker::PhantomData<S>,
+    }
+}
+
+impl<S, T: Future> SyncFuture<S, T> {
+    pub(crate) fn new(nested: T) -> Self {
+        Self {
+            nested,
+            _marker: Default::default(),
```
SomeoneToIgnore, line 21: NIT:

Phantom data could be just constructed, without trait solving magic involved:  `_marker: PhantomData,`

zenith_utils/src/sync.rs
```diff
@@ -0,0 +1,133 @@
...
+impl<T: SyncProof> SyncProof for &T {}
+impl<T: SyncProof> SyncProof for &mut T {}
+
+impl<P: SyncProof, T: Future> SyncFuture<P, T> {
+    /// Synchronously wait for future completion
+    pub fn wait(mut self) -> T::Output {
+        const RAW_WAKER: task::RawWaker = task::RawWaker::new(
+            std::ptr::null(),
+            &task::RawWakerVTable::new(
+                |_| RAW_WAKER,
```
SomeoneToIgnore, line 47: Out of curiosity, how's what working in Rust?

I seem to fully miss the point on how it's possible to

1. call the constructor function as `const RAW_WAKER: task::RawWaker = task::RawWaker::new(`
2. create one of the function arguments to the constructor function as `&task::RawWakerVTable::new(`
3. reference itself from `1.` in `2.` as `|_| RAW_WAKER` in the argument creation in a `Fn` closure that's coerced to `unsafe fn(*const ()) -> RawWaker`

Syntactically looks like a circular dependency, where one object cannot really cbe constructed without another unless a compiler does some trick?

Also, there are no `Clone` or `Copy` impls for `RawWaker`, how that closure can exist or be called when we have moved the `RAW_WAKER` later into `task::Waker::from_raw`?

SomeoneToIgnore (commented):
Looks interesting, is there any proxy feature usage example, just to see the whole spectrum of improvements?

zenith_utils/src/sync.rs
funbringer: Sure, why not.

funbringer:
After we'd discussed this with @LizardWizzard elsewhere, I decided to write more docs. Hope this clarifies things a bit.

zenith_utils/src/sync.rs
funbringer: FWIW, even the most trivial cycle is allowed syntactically (it obviously fails semantic checks, though):

```rust
error[E0391]: cycle detected when simplifying constant for the type system `FOO`
 --> const.rs:1:1
  |
1 | const FOO: u32 = FOO;
  | ^^^^^^^^^^^^^^^^^^^^^
  |
note: ...which requires simplifying constant for the type system `FOO`...
 --> const.rs:1:1
  |
1 | const FOO: u32 = FOO;
  | ^^^^^^^^^^^^^^^^^^^^^
note: ...which requires const-evaluating + checking `FOO`...
 --> const.rs:1:1
  |
1 | const FOO: u32 = FOO;
  | ^^^^^^^^^^^^^^^^^^^^^
  = note: ...which requires normalizing `FOO`...
  = note: ...which again requires simplifying constant for the type system `FOO`, completing the cycle
  = note: cycle used when running analysis passes on this crate

error: aborting due to previous error; 1 warning emitted
```

> Also, there are no Clone or Copy impls for RawWaker, how that closure can exist or be called when we have moved the RAW_WAKER later into task::Waker::from_raw

We haven't really moved anything. Consts are expanded each time we reference them, so it's not "that very same object".

I think it's not much different from recursive functions, where you can reference the function you haven't really defined yet. Here you have two mutually recursive thingies: a closure (function, really) and a value.

funbringer:
> Looks interesting, is there any proxy feature usage example, just to see the whole spectrum of improvements?

Unfortunately, no. I had to introduce this change in order to enable migration to async in proxy, but at the moment there's nothing to boast with.

zenith_utils/src/sync.rs
SomeoneToIgnore: Oh, I've missed const context entirely here, that explains things more.

Thanks for elaborating.

zenith_utils/src/sync.rs
```diff
@@ -0,0 +1,174 @@
...
+    #[inline(always)]
+    fn poll(self: Pin<&mut Self>, cx: &mut task::Context<'_>) -> task::Poll<Self::Output> {
+        self.project().nested.poll(cx)
+    }
+}
+
+/// Proves that we can call [`SyncFuture::wait`].
+/// If implementer is also a [`Future`], it should always
+/// return [`task::Poll::Ready`] from [`Future::poll`].
+pub trait SyncProof {}
```
bojanserafimov, line 74:
```suggestion
///
/// Each implementation should document which futures specifically are being declared sync-proof
pub trait SyncProof {}
```

zenith_utils/src/sync.rs
```diff
@@ -0,0 +1,174 @@
...
+    }
+}
+
+/// This wrapper turns any [`std::io::Read`] into a blocking [`tokio::io::AsyncRead`],
+/// which lets us abstract over sync & async readers in methods returning [`SyncFuture`].
+/// NOTE: you **should not** use this in async code.
+#[repr(transparent)]
+pub struct AsyncishRead<T: io::Read + Unpin>(pub T);
+
+/// This lets us call [`SyncFuture::wait`].
```
bojanserafimov, line 113:
```suggestion
/// This lets us call [`SyncFuture<AsyncishRead<_>, _>::wait`], and allows the future to await on any
/// of the AsyncRead and AsyncReadExt methods on AsyncishRead.
```

zenith_utils/src/sync.rs
```diff
@@ -0,0 +1,174 @@
...
+use std::future::Future;
+use std::marker::PhantomData;
+use std::pin::Pin;
+use std::{io, task};
+
+pin_project! {
+    /// We use this future to mark certain methods
+    /// as callable in both sync and async modes.
+    #[repr(transparent)]
+    pub struct SyncFuture<S, T: Future> {
```
bojanserafimov, line 11: The name `SyncFuture` is a bit misleading because it's only sync if `S` is SyncProof. But I couldn't think of a better name, so I can't object.

zenith_utils/src/sync.rs
```diff
@@ -0,0 +1,174 @@
...
+        #[pin]
+        nested: T,
```
bojanserafimov, line 13: I'd rename this to `inner`, since it's [somewhat](https://doc.rust-lang.org/src/std/fs.rs.html#92-94) of a convention. This is a tiny nitpick and it doesn't matter that much, but this is a difficult file to understand so anything helps.

bojanserafimov (approved):
It's a very clever solution but hard to reason about, since the `SyncFuture` contract is a lot to keep track of. I bet the second time I read this code I'll have to re-learn how it works.

I tried hard to find alternative solutions, and [this](https://github.com/bojanserafimov/rust-experiments/blob/main/sync-async-parser/src/main.rs ) is the best solution I came up with. My approach is safer and easier to reason about, but without generators and yield syntax we'll need to manually implement parser state machines, which requires a bigger refactor of the parser. This refactor is probably not worth, especially if we're going full async soon.

So let's go with your solution instead. I've added a few docstring suggestions to help with readability (see inline comments). Feel free to address the ones that make sense to you and merge.

zenith_utils/src/pq_proto.rs
```diff
@@ -93,139 +92,171 @@ impl FeMessage {
...
+
+    /// Read one message from the stream.
+    /// See documentation for `Self::read`.
+    pub fn read_fut<Reader>(
+        stream: &mut Reader,
+    ) -> SyncFuture<Reader, impl Future<Output = anyhow::Result<Option<FeMessage>>> + '_>
+    where
+        Reader: tokio::io::AsyncRead + Unpin,
+    {
+        SyncFuture::new(async move {
```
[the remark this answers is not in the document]
bojanserafimov: If we break the contract we could get a panic, but not realiably! It could pass in tests and fail in production.

zenith_utils/src/sync.rs
funbringer: No problem!

zenith_utils/src/sync.rs
funbringer: I won't mind if we rename this once we've come up with a better name :)

PR #818, opened 2021-11-01
Name walkeeper threads to make debugging more convenient

lubennikovaav:


arssher (approved):
+1

I also wanted to add timelineid to the context and ideally unique per-connection id to threads getting WAL from proposer, but can be done separately.

PR #825, opened 2021-11-02
pageserver - make `PageVersion` an enum

patins:


SomeoneToIgnore (approved):
Good, looks like we've never had the case with both version variants constructed at the same time, despite the comment.

hlinnaka (approved):

PR #827, opened 2021-11-02
WIP: Test docs update

yeputons:
Based on #813 

* [ ] Move `pre-commit.py` to the `scripts/` directory
* [ ] Point to `make fmt` in docs instead of giving out all details about `mypy` and stuff

### DIFF
--- a/test_runner/README.md
+++ b/test_runner/README.md
@@ -17,8 +17,8 @@ The tests are divided into a few batches, such that each batch takes roughly
 the same amount of time. The batches can be run in parallel, to minimize total
 runtime. Currently, there are only two batches:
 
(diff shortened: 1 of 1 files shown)


test_runner/README.md
```diff
@@ -28,38 +28,37 @@ run the tests from within the `test_runner` directory.
...
-
-`pipenv run pytest test_pgbench.py`
-
-If you want to run all tests that have the string "bench" in their names:
-
-`pipenv run pytest -k bench`
+* You can run all the tests with:
+  ```bash
+  pipenv run pytest
+  ```
```
petuhovskiy, line 34: I'd also add an example for running tests in parallel (with `-n4` for example), to execute them faster.

PR #829, opened 2021-11-03
Use PG timelineid 1 in safekeeper everywhere.

arssher:
As changing it doesn't have useful meaning in Zenith.

### DIFF
--- a/walkeeper/src/replication.rs
+++ b/walkeeper/src/replication.rs
@@ -6,7 +6,9 @@ use crate::timeline::{ReplicaState, Timeline, TimelineTools};
 use anyhow::{anyhow, Context, Result};
 use bytes::Bytes;
 use log::*;
-use postgres_ffi::xlog_utils::{get_current_timestamp, TimestampTz, XLogFileName, MAX_SEND_SIZE};
+use postgres_ffi::xlog_utils::{
+    get_current_timestamp, TimestampTz, XLogFileName, MAX_SEND_SIZE, PG_TLI,
+};
 use regex::Regex;
 use serde::{Deserialize, Serialize};
 use std::cmp::min;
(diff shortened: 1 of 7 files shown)


walkeeper/src/lib.rs
```diff
@@ -22,6 +22,9 @@ pub mod defaults {
     pub const DEFAULT_HTTP_LISTEN_ADDR: &str = formatcp!("127.0.0.1:{DEFAULT_HTTP_LISTEN_PORT}");
 }
 
+// PG timeline is always 1, changing it doesn't have useful meaning in Zenith.
+pub const PG_TLI: u32 = 1;
```
lubennikovaav, line 26: Please, move this definition to `postgres_ffi/xlog_utils.rs`, so that we could use it in pageserver too.
We have a couple of places with comments about this tli: `// FIXME: always use Postgres timeline 1`

walkeeper/src/replication.rs
```diff
@@ -244,7 +245,7 @@ impl ReplicationConn {
                 None => {
                     // Open a new file.
                     let segno = start_pos.segment_number(wal_seg_size);
-                    let wal_file_name = XLogFileName(timeline, segno, wal_seg_size);
+                    let wal_file_name = XLogFileName(PG_TLI, segno, wal_seg_size);
```
MMeent, line 248: Are there any reasons left why we would want to retain a timeline ID in the file name?

MMeent (commented):
Thanks! One extra comment added regarding file naming now that we don't use PG's TLI anymore, and I second that PG_TLI should be moved to postgres_ffi.

walkeeper/src/replication.rs
arssher: Yes, at least for consistency to have 1 everywhere. I'd expect 0 break various tooling, for example pg_waldump assumes timeline 1 by default and will search for such files.

walkeeper/src/lib.rs
arssher: done

lubennikovaav (approved):

walkeeper/src/replication.rs
MMeent: Consistency with PG's naming in that case? Because we could reasonably use our own naming scheme for XLOG segment files, one without TLI.

But, compatibility sounds like a good enough reason.

MMeent (approved):

PR #830, opened 2021-11-03
Adjust timeouts for test_restarts_under_load

petuhovskiy:
There were several `Too long with no output` errors in CircleCI jobs, for example:
- https://app.circleci.com/pipelines/github/zenithdb/zenith/3030/workflows/20e84844-3c0d-4745-a94d-750e4ee25e15/jobs/24290
- https://app.circleci.com/pipelines/github/zenithdb/zenith/3052/workflows/30525649-6b75-485f-9983-8576c427c9d2/jobs/24498

These timeouts happen because test_restarts_under_load takes a long time to complete, and pytest is not writing anything to output at that time. 

This PR changes timeouts to get timeout error from pytest rather than CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -243,6 +243,9 @@ jobs:
       - run:
           name: Run pytest
           working_directory: test_runner
+          # pytest doesn't output test logs in real time, so CI job may fail
+          # with `Too long with no output` error, if a test is running for a long time.
+          no_output_timeout: 20m
           environment:
             - ZENITH_BIN: /tmp/zenith/bin
             - POSTGRES_DISTRIB_DIR: /tmp/zenith/pg_install
--- a/test_runner/batch_others/test_wal_acceptor_async.py
+++ b/test_runner/batch_others/test_wal_acceptor_async.py
@@ -147,6 +147,9 @@ async def run_restarts_under_load(pg: Postgres, acceptors: List[Safekeeper], n_w
     period_time = 10
     iterations = 6
 
+    test_timeout = 15 * 60  # 15 minutes
+    started_at = time.time()
+
     pg_conn = await pg.connect_async()
     tenant_id = await pg_conn.fetchval("show zenith.zenith_tenant")
     timeline_id = await pg_conn.fetchval("show zenith.zenith_timeline")
@@ -162,6 +165,9 @@ async def run_restarts_under_load(pg: Postgres, acceptors: List[Safekeeper], n_w
         workers.append(asyncio.create_task(worker))
 
     for it in range(iterations):
+        if time.time() - started_at > test_timeout:
+            raise RuntimeError('test timed out')
+
         victim_idx = it % len(acceptors)
         victim = acceptors[victim_idx]
         victim.stop()
--- a/test_runner/batch_others/test_wal_acceptor_async.py
+++ b/test_runner/batch_others/test_wal_acceptor_async.py
@@ -109,7 +109,7 @@ async def wait_for_lsn(safekeeper: Safekeeper,
                        timeline_id: str,
                        wait_lsn: str,
                        polling_interval=1,
-                       timeout=600):
+                       timeout=200):
     """
     Poll flush_lsn from safekeeper until it's greater or equal than
     provided wait_lsn. To do that, timeline_status is fetched from
--- a/test_runner/batch_others/test_wal_acceptor_async.py
+++ b/test_runner/batch_others/test_wal_acceptor_async.py
@@ -147,8 +147,9 @@ async def run_restarts_under_load(pg: Postgres, acceptors: List[Safekeeper], n_w
     period_time = 10
     iterations = 6
 
-    test_timeout = 15 * 60  # 15 minutes
-    started_at = time.time()
+    # Set timeout for this test at 15 minutes. It should be enough for
+    # test to complete and less than CircleCI's no_output_timeout.
+    test_timeout_at = time.monotonic() + 15 * 60
 
     pg_conn = await pg.connect_async()
     tenant_id = await pg_conn.fetchval("show zenith.zenith_tenant")
@@ -165,8 +166,7 @@ async def run_restarts_under_load(pg: Postgres, acceptors: List[Safekeeper], n_w
         workers.append(asyncio.create_task(worker))
 
     for it in range(iterations):
-        if time.time() - started_at > test_timeout:
-            raise RuntimeError('test timed out')
+        assert time.monotonic() < test_timeout_at, 'test timed out'
 
         victim_idx = it % len(acceptors)
         victim = acceptors[victim_idx]


.circleci/config.yml
```diff
@@ -243,6 +243,9 @@ jobs:
       - run:
           name: Run pytest
           working_directory: test_runner
+          # pytest doesn't output test logs in real time, so CI job may fail
+          # with `Too long with no output` error, if a test is running for a long time.
+          no_output_timeout: 20m
```
yeputons, line 248: It seems that each `run-pytest` completes in under ten minutes, and `test_restarts_under_load` in particular takes about three minutes only under normal circumstances: https://app.circleci.com/pipelines/github/zenithdb/zenith/3061/workflows/42b1e5ca-db66-41aa-9284-f9b3059ae060 .

I'm not sure if raising `no_output_timeout` to 20 minutes is a good idea. If a test understands that it's stuck or significantly slower than expected, it should fail itself. CI timelimits should be the last resort.

test_runner/batch_others/test_wal_acceptor_async.py
```diff
@@ -109,7 +109,7 @@ async def wait_for_lsn(safekeeper: Safekeeper,
                        timeline_id: str,
                        wait_lsn: str,
                        polling_interval=1,
-                       timeout=600):
+                       timeout=200):
```
yeputons, line 112: `wait_for_lsn`'s timeout is for a single safekeeper for a single iteration only. Six iterations times two safekeepers (one is a victim) times 200s timeout is 40 minutes per test. Seems excessive to me, and it's definitely much bigger than the `no_output_timeout`, so it probably won't help.

E.g. in https://app.circleci.com/pipelines/github/zenithdb/zenith/3061/workflows/42b1e5ca-db66-41aa-9284-f9b3059ae060/jobs/24588/steps `test_restarts_under_load` starts at `2021-11-03 10:25:38.251`, followed by `waiting for lsn` at:

* `2021-11-03 10:25:44.213` and `2021-11-03 10:25:44.217`, synced at `2021-11-03 10:25:54.218`
* `2021-11-03 10:25:55.309` and `2021-11-03 10:26:04.344`, synced at `2021-11-03 10:26:14.345`
* `2021-11-03 10:26:15.424` and `2021-11-03 10:26:15.428`, synced at `2021-11-03 10:26:45.501`
* `2021-11-03 10:26:46.585` and `2021-11-03 10:26:46.589`, synced at `2021-11-03 10:27:19.670`
* `2021-11-03 10:27:20.775` and `2021-11-03 10:27:40.882`, synced at `2021-11-03 10:27:50.883`
* `2021-11-03 10:27:51.987` and `2021-11-03 10:27:51.992`, synced at `2021-11-03 10:28:22.69`

It seems that each iteration takes between 10 and 30 seconds, so I'd use actually lower timeout to, say, 45 or 60 seconds. In the latter case the maximum time spent in a test is about 12 minutes, so raising `no_output_timeout` starts looking like a good idea.

Looking at the logs, maybe timeout should actually be per-iteration, not per-safekeeper.

yeputons (commented):
Another idea: maybe run `test_restarts_under_load` separately and without `-n4`? That way pytest will print logs live.

test_runner/batch_others/test_wal_acceptor_async.py
petuhovskiy: > synced at HH:MM:SS

Hmm, I think you're using `All workers made N transactions` entry as a "synced at" moment, but sync actually completes 10 seconds before that, there's `asyncio.sleep(10)` in the code after sync.

> Looking at the logs, maybe timeout should actually be per-iteration, not per-safekeeper.

It's actually not that different. In single iteration we have 2 alive safekeepers, where one has just been started and another has been working for some time. Safekeeper which was working in previous iteration should have all WAL up to wait_lsn and we won't wait for him. Another safekeeper (which has just been started) is likely to fall behind significantly and we will spend most time waiting for him in `wait_for_lsn`.

> It seems that each iteration takes between 10 and 30 seconds

I would say this duration is very unstable. When I was adjusting timeout last time, I tried to set timeout=120 and it didn't work, because sync was too slow to complete in 120 seconds, so I set it to 600. From my recent observations, wal_proposer sync speed is very unstable, and can be slow and fast in the two consecutive test runs.

In my opinion, this unstable behavior is the root cause of timeout errors and I hope we will fix it soon. It seems that syncing always took about ~6-10 seconds when everything was stable. I would say this PR tries to mitigate CircleCI timeouts temporarily, until root cause is fixed. CircleCI timeouts are not very good because they prevent pytest logs from being shown.

petuhovskiy:
> Another idea: maybe run test_restarts_under_load separately and without -n4? That way pytest will print logs live.

It can work, but not sure if it's worth it. 

Another idea from me relates to yet another problem, seen in `test_restarts_under_load/pg.log` at job 24498.
https://24498-351806852-gh.circle-artifacts.com/0/tmp/test_output/test_restarts_under_load/repo/pgdatadirs/tenants/93a347f1870140e7a7ab3fdc86b3b99c/test_wal_acceptors_restarts_under_load/pg.log  

There is a problem where safekeeper can't start receiving WAL because of `ERROR:  invalid xl_tot_len 0 at 0/23B0018`. wal_proposer tries to reconnect and start pusing WAL every second or so, but can't. It also means that safekeeper's `flush_lsn` is not changed. Maybe we can detect stuck `flush_lsn` and mark test as failed faster. This will prevent timeout error in that case.

.circleci/config.yml
petuhovskiy: I also prefer test fails to CI timeouts. I think raising `no_output_timeout` will help with that, at least temporarily.

yeputons: How about adding some code to `test_restarts_under_load` so it ensures a timeout of, say, 15 minutes on its own?

I think checking on each iteration is fine. 15 minutes + 200 seconds < 20 minutes.

test_runner/batch_others/test_wal_acceptor_async.py
```diff
@@ -147,6 +147,9 @@ async def run_restarts_under_load(pg: Postgres, acceptors: List[Safekeeper], n_w
     period_time = 10
     iterations = 6
 
+    test_timeout = 15 * 60  # 15 minutes
```
yeputons, line 150: * Better use `time.monotonic()`, otherwise the test may be randomly affected by clock adjustment. It _probably_ won't be big, but still.

And two nits:

* Add a comment on how this constant is chosen. E.g. 20 minutes is no good.
* Maybe replace `test_timeout` and `started_at` with `timeout_at = time.monotonic() + 15 * 60`.

test_runner/batch_others/test_wal_acceptor_async.py
```diff
@@ -162,6 +165,9 @@ async def run_restarts_under_load(pg: Postgres, acceptors: List[Safekeeper], n_w
         workers.append(asyncio.create_task(worker))
 
     for it in range(iterations):
+        if time.time() - started_at > test_timeout:
+            raise RuntimeError('test timed out')
```
yeputons, line 169: `assert` should be enough

yeputons (changes requested):

test_runner/batch_others/test_wal_acceptor_async.py
```diff
@@ -147,6 +147,10 @@ async def run_restarts_under_load(pg: Postgres, acceptors: List[Safekeeper], n_w
     period_time = 10
     iterations = 6
 
+    # Set timeout for this test at 15 minutes. It should be enough for
+    # test to complete and less than CircleCI's no_output_timeout.
```
yeputons, line 151: `It` --> `It plus a single iteration timeout` or something. Because we only check once per iteration.

yeputons (approved):

.circleci/config.yml
LizardWizzard: Just to note, there is handy plugin for pytest to enable per test timeouts https://pypi.org/project/pytest-timeout/.

yeputons: Good find. How dangerous is it? There are some flashy warnings in its README. For example, I'm not sure it will gracefully terminate the test and stop all processes, though the `signal` method looks promising. On the other hand, we use some C code, and if the signal handler gets called while GIL is locked... I don't know.

LizardWizzard: I've used it in several python projects without problems. Timeout is indeed the last resort and in CI if there are any processes left they will die with the container anyway.

PR #840, opened 2021-11-04
Follow up stagin benchmarking

LizardWizzard:
* change zenith-perf-data checkout ref to be main
* set cluster id through secrets so there is no code changes required
  when we wipe out clusters on staging


LizardWizzard:
That's strange but after a failed run of benchmarks on github actions, workflow itself disappeared from checks section entirely 🤔

LizardWizzard:
Tested in branch, passed successfully. I'm merging it and see how it goes

PR #848, opened 2021-11-05
Allow downloading remote file range

SomeoneToIgnore:
Another small step to prepare for S3 archives, helps when a part of remote archive such as metadata (as in proposed timeline compression design) or a particular file (in case any other archivation approach is picked and the offsets are known).

### DIFF
--- a/pageserver/src/remote_storage/local_fs.rs
+++ b/pageserver/src/remote_storage/local_fs.rs
@@ -11,16 +11,15 @@ use std::{
     pin::Pin,
 };
 
-use anyhow::{bail, Context};
+use anyhow::{bail, ensure, Context};
 use tokio::{
     fs,
-    io::{self, AsyncWriteExt},
+    io::{self, AsyncReadExt, AsyncSeekExt, AsyncWriteExt},
 };
 use tracing::*;
 
-use crate::layered_repository::metadata::METADATA_FILE_NAME;
-
 use super::{parse_ids_from_path, strip_path_prefix, RemoteFileInfo, RemoteStorage};
+use crate::layered_repository::metadata::METADATA_FILE_NAME;
 
 pub struct LocalFs {
     pageserver_workdir: &'static Path,
@@ -145,6 +144,64 @@ impl RemoteStorage for LocalFs {
             io::copy(&mut source, to)
                 .await
                 .context("Failed to download a file from the local storage")?;
+            Ok(())
+        } else {
+            bail!(
+                "File '{}' either does not exist or is not a file",
+                file_path.display()
+            )
+        }
+    }
+
+    async fn download_range(
+        &self,
+        from: &Self::StoragePath,
+        start_inclusive: u64,
+        end_exclusive: Option<u64>,
+        to: &mut (impl io::AsyncWrite + Unpin + Send + Sync),
+    ) -> anyhow::Result<()> {
+        if let Some(end_exclusive) = end_exclusive {
+            ensure!(
+                end_exclusive > start_inclusive,
+                "Invalid range, start ({}) is bigger then end ({:?})",
+                start_inclusive,
+                end_exclusive
+            );
+            if start_inclusive == end_exclusive.saturating_sub(1) {
+                return Ok(());
+            }
+        }
+        let file_path = self.resolve_in_storage(from)?;
+
+        if file_path.exists() && file_path.is_file() {
+            let mut source = io::BufReader::new(
+                fs::OpenOptions::new()
+                    .read(true)
+                    .open(&file_path)
+                    .await
+                    .with_context(|| {
+                        format!(
+                            "Failed to open source file '{}' to use in the download",
+                            file_path.display()
+                        )
+                    })?,
+            );
+            source
+                .seek(io::SeekFrom::Start(start_inclusive))
+                .await
+                .context("Failed to seek to the range start in a local storage file")?;
+            match end_exclusive {
+                Some(end_exclusive) => {
+                    io::copy(&mut source.take(end_exclusive - start_inclusive), to).await
+                }
+                None => io::copy(&mut source, to).await,
+            }
+            .with_context(|| {
+                format!(
+                    "Failed to download file '{}' range from the local storage",
+                    file_path.display()
+                )
+            })?;
             Ok(())
         } else {
             bail!(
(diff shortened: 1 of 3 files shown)


pageserver/src/remote_storage/local_fs.rs
```diff
@@ -154,6 +153,58 @@ impl RemoteStorage for LocalFs {
...
+                            file_path.display()
+                        )
+                    })?,
+            );
+            source
+                .seek(io::SeekFrom::Start(start_inclusive))
+                .await
+                .context("Failed to seek to the range start in a local storage file")?;
+            match end_inclusive {
+                Some(end) => read_n_bytes(end + 1, &mut source.take(end + 1), to).await,
```
patins, line 192: Can you `io::copy` `source.take()`?

pageserver/src/remote_storage/rust_s3.rs
```diff
@@ -159,6 +159,31 @@ impl RemoteStorage for S3 {
...
+            .with_context(|| format!("Failed to download s3 object with key {}", from.key()))?;
+        if code != 200 {
+            Err(anyhow::format_err!(
+                "Received non-200 exit code during downloading object from directory, code: {}",
+                code
+            ))
+        } else {
+            io::copy(&mut io::BufReader::new(data.as_slice()), to)
+                .await
+                .context("Failed to write downloaded data into the destination buffer")?;
```
patins, line 182: Is streaming not allowed with get_object_range? Is that a limitation of our s3 crate?

If so, should we do `to.write_all`

pageserver/src/remote_storage/local_fs.rs
```diff
@@ -154,6 +153,58 @@ impl RemoteStorage for LocalFs {
...
+    ) -> anyhow::Result<()> {
+        if Some(start_inclusive) == end_inclusive {
+            return Ok(());
+        }
+        ensure!(
+            start_inclusive < end_inclusive.unwrap_or(u64::MAX),
+            "Invalid range, start ({}) is bigger then end ({:?})",
+            start_inclusive,
+            end_inclusive
+        );
```
patins, line 171: Would it make more sense to put both of these checks inside if let Some(end_inclusive) = end_inclusive { ... }

pageserver/src/remote_storage/rust_s3.rs
SomeoneToIgnore: Alas, `rust_s3` returns `Vec<u8>` for both [`get_object`](https://github.com/durch/rust-s3/blob/c00a18693e53da096dd6e59d6e7880f2add6039f/s3/src/bucket.rs#L526) and [`get_object_range`](https://github.com/durch/rust-s3/blob/c00a18693e53da096dd6e59d6e7880f2add6039f/s3/src/bucket.rs#L608) 

I've used the `io::copy` version, keeping `rusoto` in mind, which has a streaming HTTP body returned instead, I'll place a comment about that and won't change the code itself, but let me know if you prefer writing vectors properly 🙂

pageserver/src/remote_storage/local_fs.rs
```diff
@@ -154,6 +153,64 @@ impl RemoteStorage for LocalFs {
...
+        if let Some(end_exclusive) = end_exclusive {
+            ensure!(
+                end_exclusive > start_inclusive,
+                "Invalid range, start ({}) is bigger then end ({:?})",
+                start_inclusive,
+                end_exclusive
+            );
+            if start_inclusive == end_exclusive - 1 {
+                return Ok(());
+            }
```
patins, line 172: Would be nice to avoid panic if end_exclusive == 0

pageserver/src/remote_storage/rust_s3.rs
```diff
@@ -148,17 +148,49 @@ impl RemoteStorage for S3 {
...
             Err(anyhow::format_err!(
-                "Received non-200 exit code during downloading object from directory, code: {}",
+                "Received non-200 exit code during downloading object, code: {}",
                 code
             ))
         } else {
+            // we don't have to write vector into the destination this way, `to_write_all` would be enough.
+            // but we want to prepare for migration on `rusoto`, that has a streaming HTTP body instead here, with
+            // which it makes more sense to use `io::copy`.
             io::copy(&mut io::BufReader::new(data.as_slice()), to)
```
patins, line 158: I get wanting to use a streaming API, but should we still use BufReader? Can the slice alone be the input to io::copy

pageserver/src/remote_storage/rust_s3.rs
```diff
@@ -148,17 +148,49 @@ impl RemoteStorage for S3 {
...
+    async fn download_range(
+        &self,
+        from: &Self::StoragePath,
+        start_inclusive: u64,
+        end_exclusive: Option<u64>,
+        to: &mut (impl io::AsyncWrite + Unpin + Send + Sync),
+    ) -> anyhow::Result<()> {
+        // S3 accepts ranges as https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.35
+        // and needs both ends to be exclusive
+        let end_inclusive = end_exclusive.map(|end| end - 1);
```
patins, line 174: Same panic note here

pageserver/src/remote_storage/rust_s3.rs
```diff
@@ -148,17 +148,49 @@ impl RemoteStorage for S3 {
...
+            .await
+            .with_context(|| format!("Failed to download s3 object with key {}", from.key()))?;
+        if code != 206 {
+            Err(anyhow::format_err!(
+                "Received non-206 exit code during downloading object range, code: {}",
+                code
+            ))
+        } else {
+            // see `download` function above for the comment on why `Vec<u8>` buffer is copied this way
+            io::copy(&mut io::BufReader::new(data.as_slice()), to)
```
patins, line 187: Same BufReader note here

patins (approved):

pageserver/src/remote_storage/rust_s3.rs
SomeoneToIgnore: The culprit here is `io::copy` signature:

> `pub async fn copy<'a, R, W>(reader: &'a mut R, writer: &'a mut W) -> io::Result<u64>`

`AsyncRead` is implemented for `&[u8]` only, not its mutable counterpart, so I have to wrap it.

I will wrap it in `std::io::Cursor` then, no need to use `BufReader`, but not sure what else besides those two could be used as `&mut impl AsyncRead`. 
Maybe there's a better method that just accepts non-mutable `AsyncRead` somewhere?

pageserver/src/remote_storage/local_fs.rs
SomeoneToIgnore: Good catch, thanks, fixed.

pageserver/src/remote_storage/rust_s3.rs
patins: What about
`io::copy(&mut data.as_slice(), ...)`

SomeoneToIgnore: Oh my, indeed. Thank you 😅

PR #849, opened 2021-11-05
Feature/tag docker images

cicdteam:
- docker image `zenithdb/zenith` tagged by `latest` and tag that counts commits in repository (`git log --oneline|wc -l`, for example `zenithdb/zenith:1030`)
- added build flow for `zenithdb/compute-node` with the same tags

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -299,7 +299,28 @@ jobs:
           name: Build and push Docker image
           command: |
             echo $DOCKER_PWD | docker login -u $DOCKER_LOGIN --password-stdin
-            docker build --build-arg GIT_VERSION=$CIRCLE_SHA1 -t zenithdb/zenith:latest . && docker push zenithdb/zenith:latest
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -299,7 +299,28 @@ jobs:
...
+          docker_layer_caching: true
+      - run:
+          name: Init postgres submodule
+          command: git submodule update --init --depth 1
+      - run:
+          name: Build and push Docker image
+          command: |
+            echo $DOCKER_PWD | docker login -u $DOCKER_LOGIN --password-stdin
+            DOCKER_TAG=$(git log --oneline|wc -l)
+            docker build -t zenithdb/compute-node:latest vendor/postgres && docker push zenithdb/compute-node:latest
```
petuhovskiy, line 322: I'm not sure if we're using it at all, but `latest` tag may interfere with CI build in postgres https://github.com/zenithdb/postgres/blob/main/.circleci/config.yml

Maybe it's better not to push `latest` tag here or there.

ololobus: Yeah, and it does interfere now. We should remove CI pipeline with Docker image from `postgres` once we move all images building jobs to `zenith`

kelvich (approved):

PR #853, opened 2021-11-08
Disconnect walreceiver when no new WAL arrives

lubennikovaav:
to turn tenant into Idle state.

Compute-node <--> WAL service connection changes:

Cancel SendWalHandler that has stopped receiving WAL after idle_session_timeout.
Now this is hardcoded as 30 seconds. Maybe worth turning into config parameter.

WAL service <--> Pageserver connection changes:

Change request_callback behavior: only request callback when receiver has some new WAL to stream,
shutdown request_callback thread together at ReceiveWalConn exit

If pageserver is caught up, disconnect ReplicationConn from WAL service to pageserver after some timeout.
Note that caught up here means that pageserver received the latest WAL. We don't have to wait for it to be durably stored, because it is already stored in WAL service and can be retreived later from safekeepers in case of pageserver restart.

Pageserver changes:
If walreceiver exited because of timeout, don't try to reconnect. Rely on WAL service request_callback to initiate this connection again, when new WAL arrives.

### DIFF
--- a/walkeeper/src/receive_wal.rs
+++ b/walkeeper/src/receive_wal.rs
@@ -8,6 +8,7 @@ use log::*;
 use postgres::{Client, Config, NoTls};
 
 use std::net::SocketAddr;
+use std::sync::mpsc::{channel, Receiver, Sender, TryRecvError};
 use std::thread;
 use std::thread::sleep;
 
@@ -15,9 +16,11 @@ use crate::safekeeper::AcceptorProposerMessage;
 use crate::safekeeper::ProposerAcceptorMessage;
 
 use crate::send_wal::SendWalHandler;
-use crate::timeline::TimelineTools;
+use crate::timeline::{Timeline, TimelineTools};
 use crate::SafeKeeperConf;
+use std::sync::Arc;
 use zenith_utils::connstring::connection_host_port;
+use zenith_utils::lsn::Lsn;
 use zenith_utils::postgres_backend::PostgresBackend;
 use zenith_utils::pq_proto::{BeMessage, FeMessage};
 use zenith_utils::zid::{ZTenantId, ZTimelineId};
@@ -27,13 +30,20 @@ pub struct ReceiveWalConn<'pg> {
     pg_backend: &'pg mut PostgresBackend,
     /// The cached result of `pg_backend.socket().peer_addr()` (roughly)
     peer_addr: SocketAddr,
+    callmemeybe_sender: Option<Sender<()>>,
 }
 
 ///
 /// Periodically request pageserver to call back.
 /// If pageserver already has replication channel, it will just ignore this request
 ///
-fn request_callback(conf: SafeKeeperConf, timelineid: ZTimelineId, tenantid: ZTenantId) {
+fn request_callback(
+    conf: SafeKeeperConf,
+    rx: Receiver<()>,
+    timeline: Arc<Timeline>,
+    timelineid: ZTimelineId,
+    tenantid: ZTenantId,
+) {
     let ps_addr = conf.pageserver_addr.unwrap();
     let ps_connstr = format!(
         "postgresql://no_user:{}@{}/no_db",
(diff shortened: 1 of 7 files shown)


arssher:
So there are multiple things added here.

1. On safekeeper, stop callmemaybe thread when wal receiver which started it dies. It is ok, it fixes current unpleasant bug, I vote with three hands to merge it ASAP.
2. If nothing comes from compute to safekeeper, after 30s thread receiving WAL on safekeeper goes down. Note though (not sure it is intended) as currently coded it also influences thread sending WAL to pageserver, as it lives under postgres_backend as well -- if no feedback comes from pageserver, it will die too.
3. If pageserver is caughtup with safekeeper, safekeeper at some point will terminate the connection between them.
4. Callmemaybe thread on safekeeper: If there is no more committed WAL /since the last time it was checked/, don't request to connect.
5. Reaction to callmemaybe on pageserver: previously pageserver reconnected to the requested safekeeper infinitely, now only 10 times.

I'm not sure whether we really want 2, as we already have similar timeout logic on the higher lever: console turnes idle compute off after some timeout. Purely read only nodes are not covered by this, but how much we want to kill connections for them? And note that dying compute <-> safekeeper connection will terminate safekeeper <-> pageserver one (which is good), but it means if some chunk of WAL got to safekeeper but not to pageserver, 2 will intervene in pushing it further -- though not too successfully as walproposer will constantly reconnect anyway. So proper implementation would be more complicated.

Same for 3+4. Moreover, if we have 2 properly implemented, we get it for free: if nothing happens, again dying compute <-> safekeeper connection terminates  safekeeper <-> pageserver one. There are also implementation issues, see inline.


As for 5, it is nice to get rid of infinite reconnections, but why 10? I imagined myself it like this: one callmemaybe requests to connect 1 time. If something goes wrong, callmemaybe repeats. OTOH, if safekeeper has live pageserver <-> safekeeper connection, it stops doing callmemaybe to avoid spamming logs and network. Currently 2 of 3 safekeepers will still poll, but eventually we should have some gossip between them deciding who is the most advanced (and in the quorum) and should server WAL.

walkeeper/src/replication.rs
```diff
@@ -113,6 +118,16 @@ impl ReplicationConn {
                             let reply = StandbyReply::des(&m[1..])
                                 .context("failed to deserialize StandbyReply")?;
                             state.disk_consistent_lsn = reply.write_lsn;
+                            // Meaning of `flush_lsn` and `write_lsn`
+                            // is far from obvious, but it is what walreceiver returns.
```
arssher, line 122: oh yes

walkeeper/src/replication.rs
```diff
@@ -202,10 +220,30 @@ impl ReplicationConn {
 
         let mut end_pos = Lsn(0);
         let mut wal_file: Option<File> = None;
+        let mut retry_count = 0;
 
         loop {
+            if let Ok(lsn) = rx.try_recv() {
```
arssher, line 226: It is slightly dangerous to use channel here: currently pageserver only sends feedback on each message, but it is more efficient to do that only once in a while, like in pg walreceiver. And in that case channel might overflow which will block feedback thread which (eventually) will block pageserver walreceiver. Putting it atomic/under mutex would do.

walkeeper/src/receive_wal.rs
```diff
@@ -50,18 +60,29 @@ fn request_callback(conf: SafeKeeperConf, timelineid: ZTimelineId, tenantid: ZTe
...
-                if let Err(e) = client.simple_query(&callme) {
-                    error!("Failed to send callme request to pageserver: {}", e);
+        if let Err(TryRecvError::Disconnected) = rx.try_recv() {
+            trace!("Walstream disconnected. Exiting request_callback thread");
+            break;
+        }
+
+        // Request callback only at start or
+        // if we have some new WAL to stream
+        if let Some(lsn) = timeline.wait_for_lsn(start_pos) {
```
arssher, line 72: That's broken: even if safekeeper doesn't have new WAL since the last check, it doesn't mean pageserver is caught up.

arssher (changes requested):

walkeeper/src/replication.rs
lubennikovaav: Btw, I fixed it in the first commit of #866 : [cf275011](https://github.com/zenithdb/zenith/pull/866/commits/cf275011dc7e52763b4a4bc8be04f92d13aeab2d)

walkeeper/src/receive_wal.rs
```diff
@@ -78,6 +99,7 @@ impl<'pg> ReceiveWalConn<'pg> {
         Ok(ReceiveWalConn {
             pg_backend: pg,
             peer_addr,
+            callmemeybe_sender: None,
```
kelvich, line 102: nit: meybe -> maybe

PR #859, opened 2021-11-09
In wal_proposer_recovery, don't wait outcoming WAL to be committed.

arssher:
Otherwise we're deadlocking ourselves. Oversight of 33007cc0bb.

### DIFF
--- a/walkeeper/src/replication.rs
+++ b/walkeeper/src/replication.rs
@@ -194,8 +194,22 @@ impl ReplicationConn {
                 break;
             }
         }
-        let (_, timeline) = swh.timeline.get().get_end_of_wal();
-        info!("Start replication from {}", start_pos);
+        let (wal_end, timeline) = swh.timeline.get().get_end_of_wal();
+        // Walproposer gets special handling: safekeeper must give proposer all
+        // local WAL till the end, whether committed or not (walproposer will
+        // hang otherwise). That's because walproposer runs the consensus and
+        // synchronizes safekeepers on the most advanced one.
+        //
+        // There is a small risk of this WAL getting concurrently garbaged if
+        // another compute rises which collects majority and starts fixing log
+        // on this safekeeper itself. That's ok as (old) proposer will never be
+        // able to commit such WAL.
+        let stop_pos: Option<Lsn> = if swh.appname == Some("wal_proposer_recovery".to_string()) {
+            Some(wal_end)
+        } else {
+            None
+        };
+        info!("Start replication from {:?} till {:?}", start_pos, stop_pos);
 
         // switch to copy
         pgb.write_message(&BeMessage::CopyBothResponse)?;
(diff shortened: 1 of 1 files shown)


walkeeper/src/replication.rs
```diff
@@ -259,7 +274,7 @@ impl ReplicationConn {
 
             start_pos += send_size as u64;
 
-            debug!("sent WAL up to {}", end_pos);
+            debug!("sent WAL up to {}", start_pos);
```
yeputons, line 277: Is this change intended?

walkeeper/src/replication.rs
```diff
@@ -194,8 +194,14 @@ impl ReplicationConn {
...
         }
-        let (_, timeline) = swh.timeline.get().get_end_of_wal();
-        info!("Start replication from {}", start_pos);
+        let (wal_end, timeline) = swh.timeline.get().get_end_of_wal();
+        let stop_pos: Lsn;
+        if swh.appname == Some("wal_proposer_recovery".to_string()) {
+            stop_pos = wal_end;
+        } else {
+            stop_pos = Lsn(0);
+        }
```
yeputons, line 203: `let stop_pos: Option<Lsn> = if ... { Some(wal_end) } else { None }`?

walkeeper/src/replication.rs
```diff
@@ -204,18 +210,27 @@ impl ReplicationConn {
         let mut wal_file: Option<File> = None;
 
         loop {
-            /* Wait until we have some data to stream */
-            if let Some(lsn) = swh.timeline.get().wait_for_lsn(start_pos) {
-                end_pos = lsn
+            /* For recovery walproposer needs all WAL safekeeper has whether committed or not */
+            if stop_pos != Lsn(0) {
+                if start_pos >= stop_pos {
+                    /* recovery finished */
```
yeputons, line 216: Nit: I think these two comments belong above, where we actually initialize the `stop_pos`.

walkeeper/src/replication.rs
```diff
@@ -194,8 +194,14 @@ impl ReplicationConn {
                 break;
             }
         }
-        let (_, timeline) = swh.timeline.get().get_end_of_wal();
-        info!("Start replication from {}", start_pos);
+        let (wal_end, timeline) = swh.timeline.get().get_end_of_wal();
+        let stop_pos: Lsn;
+        if swh.appname == Some("wal_proposer_recovery".to_string()) {
```
yeputons, line 199: I'd elaborate here, e.g.:

```
// wal_proposer_recovery gets special handling because it is:
// a) part of the consensus protocol,
// b) that has to synchronize Safekeepers with each other,
// c) and needs all flushed WAL from all Safekeepers to do so even if it's not committed.
// So, we serve it all existing WAL (even uncommitted) and exit immediately.
// We cannot continue replication because WAL may be overwritten after proposer recovers and overwrites uncommitted history.
// If we receive new WAL during this replication, that WAL should originate with another proposer node. Hence, TODO TODO TODO
```

Is this accurate? I'm actually not sure why it's ok to set `stop_pos = wal_end` here and not care about sending updated WAL. Say, what if we had no uncommitted changes, but then they suddently appear from another proposer node... Not sure.

petuhovskiy, line 199: I'd add comment explaining why this `if` is necessary to prevent deadlock, it will help to not miss changes in the next review.

yeputons:
I'm also concerned that this was not caught by existing tests, but that's a separate issue.

yeputons (approved):
Anyway, it seems to be an accurate partial rollback of #777 **except for the `debug!` line**.

Maybe a better fix is needed, but I think it's ok to heal tests for now, leave some `TODO: do not remove this line`, create an issue, and carefully deal with it afterwards.

walkeeper/src/replication.rs
arssher: Yes. Done that in sep commit now.

petuhovskiy (approved):
Logic seems ok, but more comments explaining what is happening in `replication.rs` would be great.

walkeeper/src/replication.rs
arssher: I've added something along these lines. Minor correction: 'We have to exit otherwise' -- no, walproposer actually knows when it got enough WAL and will end streaming on its own. We just need to have some border till which safekeeper has WAL available, and stop_pos serves this purpose instead of `wait_for_lsn` used normally.

arssher:
> I'm also concerned that this was not caught by existing tests, but that's a separate issue.

It was, actually. Mine yesterday commit in main failed in CI; Dmitry Rodionov reported "I've run test_restarts_under_load around ten times and it passed only once.". However, it is true current tests don't catch it with 100% probability. I believe it is impossible to do so in integration tests as we need to fire up recovery in short moment when `flush_lsn` was advanced on some safekeeper but `commit_lsn` hasn't. It can be done in manually crafted schedule though, like `test_sync_safekeepers`; it would be a good idea for another PR.

arssher:
test_restarts_under_load failed on one of the builds, but I don't see anything suspicious apart from very slow apply by catching-up safekeeper; seems like this is  https://github.com/zenithdb/zenith/issues/780 , so I'm merging.

PR #867, opened 2021-11-11
Treat remote storage sync tasks more fair

SomeoneToIgnore:
Addresses https://github.com/zenithdb/zenith/pull/686/files#r732327737 and prepares for handling archives in remote storage by 
* using a single atomic (no locks and unwraps) dequeue for all tasks, no "urgent" tasks or anyhow differently ordered tasks
* track and evict tasks that fail too much
* process sync tasks in bulk, instead of using semaphores per file sync

### DIFF
--- a/pageserver/src/layered_repository.rs
+++ b/pageserver/src/layered_repository.rs
@@ -34,7 +34,7 @@ use std::time::{Duration, Instant};
 use self::metadata::{metadata_path, TimelineMetadata};
 use crate::page_cache;
 use crate::relish::*;
-use crate::remote_storage::schedule_timeline_upload;
+use crate::remote_storage::schedule_timeline_checkpoint_upload;
 use crate::repository::{GcResult, Repository, Timeline, TimelineWriter, WALRecord};
 use crate::tenant_mgr;
 use crate::walreceiver;
@@ -259,16 +259,37 @@ impl Repository for LayeredRepository {
 
         let timelines = self.timelines.lock().unwrap();
         for (timelineid, timeline) in timelines.iter() {
-            walreceiver::stop_wal_receiver(*timelineid);
-            // Wait for syncing data to disk
-            trace!("repo shutdown. checkpoint timeline {}", timelineid);
-            timeline.checkpoint(CheckpointConfig::Forced)?;
-
-            //TODO Wait for walredo process to shutdown too
+            shutdown_timeline(*timelineid, timeline.as_ref())?;
         }
 
         Ok(())
     }
+
+    fn unload_timeline(&self, timeline_id: ZTimelineId) -> Result<()> {
+        let mut timelines = self.timelines.lock().unwrap();
+        let removed_timeline = match timelines.remove(&timeline_id) {
+            Some(timeline) => timeline,
+            None => {
+                warn!("Timeline {} not found, nothing to remove", timeline_id);
+                return Ok(());
+            }
+        };
+        drop(timelines);
+        shutdown_timeline(timeline_id, removed_timeline.as_ref())?;
+
+        Ok(())
+    }
+}
+
+fn shutdown_timeline(
+    timelineid: ZTimelineId,
+    timeline: &LayeredTimeline,
+) -> Result<(), anyhow::Error> {
+    walreceiver::stop_wal_receiver(timelineid);
+    trace!("repo shutdown. checkpoint timeline {}", timelineid);
+    timeline.checkpoint(CheckpointConfig::Forced)?;
+    //TODO Wait for walredo process to shutdown too
+    Ok(())
 }
 
 /// Private functions
@@ -318,7 +339,12 @@ impl LayeredRepository {
                     .load_layer_map(disk_consistent_lsn)
                     .context("failed to load layermap")?;
                 if self.upload_relishes {
-                    schedule_timeline_upload(self.tenantid, timelineid, loaded_layers, metadata);
+                    schedule_timeline_checkpoint_upload(
+                        self.tenantid,
+                        timelineid,
+                        loaded_layers,
+                        metadata,
+                    );
                 }
 
                 // needs to be after load_layer_map
@@ -1332,7 +1358,12 @@ impl LayeredTimeline {
                 false,
             )?;
             if self.upload_relishes {
-                schedule_timeline_upload(self.tenantid, self.timelineid, layer_uploads, metadata);
+                schedule_timeline_checkpoint_upload(
+                    self.tenantid,
+                    self.timelineid,
+                    layer_uploads,
+                    metadata,
+                );
             }
 
             // Also update the in-memory copy
--- a/pageserver/src/remote_storage/storage_sync.rs
+++ b/pageserver/src/remote_storage/storage_sync.rs
@@ -13,7 +13,7 @@
 //! pageserver writes them.
 //!
 //! The list construction is currently the only place where the storage sync can return an [`Err`] to the user.
-//! New upload tasks are accepted via [`schedule_timeline_upload`] function disregarding of the corresponding loop startup,
+//! New upload tasks are accepted via [`schedule_timeline_checkpoint_upload`] function disregarding of the corresponding loop startup,
 //! it's up to the caller to avoid uploading of the new file, if that caller did not enable the loop.
 //! After the initial state is loaded into memory and the loop starts, any further [`Err`] results do not stop the loop, but rather
 //! reschedules the same task, with possibly less files to sync in it.
@@ -59,12 +59,12 @@
 //! This is a subject to change in the near future, but requires more changes to [`crate::tenant_mgr`] before it can happen.
 
 use std::{
-    cmp::Ordering,
-    collections::{hash_map, BinaryHeap, HashMap, HashSet},
+    collections::{hash_map, HashMap, HashSet},
+    num::{NonZeroU32, NonZeroUsize},
+    ops::DerefMut,
     path::{Path, PathBuf},
-    sync::Mutex,
+    sync::Arc,
     thread,
-    time::Duration,
 };
 
 use anyhow::{anyhow, ensure, Context};
@@ -73,18 +73,21 @@ use lazy_static::lazy_static;
 use tokio::{
     fs,
     io::{self, AsyncWriteExt},
-    sync::Semaphore,
+    sync::{
+        mpsc::{self, UnboundedReceiver},
+        Mutex,
+    },
     time::Instant,
 };
 use tracing::*;
 
-use super::RemoteStorage;
+use super::{RemoteStorage, TimelineSyncId};
 use crate::{
     layered_repository::{
         metadata::{metadata_path, TimelineMetadata, METADATA_FILE_NAME},
         TENANTS_SEGMENT_NAME, TIMELINES_SEGMENT_NAME,
     },
-    tenant_mgr::register_timeline_download,
+    tenant_mgr::{perform_post_timeline_sync_steps, PostTimelineSyncStep},
     PageServerConf,
 };
 use zenith_metrics::{register_histogram_vec, register_int_gauge, HistogramVec, IntGauge};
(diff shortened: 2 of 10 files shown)


pageserver/src/remote_storage/storage_sync.rs
```diff
@@ -240,82 +376,22 @@ fn storage_sync_loop<P: std::fmt::Debug, S: 'static + RemoteStorage<StoragePath
...
+        .map(SyncKind::Download)
+        .for_each(|sync_kind| SYNC_QUEUE.push(SyncTask::new(0, sync_kind)));
+}
 
-    while !crate::tenant_mgr::shutdown_requested() {
-        let mut queue_accessor = SYNC_QUEUE.lock().unwrap();
-        let next_task = queue_accessor.pop();
-        let remaining_queue_length = queue_accessor.len();
-        drop(queue_accessor);
+fn get_next_tasks(max_concurrent_sync: NonZeroUsize) -> Vec<SyncTask> {
```
SomeoneToIgnore, line 383: This is its own can of worms.

First, allowing default (100) concurrent upload/download tasks for uncompressed files feels wrong. That hopefully gets fixed soon in the archiving PR, so not super critical.

Then, in general, the tasks may overlap (there's nothing in the code that ensures they don't) and cause unneeded races.

For now, it's fine to pick tasks in batch due to incremental nature of all checkpoints and the fact that downloads are either for entire timeline or for nothing: neither uploads and downloads intersect now.

I think it's fine to leave it as is for now, later there will be more fun with archive defragmentation, where we could think on batching more, possibly add some heuristics to merge the tasks.

But for now, when thinking of remote files as archives, it seems wasteful to not to batch multiple tasks.

pageserver/src/remote_storage/storage_sync.rs
```diff
@@ -210,19 +231,134 @@ pub(super) fn spawn_storage_sync_thread<
...
+) {
+    if task.retries > max_sync_errors.get() {
+        error!(
+            "Evicting task {:?} that failed {} times, exceeding the error theshold",
+            task.kind, task.retries
+        );
+        let tenant_id = task.tenant_id();
+        let timeline_id = task.timeline_id();
+        if let Err(e) =
+            get_repository_for_tenant(tenant_id).and_then(|repo| repo.unload_timeline(timeline_id))
```
SomeoneToIgnore, line 314: Another question is how adequate it is to unload a timeline, if we fail to download its updates or upload some of its checkpoints.

Both failures may occur unpredictably and always occur in the background, so a very sudden timeline disappear might happen due to this.

pageserver/src/layered_repository.rs
```diff
@@ -230,16 +230,37 @@ impl Repository for LayeredRepository {
...
+}
+
+fn shutdown_timeline(
+    timelineid: ZTimelineId,
+    timeline: &LayeredTimeline,
+) -> Result<(), anyhow::Error> {
+    walreceiver::stop_wal_receiver(timelineid);
+    trace!("repo shutdown. checkpoint timeline {}", timelineid);
+    timeline.checkpoint(CheckpointConfig::Forced)?;
+    //TODO Wait for walredo process to shutdown too
```
patins, line 262: Shouldn't this hook into the existing `tenant_mgr` shutdown logic

pageserver/src/remote_storage/storage_sync.rs
patins: Will unload_timeline be blocking if it waits for it to unload

pageserver/src/remote_storage/storage_sync.rs
```diff
@@ -210,19 +231,134 @@ pub(super) fn spawn_storage_sync_thread<
...
+    remote_storage: Arc<S>,
+    task: SyncTask,
+    max_sync_errors: NonZeroU32,
+    remote_timelines: Arc<Mutex<HashMap<(ZTenantId, ZTimelineId), RemoteTimeline>>>,
+) {
+    if task.retries > max_sync_errors.get() {
+        error!(
+            "Evicting task {:?} that failed {} times, exceeding the error theshold",
+            task.kind, task.retries
+        );
```
patins, line 310: Does unloading the entire timeline make sense if a task for it fails?

pageserver/src/remote_storage/storage_sync.rs
```diff
@@ -210,19 +231,134 @@ pub(super) fn spawn_storage_sync_thread<
...
+        {
+            error!(
+                "Failed to remove repository for tenant {}, timeline {}: {:#}",
+                tenant_id, timeline_id, e
+            )
+        }
+        return;
+    }
+
+    let seconds_to_wait = f64::EPSILON.powf(task.retries.saturating_sub(1) as f64) - 1.0;
```
patins, line 324: Why `EPSILON`? Also I think it might be good to guard the wait behind an if for the first iteration

pageserver/src/remote_storage/storage_sync.rs
```diff
@@ -210,19 +231,134 @@ pub(super) fn spawn_storage_sync_thread<
...
+    let remote_storage = Arc::new(remote_storage);
+
+    while !crate::tenant_mgr::shutdown_requested() {
+        let next_tasks = get_next_tasks(max_concurrent_sync);
+        let remaining_queue_length = SYNC_QUEUE.len();
+
+        if next_tasks.is_empty() {
+            trace!("No storage sync tasks found");
+            thread::sleep(Duration::from_secs(1));
+            continue;
```
patins, line 262: I think `std::sync::mpsc` might make more sense then `SegQueue`, as it has blocking behavior

pageserver/src/remote_storage/storage_sync.rs
```diff
@@ -210,19 +231,134 @@ pub(super) fn spawn_storage_sync_thread<
...
-    let mut remote_timelines = runtime
+    let remote_timelines = runtime
         .block_on(fetch_existing_uploads(&remote_storage))
         .context("Failed to determine previously uploaded timelines")?;
 
-    let urgent_downloads = latest_timelines(&remote_timelines)
+    schedule_first_tasks(config, &remote_timelines);
+
+    let remote_timelines = Arc::new(Mutex::new(remote_timelines));
+    let remote_storage = Arc::new(remote_storage);
```
patins, line 253: Does it make sense to combine these Arcs?

pageserver/src/remote_storage/storage_sync.rs
```diff
@@ -240,82 +376,22 @@ fn storage_sync_loop<P: std::fmt::Debug, S: 'static + RemoteStorage<StoragePath
...
 
-    while !crate::tenant_mgr::shutdown_requested() {
-        let mut queue_accessor = SYNC_QUEUE.lock().unwrap();
-        let next_task = queue_accessor.pop();
-        let remaining_queue_length = queue_accessor.len();
-        drop(queue_accessor);
+fn get_next_tasks(max_concurrent_sync: NonZeroUsize) -> Vec<SyncTask> {
+    let mut max_concurrent_sync = max_concurrent_sync.get();
+    let mut tasks = Vec::with_capacity(max_concurrent_sync);
+    while let Some(new_task) = SYNC_QUEUE.pop() {
```
patins, line 386: Too bad none of these queues have batch reads

pageserver/src/remote_storage/storage_sync.rs
```diff
@@ -110,37 +112,59 @@ lazy_static! {
...
+            SyncKind::Upload(upload_kind) => upload_kind.tenant_id,
+        }
+    }
+
+    fn timeline_id(&self) -> ZTimelineId {
+        match &self.kind {
+            SyncKind::Download(download_kind) => download_kind.timeline_id,
+            SyncKind::Upload(upload_kind) => upload_kind.timeline_id,
+        }
+    }
```
patins, line 143: Should we just pull these into SyncTask

pageserver/src/remote_storage/storage_sync.rs
SomeoneToIgnore: I don't think it will impact much the other tasks, since we wrap every `process_task` invocation into `tokio::spawn` that has to mitigate such effects, but sure the batch processing per se may be affected or even get blocked.

We could do that in a separate thread explicitly, not sure how feasible to do.

pageserver/src/layered_repository.rs
SomeoneToIgnore: It's actually vice versa: the `tenant_mgt` calls this `shutdown_all_tenants` that gets here eventually, so not sure this is a good idea to change right now.
Originally it came with https://github.com/zenithdb/zenith/pull/671, I've moved it around and did not dare to change.

pageserver/src/remote_storage/storage_sync.rs
SomeoneToIgnore: Yes, same question here, not sure: it's lame to unload the data from the network outage indeed.

But what to do with tasks that fail for real then? We cannot allow such to accumulate, but presumably we cannot just skip one particular task and allow syncing the same timeline further later: if it goes for the upload task, missing a bunch of checkpoint files and allowing uploading another checkpoint later could render the whole timeline corrupt if I get it right.

So, we have to prohibit sync at least. That's doable, every `LayeredTimeline` carries the `uupload_relishes` flag inside that can be changed. But how to indicate that this timeline is not ok? This could be a big bad surprise too, while unloaded timeline is at least explicit.

pageserver/src/remote_storage/storage_sync.rs
SomeoneToIgnore: How could one implement batching with such model?
It appears beneficial to have one, since allows us to reorder or merge some tasks, but there's no way to know without blocking/awaiting (I think it's better to be the tokio version of the channel in the async code, but not important) if there are any more tasks left in the channel's buffer?

pageserver/src/remote_storage/storage_sync.rs
SomeoneToIgnore: Yes, https://docs.rs/crossbeam-deque/0.8.1/crossbeam_deque/ is intended to be for that, but it's a bit unusual in api-wise.

pageserver/src/remote_storage/storage_sync.rs
SomeoneToIgnore: Actually, wanted to use `E`, to use exponential back-off, thanks for spotting.
I did not think that if is necessary due to having `powf - 1` always >= 0, but sure, I can add one.

pageserver/src/remote_storage/storage_sync.rs
SomeoneToIgnore: That's quite a change actually, plenty of things are based on that.
I have a similar idea to introduce a `pub struct TimelineSyncId(ZTenantId, ZTimelineId);`, will try to bring it into with your suggestion.

SomeoneToIgnore: Done.

pageserver/src/remote_storage/storage_sync.rs
patins: `recv` for the first task, then `try_recv` for the following tasks.

pageserver/src/remote_storage/storage_sync.rs
patins: I think base 2 is pretty common. You might also want to cap the maximum backoff to something reasonable (e^6 = 403 s) seems excessive. Also some recommend using a random number of milliseconds on top of this, not sure how much that matters for this use case

pageserver/src/remote_storage/storage_sync.rs
patins: I'm worried that unload_timeline will `join()` on other threads which may take a while. This could block the storage sync executor right?

pageserver/src/layered_repository.rs
patins: Ahh I see, my bad

pageserver/src/remote_storage/storage_sync.rs
SomeoneToIgnore: Yes, partially, if we use a multithreaded tokio runtime, but that won't much of a solace.

I think, we could batch such tasks and them register them all at once, which can be done in the end, this blocking in the end of the loop. Not superb, but better than spawning a thread on the loop?

patins: Could we also use `spawn_blocking`

SomeoneToIgnore: We could, but the batch register seemed simple to implement, that's what I've ended up with 🙂 
While we can use `spawn_blocking` on it still, makes not much sense IMO: the batch register is in the end of the loop iteration and we'll have to wait at the `spawn_blocking` join handle at some point: making things more async than they are might be cumbersome.

pageserver/src/remote_storage/storage_sync.rs
SomeoneToIgnore: Ok, had come up with the changes to reflect that.
There's a possibility to drop any of the ends to stop the whole machinery, but I have not found a good way to plug that into our loop concept. In theory, I could place the whole mpsc under the lock, but that beats the purpose of the whole idea IMO.
It might be better, if we move from static functions to some wrapper for the uploaded, as it was in the very first version, but that's a separate step, I think.

pageserver/src/remote_storage/storage_sync.rs
```diff
@@ -239,73 +326,100 @@ fn storage_sync_loop<
...
-                    }
-                })
-                .collect::<FuturesUnordered<_>>();
-
-            let mut downloads_to_sync = Vec::with_capacity(task_batch.len());
-            runtime.block_on(async {
-                while let Some(sync_id) = task_batch.next().await {
-                    if let Some(sync_id) = sync_id {
-                        downloads_to_sync.push(sync_id);
+async fn loop_step<
```
patins, line 349: What's the advantage of using async mpsc over sync mpsc?

pageserver/src/remote_storage/storage_sync.rs
```diff
@@ -115,7 +119,83 @@ lazy_static! {
...
 lazy_static! {
-    static ref SYNC_QUEUE: SegQueue<SyncTask> = SegQueue::new();
+    static ref SYNC_QUEUE: SyncQueue = SyncQueue::new();
+}
+
+struct SyncQueue {
+    sender: UnboundedSender<SyncTask>,
+    receiver: Mutex<UnboundedReceiver<SyncTask>>,
+    length: AtomicUsize,
+}
```
patins: Consider using once_cell for the sender, and owning the receiver in the sync thread. Then you don't need a Mutex

pageserver/src/remote_storage/storage_sync.rs
patins: I think my concern is that we're calling into the sync world of LayeredRepo from async functions. unload_timeline blocks until the timeline is removed from memory and all associated tasks are shut down, which seems like it could take a while. Is my understanding of this wrong?

SomeoneToIgnore: No, you're absolutely right here, it's all done sync and could take a while, despite some attempts to do things in batch now.

There's no `async` world around the `register_timeline_downloads` though, that's why I think it's "fine": it's done outside of `runtime.block_on` calls.
I still agree that's not ideal, since might still take a while, even though being outside async machinery, not blocking it directly.

I don't have ready plans to tackle such "a while" things besides profiling and rewriting the registration code?
We could move off the registration into another thread/thread-wrapped-in-future and loop on, as I've mentioned.
But that would require managing two queues now, not sure if that's a nice or easy way to go with.
No good ideas on the topic from me so far.

pageserver/src/remote_storage/storage_sync.rs
SomeoneToIgnore: I wanted to avoid bloating the diff more, hence kept the `SYNC_QUEUE` api unchanged.
Both `Sender` and `Receiver` in `std::sync::mpsc` are `!Sync` (which is needed by the lasy init wrappers), so require their own wrappers to interact with, that would cause many changes. 
Async `sender` is fine and does not block either, so more convenient.

Also, with `std::sync::mpsc` it's a really blocking code in the async code this time, which IMO is more dangerous compared to the async counterpart (even though we call that blocking code before the main async loop call, which is not that bad with blocking calls too).

Otherwise, both sync and async versions are not that different: cancellable on drop of the pair and have bounded and unbounded flavors, so not that.
I can use the sync version, if you think it's worth it still, but not sure myself why it's beneficial.

pageserver/src/remote_storage/storage_sync.rs
patins: I'm thinking about the call to `unload_timeline` done inside the async fn `process_task` which is executed by the tokio::spawn call in `loop_step`.

pageserver/src/remote_storage/storage_sync.rs
patins: Interesting re `Sync`. Thanks for that insight.

I believe the earlier version had `get_next_tasks` being called from a sync context (inside `storage_sync_loop`), that's where I was thinking about using sync::mpsc.

pageserver/src/remote_storage/storage_sync.rs
SomeoneToIgnore: I like the idea, thanks. Had implemented it by moving away all queue-related things into a separate module.

pageserver/src/remote_storage/storage_sync.rs
SomeoneToIgnore: Now get it, thanks, very good catch.

Yes, that's an issue, I've moved it away, similarly to what I've done to downloads.
It gets even better now, since `tenant_mgr` has a generic method to register any future sync operations, if needed.

pageserver/src/remote_storage/storage_sync.rs
SomeoneToIgnore: So far, let's keep it this way, we should load test this all eventually, could test some failure scenarios along the way too.

PR #869, opened 2021-11-11
Use less pageserver-specific method in RemoteStorage trait

SomeoneToIgnore:
Presumably last PR, stripped off the archiving impl in the forthcoming PR.

Instead of requiring to have a method that returns a very pageserver-specific `RemoteFileInfo`, instead now return a "local path" as the main part of that info. The rest of the info can be derived in-place and dose not have to stuck in the public storage API.

### DIFF
--- a/pageserver/src/remote_storage/local_fs.rs
+++ b/pageserver/src/remote_storage/local_fs.rs
@@ -5,7 +5,6 @@
 //! volume is mounted to the local FS.
 
 use std::{
-    ffi::OsStr,
     future::Future,
     path::{Path, PathBuf},
     pin::Pin,
@@ -18,8 +17,7 @@ use tokio::{
 };
 use tracing::*;
 
-use super::{parse_ids_from_path, strip_path_prefix, RemoteFileInfo, RemoteStorage};
-use crate::layered_repository::metadata::METADATA_FILE_NAME;
+use super::{strip_path_prefix, RemoteStorage};
 
 pub struct LocalFs {
     pageserver_workdir: &'static Path,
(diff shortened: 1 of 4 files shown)


pageserver/src/remote_storage/local_fs.rs
```diff
@@ -141,9 +134,12 @@ impl RemoteStorage for LocalFs {
...
             );
-            io::copy(&mut source, to)
-                .await
-                .context("Failed to download a file from the local storage")?;
+            io::copy(&mut source, to).await.with_context(|| {
+                format!(
+                    "Failed to download file '{}' from the local storage",
+                    file_path.display()
+                )
+            })?;
```
patins, line 142: Do we also need to flush here

pageserver/src/remote_storage/storage_sync.rs
```diff
@@ -365,19 +368,41 @@ async fn fetch_existing_uploads<P: std::fmt::Debug, S: 'static + RemoteStorage<S
...
+                Ok(path) => path,
+                Err(e) => return Err((e, remote_path)),
+            };
+            let metadata = if local_path
+                .file_name()
+                .and_then(|os_str| os_str.to_str())
+                .unwrap_or_default()
+                == METADATA_FILE_NAME
+            {
+                let mut metadata_bytes = std::io::Cursor::new(Vec::new());
```
patins, line 381: I think you don't need cursor here

pageserver/src/remote_storage/storage_sync.rs
```diff
@@ -365,19 +368,41 @@ async fn fetch_existing_uploads<P: std::fmt::Debug, S: 'static + RemoteStorage<S
     let mut data_fetches = uploaded_files
         .into_iter()
         .map(|remote_path| async {
-            (
-                remote_file_info(remote_storage, &remote_path).await,
-                remote_path,
-            )
+            let local_path = match remote_storage.local_path(&remote_path) {
```
patins, line 371: It'd be cool to reduce the duplication of map_err in this function. Maybe use a helper function and `?`.

SomeoneToIgnore: Agree, but let's omit that in this PR: the code for fetching remote files is different with archives, so does not make too much sense to make this version readable.

patins (approved):

PR #881, opened 2021-11-15
[proxy] Pass PostgreSQL version and other improvements

funbringer:
I could've created 3 different PRs, but this would've been much more cumbersome.

### DIFF
--- a/proxy/src/cplane_api.rs
+++ b/proxy/src/cplane_api.rs
@@ -1,10 +1,6 @@
-use anyhow::{bail, Context, Result};
+use anyhow::{anyhow, bail, Context};
 use serde::{Deserialize, Serialize};
 use std::net::{SocketAddr, ToSocketAddrs};
-
-pub struct CPlaneApi {
-    auth_endpoint: &'static str,
-}
 
 #[derive(Serialize, Deserialize, Debug)]
 pub struct DatabaseInfo {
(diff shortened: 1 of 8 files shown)


proxy/src/cplane_api.rs
```diff
@@ -16,20 +14,21 @@ pub struct DatabaseInfo {
...
 #[derive(Serialize, Deserialize, Debug)]
-pub struct ProxyAuthResult {
-    pub ready: bool,
-    pub error: Option<String>,
-    pub conn_info: Option<DatabaseInfo>,
+#[serde(untagged)]
+enum ProxyAuthResponse {
+    Ready { conn_info: DatabaseInfo },
+    Error { error: String },
+    NotReady { ready: bool }, // TODO: get rid of `ready`
```
ololobus, line 21: Just curios, am I correct that `serde` tries to match enum variants sequentially and discards all unmatched fields?

Why I ask is that `{ ready: true, conn_info: ... }` should match both `Ready` and `NotReady`, IIUC from the docs. And we protect ourselves from matching it with `NotReady` by the order of arms in the `match` bellow

kelvich (approved):

proxy/src/cplane_api.rs
funbringer: Correct, that's how [untagged](https://serde.rs/enum-representations.html#untagged) mode works. What I'd love to see instead is the proper internally/externally tagged json, but I don't want to change too many things across two repos at the moment. Feel free to introduce such a change after we've merged this.

PR #887, opened 2021-11-17
callmemaybe refactoring

lubennikovaav:
- Don't spawn a separate thread for each connection.
Instead use one thread per safekeeper, that iterates over all connections and sends callback requests for them.

-Use tokio postgres to connect to the pageserver, to avoid spawning a new thread for each connection.

### DIFF
--- a/walkeeper/Cargo.toml
+++ b/walkeeper/Cargo.toml
@@ -32,6 +32,7 @@ signal-hook = "0.3.10"
 serde = { version = "1.0", features = ["derive"] }
 hex = "0.4.3"
 const_format = "0.2.21"
+tokio-postgres = { git = "https://github.com/zenithdb/rust-postgres.git", rev="9eb0dbfbeb6a6c1b79099b9f7ae4a8c021877858" }
 
 # FIXME: 'pageserver' is needed for ZTimelineId. Refactor
 pageserver = { path = "../pageserver" }
--- a/walkeeper/src/callmemaybe.rs
+++ b/walkeeper/src/callmemaybe.rs
@@ -200,9 +200,6 @@ impl Drop for SubscriptionStateInner {
 }
 
 pub async fn main_loop(conf: SafeKeeperConf, mut rx: Receiver<CallmeEvent>) -> Result<()> {
-    let default_timeout = Duration::from_secs(5);
-    let recall_period = conf.recall_period.unwrap_or(default_timeout);
-
     let subscriptions: Mutex<HashMap<(ZTenantId, ZTimelineId), SubscriptionState>> =
         Mutex::new(HashMap::new());
 
@@ -259,7 +256,7 @@ pub async fn main_loop(conf: SafeKeeperConf, mut rx: Receiver<CallmeEvent>) -> R
                     },
                 }
             },
-            _ = tokio::time::sleep(recall_period) => { true },
+            _ = tokio::time::sleep(conf.recall_period) => { true },
         };
 
         if call_iteration {
@@ -268,7 +265,7 @@ pub async fn main_loop(conf: SafeKeeperConf, mut rx: Receiver<CallmeEvent>) -> R
             for (&(_tenantid, _timelineid), state) in subscriptions.iter() {
                 let listen_pg_addr = conf.listen_pg_addr.clone();
 
-                state.call(recall_period, listen_pg_addr);
+                state.call(conf.recall_period, listen_pg_addr);
             }
         }
     }
--- a/walkeeper/src/callmemaybe.rs
+++ b/walkeeper/src/callmemaybe.rs
@@ -1,18 +1,22 @@
 //!
-//!  Callmemaybe thread is responsible for periodically requesting
+//!  Callmemaybe module is responsible for periodically requesting
 //!  pageserver to initiate wal streaming.
 //!
+//!  Other threads can use CallmeEvent messages to subscribe or unsubscribe
+//!  from the call list.
+//!
 use crate::SafeKeeperConf;
+use anyhow::anyhow;
 use anyhow::Result;
-use log::*;
 use std::collections::HashMap;
-use std::net::ToSocketAddrs;
 use std::sync::Mutex;
 use std::time::{Duration, Instant};
 use tokio::runtime;
-use tokio::sync::mpsc::Receiver;
+use tokio::sync::mpsc::UnboundedReceiver;
 use tokio::task;
 use tokio_postgres::NoTls;
+use tracing::*;
+use zenith_utils::connstring::connection_host_port;
 use zenith_utils::zid::{ZTenantId, ZTimelineId};
 
 async fn request_callback(
@@ -33,21 +37,14 @@ async fn request_callback(
         }
     });
 
-    // Send callmemaybe request
-    let listen_pg_addr = listen_pg_addr_str
-        .to_socket_addrs()
-        .unwrap()
-        .next()
-        .unwrap();
+    // use Config parsing because SockAddr parsing doesnt allow to use host names instead of ip addresses
+    let me_connstr = format!("postgresql://no_user@{}/no_db", listen_pg_addr_str);
+    let me_conf: postgres::config::Config = me_connstr.parse().unwrap();
+    let (host, port) = connection_host_port(&me_conf);
 
     let callme = format!(
         "callmemaybe {} {} host={} port={} options='-c ztimelineid={} ztenantid={}'",
-        tenantid,
-        timelineid,
-        &listen_pg_addr.ip().to_string(),
-        listen_pg_addr.port(),
-        timelineid,
-        tenantid
+        tenantid, timelineid, host, port, timelineid, tenantid
     );
 
     let _ = client.simple_query(&callme).await?;
@@ -55,7 +52,7 @@ async fn request_callback(
     Ok(())
 }
 
-pub fn thread_main(conf: SafeKeeperConf, rx: Receiver<CallmeEvent>) -> Result<()> {
+pub fn thread_main(conf: SafeKeeperConf, rx: UnboundedReceiver<CallmeEvent>) -> Result<()> {
     let runtime = runtime::Builder::new_current_thread()
         .enable_all()
         .build()
@@ -64,6 +61,7 @@ pub fn thread_main(conf: SafeKeeperConf, rx: Receiver<CallmeEvent>) -> Result<()
     runtime.block_on(main_loop(conf, rx))
 }
 
+/// Messages to the callmemaybe thread
 #[derive(Debug)]
 pub enum CallmeEvent {
     // add new subscription to the list
@@ -77,17 +75,13 @@ pub enum CallmeEvent {
     Resume(ZTenantId, ZTimelineId),
 }
 
-struct SubscriptionStateInner {
-    handle: Option<task::JoinHandle<()>>,
-    last_call_time: Instant,
-    paused: bool,
-}
-
 struct SubscriptionState {
     tenantid: ZTenantId,
     timelineid: ZTimelineId,
     pageserver_connstr: String,
-    inner: Mutex<SubscriptionStateInner>,
+    handle: Option<task::JoinHandle<()>>,
+    last_call_time: Instant,
+    paused: bool,
 }
 
 impl SubscriptionState {
@@ -96,63 +90,29 @@ impl SubscriptionState {
         timelineid: ZTimelineId,
         pageserver_connstr: String,
     ) -> SubscriptionState {
-        let state_inner = SubscriptionStateInner {
-            handle: None,
-            last_call_time: Instant::now(),
-            paused: false,
-        };
-
         SubscriptionState {
             tenantid,
             timelineid,
             pageserver_connstr,
-            inner: Mutex::new(state_inner),
+            handle: None,
+            last_call_time: Instant::now(),
+            paused: false,
         }
     }
 
-    fn pause(&self) {
-        let mut state_inner = self.inner.lock().unwrap();
-        state_inner.paused = true;
-
-        if let Some(handle) = state_inner.handle.take() {
-            handle.abort();
-        }
+    fn pause(&mut self) {
+        self.paused = true;
+        self.abort_handle();
     }
 
-    fn resume(&self) {
-        let mut state_inner = self.inner.lock().unwrap();
-        state_inner.paused = false;
+    fn resume(&mut self) {
+        self.paused = false;
     }
 
-    fn call(&self, recall_period: Duration, listen_pg_addr: String) {
-        let mut state_inner = self.inner.lock().unwrap();
-
-        // Ignore call request if this subscription is paused
-        if state_inner.paused {
-            debug!(
-                "ignore call request for paused subscription
-                tenantid: {}, timelineid: {}",
-                self.tenantid, self.timelineid
-            );
-            return;
-        }
-
-        // Check if it too early to recall
-        if state_inner.handle.is_some() && state_inner.last_call_time.elapsed() < recall_period {
-            debug!(
-                "too early to recall. state_inner.last_call_time.elapsed: {:?}, recall_period: {:?}
-                tenantid: {}, timelineid: {}",
-                state_inner.last_call_time, recall_period, self.tenantid, self.timelineid
-            );
-            return;
-        }
-
-        // If previous task didn't complete in recall_period, it must be hanging,
-        // so don't wait for it forever, just abort it and try again.
-        //
-        // Most likely, the task have already successfully completed
-        // and abort() won't have any effect.
-        if let Some(handle) = state_inner.handle.take() {
+    // Most likely, the task have already successfully completed
+    // and abort() won't have any effect.
+    fn abort_handle(&mut self) {
+        if let Some(handle) = self.handle.take() {
             handle.abort();
 
             let timelineid = self.timelineid;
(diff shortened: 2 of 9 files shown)


kelvich:
Mpsc changed to crossbeam due to
https://doc.rust-lang.org/nightly/std/sync/mpsc/struct.Receiver.html#known-issues

funbringer:
Thanks for the patch! I'll look into it tomorrow.

walkeeper/src/receive_wal.rs
```diff
@@ -127,19 +88,27 @@ impl<'pg> ReceiveWalConn<'pg> {
...
-            let _ = thread::Builder::new()
-                .name("request_callback thread".into())
-                .spawn(move || {
-                    request_callback(conf, timelineid, tenant_id);
-                })
-                .unwrap();
-        }
+        // as long as this wal_stream is alive, callmemaybe thread
+        // will send requests to pageserver
+        let _wal_stream = match swh.conf.pageserver_addr {
```
LizardWizzard, line 93: If I understand it correctly here we use reference count as a criteria of liveness. It might be good to add more explanations to the comment.

Also note that the same can be achieved using Guard pattern. See ReplicationConnGuard as an example.

Am I right that we will send callmemaybe requests forever until walproposer disconnects? Is it correct?

walkeeper/src/callmemaybe.rs
```diff
@@ -0,0 +1,131 @@
...
+        match rx.recv_timeout(timeout) {
+            Ok(streaminfo) => {
+                info!("callmemaybe. thread_main. request_callback for timelineid={} tenantid={} refcount {}",
+                streaminfo.timelineid, streaminfo.tenantid, Arc::strong_count(&streaminfo));
+
+                let pageserver_addr = conf.pageserver_addr.as_ref().unwrap().clone();
+                let listen_pg_addr = conf.listen_pg_addr.clone();
+                let password = conf.pageserver_auth_token.clone();
+
+                runtime.block_on(async {
```
LizardWizzard, line 76: It seems that this is equivalent to using sync version. Do we need async here? 

If we use async there it becomes possible to run multiple callmemaybe calls concurrently. 

In current implementation the following is possible: imagine that we have two callmemaybe calls in the queue and the first one timeouts or is stuck for any other reason, in that case second call must wait for the first one. This behaviour might be undesirable. Solution to this is to use runtime.spawn, or something like that ie LocalSet or FuturesUnordered. But also it makes possible to issue multiple concurrent calls for the same pageserver/tenant/timeline. It might be desirable to avoid that

walkeeper/src/callmemaybe.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    Ok(())
+}
+
+pub fn thread_main(conf: SafeKeeperConf, rx: Receiver<Arc<WalStreamInfo>>) -> Result<()> {
+    let runtime = tokio::runtime::Builder::new_current_thread()
+        .enable_all()
+        .build()?;
+
+    let default_timeout = Duration::from_secs(5);
```
LizardWizzard, line 62: Might be better to move to a top level constant

walkeeper/Cargo.toml
```diff
@@ -31,6 +31,8 @@ walkdir = "2"
 serde = { version = "1.0", features = ["derive"] }
 hex = "0.4.3"
 const_format = "0.2.21"
+tokio-postgres = "0.7.5"
```
LizardWizzard, line 34: tokio-postgres lives in the same repo as rust-postgres which we've already forked. Is it hard to point it to our forked version for less confusion in the future?

petuhovskiy (approved):
Looks good, should help with callmemaybe spamming that we can have now.

I'm not very familiar with tokio and may have missed some points. One thing that caught my eye is `current_thread::Runtime` with `block_on`. It seems that `request_callback` calls block current thread and we can have some problems with callmemaybe requests blocking each other, especially in case of timeouts and similar problems.

walkeeper/src/callmemaybe.rs
```diff
@@ -0,0 +1,131 @@
...
+    let timeout = conf.recall_period.unwrap_or(default_timeout);
+    let mut addresses = Vec::new();
+
+    loop {
+        match rx.recv_timeout(timeout) {
```
funbringer, line 67: Here's an idea: with an async channel (e.g. [`tokio::sync::mpsc`](https://docs.rs/tokio/1.14.0/tokio/sync/mpsc/index.html)) it's possible to perform wait-for-timeout and wait-for-subscription concurrently! A simplified sketch:

```rust
enum Event {
  CallMeMaybe(...),
  Timeout,
}

// Whoa, async magic!
let event = tokio::select! {
    x = rx.recv() => CallMeMaybe(x),
    _ = tokio::time::sleep(...) => Timeout,
};

match event {
  ...
}
```

Furthermore, each "callmemaybe" subscription could be represented by a struct containing not only the request string but also a "last successful call timestamp", allowing us to unconditionally loop through a list of subscriptions and call their `send_request_if_time_has_passed()` (not a real name ;) methods.

TLDR: I'd rewrite the whole `thread_main` as an async function.

---

* https://docs.rs/tokio/1.14.0/tokio/macro.select.html (see notes on cancellation safety)

walkeeper/src/callmemaybe.rs
```diff
@@ -0,0 +1,131 @@
...
+            Ok(streaminfo) => {
+                info!("callmemaybe. thread_main. request_callback for timelineid={} tenantid={} refcount {}",
+                streaminfo.timelineid, streaminfo.tenantid, Arc::strong_count(&streaminfo));
```
funbringer, line 70: @LizardWizzard has proposed a brilliant idea: we could implement "unsubscribe" as a guard, i.e. a struct with a destructor. The destructor could send an "unsubscribe" event containing `TentantId` and whatnots via a [`Sender`](https://docs.rs/tokio/1.14.0/tokio/sync/mpsc/struct.Sender.html).

It's also worth mentioning that tokio's mpsc supports [blocking send](https://docs.rs/tokio/1.14.0/tokio/sync/mpsc/struct.Sender.html#method.blocking_send).

funbringer (commented):
Thanks for the patch!

funbringer (changes requested):
Just using the scary red banner to highlight the fact that there are some concerns :)

walkeeper/src/callmemaybe.rs
```diff
@@ -0,0 +1,131 @@
...
+    timelineid: ZTimelineId,
+    tenantid: ZTenantId,
+) -> Result<()> {
+    // Connect to the pageserver.
+    let ps_addr_split: Vec<&str> = ps_addr.split(':').collect();
+    let mut config = tokio_postgres::Config::new();
+
+    config
+        .host(ps_addr_split[0])
+        .port(ps_addr_split[1].parse().unwrap());
```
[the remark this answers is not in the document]
LizardWizzard: Maybe SocketAddr from std comes in handy

walkeeper/src/callmemaybe.rs
```diff
@@ -0,0 +1,131 @@
...
+
+async fn request_callback(
+    ps_addr: String,
+    listen_pg_addr: String,
+    pageserver_auth_token: Option<String>,
+    timelineid: ZTimelineId,
+    tenantid: ZTenantId,
+) -> Result<()> {
+    // Connect to the pageserver.
+    let ps_addr_split: Vec<&str> = ps_addr.split(':').collect();
```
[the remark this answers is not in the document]
lubennikovaav: rewrote to use SocketAddr.
BTW, is there any reason we don't use it for all `*_addr` in SafeKeeperConf instead of String?

walkeeper/src/receive_wal.rs
lubennikovaav: Thank you for the idea. Rewrote to use the guard.

walkeeper/src/callmemaybe.rs
lubennikovaav: I'd rather get rid of this entirely and introduce default `recall_period` because I don't see a use case for safekeepers with `recall` disabled.
@petuhovskiy , @arssher , what do you think?

walkeeper/src/callmemaybe.rs
lubennikovaav: Thank you for the idea with `select!`. I rewrote the main to be async.

As for "last successful call timestamp" I am not so sure.  Do we need to be that accurate? I don't see any harm in sending a few extra calls.

walkeeper/src/callmemaybe.rs
arssher: Makes sense.

walkeeper/src/callmemaybe.rs
```diff
@@ -0,0 +1,131 @@
...
+                            streaminfo.timelineid, streaminfo.tenantid, e
+                        );
+                    });
+                });
+
+                if conf.recall_period.is_some() {
+                    addresses.push(streaminfo.clone());
+                }
+            }
+            Err(RecvTimeoutError::Timeout) => {
```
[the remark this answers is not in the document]
lubennikovaav: Indeed.
I rewrote it so that every incoming subscription is just added to the list, and then we call all addresses in the list.

walkeeper/src/callmemaybe.rs
```diff
@@ -0,0 +1,155 @@
...
+    Ok(())
+}
+
+pub fn thread_main(conf: SafeKeeperConf, rx: Receiver<CallmeEvent>) -> Result<()> {
+    let runtime = runtime::Builder::new_current_thread()
+        .enable_all()
+        .build()
+        .unwrap();
+
+    runtime.block_on(async { main_loop(conf, rx).await })
```
funbringer, line 75: I bet just `main_loop(conf, rx)` would do it.

walkeeper/src/callmemaybe.rs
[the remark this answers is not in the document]
LizardWizzard: I think nothing keeps us from doing that. I remember that in pageserver there are two config structures and one of them is used for toml serialization, so it may be necessary to add some custom handlers to parse socket addr from string. But worth checking. For safekeeper there is nothing similar because there is no persistent config, so we can just do it :)

walkeeper/src/receive_wal.rs
```diff
@@ -128,19 +83,27 @@ impl<'pg> ReceiveWalConn<'pg> {
...
+        // as long as this wal_stream is alive, callmemaybe thread
+        // will send requests to pageserver
+        let _guard = match swh.conf.pageserver_addr {
+            Some(_) => {
+                // Need to establish replication channel with page server.
+                // Add far as replication in postgres is initiated by receiver, we should use callme mechanism
+                let timelineid = swh.timeline.get().timelineid;
+                let tx_clone = swh.tx.clone();
+                swh.tx
+                    .blocking_send(CallmeEvent::Subscribe(tenant_id, timelineid))
```
LizardWizzard, line 95: You can use unbounded channel and then no blocking is needed

walkeeper/src/callmemaybe.rs
```diff
@@ -0,0 +1,293 @@
+//!
```
LizardWizzard, line 1: Maybe it is a good idea to have top level docs mentioning what messages are sent when so the whole picture is covered

LizardWizzard (commented):
Thanks! This is quite nice :) I've left a few suggestions here and there

walkeeper/src/callmemaybe.rs
```diff
@@ -0,0 +1,293 @@
...
+    let default_timeout = Duration::from_secs(5);
+    let recall_period = conf.recall_period.unwrap_or(default_timeout);
+
+    let subscriptions: Mutex<HashMap<(ZTenantId, ZTimelineId), SubscriptionState>> =
+        Mutex::new(HashMap::new());
+
+    loop {
+        let call_iteration = tokio::select! {
+            request = rx.recv() => {
+                match request {
```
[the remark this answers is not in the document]
funbringer: > Hm, thinking about how to reduce indentiation level.
 
+1 to decreasing indent. Frankly, I'd use question mark here, i.e. `rx.recv().ok_or_else(|| anyhow!("done!"))?` and handle the error somewhere near `thread_main`.

walkeeper/src/replication.rs
```diff
@@ -221,6 +248,27 @@ impl ReplicationConn {
         };
         info!("Start replication from {:?} till {:?}", start_pos, stop_pos);
 
+        // Don't spam pageserver with callmemaybe queries
+        // when connection is already established.
```
arssher, line 252: slightly better to do that only when connection is indeed from pageserver, i.e. appname is not wal_proposer_recovery, see several lines above

walkeeper/src/callmemaybe.rs
```diff
@@ -0,0 +1,293 @@
...
+            request_callback(
+                pageserver_addr,
+                listen_pg_addr,
+                password,
+                timelineid,
+                tenantid,
+            )
+            .await
+            .unwrap_or_else(|e| {
+                error!(
```
arssher, line 193: Seems like this error is never logged, better do that.

arssher (approved):

walkeeper/src/callmemaybe.rs
```diff
@@ -0,0 +1,293 @@
...
+struct SubscriptionStateInner {
+    handle: Option<task::JoinHandle<()>>,
+    last_call_time: Instant,
+    paused: bool,
+}
+
+struct SubscriptionState {
+    tenantid: ZTenantId,
+    timelineid: ZTimelineId,
+    inner: Mutex<SubscriptionStateInner>,
```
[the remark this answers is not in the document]
arssher: current_thread runtime is used here (and we don't need more than that), so there is nothing to block/being blocked by. It would be nice to drop the mutex altogether actually.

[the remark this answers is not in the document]
LizardWizzard: If we hold the mutex across await points it might still make sense in a single threaded runtime. To actually spawn non send futures see [LocalSet](https://docs.rs/tokio/1.14.0/tokio/task/struct.LocalSet.html). Mutex might be replaced with Rc<RefCell>. But still beware of holding RefMut across await points, compiler should issue a warning in that case.

walkeeper/src/receive_wal.rs
lubennikovaav: Why would we need it if we send the message from a sync code anyway?

walkeeper/src/callmemaybe.rs
```diff
@@ -0,0 +1,293 @@
...
+async fn request_callback(
+    ps_addr_str: String,
+    listen_pg_addr_str: String,
+    pageserver_auth_token: Option<String>,
+    timelineid: ZTimelineId,
+    tenantid: ZTenantId,
+) -> Result<()> {
+    // Connect to the pageserver.
+    let mut config = tokio_postgres::Config::new();
+    let ps_addr = ps_addr_str.to_socket_addrs().unwrap().next().unwrap();
```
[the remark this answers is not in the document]
lubennikovaav: this code is gone after rebase

walkeeper/src/replication.rs
lubennikovaav: Fixed. Do we need any similar handling in `ReceiveWalConn::run()`?

walkeeper/src/callmemaybe.rs
```diff
@@ -0,0 +1,293 @@
...
+                            CallmeEvent::Unsubscribe(tenantid, timelineid) => {
+                                let mut subscriptions = subscriptions.lock().unwrap();
+                                subscriptions.remove(&(tenantid, timelineid));
+                                debug!("callmemaybe. thread_main. unsubscribe callback request for timelineid={} tenantid={}",
+                                timelineid, tenantid);
+                                false
+                            },
+                            CallmeEvent::Pause(tenantid, timelineid) => {
+                                let subscriptions = subscriptions.lock().unwrap();
+                                if let Some(sub) = subscriptions.get(&(tenantid, timelineid))
```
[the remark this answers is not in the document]
lubennikovaav: it turned out that clippy doesn't like this style
https://rust-lang.github.io/rust-clippy/master/index.html#option_map_unit_fn

walkeeper/src/replication.rs
arssher: No, it can only receive WAL from the proposer.

walkeeper/src/callmemaybe.rs
arssher: uh, I was wrong and missed `error!` and `bail!` -- `error!` in the previous version was ok, but in the current code result of `request_callback` is unwrapped,  it will panic on failure.

walkeeper/src/callmemaybe.rs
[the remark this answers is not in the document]
LizardWizzard: That’s strange, and example doesn’t make sense to me. @funbringer what do you think about disabling this lint globally?

walkeeper/src/receive_wal.rs
```diff
@@ -136,20 +90,33 @@ impl<'pg> ReceiveWalConn<'pg> {
...
-                .name("request_callback thread".into())
-                .spawn(move || {
-                    request_callback(conf, pageserver_connstr, timelineid, tenant_id);
+        // if requested, ask pageserver to fetch wal from us
+        // as long as this wal_stream is alive, callmemaybe thread
+        // will send requests to pageserver
+        let _guard = match self.pageserver_connstr {
+            Some(ref pageserver_connstr) => {
+                // Need to establish replication channel with page server.
+                // Add far as replication in postgres is initiated by receiver, we should use callme mechanism
```
LizardWizzard, line 99: typo

walkeeper/src/receive_wal.rs
```diff
@@ -136,20 +90,33 @@ impl<'pg> ReceiveWalConn<'pg> {
...
+                let timelineid = swh.timeline.get().timelineid;
+                let tx_clone = swh.tx.clone();
+                let pageserver_connstr = pageserver_connstr.to_owned();
+                swh.tx
+                    .send(CallmeEvent::Subscribe(
+                        tenant_id,
+                        timelineid,
+                        pageserver_connstr,
+                    ))
+                    .unwrap();
```
LizardWizzard, line 109: It can fail when channel was closed or receiver was dropped, thats unlikely but lets add some context via expect or turn this to an error

walkeeper/src/replication.rs
```diff
@@ -221,6 +248,28 @@ impl ReplicationConn {
         };
         info!("Start replication from {:?} till {:?}", start_pos, stop_pos);
 
+        // Don't spam pageserver with callmemaybe queries
+        // when connection to compute node is already established.
```
LizardWizzard, line 252: Comment is a bit misleading to me, we suspend callmemaybe polling when pageserver connects to us (safekeeper) and the polling will be continued if/when replication connection was destroyed

LizardWizzard (approved):
LGTM

Left few comments on minor leftovers

walkeeper/src/callmemaybe.rs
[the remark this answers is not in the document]
funbringer: Personally, I agree with clippy (I wouldn't go as far as to impose my vision though). `map` better suits functional style, but returning unit `()` kind of implies that the statement is effectful (i.e. has side effects, or why else would you return unit?), and mixing both functional and effectful styles might lead to bugs. TLDR `if let` makes side effects stand out more.

PR #927, opened 2021-11-18
Relax back check in test_readonly_node

LizardWizzard:
It was changed in https://github.com/zenithdb/zenith/pull/877 and this
lead to unstable test behaviour

### DIFF
--- a/test_runner/batch_others/test_readonly_node.py
+++ b/test_runner/batch_others/test_readonly_node.py
@@ -86,6 +86,10 @@ def test_readonly_node(zenith_simple_env: ZenithEnv):
     assert cur.fetchone() == (1, )
 
     # Create node at pre-initdb lsn
-    with pytest.raises(Exception, match='extracting base backup failed'):
+    # This can fail in two different ways:
+    # 1. With "extracting base backup failed"
+    # 2. With "directory <repo dir> is not a database cluster directory"
+    # TODO have explicit check so the failure has constant reason
+    with pytest.raises(Exception):
         # compute node startup with invalid LSN should fail
         env.zenith_cli(["pg", "start", "test_branch_preinitdb", "test_readonly_node@0/42"])


test_runner/batch_others/test_readonly_node.py
```diff
@@ -86,6 +86,10 @@ def test_readonly_node(zenith_simple_env: ZenithEnv):
     assert cur.fetchone() == (1, )
 
     # Create node at pre-initdb lsn
-    with pytest.raises(Exception, match='extracting base backup failed'):
+    # This can fail in two different ways:
+    # 1. With "extracting base backup failed"
+    # 2. With "directory <repo dir> is not a database cluster directory"
```
lubennikovaav, line 91: This error means that we somehow got the base backup at invalid lsn. How is it possible?

LizardWizzard: For example see https://app.circleci.com/pipelines/github/zenithdb/zenith/3311/workflows/956355ea-e846-455d-bdb7-2a47e0222c85/jobs/27532/tests#failed-test-0
And https://app.circleci.com/pipelines/github/zenithdb/zenith/3325/workflows/ce77a108-4d2e-47b2-b9b0-43db42deac70/jobs/27688/tests#failed-test-0

Why this should be impossible? I do not see where we check that basebackup lsn must be greater than e.g. initdb lsn. But I think we should have this check, for all, readonly nodes, branching and basebackup. I can add directly that. Also there is new latest_gc_cutoff lsn, maybe we should check against it too.

lubennikovaav: Ah, I see. We add some stuff to basebackup even if there is no data in the repository at this Lsn.

there is already a meaningful error in pageserver.log
`ERROR page server thread exited with error err=segment checkpoint not found at 0/42`

It should be enough to check the result of send_tarball() and return `ErrorResponse` if something went wrong.

LizardWizzard: What do you think about explicit lsn checking? Error checking after send_tarball happens when data is already sent to the client (if I understand it correctly), so this is a bit confusing. IMO there is no reason in trying to send something if there is nothing valid to send.

lubennikovaav: It will work too. I don't have a strong opinion on this exact fix.

My point is that we must check `send_tarball()` return value anyway. Without it, the client cannot check if sanely looking pgdata lacks some files because of basebackup error.

note that `initdblsn` may change between postgreSQL major versions
+1 to `latest_gc_cutoff` check

LizardWizzard: > My point is that we must check send_tarball() return value anyway

Currently it uses "?" operator, which should result in ErrorResponse, so I think it is already in the right shape

> note that initdblsn may change between postgreSQL major versions 

I think it is not too much overhead to store latest lsn after initdb and check against that

LizardWizzard: Also the error you've mentioned comes from `get_page_at_lsn`, so it happens after basebackup. I was confused for some time trying to understand how it is possible to simultaneously succeed with basebackup and have a failure in it :)

lubennikovaav: > Also the error you've mentioned comes from `get_page_at_lsn`, so it happens after basebackup. 

This `get_page_at_lsn()` is called from `add_pgcontrol_file()` during basebackup.

Try command `./target/debug/zenith pg start test_branch_preinitdb main@0/42`
that's what I see in log (with extra debug messages):
```
2021-11-19T12:20:47.422491Z  INFO basebackup{timeline=67ee21803471cb242b4dfec6e2f85c8a tenant=fa38a19e93fe43cab04368a8915e276e}: taking basebackup lsn=0/42, prev_lsn=0/0    
2021-11-19T12:20:47.423395Z  INFO basebackup{timeline=67ee21803471cb242b4dfec6e2f85c8a tenant=fa38a19e93fe43cab04368a8915e276e lsn="0/42"}: list_nonrels called at 0/42
2021-11-19T12:20:47.423568Z  INFO basebackup{timeline=67ee21803471cb242b4dfec6e2f85c8a tenant=fa38a19e93fe43cab04368a8915e276e lsn="0/42"}: generating pg_control at lsn 0/42    
2021-11-19T12:20:47.423767Z ERROR page server thread exited with error err=segment checkpoint not found at 0/42
```
Unlike correct basebackup, It doesn't get to `all tarred up!` and `CopyDone sent!` lines.

>I was confused for some time trying to understand how it is possible to simultaneously succeed with basebackup and have a failure in it :)

The fact that the client considers this basebackup successful is exactly the problem

LizardWizzard: > This get_page_at_lsn() is called from add_pgcontrol_file() during basebackup.

Oh thanks, I've missed it

> The fact that the client considers this basebackup successful is exactly the problem

I digged in a bit and I found that for `FeMessage::Query` PgBackend handles errors and sends `ErrorResponse` back to the client. But this call to basebackup was performed via `FeMessage::Execute` which lacks error handling and thread just dies in case of an error. I was assuming that error should be handled for basebackup too, and this led to confusion. 
With this fixed I still manage to get both errors. This is strange to me, because it seems that tar unpack can somehow swallow `ErrorResponse`.

Maybe I'm missing something?

Nevertheless I plan to add both fixes, with lsn check and proper error reporting

LizardWizzard:
Reworked in #933 with all the needed fixes/cleanups

Issue #935, opened 2021-11-23
Possible replication failure on staging


LizardWizzard:
There is a problem with one of the clusters (325). @kelvich suggested that replication between safekeeper and pageserver might've stuck.

Observable behavior is a timeout during call to basebackup. There is a problem in pageserver error handling so basebackup error appears in pageserver logs like that:
```
2021-11-23T13:10:08.884741Z ERROR page server thread exited with error err=Timed out while waiting for WAL record at LSN 1/29D15638 to arrive, disk consistent LSN=1/29D07CD0 
```
And client (compute tools) sees this as a broken connection during archive downloading:
```
compute-node-325	Caused by:
--
  |   | compute-node-325
  |   | compute-node-325	Error: failed to iterate over archive
```

compute logs prior to the error:
```
compute-node-325	2021-11-23 16:15:20.779 UTC [compute-monitor] INFO: cannot connect to postgres: error connecting to server: Cannot assign requested address (os error 99), retrying
--
  |   | compute-node-325	2021-11-23 16:15:19.818 UTC [main] INFO: safekeepers synced at LSN 1/29D15638
  |   | compute-node-325	2021-11-23 16:15:19.779 UTC [compute-monitor] INFO: watching Postgres activity at postgresql://zenith_admin@localhost/postgres
  |   | compute-node-325	2021-11-23 16:15:19.776 UTC [main] INFO: starting safekeepers syncing
  |   | compute-node-325	2021-11-23 16:15:19.776 UTC [http-endpoint] INFO: starting HTTP server on 0.0.0.0:3080
```

@petuhovskiy could you please take a look?

petuhovskiy:
I looked at the replication thread for this tenant in pageserver and found out that it's stuck in `postgres::client::Client` destructor, which is called at the end of [walreceiver_main](https://github.com/zenithdb/zenith/blob/0650e51f0b8b27a4b2c5818bb13912111b92a99d/pageserver/src/walreceiver.rs#L178). Stacktraces of the stuck thread:
- [short](https://gist.github.com/petuhovskiy/82d70354dcba7d634a61dfe8fe4bc37c)
- [full](https://gist.github.com/petuhovskiy/bd672d8bfc60647948efb998ef0f46fd)

It looks like the issue https://github.com/sfackler/rust-postgres/issues/755 in rust-postgres.

kelvich:
[...]

PR #938, opened 2021-11-26
Shutdown replication socket on CopyFail

petuhovskiy:
A bit hacky fix for the #935

### DIFF
--- a/walkeeper/src/replication.rs
+++ b/walkeeper/src/replication.rs
@@ -14,6 +14,7 @@ use serde::{Deserialize, Serialize};
 use std::cmp::min;
 use std::fs::File;
 use std::io::{Read, Seek, SeekFrom};
+use std::net::Shutdown;
 use std::path::Path;
 use std::sync::Arc;
 use std::thread::sleep;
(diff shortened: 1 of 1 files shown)


kelvich (approved):

walkeeper/src/replication.rs
```diff
@@ -121,7 +122,12 @@ impl ReplicationConn {
                     }
                 }
                 FeMessage::Sync => {}
-                FeMessage::CopyFail => return Err(anyhow!("Copy failed")),
+                FeMessage::CopyFail => {
+                    // Shutdown the connection, because rust-postgres client cannot be dropped
+                    // when connection is alive.
+                    let _ = stream_in.shutdown(Shutdown::Both);
```
LizardWizzard, line 128: Could you please add a comment describing why it is ok to ignore shutdown error? I see that we return error anyway, but maybe it makes sense to log shutdown error too

LizardWizzard (approved):

walkeeper/src/replication.rs
petuhovskiy: https://doc.rust-lang.org/std/net/struct.TcpStream.html#method.shutdown

I think it may only return error if socket was already shut down somewhere else, which is ok for this case.

PR #939, opened 2021-11-26
Forward pageserver connection string from compute to safekeeper

LizardWizzard:
This is needed for implementation of tenant rebalancing. With this
change safekeeper becomes aware of which pageserver is supposed to be
used for replication from this particular compute.

Also this removes possibility of launching compute without pageserver. FYI @petuhovskiy This might be a problem for your safekeeper benchmarks, let me know if I should keep this option

Corresponding postgres PR: https://github.com/zenithdb/postgres/pull/104

Resolves https://github.com/zenithdb/zenith/issues/892

### DIFF
--- a/walkeeper/src/bin/safekeeper.rs
+++ b/walkeeper/src/bin/safekeeper.rs
@@ -44,6 +44,9 @@ fn main() -> Result<()> {
                 .takes_value(true)
                 .help(formatcp!("http endpoint address for metrics on ip:port (default: {DEFAULT_HTTP_LISTEN_ADDR})")),
         )
+        // FIXME this argument is no longer needed since pageserver address is forwarded from compute.
+        // However because this argument is in use by console's e2e tests lets keep it for now and remove separately.
+        // So currently it is a noop.
         .arg(
             Arg::with_name("pageserver")
                 .short("p")
@@ -99,10 +102,6 @@ fn main() -> Result<()> {
 
     if let Some(addr) = arg_matches.value_of("listen-http") {
         conf.listen_http_addr = addr.to_owned();
-    }
-
-    if let Some(addr) = arg_matches.value_of("pageserver") {
-        conf.pageserver_addr = Some(addr.to_owned());
     }
 
     if let Some(ttl) = arg_matches.value_of("ttl") {
--- a/walkeeper/src/send_wal.rs
+++ b/walkeeper/src/send_wal.rs
@@ -46,6 +46,7 @@ impl postgres_backend::Handler for SendWalHandler {
         if let Some(app_name) = sm.params.get("application_name") {
             self.appname = Some(app_name.clone());
         }
+
         Ok(())
     }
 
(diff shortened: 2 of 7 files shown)


LizardWizzard:
Hm, it segfaults somewhere in postgres only in release build only in CI, I'll continue investigation on monday

walkeeper/src/bin/safekeeper.rs
```diff
@@ -43,6 +43,9 @@ fn main() -> Result<()> {
                 .takes_value(true)
                 .help(formatcp!("http endpoint address for metrics on ip:port (default: {DEFAULT_HTTP_LISTEN_ADDR})")),
         )
+        // FIXME this argument is no longer needed since pageserver address is forwarded from compute.
+        // However because this argument is in use by console's e2e tests lets keep it for now and remove separately.
+        // So currently it is a noop.
```
LizardWizzard, line 48: It is easier to remove it from console e2e tests first and then remove from safekeeper. A bit of a circular dependency between repos

petuhovskiy:
@LizardWizzard I tried to run postgres with `zenith.page_server_connstring=''` as I do in walproposer+safekeepers setup, it didn't work for me.

pg.log: 
```
psql: error: connection to server at "127.0.0.1", port 55432 failed: FATAL:  [ZENITH_SMGR] could not establish connection
DETAIL:  invalid integer value "(null)" for connection option "port"
```

safekeeper.log
```
2021-11-29T17:11:55.893714Z  INFO requesting page server to connect to us: start postgresql://(null):@(null):(null) callmemaybe ca8b18aeab59cfff75dfcb11a1ab050f 5c57d306a73e6da6679dc92a77b1dbc7 host=localhost port=5454 options='-c ztimelineid=5c57d306a73e6da6679dc92a77b1dbc7 ztenantid=ca8b18aeab59cfff75dfcb11a1ab050f'    
2021-11-29T17:11:55.893796Z ERROR Failed to connect to pageserver postgresql://(null):@(null):(null): invalid connection string: invalid value for option `port`    
```

It seems that connstr is formatted to `postgresql://(null):@(null):(null)` in postgres code if it's set to empty string in config.

Here's [instruction](https://gist.github.com/petuhovskiy/9a118ba174c6bdc6757a8994cb3a935b) on how to run zenith without pageserver if you want to reproduce this.

walkeeper/src/send_wal.rs
```diff
@@ -75,9 +92,16 @@ impl postgres_backend::Handler for SendWalHandler {
         } else if query_string.starts_with(b"START_REPLICATION") {
             ReplicationConn::new(pgb).run(self, pgb, &query_string)?;
         } else if query_string.starts_with(b"START_WAL_PUSH") {
-            ReceiveWalConn::new(pgb)
-                .run(self)
-                .with_context(|| "failed to run ReceiveWalConn")?;
+            // this solution with pageserver connstr maybe non obvious, maybe it makes more sense to have it in START_WAL_PUSH directly
```
petuhovskiy, line 95: Is there anything against providing pageserver_connstr in START_WAL_PUSH query? I looked through the code, it seems that it would be more convenient to use something like `START_WAL_PUSH {pageserver_connstr}` and will not require some hacks like checking for "pageserver" in application_name.

petuhovskiy: There's also an option to pass connstr in ProposerGreeting, but it may be less convenient due to serialization of variable length string.

LizardWizzard: Agree. I've changed it. This is indeed looks cleaner.

LizardWizzard:
> it didn't work for me.

Hm. Do we consider these tests without pageserver as a long term ones that we will integrate in our pipeline some time? If they are kind of temporary ones it might be ok to specify fake pageserver address in config, or specify compute node's own address so connect is successful but later messages cause errors, but I'm not sure will it actually work, so it needs checking. What are the plans for these tests in general?

lubennikovaav:
looks good to me
There is one issue with vendor/postgres PR. Please, check the comment there.

LizardWizzard:
@petuhovskiy I've discussed this with @kelvich, and the outcome is that we should continue to support running postgres without pageserver, so I'll fix it. I'll keep you posted

LizardWizzard:
I've updated this to allow configuration without running pageserver. Please take a look

@petuhovskiy Using your instruction I've checked that compute works without a pageserver. I think it is a good idea to add a test on that so we won't brake it in the future. It will require some hacking around EnvBuilder in tests because currently we assume that we always have a pageserver. And also it is a good way to preserve needed configuration changes for easier reproduction of any further activities in that direction

@arssher Please take a look when you have a chance

lubennikovaav (approved):

arssher (approved):

arssher (approved):

PR #963, opened 2021-12-06
Bump safekeeper control file version and allow reading the previous one.

arssher:
Should have been a part of cba4da3f4d90e to provide upgrade for previously
existing clusters. Separates version independent header (magic + version) out of
SafeKeeperState to choose what to deserialize.

### DIFF
--- a/walkeeper/src/timeline.rs
+++ b/walkeeper/src/timeline.rs
@@ -2,6 +2,7 @@
 //! persistence and support for interaction between sending and receiving wal.
 
 use anyhow::{anyhow, bail, ensure, Context, Result};
+use byteorder::{LittleEndian, ReadBytesExt, WriteBytesExt};
 use fs2::FileExt;
 use lazy_static::lazy_static;
 use log::*;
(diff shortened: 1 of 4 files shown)


walkeeper/src/timeline.rs
```diff
@@ -156,6 +158,30 @@ impl SharedState {
         })
     }
 
+    // Check the magic/version in the on-disk data and deserialize it, if possible
+    fn deser_sk_state(buf: &mut &[u8], control_file_path: &Path) -> Result<SafeKeeperState> {
```
petuhovskiy, line 162: `control_file_path` is used only for error messages, can be applied at the caller side instead of passing as argument

walkeeper/src/timeline.rs
```diff
@@ -224,25 +250,7 @@ impl SharedState {
...
-            }
-            if state.format_version != SK_FORMAT_VERSION {
-                bail!(
-                    "Got incompatible format version, expected {}, got {}",
-                    SK_FORMAT_VERSION,
-                    state.format_version,
-                );
-            }
-            state
+            SharedState::deser_sk_state(&mut &buf[..buf.len() - CHECKSUM_SIZE], &control_file_path)?
```
petuhovskiy, line 253: SharedState seems not really fitting for deser_sk_state

arssher:
Pushed review changes.

PR #968, opened 2021-12-07
Add basic timeline info endpoint

SomeoneToIgnore:
Prepares for the s3 download on demand functionality by adding a way to query a timeline from pageserver.

Adds two endpoints:
* one that currently lists all local timelines, located on pageserver.
Ideally, remote timelines with statuses should be mentioned too, but the index code that stores the remote timeline info is not yet in main

* one that lists more detailed timeline details
A bunch of generic properties is displayed currently. This method is intended to be used later to trigger a timeline download, that's missing locally

### DIFF
--- a/pageserver/src/http/routes.rs
+++ b/pageserver/src/http/routes.rs
@@ -1,10 +1,11 @@
 use std::sync::Arc;
 
-use anyhow::Result;
+use anyhow::{Context, Result};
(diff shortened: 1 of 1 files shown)


SomeoneToIgnore:
Examples:

* list timelines
timelines

* timeline info
timeline_info

PR #975, opened 2021-12-09
Prefix unused field with underscore

SomeoneToIgnore:
With rustc 1.57.0 , when I run `./run_clippy.sh`, it complains with 

```
error: field is never read: `lock_file`
   --> walkeeper/src/timeline.rs:330:5
    |
330 |     lock_file: File,
    |     ^^^^^^^^^^^^^^^
    |
    = note: `-D dead-code` implied by `-D warnings`

error: could not compile `walkeeper` due to previous error
```

I've added a `_` to the field to satisfy clippy, but might be, that the logic needs to be adjusted also?

### DIFF
--- a/walkeeper/src/timeline.rs
+++ b/walkeeper/src/timeline.rs
@@ -327,7 +327,7 @@ impl GlobalTimelines {
 #[derive(Debug)]
 struct FileStorage {
     // file used to prevent concurrent safekeepers running on the same data
-    lock_file: File,
+    _lock_file: File,
     // save timeline dir to avoid reconstructing it every time
     timeline_dir: PathBuf,
     conf: SafeKeeperConf,
@@ -432,7 +432,7 @@ impl FileStorage {
 
         Ok((
             FileStorage {
-                lock_file,
+                _lock_file: lock_file,
                 timeline_dir,
                 conf: conf.clone(),
                 persist_control_file_seconds: PERSIST_CONTROL_FILE_SECONDS


petuhovskiy (approved):

lubennikovaav:
It looks like the logic around `lock_file` is broken.

I've tried to spawn two safekeepers in one directory and the second safekeeper does a few strange things
- throws error to the logfile of running safekeeper
`2021-12-09T10:18:04.950570Z ERROR Error, unable to lock pid file`
- spawns threads anyway
- hangs indefinitely

To reproduce it run
```
 ./target/debug/zenith init --config=/home/anastasia/zenith/zenith/control_plane/simple.conf
./target/debug/zenith start
```

with this simple.conf 
```
[pageserver]
pg_port = 64000
http_port = 9898
auth_type = 'Trust'

[[safekeepers]]
name = 's1'
pg_port = 5454
http_port = 7676

[[safekeepers]]
name = 's1'
pg_port = 5455
http_port = 7677
```

If you're not planning to fix it in this PR, create an issue, please.

funbringer:
I'd rather either remove this field for good or let it be an irritating eyesore until someone will think this over and provide a proper fix.

SomeoneToIgnore:
Thanks for explaining how to check that.

I'd rather create an issue instead of removing or silencing the issue then.

SomeoneToIgnore:
https://github.com/zenithdb/zenith/issues/976

petuhovskiy:
> It looks like the logic around `lock_file` is broken.
> 
> I've tried to spawn two safekeepers in one directory and the second safekeeper does a few strange things
> 
>     * throws error to the logfile of running safekeeper
>       `2021-12-09T10:18:04.950570Z ERROR Error, unable to lock pid file`
> 
>     * spawns threads anyway
> 
>     * hangs indefinitely
> 

Hm, isn't pidfile different from lock_file? As I understand, we use lock_file to disallow using several safekeepers for the same timeline. And it stays in FileStorage for RAII, so the fix looks good to me.

PR #984, opened 2021-12-13
Wal stream disconnect

lubennikovaav:
The final goal of this PR is to shut down all threads of an inactive tenant in both WAL service and Pageserver.
Tenant is inactive when there are no running compute nodes, serving its timelines.

WAL stream uses the 2 connections:

1. Compute node (walproposer)  -> Safekeeper (ReceiveWalConn module)

When compute node is shut down, safekeeper needs to stop the respective receiving thread. 
Prior to this PR it didn't work because PostgresBackend haven't handled disconnection properly.

2. Safekeeper (ReplicationConn module) -> pageserver (walreceiver thread)

When incoming WAL stream is gone, safekeeper can stop streaming WAL and cancel connection as soon as replica is caught up.
Note that the WAL can be streamed to multiple replicas simultaneously, only disconnect ones that are caught up to the last_recieved_lsn.

Previously, pageserver's walreceiver was trying to reestablish a replication connection in case of failure.
This logic is removed in commit [64c9b7f9](https://github.com/zenithdb/zenith/pull/984/commits/64c9b7f94dc2c581ca287a8380585839d3d8ca80) because it is not needed. 
If necessary, WAL service will send a callback request to ask pageserver to reconnect.

In pageserver, when walreceiver thread is gone, tenant threads (checkpointer & gc) will finish too as soon as they finish processing all received WAL. This logic is already present in pageserver's code, it just wasn't used before.

### DIFF
--- a/walkeeper/src/receive_wal.rs
+++ b/walkeeper/src/receive_wal.rs
@@ -7,7 +7,9 @@ use bytes::Bytes;
 use bytes::BytesMut;
 use tracing::*;
 
+use crate::timeline::Timeline;
 use std::net::SocketAddr;
+use std::sync::Arc;
 
 use crate::safekeeper::AcceptorProposerMessage;
 use crate::safekeeper::ProposerAcceptorMessage;
@@ -92,6 +94,9 @@ impl<'pg> ReceiveWalConn<'pg> {
             _ => bail!("unexpected message {:?} instead of greeting", msg),
         }
 
+        // Incoming WAL stream resumed, so reset information about the timeline pause.
+        swh.timeline.get().continue_streaming();
+
         // if requested, ask pageserver to fetch wal from us
         // as long as this wal_stream is alive, callmemaybe thread
         // will send requests to pageserver
@@ -121,6 +126,7 @@ impl<'pg> ReceiveWalConn<'pg> {
                     tx: tx_clone,
                     tenant_id,
                     timelineid,
+                    timeline: Arc::clone(swh.timeline.get()),
                 })
             }
             None => None,
@@ -144,10 +150,12 @@ struct SendWalHandlerGuard {
     tx: UnboundedSender<CallmeEvent>,
     tenant_id: ZTenantId,
     timelineid: ZTimelineId,
+    timeline: Arc<Timeline>,
 }
 
 impl Drop for SendWalHandlerGuard {
     fn drop(&mut self) {
+        self.timeline.stop_streaming();
         self.tx
             .send(CallmeEvent::Unsubscribe(self.tenant_id, self.timelineid))
             .unwrap_or_else(|e| {
--- a/walkeeper/src/timeline.rs
+++ b/walkeeper/src/timeline.rs
@@ -18,7 +18,7 @@ use zenith_utils::bin_ser::LeSer;
 use zenith_utils::lsn::Lsn;
 use zenith_utils::zid::{ZTenantId, ZTimelineId};
 
-use crate::replication::{HotStandbyFeedback, END_REPLICATION_MARKER};
+use crate::replication::HotStandbyFeedback;
 use crate::safekeeper::{
     AcceptorProposerMessage, ProposerAcceptorMessage, SafeKeeper, SafeKeeperState, ServerInfo,
     Storage, SK_FORMAT_VERSION, SK_MAGIC,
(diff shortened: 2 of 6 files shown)


kelvich:
@lubennikovaav can you please describe the basic idea of that PR here or in the commit messages. In it's current form it is hard to tell neither what it solves nor how. Also please link relevant issue (https://github.com/zenithdb/zenith/issues/945 ?)

walkeeper/src/timeline.rs
```diff
@@ -197,8 +206,44 @@ impl Timeline {
...
+
+    // Check if it's time to stop streaming to the given replica
+    pub fn check_stop_streaming(&self, replica_id: usize) -> Result<()> {
+        let shared_state = self.mutex.lock().unwrap();
+
+        // If stop_lsn is set, it's time to shutdown streaming.
+        if let Some(stop_lsn_request) = shared_state.stop_lsn {
+            let replica_state = shared_state.replicas[replica_id].unwrap();
+            // Lsn::MAX means that we don't know the latest LSN yet.
+            // XXX How should we handle this case?
```
arssher, line 238: Seems like it's fine, we just should wait until it becomes known -- and ensure KeepAlives are actually sent regularly, even if nothing happens -- which is true as I see.

Why not use Option here instead of ::MAX?

Also I haven't checked whether stop happens if commit_lsn is 0, in the very beginning.

arssher (changes requested):

arssher:
A couple of additional notes:

- Logic 'wait for pageserver catching up before killing the connection when compute disappeared' is not 100% reliable as in case of connection trouble before the catch-up it won't be re-established. I don't think we awfully need that though.
- Rarely there might be several computes concurrently talking to safekeepers, it would break 'on new connection' 'on connection end' behaviour, we should have counters there. Can be done separately, actually I was going to add it.

walkeeper/src/receive_wal.rs
```diff
@@ -135,7 +138,12 @@ impl<'pg> ReceiveWalConn<'pg> {
             if let Some(reply) = reply {
                 self.write_msg(&reply)?;
             }
-            msg = self.read_msg()?;
+            msg = self.read_msg().map_err(|e| {
+                if e.to_string().contains("connection closed") {
+                    swh.timeline.get().stop_streaming();
+                }
+                anyhow!("failed to receive message {:?}", e)
```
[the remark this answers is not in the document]
lubennikovaav: Good idea. Fixed.
As for the guard name, I was planning to do a separate refactoring PR to clean up naming and comments in walkeeper.

walkeeper/src/timeline.rs
```diff
@@ -197,8 +206,44 @@ impl Timeline {
...
+    // Notify WAL senders that it's time to stop sending WAL
+    pub fn stop_streaming(&self) {
+        let mut shared_state = self.mutex.lock().unwrap();
+        // Ensure that safekeeper sends WAL up to the last known committed LSN.
+        // It guarantees that pageserver will receive all the latest data
+        // before walservice disconnects.
+        shared_state.stop_lsn = Some(shared_state.notified_commit_lsn);
+        trace!(
+            "Stopping WAL senders. stop_lsn: {}",
+            shared_state.sk.flush_lsn
```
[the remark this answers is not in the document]
lubennikovaav: Thanks for catching. Fixed

walkeeper/src/timeline.rs
lubennikovaav: > Why not use Option here instead of ::MAX?

I used it in the same manner as ReplicaState.disk_consistent_lsn.
Maybe it's worth refactoring both, though.

> Also I haven't checked whether stop happens if commit_lsn is 0, in the very beginning.

Hmm. Could you describe this scenario in more detail?
I guess it wouldn't stop because if safekeeper hasn't streamed anything it means that it hasn't received the feedback as well.

lubennikovaav:
> 
>     * Logic 'wait for pageserver catching up before killing the connection when compute disappeared' is not 100% reliable as in case of connection trouble before the catch-up it won't be re-established. I don't think we awfully need that though.
> 

This is just an optimization, that allows streaming all received data to pageserver and removing WAL from safekeeper, and decreasing the amount of work on the next start. In normal operation, pageserver will catch up quickly.
If the safekeeper->pageserver connection is lost before the pageserver is caught up, all data is still preserved in wal service and will be streamed/replayed on the next start.

>     * Rarely there might be several computes concurrently talking to safekeepers, it would break 'on new connection' 'on connection end' behaviour, we should have counters there. Can be done separately, actually I was going to add it.

Thank you, I wasn't aware of that.

walkeeper/src/timeline.rs
arssher: Well, commit_lsn becomes non-zero only after majority of safekeepers acked WAL receival for the first time (and safekeeper hears about that); it might not happen before the disconnection. So maybe better handle that separately.

arssher:
> This is just an optimization, ...

:ok_hand:

walkeeper/src/timeline.rs
lubennikovaav: Fixed

arssher (approved):

PR #989, opened 2021-12-13
Grab safekeeper.lock on the whole directory instead of per tli.

arssher:
closes #975

### DIFF
--- a/walkeeper/src/bin/safekeeper.rs
+++ b/walkeeper/src/bin/safekeeper.rs
@@ -1,10 +1,12 @@
 //
 // Main entry point for the safekeeper executable
 //
-use anyhow::Result;
+use anyhow::{Context, Result};
 use clap::{App, Arg};
 use const_format::formatcp;
 use daemonize::Daemonize;
+use fs2::FileExt;
+use std::fs::File;
 use std::path::{Path, PathBuf};
 use std::thread;
 use tracing::*;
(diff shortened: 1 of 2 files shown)


walkeeper/src/bin/safekeeper.rs
```diff
@@ -123,6 +127,17 @@ fn start_safekeeper(conf: SafeKeeperConf) -> Result<()> {
 
     info!("version: {}", GIT_VERSION);
 
+    // Prevent running multiple safekeepers on the same directory
+    let lock_file_path = conf.workdir.join(LOCK_FILE_NAME);
+    let lock_file = File::create(&lock_file_path).with_context(|| "failed to open lockfile")?;
+    lock_file.try_lock_exclusive().map_err(|e| {
+        anyhow!(
+            "control file {:?} is locked by some other process: {}",
+            &lock_file_path,
```
SomeoneToIgnore, line 136: Few optional style nits:

* you can format `lock_file_path.display()` with `{}`, in case you prefer better control over what's formatting
* you can chain any standard error, so instead of `map_err` you can use `.context` and `.with_context` from `anyhow::Context`. This way, you get the error "hierarchy" the way anyhow provides it.

SomeoneToIgnore (approved):
Thanks for fixing this, now I can run clippy simply again.

PR #990, opened 2021-12-13
Use proper s3 archive sync order

SomeoneToIgnore:
Also adds logging spans to distinguish between different tasks


LizardWizzard (approved):
LGTM

PR #991, opened 2021-12-13
Propagate disk consistent lsn with timeline sync statuses

SomeoneToIgnore:
Part of https://github.com/zenithdb/zenith/issues/944

Adds a `disk_consistent_lsn` into `TimelineSyncState`, making it possible to know timeline's latest `dick_consistent_lsn` on the remote storage.

### DIFF
--- a/pageserver/src/layered_repository.rs
+++ b/pageserver/src/layered_repository.rs
@@ -139,9 +139,13 @@ impl Repository for LayeredRepository {
         Ok(
             match self.get_or_init_timeline(timelineid, &mut timelines)? {
                 LayeredTimelineEntry::Local(local) => RepositoryTimeline::Local(local),
-                LayeredTimelineEntry::Remote(remote_timeline_id) => {
-                    RepositoryTimeline::Remote(remote_timeline_id)
-                }
+                LayeredTimelineEntry::Remote {
+                    id,
+                    disk_consistent_lsn,
+                } => RepositoryTimeline::Remote {
+                    id,
+                    disk_consistent_lsn,
+                },
             },
         )
     }
@@ -184,7 +188,7 @@ impl Repository for LayeredRepository {
         let mut timelines = self.timelines.lock().unwrap();
         let src_timeline = match self.get_or_init_timeline(src, &mut timelines)? {
             LayeredTimelineEntry::Local(timeline) => timeline,
-            LayeredTimelineEntry::Remote(_) => {
+            LayeredTimelineEntry::Remote { .. } => {
                 bail!("Cannot branch off the timeline {} that's not local", src)
             }
         };
@@ -262,7 +266,7 @@ impl Repository for LayeredRepository {
                 info_span!("checkpoint", timeline = %timelineid, tenant = %self.tenantid).entered();
             match timeline {
                 LayeredTimelineEntry::Local(timeline) => timeline.checkpoint(cconf)?,
-                LayeredTimelineEntry::Remote(_) => debug!(
+                LayeredTimelineEntry::Remote { .. } => debug!(
                     "Cannot run the checkpoint for remote timeline {}",
                     timelineid
                 ),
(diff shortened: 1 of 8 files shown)


pageserver/src/layered_repository.rs
```diff
@@ -392,6 +392,14 @@ impl LayeredTimelineEntry {
             }
         }
     }
+
+    /// Gets a current (latest for the remote case) disk consistent Lsn for the timeline.
+    fn disk_consistent_lsn(&self) -> Option<Lsn> {
```
LizardWizzard, line 397: Option looks unnecessary in the return type

pageserver/src/repository.rs
```diff
@@ -69,7 +69,7 @@ pub enum RepositoryTimeline {
     /// Loaded into pageserver's memory and ready to be used.
     Local(Arc<dyn Timeline>),
     /// Timeline, found on the pageserver's remote storage, but not yet downloaded locally.
-    Remote(ZTimelineId),
+    Remote(ZTimelineId, Lsn),
```
LizardWizzard, line 72: What do you think about using named struct with fields here? Or at least a comment what is the meaning of this Lsn

pageserver/src/repository.rs
```diff
@@ -82,22 +82,38 @@ impl RepositoryTimeline {
...
     /// Timeline was evicted from the pageserver's local working directory due to conflicting remote and local states or too many errors during the synchronization.
-    /// Such timelines cannot have their state synchronized further.
-    Evicted,
+    /// Such timelines cannot have their state synchronized further and may not have the data about remote timeline's disk_consistent_lsn, since eviction may happen
+    /// due to errors before the remote timeline contents is known.
+    Evicted(Option<Lsn>),
+}
+
+impl TimelineSyncState {
+    pub fn synced_disk_consistent_lsn(&self) -> Option<Lsn> {
```
LizardWizzard, line 107: Just a naming suggestion: remote_disk_consistent_lsn, feel free to leave as is

pageserver/src/remote_storage/storage_sync/download.rs
```diff
@@ -43,53 +43,67 @@ pub(super) async fn download_timeline<
     sync_id: TimelineSyncId,
     mut download: TimelineDownload,
     retries: u32,
-) -> Option<bool> {
+) -> Option<(bool, Lsn)> {
```
LizardWizzard, line 46: It would be good to document what is the meaning of different return values. I e None vs Some(false)

LizardWizzard (approved):
Looks good to me! Left some minor comments

lubennikovaav (approved):
Thank you! looks good to me

pageserver/src/remote_storage/storage_sync/download.rs
```diff
@@ -43,53 +43,67 @@ pub(super) async fn download_timeline<
...
+                    }
                 }
             }
         }
     };
+    let remote_disk_consistent_lsn = match remote_timeline.checkpoints().max() {
+        Some(lsn) => lsn,
+        None => {
+            debug!("Cannot download: no disk consistent Lsn is present for the remote timeline");
+            return None;
```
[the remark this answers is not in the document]
SomeoneToIgnore: I'm also not sure how applicable those cases are, I'm using `DownloadedTimeline::Abort` instead of `None` to highlight those cases after your proposals (check it out), and there are 3 cases where I use it, all related to situations when the index data about the remote timelines is incorrect: either misses the remote timeline id, or the entire timeline for the id given.

Both at this point seem more like a programmatic error rather than a possible case, but I'm super reluctant to panicking in such cases, preferring to return some error instead, that could be logged and processed regularly.

SomeoneToIgnore:
The only CI job that fails now is https://github.com/zenithdb/zenith/runs/4534103707?check_suite_focus=true

That is caused by the internal compiler error with rustc 1.57 (that's what GItHub job uses), seems to be already reported and a fix is also submitted: https://github.com/rust-lang/rust/issues/91696
The bug is caused by the incremental compilation issues and apparently GitHub preserves the worker state between job reruns, so the same cache is used and is causing the same issue between reruns.

The rest of the tests pass, so I'll merge the branch as is, if any other GitHub errors occur, we can disable the incremental compilation with the env variable for the time being, or downgrade GitHub's toolchain version.

pageserver/src/remote_storage/storage_sync/download.rs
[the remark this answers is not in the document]
LizardWizzard: Thanks! I like the explicitness of DownloadedTimeline::Abort

Agree that these are more of a programmatic errors and Result is preferable over panics

PR #995, opened 2021-12-14
very first version of tenant migration from one pageserver to another via remote storage

### LINKED ISSUE #896: Attach tenant, part 1 -- load from s3

### LINKED ISSUE #897: Attach tenant, part 2 -- subscribe to safekeepers

### LINKED ISSUE #898: Attach tenant, part 3 -- restart compute with the new pageserver address

### LINKED ISSUE #900: Write test with tenant/pageserver rebalancing

LizardWizzard:
Currently this contains python test which makes some assumptions and uses quite hacky setup with second pageserver without zenith cli support

Resolves #896
Resolves #897
Resolves #898
Resolves #900

### DIFF
--- a/docs/pageserver-tenant-migration.md
+++ b/docs/pageserver-tenant-migration.md
@@ -1,0 +1,22 @@
+## Pageserver tenant migration
+
+### Overview
+
+This feature allows to migrate a timeline from one pageserver to another by utilizing remote storage capability.
+
+### Migration process
+
+Pageserver implements two new http handlers: timeline attach and timeline detach.
+Timeline migration is performed in a following way:
+1. Timeline attach is called on a target pageserver. This asks pageserver to download latest checkpoint uploaded to s3.
+2. For now it is necessary to manually initialize replication stream via callmemaybe call so target pageserver initializes replication from safekeeper (it is desired to avoid this and initialize replication directly in attach handler, but this requires some refactoring (probably [#997](https://github.com/zenithdb/zenith/issues/997)/[#1049](https://github.com/zenithdb/zenith/issues/1049))
+3. Replication state can be tracked via timeline detail pageserver call.
+4. Compute node should be restarted with new pageserver connection string. Issue with multiple compute nodes for one timeline is handled on the safekeeper consensus level. So this is not a problem here.Currently responsibility for rescheduling the compute with updated config lies on external coordinator (console).
+5. Timeline is detached from old pageserver. On disk data is removed.
+
+
+### Implementation details
+
+Now safekeeper needs to track which pageserver it is replicating to. This introduces complications into replication code:
+* We need to distinguish different pageservers (now this is done by connection string which is imperfect and is covered here: https://github.com/zenithdb/zenith/issues/1105). Callmemaybe subscription management also needs to track that (this is already implemented).
+* We need to track which pageserver is the primary. This is needed to avoid reconnections to non primary pageservers. Because we shouldn't reconnect to them when they decide to stop their walreceiver. I e this can appear when there is a load on the compute and we are trying to detach timeline from old pageserver. In this case callmemaybe will try to reconnect to it because replication termination condition is not met (page server with active compute could never catch up to the latest lsn, so there is always some wal tail)
--- a/pageserver/src/http/routes.rs
+++ b/pageserver/src/http/routes.rs
@@ -25,6 +25,7 @@ use zenith_utils::zid::{opt_display_serde, ZTimelineId};
 use super::models::BranchCreateRequest;
 use super::models::TenantCreateRequest;
 use crate::branches::BranchInfo;
+use crate::repository::RepositoryTimeline;
 use crate::repository::TimelineSyncState;
 use crate::{branches, config::PageServerConf, tenant_mgr, ZTenantId};
 
@@ -247,6 +248,58 @@ async fn timeline_detail_handler(request: Request<Body>) -> Result<Response<Body
     Ok(json_response(StatusCode::OK, response_data)?)
 }
 
+async fn timeline_attach_handler(request: Request<Body>) -> Result<Response<Body>, ApiError> {
+    let tenant_id: ZTenantId = parse_request_param(&request, "tenant_id")?;
+    check_permission(&request, Some(tenant_id))?;
+
+    let timeline_id: ZTimelineId = parse_request_param(&request, "timeline_id")?;
+
+    tokio::task::spawn_blocking(move || {
+        let _enter =
+            info_span!("timeline_attach_handler", tenant = %tenant_id, timeline = %timeline_id)
+                .entered();
+        let repo = tenant_mgr::get_repository_for_tenant(tenant_id)?;
+        match repo.get_timeline(timeline_id)? {
+            RepositoryTimeline::Local(_) => {
+                anyhow::bail!("Timeline with id {} is already local", timeline_id)
+            }
+            RepositoryTimeline::Remote {
+                id: _,
+                disk_consistent_lsn: _,
+            } => {
+                // FIXME (rodionov) get timeline already schedules timeline for download, and duplicate tasks can cause errors
+                //  first should be fixed in https://github.com/zenithdb/zenith/issues/997
+                // TODO (rodionov) change timeline state to awaits download (incapsulate it somewhere in the repo)
+                // TODO (rodionov) can we safely request replication on the timeline before sync is completed? (can be implemented on top of the #997)
+                Ok(())
+            }
+        }
+    })
+    .await
+    .map_err(ApiError::from_err)??;
+
+    Ok(json_response(StatusCode::ACCEPTED, ())?)
+}
+
+async fn timeline_detach_handler(request: Request<Body>) -> Result<Response<Body>, ApiError> {
+    let tenant_id: ZTenantId = parse_request_param(&request, "tenant_id")?;
+    check_permission(&request, Some(tenant_id))?;
+
+    let timeline_id: ZTimelineId = parse_request_param(&request, "timeline_id")?;
+
+    tokio::task::spawn_blocking(move || {
+        let _enter =
+            info_span!("timeline_detach_handler", tenant = %tenant_id, timeline = %timeline_id)
+                .entered();
+        let repo = tenant_mgr::get_repository_for_tenant(tenant_id)?;
+        repo.detach_timeline(timeline_id)
+    })
+    .await
+    .map_err(ApiError::from_err)??;
+
+    Ok(json_response(StatusCode::OK, ())?)
+}
+
 async fn tenant_list_handler(request: Request<Body>) -> Result<Response<Body>, ApiError> {
     // check for management permission
     check_permission(&request, None)?;
@@ -267,13 +320,13 @@ async fn tenant_create_handler(mut request: Request<Body>) -> Result<Response<Bo
 
     let request_data: TenantCreateRequest = json_request(&mut request).await?;
 
-    let response_data = tokio::task::spawn_blocking(move || {
+    tokio::task::spawn_blocking(move || {
         let _enter = info_span!("tenant_create", tenant = %request_data.tenant_id).entered();
         tenant_mgr::create_repository_for_tenant(get_config(&request), request_data.tenant_id)
     })
     .await
     .map_err(ApiError::from_err)??;
-    Ok(json_response(StatusCode::CREATED, response_data)?)
+    Ok(json_response(StatusCode::CREATED, ())?)
 }
 
 async fn handler_404(_: Request<Body>) -> Result<Response<Body>, ApiError> {
@@ -307,6 +360,14 @@ pub fn make_router(
         .get(
             "/v1/timeline/:tenant_id/:timeline_id",
             timeline_detail_handler,
+        )
+        .post(
+            "/v1/timeline/:tenant_id/:timeline_id/attach",
+            timeline_attach_handler,
+        )
+        .post(
+            "/v1/timeline/:tenant_id/:timeline_id/detach",
+            timeline_detach_handler,
         )
         .get("/v1/branch/:tenant_id", branch_list_handler)
         .get("/v1/branch/:tenant_id/:branch_name", branch_detail_handler)
--- a/pageserver/src/remote_storage/local_fs.rs
+++ b/pageserver/src/remote_storage/local_fs.rs
@@ -73,7 +73,7 @@ impl RemoteStorage for LocalFs {
     }
 
     async fn list(&self) -> anyhow::Result<Vec<Self::StoragePath>> {
-        Ok(get_all_files(&self.root).await?.into_iter().collect())
+        get_all_files(&self.root).await
     }
 
     async fn upload(
--- a/test_runner/batch_others/test_tenant_relocation.py
+++ b/test_runner/batch_others/test_tenant_relocation.py
@@ -12,6 +12,10 @@ import pytest
 from fixtures.zenith_fixtures import PgProtocol, PortDistributor, Postgres, ZenithEnvBuilder, ZenithPageserverHttpClient, zenith_binpath, pg_distrib_dir
 
 
+def assert_abs_margin_ratio(a: float, b: float, margin_ratio: float):
+    assert abs(a - b) / a < margin_ratio, (a, b, margin_ratio)
+
+
 @contextmanager
 def new_pageserver_helper(new_pageserver_dir: pathlib.Path,
                           pageserver_bin: pathlib.Path,
@@ -95,7 +99,9 @@ def load(pg: Postgres, stop_event: threading.Event, load_ok_event: threading.Eve
                     # if we recovered after failure verify that we have correct number of rows
                     log.info("recovering at %s", inserted_ctr)
                     cur.execute("SELECT count(*) FROM load")
-                    assert cur.fetchone()[0] == inserted_ctr
+                    # it seems that sometimes transaction gets commited before we can acknowledge
+                    # the result, so sometimes selected value is larger by one than we expect
+                    assert cur.fetchone()[0] - inserted_ctr <= 1
                     log.info("successfully recovered %s", inserted_ctr)
                     failed = False
                     load_ok_event.set()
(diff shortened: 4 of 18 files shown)


pageserver/src/http/routes.rs
```diff
@@ -204,6 +206,8 @@ struct TimelineInfo {
     timeline_state: Option<TimelineSyncState>,
 }
 
+// TODO should we return something meaningful about remote timelines too
+// No way currently to ckeck uploading by tenant state, it woud be just active or idle (check by s3 consistent lsn?)
```
SomeoneToIgnore, line 210:
```suggestion
// No way currently to check uploading by tenant state, it would be just active or idle (check by s3 consistent lsn?)
```

There would be a way very soon, see the last file in the PR for the API usage:
https://github.com/zenithdb/zenith/pull/991

pageserver/src/http/routes.rs
```diff
@@ -235,6 +239,36 @@ async fn timeline_detail_handler(request: Request<Body>) -> Result<Response<Body
...
+        let repo = tenant_mgr::get_repository_for_tenant(tenant_id)?;
+        match repo.get_timeline(timeline_id)? {
+            RepositoryTimeline::Local(_) => {
+                bail!("Timeline with id {} is already local", timeline_id)
+            }
+            RepositoryTimeline::Remote(_) => {
+                // FIXME this is a bit hacky because there is local_or_schedule_download on LayeredTimelineEntry, but not on RepositoryTimeline
+                // Also it is a good idea to change timeline status to AwaitsDownload (TODO probably incapsulate it somewhere in the repo)
+                repo.set_timeline_state(timeline_id, TimelineSyncState::AwaitsDownload)?;
+                schedule_timeline_download(tenant_id, timeline_id);
```
SomeoneToIgnore, line 261: I know this is a bit of a weird naming now, but `repo.get_timeline` now, being implemented by the `LayeredRepository`, always attempts to download (if not downloaded) and read from disk and initialize (if not intialized) before returning a timeline, so no need to schedule anything else on top, I assume?

Impl details:
https://github.com/zenithdb/zenith/blob/5cff7d1de9862d8a826c4454d997a193ffa26ba6/pageserver/src/layered_repository.rs#L140


> is a good idea to change timeline status to AwaitsDownload

Yes, good idea, yet not super strictly necessary now: we already return a techically correct `CloudOnly` one on the `LayeredRepository` impl: https://github.com/zenithdb/zenith/blob/5cff7d1de9862d8a826c4454d997a193ffa26ba6/pageserver/src/layered_repository.rs#L326

and it should be better when we sort out repeated downloads for already downloaded timeline.

test_runner/batch_others/test_tenant_relocation.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+@contextmanager
+def new_pageserver_helper(new_pageserver_dir: pathlib.Path,
+                          pageserver_bin: pathlib.Path,
+                          remote_storage_mock_path: pathlib.Path,
+                          pg_port: int,
+                          http_port: int):
+    """
+    cannot use ZenithPageserver yet bacause it depends on zenith cli
```
SomeoneToIgnore, line 20:
```suggestion
    cannot use ZenithPageserver yet because it depends on zenith cli
```

zenith_utils/src/zid.rs
```diff
@@ -64,6 +64,12 @@ impl fmt::Display for ZId {
     }
 }
 
+impl fmt::Debug for ZId {
```
SomeoneToIgnore, line 67: 👍

test_runner/batch_others/test_tenant_relocation.py
```diff
@@ -0,0 +1,174 @@
...
+        str(new_pageserver_dir),
+        '--postgres-distrib',
+        pg_distrib_dir,
+    ]
+
+    subprocess.check_output(cmd, text=True)
+
+    new_pageserver_config_file_path = new_pageserver_dir / 'pageserver.toml'
+
+    # add remote storage mock to pageserver config
```
SomeoneToIgnore, line 40: Maybe worth a todo, IMO, we would like to enable remote storage for every Python IT tests with one env var/command, so the file appending way is rather crude and temporary.

SomeoneToIgnore (commented):
Looks nice, actually, despite all the WIP stuff.

I'm a bit concerned by the `sleep(n)` things, but this is not a problem right now.
IMO later, we could call some http method periodically and limit the number of calls instead.

pageserver/src/http/routes.rs
LizardWizzard: Hm, I think on demand nature of get_timeline is a bit awkward, for example when you want to inspect the state without changing it. IMO decision to download missing files (or initialize the local repo) should me made by the calling code. If this is the most common pattern we can have multiple methods, like get_or_download or something.

Also there is a branch create api, so we should know upfront when new timeline is being initialized, so maybe we can  avoid this confusion by throwing an error on missing timeline instead of downloading/initialization. What do you think?

LizardWizzard:
@SomeoneToIgnore Thanks for the review!

> IMO later, we could call some http method periodically and limit the number of calls instead.

Yes, I fully agree with that, we can have some utility in our python tests to wait for certain predicate to become true with a timeout on a waiting period. When all the bits will be in place I'll remove these sleep calls

pageserver/src/http/routes.rs
SomeoneToIgnore: Ideally, yes, that would be great.
Realistically, we have to go over each such invocation now and assert its behavior, possibly amending the code around.
I think it's better to discuss it a bit first and then do the changes in a separate PR, if agreed on.

LizardWizzard: See https://github.com/zenithdb/zenith/issues/997

SomeoneToIgnore:
https://github.com/zenithdb/zenith/pull/1079 helps with hacks around the remote storage

LizardWizzard:
I think in this form this patch can be accepted and TODO's are closed with follow ups. I used links in them to corresponding issues. 

@arssher @lubennikovaav Please take a look at safekeeper changes, this mostly touches callmemaybe stuff

docs/pageserver-tenant-migration.md
```diff
@@ -0,0 +1,20 @@
...
+### Overview
+
+This feature allows to migrate a timeline from one pageserver to another by utilizing remote storage capability.
+
+### Migration process
+
+Pageserver implements two new http handlers: timeline attach and timeline detach.
+Timeline migration is performed in a following way:
+1. Timeline attach is called on a target pageserver.
+2. For now it is necessary to manually initialize replication stream via callmemaybe call so target pageserver initializes replication from safekeeper (it is desired to avoid this and initialize replication directly in attach handler, but this requires some refactoring TODO create issue)
```
SomeoneToIgnore, line 12: > TODO create issue

IMO, issues can be created without getting pushed into repo as TODOs.
There's at least one more similar TODO below that benefits from an actual issue created.

pageserver/src/http/routes.rs
```diff
@@ -267,13 +320,13 @@ async fn tenant_create_handler(mut request: Request<Body>) -> Result<Response<Bo
 
     let request_data: TenantCreateRequest = json_request(&mut request).await?;
 
-    let response_data = tokio::task::spawn_blocking(move || {
+    tokio::task::spawn_blocking(move || {
         let _enter = info_span!("tenant_create", tenant = %request_data.tenant_id).entered();
```
SomeoneToIgnore, line 324: Not your issue, but the span name is definitely odd :)
```suggestion
        let _enter = info_span!("tenant_list_handler", tenant = %request_data.tenant_id).entered();
```

pageserver/src/remote_storage/local_fs.rs
```diff
@@ -73,7 +73,7 @@ impl RemoteStorage for LocalFs {
     }
 
     async fn list(&self) -> anyhow::Result<Vec<Self::StoragePath>> {
-        Ok(get_all_files(&self.root).await?.into_iter().collect())
+        get_all_files(&self.root).await
```
SomeoneToIgnore, line 76: 👍

test_runner/fixtures/zenith_fixtures.py
```diff
@@ -637,6 +637,24 @@ def __init__(self, port: int, auth_token: Optional[str] = None) -> None:
     def check_status(self):
         self.get(f"http://localhost:{self.port}/v1/status").raise_for_status()
 
+    def timeline_detail(self, tenant_id: uuid.UUID, timeline_id: uuid.UUID):
```
SomeoneToIgnore, line 640: There's `timeline_details` already: https://github.com/zenithdb/zenith/blob/8ab4c8a050895a8e6fe5ba53708d9f3c42e59ed4/test_runner/fixtures/zenith_fixtures.py#L692

so I'd rather reuse that.

walkeeper/src/callmemaybe.rs
```diff
@@ -190,7 +195,7 @@ impl Drop for SubscriptionState {
 }
 
 pub async fn main_loop(conf: SafeKeeperConf, mut rx: UnboundedReceiver<CallmeEvent>) -> Result<()> {
-    let subscriptions: Mutex<HashMap<(ZTenantId, ZTimelineId), SubscriptionState>> =
+    let subscriptions: Mutex<HashMap<(ZTenantId, ZTimelineId, String), SubscriptionState>> =
```
SomeoneToIgnore, line 198: NIT: that `String` might be more readable with the name. Also, the joined Id could be used too, as the other components come to the same approach?

walkeeper/src/handler.rs
```diff
@@ -89,6 +90,8 @@ impl postgres_backend::Handler for SafekeeperPostgresHandler {
                 self.appname = Some(app_name.clone());
             }
 
+            self.pageserver_connstr = params.get("pageserver_connstr").map(|s| s.to_owned());
```
SomeoneToIgnore, line 93:
```suggestion
            self.pageserver_connstr = params.get("pageserver_connstr").cloned();
```

SomeoneToIgnore (approved):
I have not looked thoroughly at the Python code, but LGTM overall

docs/pageserver-tenant-migration.md
LizardWizzard: Oops, this todo is for me to jut not forget to create an issue :) This particular one is added to the checklist in rebalancing epic, so I'll just remove it here

pageserver/src/http/routes.rs
LizardWizzard: This is actually tenant create handler, I was just passing by and removed odd `response_data` represented by unit type. Diff visualization is indeed confusing

test_runner/fixtures/zenith_fixtures.py
LizardWizzard: Good catch! It seems I missed it when it came with rebase on top of you remote storage tests PR :)

LizardWizzard: I've merged both versions together, so it accepts UUID instead of strings like other client methods

LizardWizzard:
I've rebased this on top of shutdown improvements patch and added margins for some exact checks to reduce possible flakiness

walkeeper/src/callmemaybe.rs
LizardWizzard: Done, see last commit. It is weird that the value (SubscriptionState) keeps the entire key too. It should be ok since this shouldn't be a bottleneck here, but doesnt look great anyway

walkeeper/src/send_wal.rs
```diff
@@ -90,18 +90,27 @@ impl Drop for ReplicationConnGuard {
...
 }
 
 impl Drop for ReplicationStreamGuard {
     fn drop(&mut self) {
         // the connection with pageserver is lost,
         // resume callback subscription
-        debug!("Connection to pageserver is gone. Resume callmemeybe subsciption if necessary. tenantid {} timelineid {}",
-         self.tenant_id, self.timelineid);
+        debug!(
+            "Connection to pageserver is gone. Resume callmemeybe subsciption if necessary. tenantid {} timelineid {}",
```
SomeoneToIgnore, line 102:
```suggestion
            "Connection to pageserver is gone. Resume callmemaybe subscription if necessary. tenantid {} timelineid {}",
```

walkeeper/src/send_wal.rs
```diff
@@ -194,8 +203,9 @@ impl ReplicationConn {
         spg: &mut SafekeeperPostgresHandler,
         pgb: &mut PostgresBackend,
         mut start_pos: Lsn,
+        pageserver_connstr: Option<String>,
     ) -> Result<()> {
-        let _enter = info_span!("WAL sender", timeline = %spg.ztimelineid.unwrap()).entered();
+        let _enter = info_span!("WAL sender", timeline = %spg.ztimelineid.unwrap(), pageserver_connstr = %pageserver_connstr.clone().unwrap_or_default()).entered();
```
SomeoneToIgnore, line 208:
```suggestion
        let _enter = info_span!("WAL sender", timeline = %spg.ztimelineid.unwrap(), pageserver_connstr = %pageserver_connstr.ad_deref().unwrap_or_default()).entered();
```

SomeoneToIgnore (commented):
If I compile walkeeper package alone, I get an error:

```
❯ cargo check --package walkeeper --all-targets
    Checking walkeeper v0.1.0 (/Users/someonetoignore/Work/zenith/zenith/walkeeper)
error[E0433]: failed to resolve: could not find `select` in `tokio`
   --> walkeeper/src/callmemaybe.rs:220:16
    |
220 |         tokio::select! {
    |                ^^^^^^ could not find `select` in `tokio`
```

feels like `tokio` is missing a feature there?

Otherwise, the same "LGTM but not sure about Python magic code" impression.

LizardWizzard:
Thanks for the review @SomeoneToIgnore!

> If I compile walkeeper package alone, I get an error:

It seems that the error appears on main too. I wonder should we pin somewhere (in zeinth_utils?) tokio dependency with all the needed features. This should simplify things a bit, and AFAIK cargo uses union of all features needed by crate/it's dependencies so this shouldn't affect compile times, but worth checking

docs/pageserver-tenant-migration.md
```diff
@@ -0,0 +1,20 @@
...
+This feature allows to migrate a timeline from one pageserver to another by utilizing remote storage capability.
+
+### Migration process
+
+Pageserver implements two new http handlers: timeline attach and timeline detach.
+Timeline migration is performed in a following way:
+1. Timeline attach is called on a target pageserver.
+2. For now it is necessary to manually initialize replication stream via callmemaybe call so target pageserver initializes replication from safekeeper (it is desired to avoid this and initialize replication directly in attach handler, but this requires some refactoring (probably [#997](https://github.com/zenithdb/zenith/issues/997)/[#1049](https://github.com/zenithdb/zenith/issues/1049))
+3. Replication state can be tracked via timeline detail pageserver call.
+4. Compute node should be restarted with new pageserver connection string.
```
lubennikovaav, line 14: Is it possible that old and new compute nodes exist simultaneously?
Which component is responsible for choosing the primary node - the one that serves write queries and streams WAL to safekeepers?
It's worth describing it in README.

LizardWizzard: It is solved on the safekeepers level. Two computes cannot simultaneously exist. If we support config change on the fly there shouldn't be even a pod restart. Also in the future it is planned to support this change without compute restart whatsoever. In its current state responsibility for rescheduling the compute with updated config is on console.

Thanks for pointing that out, I'll add this to the doc

docs/pageserver-tenant-migration.md
```diff
@@ -0,0 +1,20 @@
...
+5. Timeline is detached from old pageserver.
+
+
+### Implementation details
+
+Now safekeeper needs to track which pageserver it is replicating to. This introduces complications into replication code and callmemaybe subscription state handling.
```
lubennikovaav, line 20: Could you elaborate on the "complications in the replication code"? 

I wonder if we need special treatment of pageserver's feedback in `get_replicas_state()` on safekeeper, given that two streams can exist for the same timeline?

test_runner/batch_others/test_tenant_relocation.py
```diff
@@ -0,0 +1,266 @@
...
+    log.info('load thread stopped')
+
+
+def assert_local(pageserver_http_client: ZenithPageserverHttpClient, tenant: str, timeline: str):
+    timeline_detail = pageserver_http_client.timeline_detail(UUID(tenant), UUID(timeline))
+    assert timeline_detail.get('type') == "Local", timeline_detail
+    return timeline_detail
+
+
+@pytest.mark.parametrize('with_load', ['with_load', 'without_load'])
```
lubennikovaav, line 117: why not just `'with_load', [False, True]`?

test_runner/batch_others/test_tenant_relocation.py
```diff
@@ -0,0 +1,266 @@
...
+                               pageserver_bin,
+                               remote_storage_mock_path,
+                               new_pageserver_pg_port,
+                               new_pageserver_http_port):
+
+        # call to attach timeline to new timeline
+        new_pageserver_http_client.timeline_attach(UUID(tenant), UUID(timeline))
+        # FIXME cannot handle duplicate download requests, subject to fix in https://github.com/zenithdb/zenith/issues/997
+        time.sleep(5)
+        # new pageserver should be completely in sync with the old one (because there was no new writes since checkpoint)
```
lubennikovaav, line 199: Can `autovacuum` break this assumption?
I don't see if we disable it anywhere.

test_runner/batch_others/test_tenant_relocation.py
LizardWizzard: Because when this is encoded in the test name it becomes just `test_tenant_relocation[True]` without a parameter name. Which is harder to read. The same is already used in one of our tests, namely test_bulk_tenant_create

docs/pageserver-tenant-migration.md
LizardWizzard: By complications I mean additional handling of pageserver connection string and the requirement to track which pageserver is actually a primary one, to avoid reconnections to it (which is not currently implemented but is present in the epic in the follow ups section)

test_runner/batch_others/test_tenant_relocation.py
LizardWizzard: I think I need to fix this comment :) Code below already uses non exact assertions

lubennikovaav (approved):

docs/pageserver-tenant-migration.md
LizardWizzard: > given that two streams can exist for the same timeline?

At a first glance everything looks OK because we keep a Vec of replicas for particular timeline. Do you have some invariants in mind that should be checked?

walkeeper/src/send_wal.rs
```diff
@@ -293,12 +307,18 @@ impl ReplicationConn {
                 if let Some(lsn) = lsn {
                     end_pos = lsn;
                 } else {
-                    // Is is time to end streaming to this replica?
+                    // Is it time to end streaming to this replica?
                     if spg.timeline.get().check_stop_streaming(replica_id) {
-                        let timelineid = spg.timeline.get().timelineid;
+                        // TODO this expect is ugly, will it be better with node id?
+                        //   can we guarantee that this code is not executed during walproposer recovery?
+                        let pageserver_connstr = pageserver_connstr
```
LizardWizzard, line 314: @lubennikovaav what do you think about this `expect` call? It was not triggered in tests so I assume this code path it not followed in walproposer recovery but can we guarantee somehow that recovery won't touch this?

docs/pageserver-tenant-migration.md
LizardWizzard: Added to the doc, feel free to unresolve if something is still missing

test_runner/batch_others/test_tenant_relocation.py
LizardWizzard: I've changed the comment

walkeeper/src/send_wal.rs
lubennikovaav: If walproposer is in recovery `stop_lsn` is set and this is checked in the code above this line.

docs/pageserver-tenant-migration.md
```diff
@@ -0,0 +1,22 @@
...
+1. Timeline attach is called on a target pageserver. This asks pageserver to download latest checkpoint uploaded to s3.
+2. For now it is necessary to manually initialize replication stream via callmemaybe call so target pageserver initializes replication from safekeeper (it is desired to avoid this and initialize replication directly in attach handler, but this requires some refactoring (probably [#997](https://github.com/zenithdb/zenith/issues/997)/[#1049](https://github.com/zenithdb/zenith/issues/1049))
+3. Replication state can be tracked via timeline detail pageserver call.
+4. Compute node should be restarted with new pageserver connection string. Issue with multiple compute nodes for one timeline is handled on the safekeeper consensus level. So this is not a problem here.Currently responsibility for rescheduling the compute with updated config lies on external coordinator (console).
+5. Timeline is detached from old pageserver. On disk data is removed.
+
+
+### Implementation details
+
+Now safekeeper needs to track which pageserver it is replicating to. This introduces complications into replication code:
```
kelvich, line 20: BTW do you think we can address both of that with https://github.com/zenithdb/rfcs/pull/16 ?

LizardWizzard: For the first one, if callmemaybe goes away there is no need to weak it for these changes.

The second one can go away too if safekeeper doesnt initiate connections (and doesnt reconnect) to pageserver. As you've answered this is exactly how it is proposed in the RFC, so yeah, both problems are  absent with this RFC being implemented

PR #1001, opened 2021-12-15
Disable rustc incremental compilation to avoid ICEs

SomeoneToIgnore:


funbringer (approved):
I'd add this even if it didn't fix the ICE :) This is what we do in circle's pipeline anyway.

funbringer:
However, I'd also pin rust to 1.56.1 to match the one used in circleci. I would expect this to fix the problem, as well as add to the build consistency. If we'd like to have a separate pipeline for highlighting upcoming rustc compatibility issues, we'd better make it stand out (give it a descriptive name etc). However, you don't have to do any of these in this patch.

SomeoneToIgnore:
I like the idea of pinning, but would want to do that when getting another bump on CircleCI.

It feels like having one less CI system to deal with might simplify our work on that front though?

funbringer:
> It feels like having one less CI system to deal with might simplify our work on that front though?

Agreed. As far as I know, @LizardWizzard would also prefer to have a single CI system, but I haven't heard of any concrete plans on that matter. That said, I'd find myself in a great trouble If I were to choose between GH actions and CircleCI, since they both have so many nuances and shortcomings that I'm unable to tell which is worse.

PR #1003, opened 2021-12-15
storage CI flow for staging environment

cicdteam:
This PR add simple deployment flow for staging environment:
- discover storage nodes in staging environment (by ec2 instance tags)
- extract new binaries and copy them to storage nodes
- restart safekeepers and pageserver(s)

These variables used and  should be added to CircleCI
- `STAGING_AWS_REGION` - AWS Region where storage nodes resides
- `STAGING_AWS_ACCESS_KEY_ID` and `STAGING_AWS_SECRET_ACCESS_KEY` - AWS credentials with EC2ReadOnlyAccess (to discover EC2 instances by tags)
- `STAGING_SSH_KEY` - base64 encoded private ssh key to access storage nodes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -465,8 +465,9 @@ jobs:
           command: |
             rm -rf zenith_install postgres_install.tar.gz zenith_install.tar.gz
             mkdir zenith_install
-            docker pull --quiet zenithdb/zenith:latest
-            ID=$(docker create zenithdb/zenith:latest)
+            DOCKER_TAG=$(git log --oneline|wc -l)
+            docker pull --quiet zenithdb/zenith:${DOCKER_TAG}
+            ID=$(docker create zenithdb/zenith:${DOCKER_TAG})
             docker cp $ID:/data/postgres_install.tar.gz .
             tar -xzf postgres_install.tar.gz -C zenith_install && rm postgres_install.tar.gz
             docker cp $ID:/usr/local/bin/pageserver zenith_install/bin/
@@ -474,6 +475,7 @@ jobs:
             docker cp $ID:/usr/local/bin/proxy zenith_install/bin/
             docker cp $ID:/usr/local/bin/postgres zenith_install/bin/
             docker rm -v $ID
+            echo ${DOCKER_TAG} | tee zenith_install/.zenith_current_version
             tar -czf zenith_install.tar.gz -C zenith_install .
             ls -la zenith_install.tar.gz
       - run:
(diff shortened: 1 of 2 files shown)


kelvich (approved):
Looks great, thanks!

.circleci/config.yml
```diff
@@ -454,6 +454,45 @@ jobs:
...
+      - checkout
+      - setup_remote_docker
+      - run:
+          name: Get Zenith binaries
+          command: |
+            rm -rf zenith_install postgres_install.tar.gz zenith_install.tar.gz
+            mkdir zenith_install
+            docker pull --quiet zenithdb/zenith:latest
+            ID=$(docker create zenithdb/zenith:latest)
+            docker cp $ID:/data/postgres_install.tar.gz .
```
ololobus, line 470: I think that here we can have a race if two concurrent builds occur. If there are two builds B1 (older) and B2 (newer). B2 may build and push `zenithdb/zenith:latest` faster than B1, then B1 will override it with an older version of image. Because, IIRC, we do not cancel previous build if new one arrives currently.

That way, we should at least use the tag of the current build here, not `latest`, IMO

ololobus: And another thing is that older builds (e.g. after re-run or concurrent run) shouldn't actually override the `latest` if it's newer. I don't know how to check it though

ololobus: > That way, we should at least use the tag of the current build here, not latest, IMO

Hm, and anyway, this doesn't protect us from the situation when we deploy older binaries after newer... I cannot find how to prevent two concurrent workflows to run in parallel in circle. Maybe we can protect from this situation in the console versions update API later

kelvich: May be always use version? That way concurrent build can change docker images order but if during deploy we'll check that we are not downgrading that should be okay.

ololobus: Yes, I meant version in the 'tag of the current build'. And deploy script should check that it isn't deploying something older

cicdteam: Race could be only when two PRs merged into main branch at the same time. And yes, CircleCI hasn't ability to prevent concurrent builds. It can cancel redundant builds:

>To save time, you can configure CircleCI to automatically cancel any non-terminated workflows when a newer pipeline is triggered on that same branch.

but I highly not recommend to use this feature. Canceling deploy task in the middle could broke storage nodes (as example - partially deployed safekeepers and not deployed page server and so on).

I'll update CircleCI config to use docker tag in the binaries extraction step.

cicdteam:
@kelvich @ololobus 

I've updated PR:
- binaries taken from tagged docker image
- store version (docker tag) on the storage nodes and prevent redeploy if we try deploy version older than already deployed (it could happen in race of two PRs merged to main at the same moment).

> reviews re-requested

ololobus (approved):
Cool, thanks! Let's give it a try

PR #1032, opened 2021-12-17
Make safekeeper namings more consistent with reality.

arssher:
s/send_wal.rs/sk_postgres.rs
s/SendWalHandler/SafekeeperPostgresHandler
s/replication.rs/send_wal.rs


petuhovskiy (approved):

lubennikovaav (approved):

lubennikovaav (approved):

LizardWizzard:
IMO these short names/abbreviations do not add much clarity in general. I tend to like descriptive long names more than short ones with abbreviations. 

> SafekeeperPostgresHandler

Does not add much context to me. Does it handle connection with compute postgres or with replica? (postgres readonly replica or pageserver). Maybe it would be better if described in terms of Send/Receive. The same about file name changes. It may be possible to split something into zenith_utils (maybe some generic replication handling). This way we can have more safekeeper specific names and generic code separately. Thoughts?

arssher:
> Does not add much context to me. Does it handle connection with compute postgres or with replica? (postgres readonly replica or pageserver). Maybe it would be better if described in terms of Send/Receive. 

No, it just means it handles Postgres commands which safekeeper supports. That includes `START_REPLICATION` which pulls WAL from safekeeper and `START_WAL_PUSH` which pushes it, so send/receive doesn't provide meaningful boundary here.
So perhaps these renames are not ideal, but I'm open to concrete proposals to do it better.

I want to merge this soon as it makes rebases quite unpleasant.

LizardWizzard:
> No, it just means it handles Postgres (Postgres basically means protocol here) commands which safekeeper supports

Got it. One of the options is to be consistent with pageserver's `PageServerHandler`. So this would be `SafekeeperHandler`. Filename can be just `handler.rs` or `safekeeper.rs` and previous `safekeeper.rs` renamed to `acceptor.rs`.

> I want to merge this soon as it makes rebases quite unpleasant.

Feel free to merge, though `sk` in the filename still bothers me. We can discuss this further somewhere else

Issue #944, opened 2021-11-29
Trim safekeeper WAL based on s3_lsn feedback from pageserver


kelvich:

stepashka:
@SomeoneToIgnore , can you help with this?

SomeoneToIgnore:
I know what to do now, will deal with my current task and add this, before Friday.

SomeoneToIgnore:
https://github.com/zenithdb/zenith/pull/991

stepashka:
note, without #1049 our trimming isn't always correct

PR #1037, opened 2021-12-20
Extend replication protocol with ZenithFeedback message

lubennikovaav:
Extend replication protocol with ZenithFeedback message to pass custom feedback fields and make this message extensible.

**This PR breaks the API. Inform console team before merging.**

Instead of standard SatndbyStatusUpdate, now walreceiver sends ZenithFeedback message, which includes standby_status_update fields and also contains extra fields. Currently, the only custom field is `current_instance_size`.
The message can be extended, retaining backward compatibility.

This is needed for #320.

 #944 and backpressure lsns are fit into standard standby_status_update lsn fields, but in future, we may decide to send them as custom fields if necessary.

Note that this PR uses a patched version of rust-postgres. Respective PR is [here](https://github.com/zenithdb/rust-postgres/pull/3)
vendor/postgres PR is https://github.com/zenithdb/postgres/pull/108

### DIFF
--- a/walkeeper/src/safekeeper.rs
+++ b/walkeeper/src/safekeeper.rs
@@ -22,6 +22,7 @@ use zenith_metrics::{
 use zenith_utils::bin_ser::LeSer;
 use zenith_utils::lsn::Lsn;
 use zenith_utils::pq_proto::SystemId;
+use zenith_utils::pq_proto::ZenithFeedback;
 use zenith_utils::zid::{ZTenantId, ZTimelineId};
 
 pub const SK_MAGIC: u32 = 0xcafeceefu32;
@@ -278,9 +279,8 @@ pub struct AppendResponse {
     // We report back our awareness about which WAL is committed, as this is
     // a criterion for walproposer --sync mode exit
     pub commit_lsn: Lsn,
-    // Min disk consistent lsn of pageservers (portion of WAL applied and written to the disk by pageservers)
-    pub disk_consistent_lsn: Lsn,
     pub hs_feedback: HotStandbyFeedback,
+    pub zenith_feedback: ZenithFeedback,
 }
 
 impl AppendResponse {
@@ -289,8 +289,8 @@ impl AppendResponse {
             term,
             flush_lsn: Lsn(0),
             commit_lsn: Lsn(0),
-            disk_consistent_lsn: Lsn(0),
             hs_feedback: HotStandbyFeedback::empty(),
+            zenith_feedback: ZenithFeedback::empty(),
         }
     }
 }
@@ -395,10 +395,11 @@ impl AcceptorProposerMessage {
                 buf.put_u64_le(msg.term);
                 buf.put_u64_le(msg.flush_lsn.into());
                 buf.put_u64_le(msg.commit_lsn.into());
-                buf.put_u64_le(msg.disk_consistent_lsn.into());
                 buf.put_i64_le(msg.hs_feedback.ts);
                 buf.put_u64_le(msg.hs_feedback.xmin);
                 buf.put_u64_le(msg.hs_feedback.catalog_xmin);
+
+                msg.zenith_feedback.serialize(buf)?
             }
         }
 
@@ -614,9 +615,9 @@ where
             term: self.s.acceptor_state.term,
             flush_lsn: self.flush_lsn,
             commit_lsn: self.s.commit_lsn,
-            disk_consistent_lsn: Lsn(0),
             // will be filled by the upper code to avoid bothering safekeeper
             hs_feedback: HotStandbyFeedback::empty(),
+            zenith_feedback: ZenithFeedback::empty(),
         }
     }
 
--- a/walkeeper/src/timeline.rs
+++ b/walkeeper/src/timeline.rs
@@ -27,6 +27,7 @@ use crate::upgrade::upgrade_control_file;
 use crate::SafeKeeperConf;
 use postgres_ffi::xlog_utils::{XLogFileName, XLOG_BLCKSZ};
 use std::convert::TryInto;
+use zenith_utils::pq_proto::ZenithFeedback;
 
 // contains persistent metadata for safekeeper
 const CONTROL_FILE_NAME: &str = "safekeeper.control";
@@ -35,17 +36,17 @@ const CONTROL_FILE_NAME_PARTIAL: &str = "safekeeper.control.partial";
 const POLL_STATE_TIMEOUT: Duration = Duration::from_secs(1);
 pub const CHECKSUM_SIZE: usize = std::mem::size_of::<u32>();
 
-/// Replica status: host standby feedback + disk consistent lsn
+/// Replica status update + hot standby feedback
 #[derive(Debug, Clone, Copy)]
 pub struct ReplicaState {
     /// last known lsn received by replica
     pub last_received_lsn: Lsn, // None means we don't know
-    /// combined disk_consistent_lsn of pageservers
-    pub disk_consistent_lsn: Lsn,
     /// combined remote consistent lsn of pageservers
     pub remote_consistent_lsn: Lsn,
     /// combined hot standby feedback from all replicas
     pub hs_feedback: HotStandbyFeedback,
+    /// Zenith specific feedback received from pageserver, if any
+    pub zenith_feedback: Option<ZenithFeedback>,
 }
 
 impl Default for ReplicaState {
@@ -58,13 +59,13 @@ impl ReplicaState {
     pub fn new() -> ReplicaState {
         ReplicaState {
             last_received_lsn: Lsn::MAX,
-            disk_consistent_lsn: Lsn(u64::MAX),
-            remote_consistent_lsn: Lsn(u64::MAX),
+            remote_consistent_lsn: Lsn(0),
             hs_feedback: HotStandbyFeedback {
                 ts: 0,
                 xmin: u64::MAX,
                 catalog_xmin: u64::MAX,
             },
+            zenith_feedback: None,
         }
     }
 }
@@ -109,13 +110,40 @@ impl SharedState {
             acc.hs_feedback.xmin = min(acc.hs_feedback.xmin, state.hs_feedback.xmin);
             acc.hs_feedback.catalog_xmin =
                 min(acc.hs_feedback.catalog_xmin, state.hs_feedback.catalog_xmin);
-            acc.disk_consistent_lsn = Lsn::min(acc.disk_consistent_lsn, state.disk_consistent_lsn);
-            // currently not used, but update it to be consistent
-            acc.last_received_lsn = Lsn::min(acc.last_received_lsn, state.last_received_lsn);
-            // When at least one replica has preserved data up to remote_consistent_lsn,
-            // safekeeper is free to delete it, so chose max of all replicas.
-            acc.remote_consistent_lsn =
-                Lsn::max(acc.remote_consistent_lsn, state.remote_consistent_lsn);
+
+            // FIXME
+            // If multiple pageservers are streaming WAL and send feedback for the same timeline simultaneously,
+            // this code is not correct.
+            // Now the most advanced feedback is used.
+            // If one pageserver lags when another doesn't, the backpressure won't be activated on compute and lagging
+            // pageserver is prone to timeout errors.
+            //
+            // To choose what feedback to use and resend to compute node,
+            // we need to know which pageserver compute node considers to be main.
+            // See https://github.com/zenithdb/zenith/issues/1171
+            //
+            if let Some(zenith_feedback) = state.zenith_feedback {
+                if let Some(acc_feedback) = acc.zenith_feedback {
+                    if acc_feedback.ps_writelsn < zenith_feedback.ps_writelsn {
+                        warn!("More than one pageserver is streaming WAL for the timeline. Feedback resolving is not fully supported yet.");
+                        acc.zenith_feedback = Some(zenith_feedback);
+                    }
+                } else {
+                    acc.zenith_feedback = Some(zenith_feedback);
+                }
+
+                // last lsn received by pageserver
+                // FIXME if multiple pageservers are streaming WAL, last_received_lsn must be tracked per pageserver.
+                // See https://github.com/zenithdb/zenith/issues/1171
+                acc.last_received_lsn = Lsn::from(zenith_feedback.ps_writelsn);
+
+                // When at least one pageserver has preserved data up to remote_consistent_lsn,
+                // safekeeper is free to delete it, so choose max of all pageservers.
+                acc.remote_consistent_lsn = max(
+                    Lsn::from(zenith_feedback.ps_applylsn),
+                    acc.remote_consistent_lsn,
+                );
+            }
         }
         acc
     }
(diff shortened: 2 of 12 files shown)


kelvich:
Ok, let's add our custom `z` response type to the protocol. We have enough space in Reply/Feedback messages to fit size in (4 bytes replyTime, 16 bytes for xmin and catalogXmin), but that would be confusing. I would rather add some connection option in callmemaybe to indicate that caller is capable of handling that message types.

> Now setup with direct safekeeper->compute_node

you meant compute->pageserver?

> write_lsn = last_lsn_received by_pageserver;
> flush_lsn = write_lsn;
> apply_lsn = disk_consistent_lsn; // used for backpressure

So `disk_consistent_lsn` is used for backpressure and `write_lsn=flush_lsn=last_lsn_received by_pageserver`. Does safekeeper or compute use `write_lsn/flush_lsn`?

lubennikovaav:
> I would rather add some connection option in callmemaybe to indicate that caller is capable of handling that message types.

Is there any kind of callerthat will use callmemaybe and not use our extended protocol?

> > Now setup with direct safekeeper->compute_node
> 
> you meant compute->pageserver?
> 

Yep, that's a typo.

> > write_lsn = last_lsn_received by_pageserver;
> > flush_lsn = write_lsn;
> > apply_lsn = disk_consistent_lsn; // used for backpressure
> 
> So `disk_consistent_lsn` is used for backpressure and `write_lsn=flush_lsn=last_lsn_received by_pageserver`. Does safekeeper or compute use `write_lsn/flush_lsn`?

`write_lsn` is used by safekeeper to check if pageserver is caught up.
`flush_lsn` is used by `synchronous_commit`. In pageserver's reply flush_lsn it is the same as write_lsn.
In safekeeper's reply `flush_lsn` means "Locally flushed part of WAL".

walkeeper/src/safekeeper.rs
```diff
@@ -282,6 +282,8 @@ pub struct AppendResponse {
     pub commit_lsn: Lsn,
     // Min disk consistent lsn of pageservers (portion of WAL applied and written to the disk by pageservers)
     pub disk_consistent_lsn: Lsn,
+    // instance size of the timeline
+    pub instance_size: u64,
```
kelvich, line 286: That breaks protocol =/ @arssher @petuhovskiy may be switch hs_feedback to some binary json format so we can add new fields easily?

kelvich: Also `instance_size` sounds like a static limit value. May be `current_size` (we also have a total one which includes sliced WAL) is better?

lubennikovaav: It is already a custom message, isn't it?
AppendResponse on postgres side is handled by `HandleWalKeeperResponse()`

zenith_utils/src/pq_proto.rs
```diff
@@ -768,3 +770,14 @@ impl<'a> BeMessage<'a> {
...
     }
 }
+
+// Zenith extension of postgres replication protocol
+// See ZENITH_STATUS_UPDATE_TAG_BYTE
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ZenithStatusUpdate {
+    // The last LSN that is synced to remote storage and is guaranteed to survive pageserver crash
+    // Used by safekeepers to remove WAL preceding `remote_consistent_lsn`.
+    pub remote_consistent_lsn: Lsn,
```
kelvich, line 780: I still don't get why can't we put lsn in a usual feedback message to keep them together. IMO it would make more sense to move backpressure-related lsn here since it is semantically different from usual ones

kelvich:
> > I would rather add some connection option in callmemaybe to indicate that caller is capable of handling that message types.
> 
> Is there any kind of caller that will use callmemaybe and won't use our extended protocol?
>

Postgres when connected directly to the pageserver

kelvich:
We've chatted a bit with @lubennikovaav about formats and lsns here. Short summary:

* 3 LSN in StandbyReplyMessage are enough. Let's make they be last_receiver_lsn / disk_consistent_lsn / s3_consistent_lsn. If we need more lsn's we may put them to zenith_response
* that should work both for postgres<->pageserver and for safekeeper<->pageserver
* postgres<->safekeeper comms use they own protocol and won't be affected
* zenith_response should be flat key-value string format like -- like ErrorResponse in libpq, but let's use keys as strings (instead of bytes). That way we can easily parse it in postgres code and maintain backward-compatibility. * makes sense to switch hs_feedback in safekeeper's AppendResponse to zenith_response -- that way we can proxy pass it, enrich with some other key and maintain compatibility.

cc @arssher @petuhovskiy

arssher:
Ok.

lubennikovaav:
CI failute is unrelated to this PR and covered in issue #1068

pytest.ini
```diff
@@ -6,4 +6,4 @@ markers =
 minversion = 6.0
 log_format = %(asctime)s.%(msecs)-3d %(levelname)s [%(filename)s:%(lineno)d] %(message)s
 log_date_format = %Y-%m-%d %H:%M:%S
-log_cli = true
+log_cli = false
```
petuhovskiy, line 9: Accidentally?

lubennikovaav: yes
I reverted it.

zenith_utils/src/pq_proto.rs
```diff
@@ -892,7 +907,12 @@ impl ZenithFeedback {
                     }
                 }
                 _ => {
-                    info!("ZenithFeedback parse. unknown key {}", key)
+                    let len = buf.get_i32();
+                    info!(
+                        "ZenithFeedback parse. unknown key {} of len {}. Skip it.",
+                        key, len
+                    );
+                    buf.advance(len as usize);
```
petuhovskiy, line 915: [advance](https://docs.rs/bytes/0.4.7/bytes/buf/trait.Buf.html#tymethod.advance) takes cnt in bytes, not bits, so I'd pass len as bytes count in serialize(). I also think it's better to add unit tests here with encode/decode in different cases: with and without unknown fields.

petuhovskiy (approved):

walkeeper/src/timeline.rs
```diff
@@ -109,13 +115,18 @@ impl SharedState {
...
                 min(acc.hs_feedback.catalog_xmin, state.hs_feedback.catalog_xmin);
-            acc.disk_consistent_lsn = Lsn::min(acc.disk_consistent_lsn, state.disk_consistent_lsn);
-            // currently not used, but update it to be consistent
-            acc.last_received_lsn = Lsn::min(acc.last_received_lsn, state.last_received_lsn);
-            // When at least one replica has preserved data up to remote_consistent_lsn,
-            // safekeeper is free to delete it, so chose max of all replicas.
-            acc.remote_consistent_lsn =
-                Lsn::max(acc.remote_consistent_lsn, state.remote_consistent_lsn);
+
+            // Here we assume that only one active pageserver
```
LizardWizzard, line 119: What is the definition of "active" here? How this plays with tenant relocation/possible second pageserver for failover? Why we cant continue using value calculated across all replicas?

zenith_utils/src/pq_proto.rs
```diff
@@ -798,3 +813,156 @@ impl<'a> BeMessage<'a> {
         Ok(())
     }
 }
+
+// Zenith extension of postgres replication protocol
+// See ZENITH_STATUS_UPDATE_TAG_BYTE
+#[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
+pub struct ZenithFeedback {
+    // Last known size of the timeline. Used to enforce timeline size limit.
+    pub current_instance_size: u64,
```
LizardWizzard, line 822: Nit: If the only meaning is timeline size, can we rename it to something like current_timeline_size? IMO instance size is too broad

zenith_utils/src/pq_proto.rs
```diff
@@ -798,3 +813,156 @@ impl<'a> BeMessage<'a> {
...
+    // Parts of StandbyStatusUpdate we resend to compute via safekeeper
+    pub ps_writelsn: u64,
+    pub ps_applylsn: u64,
+    pub ps_flushlsn: u64,
+    pub ps_replytime: SystemTime,
+}
+pub const ZENITH_FEEDBACK_FIELDS_NUMBER: u8 = 5;
+
+impl ZenithFeedback {
+    pub fn empty() -> ZenithFeedback {
```
LizardWizzard, line 832: This shouldn't be needed it enum is used to store it in the `ReplicaState`. In its current form It looks like a placeholder value which possible can be misinterpreted, so I think solution with enum is preferable

zenith_utils/src/pq_proto.rs
```diff
@@ -798,3 +813,156 @@ impl<'a> BeMessage<'a> {
...
+        write_cstr(&Bytes::from("ps_flushlsn"), buf)?;
+        buf.put_i32(8);
+        buf.put_u64(self.ps_flushlsn);
+        write_cstr(&Bytes::from("ps_applylsn"), buf)?;
+        buf.put_i32(8);
+        buf.put_u64(self.ps_applylsn);
+
+        let timestamp = match self.ps_replytime.duration_since(*PG_EPOCH) {
+            Ok(d) => d.as_micros() as i64,
+            Err(e) => -(e.duration().as_micros() as i64),
```
LizardWizzard, line 861: Does it mean that we can be before PG_EPOCH? I think in that case we can bubble up the error but I cant imagine how can we actually trigger this case

zenith_utils/src/pq_proto.rs
```diff
@@ -798,3 +813,156 @@ impl<'a> BeMessage<'a> {
...
+pub struct ZenithFeedback {
+    // Last known size of the timeline. Used to enforce timeline size limit.
+    pub current_instance_size: u64,
+    // Parts of StandbyStatusUpdate we resend to compute via safekeeper
+    pub ps_writelsn: u64,
+    pub ps_applylsn: u64,
+    pub ps_flushlsn: u64,
+    pub ps_replytime: SystemTime,
+}
+pub const ZENITH_FEEDBACK_FIELDS_NUMBER: u8 = 5;
```
LizardWizzard, line 829: Lets and some note here. Something like: 
```
NOTE: Do not forget to increment this number when adding new fields to ZenithFeedback. Do not remove previously available fields because this might be backwards incompatible.
```

zenith_utils/src/pq_proto.rs
```diff
@@ -798,3 +813,156 @@ impl<'a> BeMessage<'a> {
...
+
+    pub fn parse(mut buf: Bytes) -> ZenithFeedback {
+        let mut zf = ZenithFeedback::empty();
+        let nfields = buf.get_u8();
+        let mut i = 0;
+        while i < nfields {
+            i += 1;
+            let key_cstr = read_null_terminated(&mut buf).unwrap();
+            let key = cstr_to_str(&key_cstr).unwrap();
+            match key {
```
LizardWizzard, line 878: Does it make sense to validate that all values have been successfully initialized with correct values? It might cause issues if some of the fields are missing and then someone assumes that all of them were initialized and have non zero values.

LizardWizzard (commented):
Concept looks good to me, though I'm a bit concerned about custom binary format and possible issues with it. May it be an option to use similar encoding like for parameters inside startup packet? Everything is optional and we can deserialize without giving out possibly partially initialized state. Code for managing this encoding is already available in pg backend on the rust side and may be reachable on the postgres side too. Please correct me if I'm wrong

walkeeper/src/timeline.rs
lubennikovaav: "Active" here means pageserver that is connected to the safekeeper and sends the feedback to it.
I'll reword the comment to clarify that.

If there is more than one active pageserver, we must choose what feedback to pass to the compute node.
So we need to identify the pageserver to which the compute node is connected, and use its feedback.

`remote_consistent_lsn` can be max of all values, though.
As for 'last_received_lsn' we need to track it for each pageserver separately, to ensure that they are caught up. 
I recall, your PR addressed that.

zenith_utils/src/pq_proto.rs
lubennikovaav: Well, there's not much to validate.

Feedback messages also serve as keepalive responses, so walsender may send empty values before something was actually replicated. Postgres walsender fine with receiving invalid lsns in a feedback too:
```
/*
	 * The xlog locations that have been written, flushed, and applied by
	 * standby-side. These may be invalid if the standby-side has not offered
	 * values yet.
	 */
```

If some fields are missing, it may be just a different version of the protocol.
Do you think it's worth adding an API version number to the message to control this?

zenith_utils/src/pq_proto.rs
lubennikovaav: We do need some default feedback values to fill AppendResponse before any real feedback is received from pageserver's walreceiver.
These 0 values are interpreted as invalid by Postgres walsender. See comment below

zenith_utils/src/pq_proto.rs
lubennikovaav: It is a copy-paste it from rust-postgres `standby_status_update()` implementation.
But, I don't see how can we get here, either.
So I've changed it to throwing error.

lubennikovaav:
> Concept looks good to me, though I'm a bit concerned about custom binary format and possible issues with it. May it be an option to use similar encoding like for parameters inside startup packet? Everything is optional and we can deserialize without giving out possibly partially initialized state. Code for managing this encoding is already available in pg backend on the rust side and may be reachable on the postgres side too. Please correct me if I'm wrong

startup packet uses null-terminated strings to pass key-value pairs and converting lsns to Cstrings and back is quite a mess.
Besides, there's not much code to reuse - parsing is scattered around different functions and not even close to a set of clear ser-deser functions.

So, I'd go with a custom fornat. it's clear how to read and write it on both rust and postgres sides.

zenith_utils/src/pq_proto.rs
```diff
@@ -798,3 +813,160 @@ impl<'a> BeMessage<'a> {
...
+        buf.put_i32(8);
+        buf.put_u64(self.ps_flushlsn);
+        write_cstr(&Bytes::from("ps_applylsn"), buf)?;
+        buf.put_i32(8);
+        buf.put_u64(self.ps_applylsn);
+
+        let timestamp = self
+            .ps_replytime
+            .duration_since(*PG_EPOCH)
+            .unwrap()
```
LizardWizzard, line 865: Let's use expect here instead of unwrap

zenith_utils/src/pq_proto.rs
```diff
@@ -798,3 +813,160 @@ impl<'a> BeMessage<'a> {
...
+                        zf.ps_replytime = *PG_EPOCH - Duration::from_micros(-raw_time as u64);
+                    }
+                }
+                _ => {
+                    let len = buf.get_i32();
+                    info!(
+                        "ZenithFeedback parse. unknown key {} of len {}. Skip it.",
+                        key, len
+                    );
+                    println!("{}", key);
```
LizardWizzard, line 919: I think log message should be enough

zenith_utils/src/pq_proto.rs
```diff
@@ -7,10 +7,14 @@ use byteorder::{BigEndian, ByteOrder};
 use byteorder::{ReadBytesExt, BE};
 use bytes::{Buf, BufMut, Bytes, BytesMut};
 // use postgres_ffi::xlog_utils::TimestampTz;
```
LizardWizzard, line 9: While we are here, I think it can be removed

LizardWizzard (approved):

LizardWizzard:
> So, I'd go with a custom fornat. it's clear how to read and write it on both rust and postgres sides.

Agreed. Thats indeed a better option then

PR #1043, opened 2021-12-22
Failpoints

knizhnik:
See https://discord.com/channels/869525774699462656/922518319637090354

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -126,7 +126,7 @@ jobs:
             fi
 
             export CARGO_INCREMENTAL=0
-            "${cov_prefix[@]}" cargo build $CARGO_FLAGS --bins --tests
+            "${cov_prefix[@]}" cargo build $CARGO_FLAGS --features fail/failpoints --bins --tests
 
       - save_cache:
           name: Save rust cache
--- a/control_plane/src/local_env.rs
+++ b/control_plane/src/local_env.rs
@@ -72,6 +72,18 @@ pub struct PageServerConf {
 
     // jwt auth token used for communication with pageserver
     pub auth_token: String,
+
+    // GC activation period
+    pub gc_period: String,
+
+    // GC horizon
+    pub gc_horizon: u64,
+
+    // Checkpoint period
+    pub checkpoint_period: String,
+
+    // Checkpoint distance
+    pub checkpoint_distance: u64,
 }
 
 impl Default for PageServerConf {
@@ -81,6 +93,10 @@ impl Default for PageServerConf {
             http_port: 0,
             auth_type: AuthType::Trust,
             auth_token: "".to_string(),
+            gc_period: "1s".to_string(),
+            gc_horizon: 64 * 1024,
+            checkpoint_period: "1s".to_string(),
+            checkpoint_distance: 1024 * 1024,
         }
     }
 }
--- a/control_plane/src/local_env.rs
+++ b/control_plane/src/local_env.rs
@@ -93,10 +93,10 @@ impl Default for PageServerConf {
             http_port: 0,
             auth_type: AuthType::Trust,
             auth_token: "".to_string(),
-            gc_period: "1s".to_string(),
-            gc_horizon: 64 * 1024,
+            gc_period: "100s".to_string(),
+            gc_horizon: 64 * 1024 * 1024,
             checkpoint_period: "1s".to_string(),
-            checkpoint_distance: 1024 * 1024,
+            checkpoint_distance: 256 * 1024 * 1024,
         }
     }
 }
--- a/pageserver/src/bin/pageserver.rs
+++ b/pageserver/src/bin/pageserver.rs
@@ -10,6 +10,7 @@ use std::{
     str::FromStr,
     thread,
 };
+
 use tracing::*;
 use zenith_utils::{auth::JwtAuth, logging, postgres_backend::AuthType, tcp_listener, GIT_VERSION};
 
@@ -28,6 +29,7 @@ use zenith_utils::shutdown::exit_now;
 use zenith_utils::signals::{self, Signal};
 
 use const_format::formatcp;
+use fail::FailScenario;
 
 /// String arguments that can be declared via CLI or config file
 #[derive(Serialize, Deserialize, PartialEq, Eq, Clone)]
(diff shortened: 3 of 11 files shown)


control_plane/src/local_env.rs
```diff
@@ -72,6 +72,18 @@ pub struct PageServerConf {
...
+    pub gc_period: String,
+
+    // GC horizon
+    pub gc_horizon: u64,
+
+    // Checkpoint period
+    pub checkpoint_period: String,
+
+    // Checkpoint distance
+    pub checkpoint_distance: u64,
```
funbringer, line 86: Looks like this change is unrelated to failpoints. Could you extract this into a separate patch?

pageserver/src/bin/pageserver.rs
```diff
@@ -488,6 +490,13 @@ fn main() -> Result<()> {
     // as a ref.
     let conf: &'static PageServerConf = Box::leak(Box::new(conf));
 
+    let scenario = FailScenario::setup();
+    if fail::has_failpoints() {
+        std::panic::set_hook(Box::new(|_| {
+            std::process::exit(1);
+        }));
+    }
+
```
lubennikovaav, line 499: I don't quite understand the purpose of this code.
`FailScenario::setup()` configures fail points specified in the `FAILPOINTS` environment variable. 
As far as I see it is never used in the tests.

Have you set it manually?

test_runner/fixtures/zenith_fixtures.py
```diff
@@ -431,6 +431,10 @@ def __init__(self, config: ZenithEnvBuilder):
 pg_port = {pageserver_port.pg}
 http_port = {pageserver_port.http}
 auth_type = '{pageserver_auth_type}'
+gc_period = '1s'
+gc_horizon = 67108864
+checkpoint_period = '1s'
+checkpoint_distance = 1048576
```
lubennikovaav, line 437: Using non-default settings in all tests doesn't look right to me.
Let's tune them for one test, if necessary. Or change defaults if we believe that they are bad.

test_runner/batch_others/test_recovery.py
```diff
@@ -0,0 +1,51 @@
...
+                    # Sleep for some time to let checkpoint create image layers
+                    time.sleep(2)
+
+                    # Configure failpoints
+                    pscur.execute(
+                        "failpoints checkpoint-before-sync=sleep(2000);checkpoint-after-sync=panic")
+
+                    # Do some updates until pageserver is crashed
+                    try:
+                        while True:
```
lubennikovaav, line 39: Maybe add some timeout in case checkpointer fails to fail?

control_plane/src/local_env.rs
```diff
@@ -81,6 +93,10 @@ impl Default for PageServerConf {
             http_port: 0,
             auth_type: AuthType::Trust,
             auth_token: "".to_string(),
+            gc_period: "100s".to_string(),
+            gc_horizon: 64 * 1024 * 1024,
+            checkpoint_period: "1s".to_string(),
+            checkpoint_distance: 256 * 1024 * 1024,
```
lubennikovaav, line 99: Why not use DEFAULT_CHECKPOINT_PERIOD and other `defaults` here?

.circleci/config.yml
```diff
@@ -126,7 +126,7 @@ jobs:
             fi
 
             export CARGO_INCREMENTAL=0
-            "${cov_prefix[@]}" cargo build $CARGO_FLAGS --bins --tests
+            "${cov_prefix[@]}" cargo build $CARGO_FLAGS --features fail/failpoints --bins --tests
```
lubennikovaav, line 129: Do we need a similar change for `cargo build` in Makefile?

lubennikovaav (changes requested):

pageserver/src/bin/pageserver.rs
funbringer: I suspect @knizhnik wanted to make it testable/usable in manual mode while developing this feature. Either way, I'd leave it be since it's useful.

test_runner/fixtures/zenith_fixtures.py
funbringer: I'd argue the change doesn't belong to this PR.

knizhnik: This changes are needed for test_recovery.py test included in this PR.

pageserver/src/bin/pageserver.rs
knizhnik: The purpose of this code is to cause pageserver termination in case of panic. `failpoint` module doesn't provide some builtin action to terminate process. It can only thrown panic which terminates just current thread. And if we want to test abnormal process termination we need to terminate the whole process.

control_plane/src/local_env.rs
knizhnik: This changes are needed for test_recovery.py test included in this PR.

test_runner/batch_others/test_recovery.py
knizhnik: It seems to be no normal reasons for pageserver to "fail to fail". If it happen, then it is bug as hanging of any other test. There is CI timeout to terminate test run in this case. Do we need some extra timeout in each test? Certainly if this test hangs and CI run is terminated by timeout, then we will not get results for other tests which are not launched. If it will happen frequently, then it may be a problem and we really need to somehow explicitly prevent hanging in  test_recovery test. By right now I prefer not to solve this problem.

control_plane/src/local_env.rs
knizhnik: 1 sec and 256Mb seems to be default checkpoint values.
I just prefer to specify them explicitly in this file, because setting for tests may be different with one used in productions. So not want that checkpoint interval for tests depends on default setting for production.

.circleci/config.yml
knizhnik: I do not want failpoints to be linked in production. It should be used only for tests. When we are doing "cargo build" there is no information whether we are doing it to run tests or we are building images for production. Certainly we can create special make target for building zenith with safepoints. For me  manual specification of `--features fail/failpoints` works fine. Is there any other way except creating special make target for  builds with safepoints?

Including safepoints only in debug builds IMHO is not also good option, because debug version is soooo slow.

control_plane/src/local_env.rs
funbringer: I'm trying to say that IMO failpoints infra patch should include neither the test nor those settings.

knizhnik: IMHO it is strange to commit some feature without test.

lubennikovaav:
I've opened a new PR with rebased version of this one: #1324

PR #1045, opened 2021-12-22
Use parking_lot::Mutex instead of std::Mutex in walreceiver

bojanserafimov:
Making progress towards https://github.com/zenithdb/zenith/issues/956

Note that `parking_lot` mutexes don't poison. If a thread panics while holding a mutex, the mutex is **released**, without crashing the next thread to ask for it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1146,6 +1146,7 @@ dependencies = [
  "log",
  "nix",
  "once_cell",
+ "parking_lot",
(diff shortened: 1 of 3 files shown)


kelvich:
Huh, so that is fully API-compatible -- I expected more problems) Let's switch all mutexes to parking_lot then

funbringer (approved):

PR #1048, opened 2021-12-23
Refactor Postgres command parsing in safekeeper.

arssher:
Do it separately with SafekeeperPostgresCommand enum as a result. Since query is
always C string, switch postgres_backend process_query argument from Bytes to
&str.

Make passing ztli/ztenant id in safekeeper connection string optional; this is
needed for upcoming intra-safekeeper heartbeat cmd which is not bound to any
timeline.

### DIFF
--- a/walkeeper/src/handler.rs
+++ b/walkeeper/src/handler.rs
@@ -1,16 +1,18 @@
 //! Part of Safekeeper pretending to be Postgres, i.e. handling Postgres
 //! protocol commands.
 
-use crate::json_ctrl::handle_json_ctrl;
+use crate::json_ctrl::{handle_json_ctrl, AppendLogicalMessage};
 use crate::receive_wal::ReceiveWalConn;
 use crate::send_wal::ReplicationConn;
 use crate::timeline::{Timeline, TimelineTools};
 use crate::SafeKeeperConf;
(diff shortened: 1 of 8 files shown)


walkeeper/src/handler.rs
```diff
@@ -53,51 +91,52 @@ impl postgres_backend::Handler for SafekeeperPostgresHandler {
...
+            SafekeeperPostgresCommand::StartWalPush { .. }
+            | SafekeeperPostgresCommand::StartReplication { .. }
+            | SafekeeperPostgresCommand::IdentifySystem
+            | SafekeeperPostgresCommand::JSONCtrl { .. } => {
+                let tenantid = self
+                    .ztenantid
+                    .ok_or_else(|| anyhow!("tenantid is required"))?;
+                let timelineid = self
+                    .ztimelineid
+                    .ok_or_else(|| anyhow!("timelineid is required"))?;
```
lubennikovaav, line 108: So, now ztenatntid and ztimelineid are required for all 4 commands, right?
It seems that this code can be simplified.

arssher: Yes, but I'm going to change this soon.

lubennikovaav (approved):

PR #1060, opened 2021-12-27
Parse cancel message in pq_proto

bojanserafimov:
Implmeent cancel message parsing for https://github.com/zenithdb/zenith/pull/1052

In order to implement parsing for `CancelRequest` I needed to refactor a bit:
1. Add `backend_pid` and `cancel_key` args of `CancelRequest`
2. Remove `version` and `params` args of `CancelRequest`, making `FeStartupMessage` an enum instead of struct
3. Rename `FeStartupMessage` to `FeInitialMessage` to avoid overloading a pg term and causing confusion. See pg [docs](https://www.postgresql.org/docs/14/protocol-message-formats.html) for reference

I recommend first reviewing `pub enum FeInitialMessage` and the parsing logic for it. Everything else is just refactor.


zenith_utils/src/pq_proto.rs
```diff
@@ -37,19 +38,13 @@ pub enum FeMessage {
...
-    pub params: HashMap<String, String>,
-}
-
-#[derive(Debug)]
-pub enum StartupRequestCode {
-    Cancel,
-    NegotiateSsl,
-    NegotiateGss,
-    Normal,
+pub enum FeInitialMessage {
```
patins, line 41: Consider `CancelRequest { backend_pid: u32, cancel_key: u32, }`
Same for `StartupMessage`

zenith_utils/src/pq_proto.rs
```diff
@@ -169,44 +164,53 @@ impl FeStartupMessage {
...
+        let message = match request_code {
+            CANCEL_REQUEST_CODE => {
+                let parse_i32_big_endian = |byte_slice: &[u8]| {
+                    let byte_array: [u8; 4] = byte_slice.try_into().unwrap();
+                    i32::from_be_bytes(byte_array)
+                };
+
+                ensure!(params_len == 8, "expected 8 bytes for CancelRequest params");
+                let backend_pid = parse_i32_big_endian(&params_bytes[0..4]);
+                let cancel_key = parse_i32_big_endian(&params_bytes[4..8]);
```
patins, line 184: Try using `byteorder`'s `read_u32::<BE>` on the buffer (might need to wrap with a `Cursor`?)

zenith_utils/src/pq_proto.rs
```diff
@@ -169,44 +164,53 @@ impl FeStartupMessage {
...
+                FeInitialMessage::CancelRequest(backend_pid, cancel_key)
+            }
+            NEGOTIATE_SSL_CODE => FeInitialMessage::SSLRequest,
+            NEGOTIATE_GSS_CODE => FeInitialMessage::GSSENCRequest,
+            _ => {
```
patins, line 189: Is there an explicit code for startup message? If so, I think we should error out in the catch all case

zenith_utils/src/pq_proto.rs
```diff
@@ -169,44 +164,53 @@ impl FeStartupMessage {
...
+                // Then null-terminated (String) pairs of param name / param value go.
+                let params_str = str::from_utf8(&params_bytes).unwrap();
+                let params = params_str.split('\0');
+                let mut params_hash: HashMap<String, String> = HashMap::new();
+                for pair in params.collect::<Vec<_>>().chunks_exact(2) {
+                    let name = pair[0];
+                    let value = pair[1];
```
patins, line 196: I think you can simplify this with something like
```
while let Some(name) = params.next() {
    let value = params.next().ok_or_else(|| anyhow!("bad params"))?;
    //insert into params_hash
}

File: zenith_utils/src/pq_proto.rs (hunk 3 above)
Comment (bojanserafimov) [reply zenith_utils/src/pq_proto.rs:189]: There isn't https://www.postgresql.org/docs/14/protocol-message-formats.html

For startup message the code is the version, and for the other messages the code is designed not to collide with any version.

I can leave a comment to clarify.

Comment (bojanserafimov) [reply zenith_utils/src/pq_proto.rs:196]: Hm actually I can parse the major and minor versions and only if they make sense, conclude that we have a StartupMessage. It's less error-prone

File: zenith_utils/src/pq_proto.rs (hunk 4 above)
Comment (bojanserafimov) [reply zenith_utils/src/pq_proto.rs]: It makes the code shorter but turns the iterator into a mutable var, which adds a different kind of complexity. How about `for (name, value) in params.tuples()`? Is `Itertools` [stable enough](https://github.com/rust-itertools/itertools/blob/master/CHANGELOG.md) to depend on?

File: zenith_utils/src/pq_proto.rs [hunk 5]
```diff
@@ -37,19 +37,18 @@ pub enum FeMessage {
...
-#[derive(Debug)]
-pub enum StartupRequestCode {
-    Cancel,
-    NegotiateSsl,
-    NegotiateGss,
-    Normal,
+pub enum FeInitialMessage {
+    CancelRequest {
+        backend_pid: i32,
+        cancel_key: i32,
```
patins, line 43: Do you think these make more sense as i32s or u32s? (Also applies to #1063)

zenith_utils/src/pq_proto.rs
```diff
@@ -147,13 +146,14 @@ impl FeMessage {
...
 
-impl FeStartupMessage {
+impl FeInitialMessage {
     /// Read startup message from the stream.
     pub fn read(stream: &mut impl std::io::Read) -> anyhow::Result<Option<FeMessage>> {
         const MAX_STARTUP_PACKET_LENGTH: usize = 10000;
-        const CANCEL_REQUEST_CODE: u32 = (1234 << 16) | 5678;
-        const NEGOTIATE_SSL_CODE: u32 = (1234 << 16) | 5679;
-        const NEGOTIATE_GSS_CODE: u32 = (1234 << 16) | 5680;
+        const RESERVED_INVALID_MAJOR_VERSION: u32 = 1234;
```
patins, line 153: Nice!

zenith_utils/src/pq_proto.rs
patins: Seems like a cool crate, but is there a way to use it so we can detect `num_params % 2 == 1` with it? Also adding new dependencies is :(

zenith_utils/src/pq_proto.rs
```diff
@@ -169,44 +169,58 @@ impl FeStartupMessage {
...
+        // Parse params depending on request code
+        let most_sig_16_bits = request_code >> 16;
+        let least_sig_16_bits = request_code & ((1 << 16) - 1);
+        let message = match (most_sig_16_bits, least_sig_16_bits) {
+            (RESERVED_INVALID_MAJOR_VERSION, CANCEL_REQUEST_CODE) => {
+                ensure!(params_len == 8, "expected 8 bytes for CancelRequest params");
+                FeInitialMessage::CancelRequest {
+                    backend_pid: byteorder::BigEndian::read_i32(&params_bytes[0..4]),
+                    cancel_key: byteorder::BigEndian::read_i32(&params_bytes[4..8]),
+                }
```
patins, line 188: I think you might be able to make this cleaner with something like

```
let cursor = Cursor::new(param_bytes);
cursor.read_i32::<BE>()
```

I'm not 100% sure on that exact syntax though

zenith_utils/src/pq_proto.rs
bojanserafimov: Postgres keeps them in i32 so I'm copying that. For cancel_key it technically doesn't matter since we're only sending/receiving it in binary, but the pid we get from a pg query so it matters.

bojanserafimov:
I wonder what's making tests fail. My last 2 changes are refactors, and I have 1 pass and 2 different fails

zenith_utils/src/pq_proto.rs
bojanserafimov: Nope :(

So, I wasn't planning on changing this code (my change is just indentation) but now I notice that `chunks_exact` just means "if there's leftover, we'll keep that in the iterator so you can check later yourself", and we're not checking that now. I think your suggestion is the simplest way to actually check that :)

zenith_utils/src/pq_proto.rs
```diff
@@ -169,44 +172,60 @@ impl FeStartupMessage {
...
+                bail!("Unrecognized request code {}", unrecognized_code)
+            }
+            (major_version, minor_version) => {
+                // TODO bail if protocol major_version is not 3?
+                // Parse null-terminated (String) pairs of param name / param value
+                let params_str = str::from_utf8(&params_bytes).unwrap();
+                let mut params_tokens = params_str.split('\0');
+                let mut params: HashMap<String, String> = HashMap::new();
+                while let Some(name) = params_tokens.next() {
+                    let value = params_tokens.next().ok_or_else(|| {
```
funbringer, line 206: We already have itertools in Cargo.lock, which has a nice method [`tuple_windows`](https://docs.rs/itertools/0.10.3/itertools/trait.Itertools.html#method.tuple_windows). Perhaps you could use it.

zenith_utils/src/pq_proto.rs
```diff
@@ -37,19 +37,21 @@ pub enum FeMessage {
...
 #[derive(Debug)]
-pub struct FeStartupMessage {
-    pub version: u32,
-    pub kind: StartupRequestCode,
-    // optional params arriving in startup packet
-    pub params: HashMap<String, String>,
+pub enum FeInitialMessage {
+    CancelRequest(CancelKeyData),
+    SSLRequest,
+    GSSENCRequest,
```
funbringer, line 43: nit: I'd spell those as `SslRequest` and `GssEncRequest`, cause it's easier on the eyes. I know this is how they're spelled in the docs, but I'm not sure we're after one-to-one correspondence here.

funbringer (approved):
LGTM (there's a couple of nits).

Speaking of `StartupMessage` -> `InitialMessage` rename, I think PG's naming is somewhat inconsistent, since the function handling this packet is called `ProcessStartupPacket` (which is quite similar to `StartupMessage`), yet there's a whole lot of different message types that are also covered by subsequent calls to `processCancelRequest` etc.

zenith_utils/src/pq_proto.rs
funbringer: And by the way, there's also https://docs.rs/itertools/0.10.3/itertools/trait.Itertools.html#method.next_tuple, which means we'll keep an iterator in case we wanted to check for leftovers.

bojanserafimov:
> G's na

I'd be fine with renaming `InitialMessage` to `StartupPacket`, if that's what you mean. It's different enough from `StartupMessage` to avoid confusion, and also sounds like the right input to a function `ProcessStartupPacket` or `handle_startup` (in `proxy.rs`).

funbringer:
@bojanserafimov 

> I'd be fine with renaming InitialMessage to StartupPacket, if that's what you mean.

No, there's no need, I guess. Both names are fine. I just wanted to point out that PG's naming is already confusing :)

PR #1069, opened 2021-12-28
Fix wal receiver shutdown

LizardWizzard:
This patch allows to shutdown wal receiver when there are no messages
and wal receiver is blocked inside tokio-postgres. In this case it
cannot check the shutdown flag.

This patch switches to use async interface of tokio-postgres directly
without sync wrappers. It opens the possibility to use tokio::select!
between the phsycal_stream.next() and a shutdown channel readiness to
interrupt replication process.

Also this allows to shutdown only particular wal receiver without
using global shutdown_requested flag.

### DIFF
--- a/pageserver/src/walreceiver.rs
+++ b/pageserver/src/walreceiver.rs
@@ -13,9 +13,6 @@ use crate::walingest::WalIngest;
 use anyhow::{bail, Context, Error, Result};
 use lazy_static::lazy_static;
 use parking_lot::Mutex;
-use postgres::fallible_iterator::FallibleIterator;
-use postgres::replication::ReplicationIter;
-use postgres::{Client, NoTls, SimpleQueryMessage, SimpleQueryRow};
 use postgres_ffi::waldecoder::*;
 use postgres_protocol::message::backend::ReplicationMessage;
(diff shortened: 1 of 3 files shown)


pageserver/src/walreceiver.rs
```diff
@@ -218,14 +264,24 @@ fn walreceiver_main(
...
-    while let Some(replication_message) = physical_stream.next()? {
+    while let Some(replication_message) = runtime.block_on(async {
+        tokio::select! {
+            replication_message = physical_stream.next() => {replication_message},
+            _ = &mut interrupt_receiver => {
+                info!("walreceiver interrupted");
+                None
+            }
+        }
+    }) {
```
funbringer, line 283: I wish we had a `CancellableReplicationStream: Iterator` which would hide all of this complexity, but it occurs to me that's too much to ask for.

funbringer (approved):
LGTM

pageserver/src/walreceiver.rs
LizardWizzard: This feature looks generic enough and I've found a crate for that: https://github.com/jonhoo/stream-cancel Not sure that we should bring it for just one use case

PR #1081, opened 2022-01-03
Generate safekeeper unique IDs.

arssher:
Immediate goal is to distinguish neighbours in peer-to-peer communication, but
this is generally useful.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2335,6 +2335,7 @@ dependencies = [
  "postgres",
  "postgres-protocol",
  "postgres_ffi",
+ "rand",
  "regex",
  "routerify",
  "rust-s3",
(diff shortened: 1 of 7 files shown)


arssher:
@lubennikovaav Please review the second patch.

LizardWizzard:
Some time ago we discussed introducing ids to pageservers too. Maybe worth it to share some code (at least keep that in mind).

Also I think it is ok to use ids created in console, i e when new safekeeper is added to control plane it is being assigned an id and for ease of debugging we can use this id. Thoughts?

I think it is preferable to keep this id in config for simplicity and to avoid unsafe.

arssher:
Sure, code can be reused for pageserver, perhaps with edits.

Yes, assigning from console sounds handy, but I'd still keep the code for generating it internally for tests and development.

Well, keeping it in config means config must be passed everywhere, I wouldn't call it simpler. But would work too.

LizardWizzard:
> keep the code for generating it internally for tests and development

Agree, something like generating the id if it is missing.

> means config must be passed everywhere

I think there wouldn't be too many places. Just like with no-sync option. Also I think this is cleaner because this way we keep everything config like in config.

Another way to avoid unsafe is to use lazy_static or once_cell (or Box::leak as it is done for config in pageserver)

arssher:
Superceded by https://github.com/zenithdb/zenith/pull/1228

PR #1082, opened 2022-01-03
Don't call immediately on resume in callmemaybe.

arssher:
It creates busy loop if pageserver <-> safekeeper connection fails after it was
established (e.g. currently due to 'segment checkpoint not found' error on
pageserver).

Also wake up callmemaybe thread regularly once in recall_period regardless of
channel activity.


petuhovskiy (approved):
LGTM. Wondering if we can use https://docs.rs/tokio/latest/tokio/time/struct.Interval.html to create channel with fixed rate messages for wakeup?

arssher:
Sure, that's much better.

petuhovskiy (approved):

PR #1088, opened 2022-01-06
Do not block on pageserver's shutdown

SomeoneToIgnore:
Part of https://github.com/zenithdb/zenith/issues/1036
I think I'll be able to cover the rest of the cases in a similar manner, but I'd like to discuss it a bit first.
Note that the current PR's base is https://github.com/zenithdb/zenith/pull/1079 to reuse a fraction of its code, but I can move over that particular commit I need from that PR here. Here's the commit I need from there: https://github.com/zenithdb/zenith/pull/1079/commits/a0c896dae85302373bff2d5203076e011ea694e2

I'm also not sure how to write tests on that: the case I've fixed now is very easy to reproduce manually: you have to launch  a single pageserver only (no safekeepers or anything else, just the pageserver) and try to terminate it.
I've tested it manually and can confirm that this termination now blocks on current `main` and does not on my PR.

------------

Reuses the same `join!` approach used in https://github.com/zenithdb/zenith/pull/1079 to force the pageserver to be able to react on external signals, while it waits for certain blocking things like `accept()`.

For the rest of the code, `tokio` `TcpListener` is returned now, requiring additional step to convert it back to a sync version (also added) in the places where it's needed still.
This way, we're able to use async in the places where we really benefit from it, without rewriting everything to async.

Also to note, safekeeper uses similar approach with `tokio::oneshot` channel. Yet, it does not implement `Clone` and we need to unblock pageserver in more than one place, so I did not use that.

### DIFF
--- a/pageserver/src/bin/pageserver.rs
+++ b/pageserver/src/bin/pageserver.rs
@@ -161,7 +161,8 @@ fn start_pageserver(conf: &'static PageServerConf, daemonize: bool) -> Result<()
         "Starting pageserver http handler on {}",
         conf.listen_http_addr
     );
-    let http_listener = tcp_listener::bind(conf.listen_http_addr.clone())?;
+    let http_listener =
+        tcp_listener::to_blocking_listener(tcp_listener::bind(conf.listen_http_addr.clone())?)?;
 
     info!(
         "Starting pageserver pg protocol handler on {}",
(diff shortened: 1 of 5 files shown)


SomeoneToIgnore:
Well, looks like this approach is not really working judging from the test failures, will try to fix those and reopen the PR.

PR #1091, opened 2022-01-10
Redo set vm bits

knizhnik:


lubennikovaav:
The fix in the last 2 commits looks correct to me,
but this PR contains some unrelated commits. Could you please fix it?

knizhnik:
> The fix in the last 2 commits looks correct to me, but this PR contains some unrelated commits. Could you please fix it?

Done.

knizhnik:
Looks like special handling of XLH_INSERT_ALL_FROZEN_SET bit is really not needed.
I have tested VACUUM FREEZE and COPY (FREEZE ON) commands and in both cases VM
is properly updated without this fix. In both cases visibility map is explicitly updated by Heap2/VISIBLE
WAL record.

So looks like SetVisibilityMapFlags is really not needed.
But I am still going to remove check for XLH_INSERT_ALL_FROZEN_SET  from walingest, because
clearing VM on XLH_INSERT_ALL_FROZEN_SET seems to be really strange behavior.

knizhnik:
Replaced with #1126

PR #1093, opened 2022-01-10
Add proxy metrics

bojanserafimov:
Addressing https://github.com/zenithdb/console/issues/373

I added a http server thread on `:7001` with routes `/v1/status` and `/metrics`. I tested this locally and both routes work.

Plan for after this PR:
- add actuall metrics
- make this discoverable by prometheus by exposing http to 9090


proxy/src/state.rs
```diff
@@ -10,6 +10,10 @@ pub struct ProxyConfig {
     /// main entrypoint for users to connect to
     pub proxy_address: SocketAddr,
 
+    /// internally used for status and prometheus metrics
+    pub http_address: SocketAddr,
+
+    // TODO it's not http? it's postgres protocol over tcp.
```
bojanserafimov, line 16: It says it's http but looks like postgres over tcp. Did I misunderstand the code or should I fix the docs? Why would we use a pg backend for auth callbacks from cplane?

kelvich: Comment is wrong. That pg endpoint was a quick hack and coded before we had an http endpoints in our rust code. Deserves to be migrated to http. Let's update the comment for now (and may be create an issue to migrate it to http).

proxy/src/main.rs
```diff
@@ -9,15 +9,18 @@ use anyhow::bail;
...
 
 mod cplane_api;
+mod http;
 mod mgmt;
 mod proxy;
 mod state;
 mod waiters;
 
 fn main() -> anyhow::Result<()> {
+    zenith_metrics::set_common_metrics_prefix("proxy");
```
cicdteam, line 23: @bojanserafimov 
btw, I suggest use `zenith_proxy` as metrics prefix to avoid any mix up with other products, As example [prometheus-proxy](https://github.com/pambrose/prometheus-proxy) has the same prefix `proxy`

cicdteam (changes requested):

cicdteam (approved):

PR #1095, opened 2022-01-11
Add safekeeper --dump-control-file option.

arssher:
Hexalize zids there for better output; since Serde doesn't support several
formats for one struct, on-disk representation is changed as well, make
upgrade.rs cope with it.


petuhovskiy (approved):
Looks ok, but test_restarts_under_load failed in other-tests-release. That's interesting, haven't seen this in a month or so.

arssher:
Yeah, I'm inspecting it right now. The reason is unrelated to PR, but there is clearly a bug. Seems like yet another LSN mismanagement in the queue...

Issue #997, opened 2021-12-14
Refactor on demand nature of Repository::get_timeline


LizardWizzard:
Currently `get_timeline` uses `get_or_init_timeline` which will download remote timeline if it is missing. I think this behaviour is confusing and makes it non trivial to retrieve current state of the timeline without changing something.

The question is do we need this on-demand behaviour at all? Because there is an explicit API to create a branch (new timeline). So we can make all the other code more deterministic to return errors on missing timeline or explicitly create one in place if needed.

Originates from this discussion: https://github.com/zenithdb/zenith/pull/995#discussion_r768776398

Thoughts? @SomeoneToIgnore @lubennikovaav @hlinnaka

PR #1098, opened 2022-01-12
avoid duplication of timeline download tasks

LizardWizzard:
duplication can lead to timeline state become permanently invalid in
the following scenario:
 1. schedule download
 2. while download is not completed schedule another one
 3. first download completes and sets timeline state to Local/Ready
 4. duplicate task can fail and set state back to AwaitsDownload or
    Evicted if error persists

Error in the duplicate download can appear for example when there are
multiple checkpoint archives and after successful execution of the first
task second one will fail to apply checkpoints because they are older
than current tip of the branch.

### DIFF
--- a/pageserver/src/layered_repository.rs
+++ b/pageserver/src/layered_repository.rs
@@ -149,9 +149,11 @@ impl Repository for LayeredRepository {
                 LayeredTimelineEntry::Remote {
                     id,
                     disk_consistent_lsn,
+                    awaits_download,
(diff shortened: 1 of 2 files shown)


LizardWizzard:
This is not working as expected :( because it requires &mut for LayeredTimelineEntry to set the flag. And even if I use AtomicBool it doesn't help because we copy LayeredTimelineEntry around and AtomicBool doesnt implement clone. I can use Arc but this becomes too messy, so I think it is better to be handled in #997

PR #1102, opened 2022-01-12
Add test demostrating write amplification for layered storage

knizhnik:
Closes https://github.com/zenithdb/zenith/issues/1116

### DIFF
--- a/test_runner/batch_others/test_write_amplification.py
+++ b/test_runner/batch_others/test_write_amplification.py
@@ -1,0 +1,62 @@
+from contextlib import closing
+import psycopg2.extras
+import random
+import time
+from fixtures.utils import print_gc_result
+from fixtures.zenith_fixtures import ZenithEnv
+from fixtures.log_helper import log
+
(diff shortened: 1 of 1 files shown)


knizhnik:
Yes, this test looks very similar. But not exactly. My intention was to cause rite amplification with standard (default) parameter.
It seems to be obvious that if we create 10Mb table and then update one 100 bytes record in it and perform checkpoint, then we will get x100000 write amplification. But it requires forced (artificial) checkpoint. Default checkpoint distance is 256Mb, so performing one insert will not trigger checkpoint. And if we do 2000000 inserts (to reach checkpoint distance) then ... there will be almost no write amplification...

This is why I used two tables: bug one for update of single page in segment (causing x1000 write amplification) and second small one with size 256Mb which update generates enough WAL to trigger checkpoint.

Also I have found several different problems not related with write amplification (I have created tickets for it).
So this test seems to be useful. May be it should be renamed and moved to performance. Not sure.

PR #1111, opened 2022-01-12
Use parking lot everywhere in pageserver

SomeoneToIgnore:
Closes https://github.com/zenithdb/zenith/issues/956

### DIFF
--- a/pageserver/src/layered_repository.rs
+++ b/pageserver/src/layered_repository.rs
@@ -15,6 +15,7 @@ use anyhow::{anyhow, bail, ensure, Context, Result};
 use bookfile::Book;
 use bytes::Bytes;
 use lazy_static::lazy_static;
+use parking_lot::{Mutex, MutexGuard};
 use postgres_ffi::pg_constants::BLCKSZ;
 use tracing::*;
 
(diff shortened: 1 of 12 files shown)


SomeoneToIgnore:
This breaks more than is already broken, so sorry for the preliminary noise.

LizardWizzard:
Could you please elaborate, what was the problem?

SomeoneToIgnore:
I've only noticed the general symptoms: regress test job failed, but did not look into details yet.

Will check more thoroughly later and hopefully reopen the adjusted PR.

SomeoneToIgnore:
If I remove the RwLock commit, the tests fail the same way the do in `main`:
https://app.circleci.com/pipelines/github/zenithdb/zenith/4007/workflows/f1941c7d-7f81-4ffa-8de5-35cb7167b6f8

The only arguably non-mechanical change related in the dropped commit is
https://github.com/zenithdb/zenith/pull/1111/files#diff-840e039538d6fae00a64dd5896f48c3132fbf7d3f82c84fd1aed3f76ee62fd6aR129

I'll see more into those changes, but good that we can use Mutexes already.

LizardWizzard:
There are some flaky tests on main so failures shouldn't be a problem with this PR

PR #1131, opened 2022-01-14
remove incorrect check in get_page_at_lsn

LizardWizzard:
because it is not an error to request pages older than latest_gc_cutoff

Mentioned [here](https://github.com/zenithdb/zenith/issues/1122#issuecomment-1012975852) and [here](https://github.com/zenithdb/zenith/issues/1047#issuecomment-1013100568)

### DIFF
--- a/pageserver/src/layered_repository.rs
+++ b/pageserver/src/layered_repository.rs
@@ -204,7 +204,7 @@ impl Repository for LayeredRepository {
         };
 
         src_timeline
-            .check_lsn_is_in_scope(start_lsn)
+            .check_lsn_is_in_pitr_interval(start_lsn)
             .context("invalid branch start lsn")?;
 
         let RecordLsn {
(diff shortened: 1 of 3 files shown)


knizhnik (changes requested):
There is also comparison with latest_gc_cutoff_lsn in check_lsn_is_in_scope (layered_repository.rs:1023). It also should be removed.

LizardWizzard:
@knizhnik why do you thinks so? If we create new branch then gc is not aware of it's branch point in advance so I think it is possible that some data might be found missing

knizhnik:
> 

Sorry, I didn't notice that `check_lsn_is_in_scope`  is used only for `branch_timeline`.
IMHO it should be better renamed to something like `check_lsn_is_in_pitr_iteraval`

knizhnik (approved):

hlinnaka (changes requested):
-1 on this, see comment at https://github.com/zenithdb/zenith/issues/1047#issuecomment-1013241372

LizardWizzard:
@knizhnik note that `check_lsn_is_in_scope` is used in basebackup too. 

> IMHO it should be better renamed to something like check_lsn_is_in_pitr_interval

Done

LizardWizzard:
Per discussion in #1047 this check is indeed valid

PR #1135, opened 2022-01-14
Integrate compute_tools into zenith workspace and improve logging

ololobus:
Resolve zenithdb/console#487
End-to-end tests are expected to fail in PR and after merge, until zenithdb/console#499 is merged, where they pass. Files layout changed, so we have to merge both of them


compute_tools/src/bin/zenith_ctl.rs
```diff
@@ -83,7 +83,12 @@ fn prepare_pgdata(state: &Arc<RwLock<ComputeState>>) -> Result<()> {
     let lsn = sync_safekeepers(&state.pgdata, &state.pgbin)?;
     info!("safekeepers synced at LSN {}", lsn);
 
+    info!(
+        "getting basebackup@{} from pageserver {}",
+        lsn, pageserver_connstr
+    );
     get_basebackup(&state.pgdata, &pageserver_connstr, &tenant, &timeline, &lsn)?;
```
kelvich, line 90: I would also add context right on this error:
```rust
get_basebackup(&state.pgdata, &pageserver_connstr, &tenant, &timeline, &lsn)
    .with_context(|| format!("failed to get basebackup@{} from pageserver {}", lsn, pageserver_connstr)))?;
```

Dockerfile.compute-tools
```diff
@@ -0,0 +1,13 @@
+# First transient image to build compute_tools binaries
+FROM rust:slim-buster AS rust-build
```
SomeoneToIgnore, line 2: Not the problem of this particular PR, but interesting to note that here's another place where rustc is not pinned.
Not a big issue to build a submodule, I guess, but still a bit odd.

.circleci/config.yml
```diff
@@ -571,55 +571,6 @@ jobs:
...
-          key: v03-rust-cache-deps-debug-{{ checksum "Cargo.lock" }}
-          paths:
-            - ~/.cargo/registry
-            - ~/.cargo/git
-            - target
-
-      # Run Rust formatting checks
-      - run:
-          name: cargo fmt check
-          command: cargo fmt --all -- --check
```
SomeoneToIgnore, line 613: Hope something else lints and tests this submodule 🙂

SomeoneToIgnore (approved):

.circleci/config.yml
ololobus: Yes, as it's now part of the workspace, `./run_clippy.sh` and `cargo fmt / test` do the same

Dockerfile.compute-tools
ololobus: Pinned 1.56.1 everywhere in dockerfiles and put some reference comments to help keeping all these places in sync

Issue #1070, opened 2021-12-28
Epic: Refactoring of compute_tools


ololobus:
Do some housekeeping after #1055:
- [x] Integrate into zenith workspace
- [x] zenithdb/console#487
- [x] ~~Get use of zenith_utils::logging~~ tracing is used instead
- [x] ~~Isolate volatile fields from ComputeState into a separate struct and use only it under lock. [See also](https://github.com/zenithdb/zenith/pull/1055#discussion_r775980288)~~
- [ ] Try to re-use [http server from libs/utils/src/http](https://github.com/neondatabase/neon/blob/9860d59aa2b91c5b30aa841a5496af08782027da/libs/utils/src/http/endpoint.rs)
- [ ] Try to get use of a control_plane::postgresql_conf
- [ ] Solve orphaned postgres issues with `prctl(PR_SET_PDEATHSIG, <SIGNAL>)` or anything else. See https://github.com/zenithdb/zenith/pull/1148#issuecomment-1016503120 for the context

PR #1148, opened 2022-01-18
Fix macOs compilation

SomeoneToIgnore:
https://docs.rs/libc/latest/src/libc/unix/linux_like/linux/mod.rs.html#1

`prctl` and `PR_SET_PDEATHSIG` are Linux-specific and do not work, according to `XXX` comment, so I've removed them to restore macOs compilation capabilities

### DIFF
--- a/compute_tools/src/bin/zenith_ctl.rs
+++ b/compute_tools/src/bin/zenith_ctl.rs
@@ -34,7 +34,6 @@ use std::sync::{Arc, RwLock};
 
 use anyhow::{Context, Result};
 use chrono::Utc;
-use libc::{prctl, PR_SET_PDEATHSIG, SIGINT};
 use log::info;
 use postgres::{Client, NoTls};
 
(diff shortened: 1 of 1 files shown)


ololobus (approved):
Oh, my bad, thanks! I now switched to Linux as my primary system, so haven't noticed this.

It seems that you committed some different version of `vendor/postgres` is it intentional?

BTW, it's still not clear for me what is the best practice to kill the child process (postgres), when main exits / panics. Previously I had a guard wrapper, which were killing postgres by stored pid, when main panics, but then @funbringer told me about PR_SET_PDEATHSIG. Which seemed to me as a god option to use :)

SomeoneToIgnore:
Thanks, no postgres changes slip unnoticed :) 

Yes, not clear for me either how to properly do it, alas.

funbringer:
Well, the reason it didn't work out is that it should be used in a **child**, not parent; we should understand this option as "send me a signal when my parent dies". Note that (per `man 2 prctl`) **"The parent-death signal setting is cleared for the child of a fork(2)."**. Hence, one's supposed to do the following:

1. `fork` the process.
2. Inside a fork, call `prctl(PR_SET_PDEATHSIG, <SIGNAL>)`
3. `execve` into a new program.

I'm not going to pretend that this is easy to code up, as there definitely are race conditions to look out for, e.g. (per `man 2 prctl`) "If the parent thread and all ancestor subreapers have already terminated by the time of the PR_SET_PDEATHSIG operation, then no  parent-death signal is sent to the caller.". This one should be mitigated by a pipe which the child process should read from upon calling the `prctl` syscall; if the pipe is dangling, the parent must have died, so we should exit immediately.

If you ask me, I believe there's two things we could learn from this:

* We definitely should `cfg`-out OS-specific features. In this particular case the removal is **completely justified**, since the code was broken, but it's not the *best* way to manage things in the long run. We have a diverse team of people with different backgrounds, and you never know who and when might come up with a good idea. When you lose a grip of a stream of PRs, the magical disappearances of interesting pieces are almost guaranteed to catch you off guard. I'd love to see a "justification" procedure for such "code cleanups", or at least a "final commit period" (24 hours) to give everyone time to react. By the way, @ololobus thanks for pinging me!

* I'm reasonably sure that in most cases we shouldn't remove anything unless we're able to prove the claim set by an unsure comment.

ololobus:
> Well, the reason it didn't work out is that it should be used in a child, not parent; we should understand this option as "send me a signal when my parent dies". Note that (per man 2 prctl) "The parent-death signal setting is cleared for the child of a fork(2)."

Ugh, I've eyeballed the docs a couple of times, but got it in a way that fork clears this up, but if you exec a new child it's applied to it as well. Anyway, we should definitely get back to it, as I test zenith_ctl locally from time to time and I don't like this orphaned Postgres process it leaves behind. Thanks for the comment. Put it on todo into #1070

PR #1150, opened 2022-01-19
Switch our python package management solution to poetry.

LizardWizzard:
Mainly because it has better support for installing the packages from
different python versions.

It also has better dependency resolver than Pipenv. And supports modern
standard for python dependency management. This includes usage of
pyproject.toml for project specific configuration instead of per
tool conf files. See following links for details:
 https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/
 https://www.python.org/dev/peps/pep-0518/


LizardWizzard:
e2e seems to fail because of some recent compute tools changes, so shouldn't be related to this PR

@hlinnaka Please take a look how it works for your setup. I saw you tried poetry in console e2e tests, this should work similarly.

I've added `./scripts/pysync` to update packages so it is simpler to always run this script without diving into python dependency management too much. There is also a complementary wrapper around pytest: `./scripts/pytest`

ololobus:
@LizardWizzard yes, and if you rebase it on top of the fresh main, everything should pass

funbringer (approved):
I like this!

The only thing I'm not sure about is those helper scripts (e.g. `scripts/pytest`). Considering their simplicity do you think we need them?

hlinnaka (approved):

SomeoneToIgnore (commented):
`poetry` letters are all on the same row for me, feels somewhat odd 🙂 

When I try to run `poetry install` it rejects me with the 

```
Work/zenith/zenith (3c38423) HEAD*​  ❯ poetry install
Installing dependencies from lock file

No dependencies to install or update

  ValueError

  zenith is not a package.

  at /opt/homebrew/Cellar/poetry/1.1.12/libexec/lib/python3.10/site-packages/poetry/core/masonry/utils/package_include.py:80 in check_elements
      76│                 self._package = root.name
      77│                 self._elements = sorted(list(root.glob("**/*")))  # type: List[Path]
      78│ 
      79│                 if not self.is_stub_only() and not self.has_modules():
    → 80│                     raise ValueError("{} is not a package.".format(root.name))
      81│ 
      82│                 self._is_package = True
      83│             else:
      84│                 self._package = root.stem
```

is that ok? 
The `./scripts/pysync` one works fine and the tests seem to be run fine also, apparently, but I did not try to add a regular dependency and use it in the test to see if that fails (I'd need `boto3` at least for my new s3 test that comes eventually)

LizardWizzard:
@SomeoneToIgnore by default poetry tries to install current package too (which fails because zenith's root dir doesnt look like a python package), so with pysync (which contains --no-root option to avoid this) I wanted to eliminate this confusion but it seems I've failed :)

@funbringer Probably we won't change this too frequently but I think it is good to have simple wrappers that do not change at all. I wanted to avoid broadcasting to people updates like "now to run tests you have to use --foo=bar". I don't insist on that though. It is ok either way

SomeoneToIgnore (approved):

PR #1152, opened 2022-01-19
Add a mock S3 server in Python tests

SomeoneToIgnore:
Closes https://github.com/zenithdb/zenith/issues/755

Leverages our existing Python infra and uses https://github.com/spulec/moto/ as a background mock server for IT tests, extends an existing remote storage test with the S3 one.

Also contains two small changes in pageserver as different commits, since very related and discovered along the mock server integration.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -277,6 +277,10 @@ jobs:
           name: Install deps
           command: pipenv --python 3.7 install
       - run:
+          name: Install dev deps
+          # we need to run moto s3 mock server, we start is as a separate process, so it's in the dev deps
+          command: pipenv --python 3.7 install --dev
+      - run:
           name: Run pytest
           # pytest doesn't output test logs in real time, so CI job may fail with
           # `Too long with no output` error, if a test is running for a long time.
--- a/pageserver/src/config.rs
+++ b/pageserver/src/config.rs
@@ -144,6 +144,13 @@ pub struct S3Config {
     pub access_key_id: Option<String>,
     /// "Password" to use when connecting to bucket.
     pub secret_access_key: Option<String>,
+    /// A base URL to send S3 requests to.
+    /// By default, the endpoint is derived from a region name, assuming it's
+    /// an AWS S3 region name, erroring on wrong region name.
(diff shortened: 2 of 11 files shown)


.circleci/config.yml
```diff
@@ -276,6 +276,10 @@ jobs:
       - run:
           name: Install deps
           command: pipenv --python 3.7 install
+      - run:
+          name: Install dev deps
+          # we need to run moto s3 mock server, we start is as a separate process, so it's in the dev deps
```
LizardWizzard, line 281: I think it is ok to keep it in the main deps. Since our test runner is entirely about tests I think only misc/linting packages might be considered as development ones. (I think we can merge with that, and I'll place my poetry transition PR on top and fix it there)

pageserver/src/config.rs
```diff
@@ -144,6 +144,7 @@ pub struct S3Config {
     pub access_key_id: Option<String>,
     /// "Password" to use when connecting to bucket.
     pub secret_access_key: Option<String>,
+    pub endpoint: Option<String>,
```
LizardWizzard, line 147: What does it mean to have None endpoint?

LizardWizzard (approved):
Looks good to me

pageserver/src/config.rs
SomeoneToIgnore: Right, documented it.
This is a way to use non-AWS S3 set-ups like our mocking.

.circleci/config.yml
SomeoneToIgnore: I've found poetry lockfile changes more reliable compared to pipenv, so I'd keep it this way and wait for your PR to rebase, thank you :)

SomeoneToIgnore:
I've reverted the "use anonymous creds" commit due to two reasons:
* we don't need anonymous creds now, using `testing/testing` pair for our mock server comms
* so far the plan is to actually reuse AWS metaservice to get the bucket creds automatically for our pageserver deployments (https://aws.amazon.com/ru/premiumsupport/knowledge-center/ec2-instance-access-s3-bucket/)
So now we actually need to pass `None/None` pair in `rust_s3` `Bucket` to trigger the metaservice queries.

Issue #977, opened 2021-12-09
Epic: s3 synchronisation follow ups


stepashka:
This Epic is a collection of follow ups for the S3 integration that was implemented in #340 

Here and below, "remote storage" is a synonym of "s3": we implement the logic for s3, but keep in mind, that other types of storages are involved and may require support later, so the solutions should be generic enough.

no specific order is required for these:

- [ ] #1557
- [ ] #987
- [ ] #1558
- [ ] #1559
- [x] #1049
- [x] serialize and write remote timeline index to s3 periodically to mitigate potential `s3 list` issues (done via https://github.com/neondatabase/neon/issues/1493, see https://github.com/neondatabase/neon/compare/kb/remove-remote-storage-archiving?expand=1 branch for the details)
- [x] migrate to rusoto s3 client from rust-s3, to prepeare for a future `aws-sdk-rust` migration when it matures ("official" but immature crate, it has a similar api to rusoto, more beneficial for our streaming api in many ways). Rusoto was picked as the new library to use: [the PR]( https://github.com/zenithdb/zenith/pull/1483.) Also there's an alternative branch for [aws-sdk-s3](https://github.com/zenithdb/zenith/tree/kb/rusoto-s3).
- [ ] sort out issues with pageserver checkpoints diverging local state from the remote one: #971
- [x] #755 (intersects with @LizardWizzard  's migration work)
- [x] https://github.com/zenithdb/zenith/pull/1080 implement pageserver's s3 prefixes to accomodate multiple pageservers on an s3 bucket
- [x] some preparation for IT tests is needed: s3 parameters ergonomics, refactor pageserver CLI params (currently working on that in #871 )
- [ ] In pageserver recovery we may push whole timeline dir as a checkpoint not only the necessary data. See: https://github.com/zenithdb/zenith/blob/adb0b3dadaf5a4f099e13adeaf8291c2d4b94550/pageserver/src/remote_storage/storage_sync.rs#L725. Also this data may contain garbage as we do not have checksums right now, and we probably should avoid checking every file in timeline dir on startup. Only files which were created after disk consistent lsn. Or we should just remove possibly incorrect files and rebuild the delta from wal
- [x] ?? whatever else @LizardWizzard  finds needed and useful during his tenant migration work, bugs, consistency bugs and performance issues presumably
- [x] randomly allocated work would be needed on refactoring and general logging, metrics and tracing span placement around the s3 sync code to make it observable

...... Some time later: 
[...]

PR #1163, opened 2022-01-21
refactoring of timeline memory state management in layered repo

LizardWizzard:
In this PR I'm exploring possible refactorings regarding timeline state management and corresponding layered repo APIs. I've gone quite far but this is still rough (many TODO's and leftovers), so I decided to ask for feedback before investing more effort into finishing this.

Overview of the changes:
LayeredRepository.timelines now contain entries for all locally available and remote timelines. Even ones that are not loaded into memory with layer map. For those only metadata is kept (And should be read on startup). I've tested that walking the directory and loading 3000 timelines this way took ~0.8 seconds on my laptop). For more details see what happened to LayeredTimelineEntry.

Api for timeline retrieval has significantly changed. get_timeline is now free of side effects, it won't try to load timeline from filesystem or download it from s3. Sort of equivalent of old API is get_local_timeline_load. Which will load local timeline if it is not loaded. In my opinion it is better to be explicit about that, because loading is not free especially for big databases (see https://github.com/zenithdb/zenith/issues/1155). Also what is not solved yet is that we load timelines under a mutex and this has to be reworked too. (There is a little bit more work because load is recursive regarding to ancestors.)

This is highly related to https://github.com/zenithdb/zenith/issues/1049 and intended to resolve https://github.com/zenithdb/zenith/issues/997. Also tries to fix the behaviour of remote_consistent_lsn which currently uses local disk_consistent_lsn instead of the remote one.

Some other thoughts.
1. I think we should avoid storing info regarding remote timelines at all. It can easily become out of sync when we have a dozen pageservers.
2. Details about GC and sync of branched timelines are vague right now. IMO current safe spot is to assume that we always have timeline with all its ancestors locally. (Currently it is not enforced by the typesystem and creates noise of extra error handling, but it is easy fix)
3. Timeline download from s3 should only happen through attach http handler introduced in tenant rebalancing PR https://github.com/zenithdb/zenith/pull/995 (This means that in this PR's branch there are no calls to schedule_timeline_download)

Remaining bits if we want to go in this direction:
1. Rework pageserver startup a bit so we have clear picture of what is on the filesystem (fill timelines map on startup). And deal with remote storage sync accordingly. (Most of the tests are failing because of that)
2. Change routines that previously inspected filesystem to get all locally available timelines to use timelines map.
3. Get rid of most of the TODO's

### DIFF
--- a/control_plane/src/storage.rs
+++ b/control_plane/src/storage.rs
@@ -1,4 +1,3 @@
-use std::convert::TryFrom;
 use std::io::Write;
 use std::net::TcpStream;
 use std::path::PathBuf;
@@ -10,7 +9,7 @@ use anyhow::{bail, Context};
 use nix::errno::Errno;
 use nix::sys::signal::{kill, Signal};
 use nix::unistd::Pid;
-use pageserver::http::models::{TenantCreateRequest, TimelineCreateRequest, TimelineInfoResponse};
+use pageserver::http::models::{TenantCreateRequest, TimelineCreateRequest};
 use pageserver::timelines::TimelineInfo;
 use postgres::{Config, NoTls};
 use reqwest::blocking::{Client, RequestBuilder, Response};
@@ -358,7 +357,7 @@ impl PageServerNode {
     }
 
     pub fn timeline_list(&self, tenant_id: &ZTenantId) -> anyhow::Result<Vec<TimelineInfo>> {
-        let timeline_infos: Vec<TimelineInfoResponse> = self
+        let timeline_infos: Vec<TimelineInfo> = self
             .http_request(
                 Method::GET,
                 format!("{}/tenant/{}/timeline", self.http_base_url, tenant_id),
@@ -367,10 +366,7 @@ impl PageServerNode {
             .error_from_body()?
             .json()?;
 
-        timeline_infos
-            .into_iter()
-            .map(TimelineInfo::try_from)
-            .collect()
+        Ok(timeline_infos)
     }
 
     pub fn timeline_create(
@@ -392,10 +388,8 @@ impl PageServerNode {
             })
             .send()?
             .error_from_body()?
-            .json::<Option<TimelineInfoResponse>>()?;
+            .json::<Option<TimelineInfo>>()?;
 
-        timeline_info_response
-            .map(TimelineInfo::try_from)
-            .transpose()
+        Ok(timeline_info_response)
     }
 }
--- a/pageserver/src/http/routes.rs
+++ b/pageserver/src/http/routes.rs
@@ -3,6 +3,7 @@ use std::sync::Arc;
 use anyhow::Result;
 use hyper::StatusCode;
 use hyper::{Body, Request, Response, Uri};
+use tokio::sync::RwLock;
 use tracing::*;
 use zenith_utils::auth::JwtAuth;
 use zenith_utils::http::endpoint::attach_openapi_ui;
@@ -16,24 +17,32 @@ use zenith_utils::http::{
     request::parse_request_param,
 };
 use zenith_utils::http::{RequestExt, RouterBuilder};
-use zenith_utils::zid::{HexZTenantId, ZTimelineId};
+use zenith_utils::zid::{HexZTenantId, ZTenantTimelineId, ZTimelineId};
 
 use super::models::{
-    StatusResponse, TenantCreateRequest, TimelineCreateRequest, TimelineInfoResponse,
+    StatusResponse, TenantCreateRequest, TimelineCreateRequest, TimelineInfoResponseV1,
+    TimelineInfoV1,
 };
-use crate::repository::RepositoryTimeline;
-use crate::timelines::TimelineInfo;
+use crate::remote_storage::{schedule_timeline_download, RemoteTimelineIndex};
+use crate::timelines::{
+    extract_remote_timeline_info, LocalTimelineInfo, RemoteTimelineInfo, TimelineInfo,
+};
 use crate::{config::PageServerConf, tenant_mgr, timelines, ZTenantId};
 
 #[derive(Debug)]
 struct State {
     conf: &'static PageServerConf,
     auth: Option<Arc<JwtAuth>>,
+    remote_index: Arc<RwLock<RemoteTimelineIndex>>,
     allowlist_routes: Vec<Uri>,
 }
 
 impl State {
-    fn new(conf: &'static PageServerConf, auth: Option<Arc<JwtAuth>>) -> Self {
+    fn new(
+        conf: &'static PageServerConf,
+        auth: Option<Arc<JwtAuth>>,
+        remote_index: Arc<RwLock<RemoteTimelineIndex>>,
+    ) -> Self {
         let allowlist_routes = ["/v1/status", "/v1/doc", "/swagger.yml"]
             .iter()
             .map(|v| v.parse().unwrap())
@@ -42,6 +51,7 @@ impl State {
             conf,
             auth,
             allowlist_routes,
+            remote_index,
         }
     }
 }
@@ -88,7 +98,7 @@ async fn timeline_create_handler(mut request: Request<Body>) -> Result<Response<
     .map_err(ApiError::from_err)??;
 
     Ok(match new_timeline_info {
-        Some(info) => json_response(StatusCode::CREATED, TimelineInfoResponse::from(info))?,
+        Some(info) => json_response(StatusCode::CREATED, info)?,
         None => json_response(StatusCode::CONFLICT, ())?,
     })
 }
@@ -97,15 +107,24 @@ async fn timeline_list_handler(request: Request<Body>) -> Result<Response<Body>,
     let tenant_id: ZTenantId = parse_request_param(&request, "tenant_id")?;
     check_permission(&request, Some(tenant_id))?;
     let include_non_incremental_logical_size = get_include_non_incremental_logical_size(&request);
-    let response_data: Vec<TimelineInfoResponse> = tokio::task::spawn_blocking(move || {
+    let local_timeline_infos = tokio::task::spawn_blocking(move || {
         let _enter = info_span!("timeline_list", tenant = %tenant_id).entered();
-        crate::timelines::get_timelines(tenant_id, include_non_incremental_logical_size)
+        crate::timelines::get_local_timelines(tenant_id, include_non_incremental_logical_size)
     })
     .await
-    .map_err(ApiError::from_err)??
-    .into_iter()
-    .map(TimelineInfoResponse::from)
-    .collect();
+    .map_err(ApiError::from_err)??;
+
+    let remote_index = get_state(&request).remote_index.read().await;
+    let mut response_data = Vec::with_capacity(local_timeline_infos.len());
+    for (timeline_id, local_timeline_info) in local_timeline_infos {
+        response_data.push(TimelineInfo {
+            tenant_id,
+            timeline_id,
+            local: Some(local_timeline_info),
+            remote: extract_remote_timeline_info(tenant_id, timeline_id, &remote_index),
+        })
+    }
+
     Ok(json_response(StatusCode::OK, response_data)?)
 }
 
@@ -124,30 +143,76 @@ fn get_include_non_incremental_logical_size(request: &Request<Body>) -> bool {
         .unwrap_or(false)
 }
 
-async fn timeline_detail_handler(request: Request<Body>) -> Result<Response<Body>, ApiError> {
+// common part for v1 and v2 handlers
+async fn timeline_detail_common(request: Request<Body>) -> Result<TimelineInfo, ApiError> {
     let tenant_id: ZTenantId = parse_request_param(&request, "tenant_id")?;
     check_permission(&request, Some(tenant_id))?;
 
     let timeline_id: ZTimelineId = parse_request_param(&request, "timeline_id")?;
+    let include_non_incremental_logical_size = get_include_non_incremental_logical_size(&request);
 
-    let response_data = tokio::task::spawn_blocking(move || {
-        let _enter =
-            info_span!("timeline_detail_handler", tenant = %tenant_id, timeline = %timeline_id)
-                .entered();
+    let span = info_span!("timeline_detail_handler", tenant = %tenant_id, timeline = %timeline_id);
+
+    let (local_timeline_info, span) = tokio::task::spawn_blocking(move || {
+        let entered = span.entered();
         let repo = tenant_mgr::get_repository_for_tenant(tenant_id)?;
-        let include_non_incremental_logical_size =
-            get_include_non_incremental_logical_size(&request);
-        Ok::<_, anyhow::Error>(TimelineInfo::from_repo_timeline(
-            tenant_id,
-            repo.get_timeline(timeline_id)?,
-            include_non_incremental_logical_size,
-        ))
+        let local_timeline = {
+            repo.get_timeline(timeline_id)
+                .map(|timeline| {
+                    LocalTimelineInfo::from_repo_timeline(
+                        timeline,
+                        include_non_incremental_logical_size,
+                    )
+                })
+                .transpose()?
+        };
+        Ok::<_, anyhow::Error>((local_timeline, entered.exit()))
     })
     .await
-    .map_err(ApiError::from_err)?
-    .map(TimelineInfoResponse::from)?;
+    .map_err(ApiError::from_err)??;
 
-    Ok(json_response(StatusCode::OK, response_data)?)
+    let remote_timeline_info = {
+        let remote_index_read = get_state(&request).remote_index.read().await;
+        remote_index_read
+            .timeline_entry(&ZTenantTimelineId {
+                tenant_id,
+                timeline_id,
+            })
+            .map(|remote_entry| RemoteTimelineInfo {
+                remote_consistent_lsn: remote_entry.disk_consistent_lsn(),
+                awaits_download: remote_entry.get_awaits_download(),
+            })
+    };
+
+    let _enter = span.entered();
+
+    if local_timeline_info.is_none() && remote_timeline_info.is_none() {
+        return Err(ApiError::NotFound(
+            "Timeline is not found neither locally nor remotely".to_string(),
+        ));
+    }
+
+    Ok(TimelineInfo {
+        tenant_id,
+        timeline_id,
+        local: local_timeline_info,
+        remote: remote_timeline_info,
+    })
+}
+
+// TODO remove when console adopts v2
+async fn timeline_detail_handler_v1(request: Request<Body>) -> Result<Response<Body>, ApiError> {
+    let timeline_info = timeline_detail_common(request).await?;
+    Ok(json_response(
+        StatusCode::OK,
+        TimelineInfoResponseV1::from(TimelineInfoV1::from(timeline_info)),
+    )?)
+}
+
+async fn timeline_detail_handler_v2(request: Request<Body>) -> Result<Response<Body>, ApiError> {
+    let timeline_info = timeline_detail_common(request).await?;
+
+    Ok(json_response(StatusCode::OK, timeline_info)?)
 }
 
 async fn timeline_attach_handler(request: Request<Body>) -> Result<Response<Body>, ApiError> {
@@ -155,30 +220,36 @@ async fn timeline_attach_handler(request: Request<Body>) -> Result<Response<Body
     check_permission(&request, Some(tenant_id))?;
 
     let timeline_id: ZTimelineId = parse_request_param(&request, "timeline_id")?;
+    let span = info_span!("timeline_attach_handler", tenant = %tenant_id, timeline = %timeline_id);
 
-    tokio::task::spawn_blocking(move || {
-        let _enter =
-            info_span!("timeline_attach_handler", tenant = %tenant_id, timeline = %timeline_id)
-                .entered();
-        let repo = tenant_mgr::get_repository_for_tenant(tenant_id)?;
-        match repo.get_timeline(timeline_id)? {
-            RepositoryTimeline::Local { .. } => {
-                anyhow::bail!("Timeline with id {} is already local", timeline_id)
-            }
-            RepositoryTimeline::Remote {
-                id: _,
-                disk_consistent_lsn: _,
-            } => {
-                // FIXME (rodionov) get timeline already schedules timeline for download, and duplicate tasks can cause errors
-                //  first should be fixed in https://github.com/zenithdb/zenith/issues/997
-                // TODO (rodionov) change timeline state to awaits download (incapsulate it somewhere in the repo)
-                // TODO (rodionov) can we safely request replication on the timeline before sync is completed? (can be implemented on top of the #997)
-                Ok(())
-            }
-        }
+    let span = tokio::task::spawn_blocking(move || {
+        let entered = span.entered();
+        if tenant_mgr::get_timeline_for_tenant_load(tenant_id, timeline_id).is_ok() {
+            anyhow::bail!("Timeline is already present locally")
+        };
+        Ok(entered.exit())
     })
     .await
     .map_err(ApiError::from_err)??;
+
+    let mut remote_index_write = get_state(&request).remote_index.write().await;
+
+    let _enter = span.entered(); // entered guard cannot live across awaits (non Send)
+    let index_entry = remote_index_write
+        .timeline_entry_mut(&ZTenantTimelineId {
+            tenant_id,
+            timeline_id,
+        })
+        .ok_or_else(|| ApiError::BadRequest("Unknown remote timeline".to_string()))?;
+
+    if index_entry.get_awaits_download() {
+        return Err(ApiError::NotFound(
+            "Timeline download is already in progress".to_string(),
+        ));
+    }
+
+    index_entry.set_awaits_download(true);
+    schedule_timeline_download(tenant_id, timeline_id);
 
     Ok(json_response(StatusCode::ACCEPTED, ())?)
 }
@@ -221,13 +292,17 @@ async fn tenant_create_handler(mut request: Request<Body>) -> Result<Response<Bo
     check_permission(&request, None)?;
 
     let request_data: TenantCreateRequest = json_request(&mut request).await?;
+    let remote_index = Arc::clone(&get_state(&request).remote_index);
+
+    let target_tenant_id = request_data
+        .new_tenant_id
+        .map(ZTenantId::from)
+        .unwrap_or_else(ZTenantId::generate);
 
     let new_tenant_id = tokio::task::spawn_blocking(move || {
-        let _enter = info_span!("tenant_create", tenant = ?request_data.new_tenant_id).entered();
-        tenant_mgr::create_tenant_repository(
-            get_config(&request),
-            request_data.new_tenant_id.map(ZTenantId::from),
-        )
+        let _enter = info_span!("tenant_create", tenant = ?target_tenant_id).entered();
+
+        tenant_mgr::create_tenant_repository(get_config(&request), target_tenant_id, remote_index)
     })
     .await
     .map_err(ApiError::from_err)??;
@@ -248,6 +323,7 @@ async fn handler_404(_: Request<Body>) -> Result<Response<Body>, ApiError> {
 pub fn make_router(
     conf: &'static PageServerConf,
     auth: Option<Arc<JwtAuth>>,
+    remote_index: Arc<RwLock<RemoteTimelineIndex>>,
 ) -> RouterBuilder<hyper::Body, ApiError> {
     let spec = include_bytes!("openapi_spec.yml");
     let mut router = attach_openapi_ui(endpoint::make_router(), spec, "/swagger.yml", "/v1/doc");
@@ -263,7 +339,7 @@ pub fn make_router(
     }
 
     router
-        .data(Arc::new(State::new(conf, auth)))
+        .data(Arc::new(State::new(conf, auth, remote_index)))
         .get("/v1/status", status_handler)
         .get("/v1/tenant", tenant_list_handler)
         .post("/v1/tenant", tenant_create_handler)
@@ -271,7 +347,11 @@ pub fn make_router(
         .post("/v1/tenant/:tenant_id/timeline", timeline_create_handler)
         .get(
             "/v1/tenant/:tenant_id/timeline/:timeline_id",
-            timeline_detail_handler,
+            timeline_detail_handler_v1,
+        )
+        .get(
+            "/v2/tenant/:tenant_id/timeline/:timeline_id",
+            timeline_detail_handler_v2,
         )
         .post(
             "/v1/tenant/:tenant_id/timeline/:timeline_id/attach",
--- a/pageserver/src/layered_repository.rs
+++ b/pageserver/src/layered_repository.rs
@@ -35,9 +35,9 @@ use self::metadata::{metadata_path, TimelineMetadata, METADATA_FILE_NAME};
 use crate::config::PageServerConf;
 use crate::page_cache;
 use crate::relish::*;
-use crate::remote_storage::{schedule_timeline_checkpoint_upload, schedule_timeline_download};
+use crate::remote_storage::{schedule_timeline_checkpoint_upload, RemoteTimelineIndex};
 use crate::repository::{
-    BlockNumber, GcResult, Repository, RepositoryTimeline, Timeline, TimelineSyncState,
+    BlockNumber, GcResult, Repository, RepositoryTimeline, Timeline, TimelineSyncStatusUpdate,
     TimelineWriter, ZenithWalRecord,
 };
 use crate::thread_mgr;
@@ -129,27 +129,46 @@ pub struct LayeredRepository {
     // timeout...
     gc_cs: Mutex<()>,
     walredo_mgr: Arc<dyn WalRedoManager + Send + Sync>,
+
+    // provides access to timeline data sitting in the remote storage
+    // supposed to be used for retrieval of remote consistent lsn in walreceiver
+    remote_index: Arc<tokio::sync::RwLock<RemoteTimelineIndex>>,
+
     /// Makes every timeline to backup their files to remote storage.
     upload_relishes: bool,
 }
 
 /// Public interface
 impl Repository for LayeredRepository {
-    fn get_timeline(&self, timelineid: ZTimelineId) -> Result<RepositoryTimeline> {
-        Ok(RepositoryTimeline::from(self.get_or_init_timeline(
-            timelineid,
-            &mut self.timelines.lock().unwrap(),
-        )?))
+    fn get_timeline(&self, timelineid: ZTimelineId) -> Option<RepositoryTimeline> {
+        let timelines = self.timelines.lock().unwrap();
+        self.get_timeline_internal(timelineid, &timelines)
+            .map(RepositoryTimeline::from)
     }
 
-    fn list_timelines(&self) -> Result<Vec<RepositoryTimeline>> {
-        Ok(self
-            .timelines
+    fn get_timeline_load(&self, timelineid: ZTimelineId) -> Result<Arc<dyn Timeline>> {
+        let mut timelines = self.timelines.lock().unwrap();
+        match self.get_timeline_load_internal(timelineid, &mut timelines)? {
+            Some(local_loaded_timeline) => Ok(local_loaded_timeline as _),
+            None => anyhow::bail!(
+                "cannot get local timeline: unknown timeline id: {}",
+                timelineid
+            ),
+        }
+    }
+
+    fn list_timelines(&self) -> Vec<(ZTimelineId, RepositoryTimeline)> {
+        self.timelines
             .lock()
             .unwrap()
-            .values()
-            .map(|timeline_entry| RepositoryTimeline::from(timeline_entry.clone()))
-            .collect())
+            .iter()
+            .map(|(timeline_id, timeline_entry)| {
+                (
+                    *timeline_id,
+                    RepositoryTimeline::from(timeline_entry.clone()),
+                )
+            })
+            .collect()
     }
 
     fn create_empty_timeline(
@@ -176,10 +195,16 @@ impl Repository for LayeredRepository {
             self.upload_relishes,
         );
 
-        let timeline_rc = Arc::new(timeline);
-        let r = timelines.insert(timelineid, LayeredTimelineEntry::Local(timeline_rc.clone()));
-        assert!(r.is_none());
-        Ok(timeline_rc)
+        let timeline = Arc::new(timeline);
+        let r = timelines.insert(
+            timelineid,
+            LayeredTimelineEntry::Loaded(Arc::clone(&timeline)),
+        );
+        ensure!(
+            r.is_none(),
+            "assertion failure, inserted duplicate timeline"
+        );
+        Ok(timeline)
     }
 
     /// Branch a timeline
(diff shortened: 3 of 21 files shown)


pageserver/src/layered_repository.rs
```diff
@@ -1004,11 +1099,11 @@ impl Timeline for LayeredTimeline {
...
-        self.ancestor_timeline
-            .as_ref()
-            .and_then(|ancestor_entry| ancestor_entry.local_or_schedule_download(self.tenantid))
-            .map(Timeline::get_start_lsn)
-            .unwrap_or(self.ancestor_lsn)
+        // TODO ask @knizhnik about this change,
+        // previously get_start_lsn always returned zero because it was recursive
+        // and timeline without an ancestor has 0 ancestor lsn
+        // Can it be removed in favor of self.get_ancestor_lsn?
+        self.ancestor_lsn
```
LizardWizzard, line 1106: @knizhnik Please see the comment in code. Am I correct?

knizhnik: It seems to be correct.
But I wonder why you are asking me - it's not my code:)

LizardWizzard: Sorry if I've misinterpreted something, but git blame points to this commit: [#b227c63ed](https://github.com/zenithdb/zenith/commit/b227c63edf1c7d3622daeb3b47ba693c84430052#) this is when this method was introduced

knizhnik: O, I see. get_start_lsn() method was really introduced by me.
But then this code was refactored so that I even do not recognize it as my code.

But in any case, I think it is safe to use ancestor_lsn without recursion.

pageserver/src/layered_repository.rs
```diff
@@ -195,12 +201,11 @@ impl Repository for LayeredRepository {
...
-            LayeredTimelineEntry::Local(timeline) => timeline,
-            LayeredTimelineEntry::Remote { .. } => {
-                bail!("Cannot branch off the timeline {} that's not local", src)
-            }
-        };
+        let src_timeline = self
+            .get_local_timeline_load_internal(src, &mut timelines)
+            // message about timeline being remote is one .context up in the stack
+            .context("failed to load timeline for branching")?
+            .ok_or(anyhow::anyhow!("unknown timeline id: {}", &src))?;
```
SomeoneToIgnore, line 208: NIT: might be more "right" to use `.ok_or_else(|| ...)` to defer the formatting, there are a few more placed like that. Also worth importing `anyhow::anyhow` at this point too, I guess.

pageserver/src/http/routes.rs
```diff
@@ -224,22 +224,21 @@ async fn timeline_detail_handler(request: Request<Body>) -> Result<Response<Body
...
         let repo = tenant_mgr::get_repository_for_tenant(tenant_id)?;
-        Ok::<_, anyhow::Error>(match repo.get_timeline(timeline_id)?.local_timeline() {
-            None => TimelineInfo::Remote {
-                timeline_id,
-                tenant_id,
-            },
-            Some(timeline) => TimelineInfo::Local {
+        let local_timeline = {
+            // TODO (current) handle other timeline states
+            let timeline = repo.get_local_timeline_load(timeline_id)?;
```
SomeoneToIgnore, line 229: So that will implicitly load the timeline into pageserver's memory.

Curious a bit: where we draw the line between loading into the memory and S3 downloading?
`zenith init` timeline state can be downloaded and loaded into memory in a matter of a couple of seconds now, so from my perspective, both download and load into memory could be on par speed-wise, so we could in theory treat them as different stages of the timeline loading and require both to be done explicitly?

pageserver/src/http/routes.rs
```diff
@@ -224,22 +224,21 @@ async fn timeline_detail_handler(request: Request<Body>) -> Result<Response<Body
...
                 .entered();
         let repo = tenant_mgr::get_repository_for_tenant(tenant_id)?;
-        Ok::<_, anyhow::Error>(match repo.get_timeline(timeline_id)?.local_timeline() {
-            None => TimelineInfo::Remote {
-                timeline_id,
-                tenant_id,
-            },
-            Some(timeline) => TimelineInfo::Local {
+        let local_timeline = {
+            // TODO (current) handle other timeline states
```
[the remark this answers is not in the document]
SomeoneToIgnore: Also interesting to know, how do we want to deal with `timeline_attach_handler` http requests, that are synchronous at their current definition, yet the S3 sync machinery is now very async and does not give any latency guarantees.

pageserver/src/layered_repository.rs
```diff
@@ -289,131 +304,203 @@ impl Repository for LayeredRepository {
...
-            TimelineSyncState::Evicted(_) => timelines_accessor.remove(&timeline_id),
-            TimelineSyncState::AwaitsDownload(disk_consistent_lsn)
-            | TimelineSyncState::CloudOnly(disk_consistent_lsn) => timelines_accessor.insert(
-                timeline_id,
-                LayeredTimelineEntry::Remote {
-                    id: timeline_id,
-                    disk_consistent_lsn,
-                },
-            ),
+            TimelineState::Evicted(_) => timelines_accessor.remove(&timeline_id), // FIXME Should we shutdown the timeline?
```
SomeoneToIgnore, line 327: Before that, we at least disabled its S3 sync capabilities at this moment.
That was removed in https://github.com/zenithdb/zenith/pull/1108#discussion_r783391549 , so might not be need with the approach?

pageserver/src/repository.rs
```diff
@@ -66,60 +65,103 @@ pub trait Repository: Send + Sync {
...
-    pub fn remote_disk_consistent_lsn(&self) -> Option<Lsn> {
-        Some(match self {
-            TimelineSyncState::Evicted(None) => return None,
-            TimelineSyncState::Ready(lsn) => lsn,
-            TimelineSyncState::AwaitsDownload(lsn) => lsn,
-            TimelineSyncState::CloudOnly(lsn) => lsn,
-            TimelineSyncState::Evicted(Some(lsn)) => lsn,
-        })
-        .copied()
+impl From<RepositoryTimeline> for TimelineState {
```
SomeoneToIgnore, line 141: NIT: feels a bit odd to _move_ an entire timeline to get its state, as if it could be a method on the `RepositoryTimeline` instead.

pageserver/src/remote_storage/storage_sync/upload.rs
```diff
@@ -127,6 +128,12 @@ pub(super) async fn upload_timeline_checkpoint<
                     index_write.add_timeline_entry(sync_id, TimelineIndexEntry::Full(new_timeline));
                 }
             }
+            // TODO (current) Timeline should be available.
+            // We cant just pass Timeline from schedule_checkpoint_upload because checkpoint is defined on a Timeline rather than on Arc<dyn Timeline>.
+            let timeline = tenant_mgr::get_repository_for_tenant(tenant_id)
+                .and_then(|repo| repo.get_local_timeline_load(timeline_id))
+                .expect("timeline should be available because just did a checkpoint");
+            timeline.set_remote_consistent_lsn(new_checkpoint.metadata.disk_consistent_lsn());
```
SomeoneToIgnore, line 136: It looks a bit odd that we now have two mechanisms to pass the remote_consistent_lsn:
* the old way via `new_timeline_states` (where we pass on the latest synced timeline_id)
* this way, where we potentially block the executor by locking the timelines

IMO, there has to be one way only to do so, maybe with locking, but not sure if the timeline lock is a good idea: I'd store the data about timeline availability and LSN's somewhere in a separately managed index in memory.

pageserver/src/layered_repository.rs
```diff
@@ -35,10 +35,10 @@ use self::metadata::{metadata_path, TimelineMetadata, METADATA_FILE_NAME};
 use crate::config::PageServerConf;
 use crate::page_cache;
 use crate::relish::*;
-use crate::remote_storage::{schedule_timeline_checkpoint_upload, schedule_timeline_download};
+use crate::remote_storage::schedule_timeline_checkpoint_upload;
```
SomeoneToIgnore, line 38: I've not noticed anything changing to the upload checks: does that mean, that we now can schedule whatever arbitrary upload for the same timeline, even if it's already downloaded/not loaded, etc?

One of the main benefits of commonly accessible timeline sync state should be that IMO, so could be reused.

SomeoneToIgnore (commented):
I like the direction of the main part of this PR, about splitting away the effects of the `get_timeline` method.
I'm not fond that it's pretty vague and nobody knows what to expect from those methods and their effects, so noted some contradictions on the http API.
Feels like we would benefit greatly from some baseline, like an example on how console(?) uses this http API to download/create/whatever_else timeline.

Another part of the PR, related to the timeline status changes looks quite contradictory to me. In particular, I've found odd: two mechanisms to propagate parts of the state, synchronization that requires locking all timelines again and still zero proper synchronization (or checks at least) between the S3 sync loop and the rest of the code.

I've shared my idea about having a version of S3 sync loop's `index.rs` in some form under the lock in the place, accessible by both the S3 sync loop and the rest of the code, have you checked on that? 
I have no strong preference over [not]storing the remote timeline data there in the end, but have a feeling that isolating all the sync and state logic in a separate place might be helpful with unit testing that, also with simplifying the code.

pageserver/src/http/routes.rs
SomeoneToIgnore: To paraphrase: right now, it looks like we still have some implicitness on how we treat the not-loaded-in-memory layers, mainly we seem to load such layers implicitly as here. Feels odd, compared to what we've just done to separate the download into a separate manual step.

pageserver/src/remote_storage/storage_sync/upload.rs
LizardWizzard: Yeah, I tried to pass Arc<dyn TImeline> and atomically update the lsn without locking, but that approach failed with the reason described in the comment, thanks for pointing that out

pageserver/src/branches.rs
```diff
@@ -56,39 +58,69 @@ impl BranchInfo {
...
+                    timeline_id,
+                    latest_valid_lsn: timeline.get_last_record_lsn(),
+                    ancestor_id,
+                    ancestor_lsn,
+                    current_logical_size: Some(timeline.get_current_logical_size()),
+                    current_logical_size_non_incremental,
+                }
+            }
+            LocalRepositoryTimeline::Unloaded { metadata } => {
+                // TODO is there some generic enough way to avoid this duplication? generic function with common trait?
```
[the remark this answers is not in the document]
LizardWizzard: Good Idea! It would also help to avoid error checking noise around ancestor fields, because ancestor id is Option, but ancestor lsn isnt and is set to zero when ancestor_id is None

pageserver/src/http/routes.rs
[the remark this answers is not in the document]
LizardWizzard: I think there is no need to list remote timelines at all. In our cloud setup console has all the knowledge about tenants and timelines. Requirement to know about all remote timelines on each pageserver is quite hard to implement correctly because this involves global (even distributed) state. Result of the list call can quite quickly become out of sync and hence cannot provide correct answer to the question "what remote timelines are currently available". That means that we need to introduce some background polling to update the inmemory index, and even with that 404 answer on timeline detail can mean two completely different things. It is either permanent absence of the timeline or it was just not synced yet. This leads to possible retries from clients, timeout management and all the ~~delights~~ consequences of eventual consistency. So I think we should avoid that coordination as much as possible or use separate metadata service for discovery/management of this state. See https://github.com/zenithdb/rfcs/pull/16 for more details on metadata service approach. 

> how do we want to deal with timeline_attach_handler http requests

I think the way with that is how it is done in tenant relocation test: https://github.com/zenithdb/zenith/pull/995/files#diff-28d93314ea7a547530e9c963649b70dedbbabe0cc9cb60e07f797f432cd75f6cR196. We call attach, and then wait by polling timeline detail for status to become local. This can also change with the presence of the metadata server. When pageserver can advertise statuses of the timelines it knows about. Metadata service can also provide a way to subscribe for the changes. (AFAIK both etcd and consul support that)

pageserver/src/layered_repository.rs
LizardWizzard: It is possible to implement this check on top of what is provided by this PR. It is enough to check the flag inside remote timeline state. It is missing intentionally, and can be done as part of remaining TODO's or separately

pageserver/src/http/routes.rs
LizardWizzard: > could be on par speed-wise

I think this highly depends on how our storage is organized and what is the size of the timeline. For large databases and current storage design with many relatively small layers loading of the layer map can take an hour to complete. But I think with upcoming LSM approach loading will become significantly faster.

> to load such layers implicitly as here

See TODO above :) Timeline detail should be free of side effects.

Regarding different stages of load, I have couple of open questions which require additional investigation. In general for timeline to be ready we need both, locally available data and basic index in memory. 
* How to match SLO on new compute start? Do we always keep timeline assigned to at least one pageserver or we can completely unload it. In that case cold start can take quite a long time especially for large timelines. IMO we should keep timeline on at least one pageserver
* In the upcoming storage design load from disk should not take too much time so I expect loading from s3 to take more time (but for now this is just speculation). I think we can require timeline to be loaded from disk only in the places it is actually needed (e.g. when compute connected to us or we've received callmemaybe request). And this PR opens possibility to do so. Returning an error in the above cases when timeline is not available locally is ok to me because it is actually an error to direct traffic to the pageserver which is not ready to accept it. This error indicates that we cannot start processing in a reasonable time (i e without violating our SLO). That may be the reason why we need to treat s3 sync differently from load from disk.
* Can we/should we interleave download with loading? I e download some index parts, and begin bootstrapping it before all the data is available locally. It might be quite hard to organize given our mix of sync/async parts of repository/remote storage sync. 

It is still an open question how to organize it better (especially for our future demands) and answer all the questions at once. But anyway we should start to evolve it in some promising direction

pageserver/src/http/routes.rs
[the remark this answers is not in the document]
SomeoneToIgnore: Sure, got the idea of no remote sync state.

> I think the way with that is how it is done in tenant relocation test

I was mainly confused by the fact that there we somehow store that timeline id within the same test scope, while in reality, without listing the remote timeline ids, there could be obvious way to recover id we need to use.
But if the metadata server is present, presumably that should also be solved by it.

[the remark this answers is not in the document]
LizardWizzard: > there could be obvious way to recover id we need to use

Yeah, either metadata service or console should help with that. But I agree that this is not obvious, and we should think about what to do for self hosted installations

pageserver/src/repository.rs
LizardWizzard: What do you think about implementing From for a reference? 

Something like that:
```rust
struct Foo {
    a: u32
}

struct Bar {
    b: u32
}

impl<'a> From<&'a Foo> for Bar {
    fn from(_: &Foo) -> Self { todo!() }
}
```

Maybe usual method is more readable

SomeoneToIgnore: No objections against lifetimes per se, so why not.
A small concern is a readability of that `From` usage, but we can always replace it with a plain function in case of any issues.

LizardWizzard:
This is the second iteration. The main change is separation of everything related to remote statuses out of repository into RemoteTimelineIndex. I decided to avoid storing it in global variable so it is passed around in an Arc.

Changes to repository internals are the following (combined with changes from first iteration):
* During startup we read metadata for every timeline and store it in timelines map as Unloaded. This should not significantly affect startup time, on my laptop loading of 3000 timelines with only metadata took ~0.8 second. 
* Timeline retrieval API changed to reflect two states, loaded and unloaded. So caller can 
* Loaded timeline is a usual LayeredTimeline with loaded layer map etc. Unloaded timeline contains only metadata.
* No on-demand timeline downloading. The only place where timeline is scheduled for download is timeline_attach http handler.

Http api also changed so timeline state can be observed externally. (by tests or by console)

This still has rough edges. For example Repository creation does not accept list of locally complete timelines, so on startup I call apply_timeline_remote_sync_status_update(Downloaded) to register locally available timeline. I plan to polish this api further.

Another planned improvement is avoiding timelines directory scanning by gc. Gc should use repository timelines map. This may remove the need for gc lock. This way we can avoid the problem when gc tries to pickup half extracted remote timeline from the filesystem. apply_timeline_remote_sync_status_update registers timeline in repository only when it is complete locally so repository cannot observe some intermediate state.

Note that there are some places with TODO (current) these ones will be addressed before merge.

pageserver/src/remote_storage/storage_sync/index.rs
```diff
@@ -109,23 +113,49 @@ impl RemoteTimelineIndex {
...
+            .ok_or_else(|| anyhow::anyhow!("unknown timeline sync {}", id))?
+            .set_awaits_download(awaits_download);
+        Ok(())
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Default)]
+pub struct DescriptionTimelineIndexEntry {
+    pub description: BTreeMap<ArchiveId, ArchiveDescription>,
+    pub awaits_download: bool,
```
SomeoneToIgnore, line 132: I wonder, if we need more than this status in general?
For instance, cases when the timeline was evicted on the error threshold from the sync queue, we can remember that and disallow further synchronization?

NIT: Also feels like we could move this one level up in a parent enum, to ease its access?

pageserver/src/remote_storage/storage_sync.rs
```diff
@@ -545,29 +561,23 @@ async fn loop_step<
...
-            remote_assets
-                .as_ref()
-                .1
-                .read()
-                .await
-                .timeline_entry(&task.sync_id)
-                .and_then(TimelineIndexEntry::disk_consistent_lsn),
-        ));
+        FATAL_TASK_FAILURES.inc();
+        return None;
```
SomeoneToIgnore, line 580: If we ignore an upload, we would never be able to retrieve that lost checkpoint file list again, AFAIK.
We would allow to continue uploading further checkpoints for such timelines, which might lead to us having timelines with "holes" we never detect.

I would disallow further uploads for now, to make it more explicit? And hold such "evicted" status in the index, maybe?

pageserver/src/remote_storage/storage_sync/download.rs
```diff
@@ -69,31 +69,36 @@ pub(super) async fn download_timeline<
             return DownloadedTimeline::Abort;
         }
         Some(index_entry) => match index_entry {
-            TimelineIndexEntry::Full(remote_timeline) => Cow::Borrowed(remote_timeline),
+            TimelineIndexEntry::Full(full_entry) => Cow::Borrowed(&full_entry.remote_timeline),
             TimelineIndexEntry::Description(_) => {
                 let remote_disk_consistent_lsn = index_entry.disk_consistent_lsn();
+                let timeline_awaits_download = index_entry.get_awaits_download();
```
SomeoneToIgnore, line 75: Is it so, that we don't actually prevent concurrent downloads of the same timeline with this field?
Feels like we should abort here on `true`, shouldn't we? 

In a way, we're protected by the `attach_timeline` handle logic if it's implemented right, but such data races are IMO bad enough to protect from them here too.

SomeoneToIgnore (commented):
I would expect that this shared index would allow us to fix the remote storage test
https://github.com/zenithdb/zenith/blob/ad262a46ad7966e479c3e2510a48867cac96ba61/test_runner/batch_others/test_remote_storage.py#L29-L31

and that's why got a bit surprised to see no such checks? Same feeling to see no such checks for critical failures and so, to disallow sending bad data to S3, but maybe I'm missing something.

Otherwise LGTM: besides obvious TODOs, feels like a good step forward to unification of both worlds. I think, if we get any practical result such as a fixed test in here, all those TODOs won't matter much 🙂

pageserver/src/branches.rs
```diff
@@ -56,39 +58,63 @@ impl BranchInfo {
...
-        let timeline = match repo.get_timeline(timeline_id)? {
-            RepositoryTimeline::Local(local_entry) => local_entry,
-            RepositoryTimeline::Remote { .. } => {
-                bail!("Timeline {} is remote, no branches to display", timeline_id)
+        let timeline = repo
+            .get_timeline(timeline_id)
+            .ok_or_else(|| anyhow::anyhow!("unknown timeline {}", timeline_id))?;
+
+        let branch_info = match timeline {
+            RepositoryTimeline::Loaded(timeline) => {
```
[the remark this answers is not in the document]
LizardWizzard: It is used only in branch_detail, which is supposed to be removed in your #1286. See actual timeline_detail handler for details. It tries to provide both local and remote data.  We'll have to merge our representations though :)

pageserver/src/layered_repository.rs
```diff
@@ -134,26 +135,32 @@ pub struct LayeredRepository {
...
     upload_relishes: bool,
 }
 
 /// Public interface
 impl Repository for LayeredRepository {
-    fn get_timeline(&self, timelineid: ZTimelineId) -> Result<RepositoryTimeline> {
+    fn get_timeline(&self, timelineid: ZTimelineId) -> Option<RepositoryTimeline> {
+        let timelines = self.timelines.lock().unwrap();
+        self.get_timeline_internal(timelineid, &timelines)
+            .map(Into::into)
```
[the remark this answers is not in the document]
LizardWizzard: Yeah, thanks! I already replaced my Into impl with a from but forgot to change the actual usage

pageserver/src/remote_storage/storage_sync.rs
LizardWizzard: What do you think about rescheduling instead of ignoring and an additional check for uploads, so there is no upload with smaller lsn in the queue?

pageserver/src/repository.rs
```diff
@@ -66,62 +82,55 @@ pub trait Repository: Send + Sync {
     /// perform one checkpoint iteration, flushing in-memory data on disk.
     /// this function is periodically called by checkponter thread.
     fn checkpoint_iteration(&self, cconf: CheckpointConfig) -> Result<()>;
+
+    /// detaches locally available timeline by stopping all threads and removing all the data.
+    fn detach_timeline(&self, timeline_id: ZTimelineId) -> Result<()>;
+
+    // Allows to retrieve remote timeline index from the repo. Used in walreceiver to grab remote consistent lsn.
+    fn get_remote_index(&self) -> &Arc<tokio::sync::RwLock<RemoteTimelineIndex>>;
```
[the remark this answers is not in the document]
LizardWizzard: To do that I have to mark the method as async (to retrieve the entry from index), and currently it is supported because we use dyn Repository

pageserver/src/remote_storage/storage_sync.rs
SomeoneToIgnore: That could bloat our queue with tasks that will always fail, couldn't it?
> so there is no upload with smaller lsn in the queue

Such check IMO also a bit dangerous, since someone could implement the checkpointing diffrently.
This also worsens the "constant fails" scenario in case new checkpoints will arrive after the problematic one.

I have no good solution to the issue, shutting down timeline uploads entirely does not seem right either.

------
We might allow uploading checkpoints in any order, yet keep track of "holes" and the files that were lost: then we can easily track such timelines and never touch their metadata files (yet allowing to sync other files)

pageserver/src/repository.rs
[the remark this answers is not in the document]
SomeoneToIgnore: One last idea: what if we hide another runtime there, to do the `runtime.block_on` and keep it sync?

pageserver/src/layered_repository.rs
```diff
@@ -324,140 +316,120 @@ impl Repository for LayeredRepository {
...
-                .local_or_schedule_download(self.tenantid)
-                .is_some()
-            {
-                TimelineSyncState::Ready(timeline_entry.disk_consistent_lsn())
-            } else {
-                TimelineSyncState::CloudOnly(timeline_entry.disk_consistent_lsn())
-            },
-        )
+    fn get_remote_index(&self) -> &Arc<tokio::sync::RwLock<RemoteTimelineIndex>> {
+        &self.remote_index
```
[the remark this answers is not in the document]
LizardWizzard: +1 for as_ref

All other methods in the repo use `get_` prefix so I'll leave it as is for now

pageserver/src/remote_storage/storage_sync.rs
LizardWizzard: > That could bloat our queue with tasks that will always fail, couldn't it?

Thats true, but I do not see a better option

> This also worsens the "constant fails" 

Could you elaborate on that? I see that there would be more failures because of the incorrect order indeed

> shutting down timeline uploads entirely does not seem right either 

Yeah, I think that's even worse than having a big queue. I think in that case backpressure should kick in because if checkpoints are not uploaded then we do not advance remote consistent lsn. We cannot stop checkpointing either. Maybe we can introduce some limit for number of checkpoints that are in progress of uploading, not sure

pageserver/src/remote_storage/storage_sync/download.rs
LizardWizzard: Added a check here

pageserver/src/repository.rs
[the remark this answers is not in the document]
LizardWizzard: There is even [blocking_read](https://docs.rs/tokio/latest/tokio/sync/struct.RwLock.html#method.blocking_read) method for RwLock, but I'm not sure that we need to add another level of sync/async indirection. Wal receiver already uses a runtime, so we anyway will call this function from async context, so I would prefer to keep it as is, though I agree that this API is not ideal, thoughts/objections?

pageserver/src/remote_storage/storage_sync/index.rs
LizardWizzard: > Also feels like we could move this one level up in a parent enum, to ease its access?

Done, thanks!

> was evicted on the error threshold

I think if we decide to go with this eviction marker it should be placed in timeline metadata, so this eviction marker can survive restart.

> of the same checkpoint

Indeed this behavior is not desirable, but it should be harmless. It would also be different from awaits download because awaits download is scoped for a timeline, not for a particular archive, but for checkpoints it should be ok to have more than one in a queue, so IMO we shouldn't restrict this behavior. I do not see particular case when this could happen either, so I would rather keep it outside from the scope for this PR if you agree with that

SomeoneToIgnore: Sure.

Let's have the python test reenabled as a proof that we're able to use S3 normally (to enable and test real S3 without major issues) and the rest can come any time later.

pageserver/src/remote_storage/storage_sync.rs
SomeoneToIgnore: > Could you elaborate on that?

Just thought that with the 

> an additional check for uploads, so there is no upload with smaller lsn in the queue

approach we're either making too many gaps in the remote checkpoint collection or forced to reschedule those tasks, they will accumulate and task processing will degrade over time, since required to go through more and more tasks, checking.

Either way, no good solution now is fine, we don't have to force it now, let's merge the test fix.

pageserver/src/remote_storage/storage_sync/download.rs
LizardWizzard: Actually check here is wrong, because this code can be is triggered for the same first download. In attach call we set awaits download to true and then when this code is executed awaits download here being true doesnt stand for a concurrent download

pageserver/src/remote_storage/storage_sync.rs
LizardWizzard: Mentioned this in the #977 plan to work on that in a follow up

pageserver/src/layered_repository.rs
LizardWizzard: added to the follow up list in #977

pageserver/src/remote_storage/storage_sync/index.rs
LizardWizzard: tests are reenabled and green, eviction or other solution to critical error handling is added to #977 as a follow up

pageserver/src/remote_storage/storage_sync/upload.rs
LizardWizzard: This was changed in a second iteration, so no longer applicable

pageserver/src/layered_repository.rs
LizardWizzard: Evicted state is removed for now, will see how it works out in a follow up

LizardWizzard:
So cool to see every check with a green checkmark, haven't seen that for a long time :)

I think it is good enough to merge as is to avoid more rebases (this one for removing branch files PR was a tough one)

Tests for tenant relocation and remote storage are reenabled. I think there shouldnt be any major problems with that anymore.

I did more changes to http api, but provided a backwards compatible endpoint for timeline_detail. If everything goes as planned old handler can be removed after changes are supported in console. e2e tests are green so I assume that console is fine with my changes :)

- [ ] Check that swagger spec matches the api, add timeline_detail_v2

For sure will conflict with #1350. As I can see it shouldnt be too bad because we touch different parts. For layered repo this PR touches only get_timeline* methods and changes in `TimelineInfo` struct (and few other minor things). @hlinnaka please take a look at the API changes

pageserver/src/remote_storage/storage_sync/download.rs
```diff
@@ -80,16 +87,19 @@ pub(super) async fn download_timeline<
                     Ok(remote_timeline) => Cow::Owned(remote_timeline),
                     Err(e) => {
                         error!("Failed to download full timeline index: {:?}", e);
+                        // TODO (current) confusing place, we save remote_disk_consistent_lsn
```
LizardWizzard, line 90: @SomeoneToIgnore WDYT?

pageserver/src/layered_repository.rs
```diff
@@ -606,88 +608,70 @@ impl LayeredRepository {
         let now = Instant::now();
 
         // grab mutex to prevent new timelines from being created here.
-        let _gc_cs = self.gc_cs.lock().unwrap();
+        let _gc_cs = self.gc_cs.lock().unwrap(); // BLOCKED HERE
```
SomeoneToIgnore, line 611: Some debugging leftovers? Worth a bit bigger elaboration, if not.

SomeoneToIgnore: Just noticed another one below, so feels like a leftover after all.

pageserver/src/tenant_mgr.rs
```diff
@@ -255,19 +252,27 @@ pub fn get_repository_for_tenant(tenantid: ZTenantId) -> Result<Arc<dyn Reposito
...
     let tenant = m
         .get(&tenantid)
-        .with_context(|| format!("Tenant not found for tenant {}", tenantid))?;
+        .with_context(|| format!("Tenant {} not found", tenantid))?;
 
     Ok(Arc::clone(&tenant.repo))
 }
 
+// Retrieve timeline for tenant without loading whole timeline state into memory
 pub fn get_timeline_for_tenant(
```
SomeoneToIgnore, line 261: Interesting, looks like we don't use this function nowhere in the code? 
If we apply YAGNI and remove it, we'll get back to "one method gets and loads the timeline" situation that we've had before, but with a good name now.

Not sure if we really need those trait adjustments now, but not insisting on any changes now either.

pageserver/src/timelines.rs
```diff
@@ -189,15 +168,45 @@ pub fn init_pageserver(
     Ok(())
 }
 
+pub enum CreateRepo {
+    Real {
+        wal_redo_manager: Arc<dyn WalRedoManager + Send + Sync>,
+        remote_index: Arc<tokio::sync::RwLock<RemoteTimelineIndex>>,
```
SomeoneToIgnore, line 174: `Arc<tokio::sync::RwLock<RemoteTimelineIndex>>`

IMO, that could be a newtype like `Index()` or a `pub type Index = ` for simplicity, that would hide tokio and our hacks around better, we can even provide a sync api for the parts that we need.
Another non-big follow-up task, I guess.

pageserver/src/timelines.rs
```diff
@@ -299,30 +309,24 @@ fn bootstrap_timeline(
...
-        .filter_map(|timeline| match timeline {
-            RepositoryTimeline::Local { timeline, id } => Some((id, timeline)),
-            RepositoryTimeline::Remote { .. } => None,
-        })
-        .map(|(timeline_id, timeline)| {
-            TimelineInfo::from_dyn_timeline(
-                tenant_id,
-                timeline_id,
-                timeline.as_ref(),
+    let mut local_timelines = vec![];
```
SomeoneToIgnore, line 319: NIT: `Vec::with_capacity()` of the listed timelines would be slightly more reasonable.

test_runner/batch_others/test_remote_storage.py
```diff
@@ -54,13 +55,21 @@ def test_remote_storage_backup_and_restore(zenith_env_builder: ZenithEnvBuilder,
...
+            cur.execute("SELECT pg_current_wal_flush_lsn()")
+            current_lsn = int(cur.fetchone()[0].split('/')[1], base=16)
+
+    # wait until pageserver receives that data
+    wait_for_last_record_lsn(client, UUID(tenant_id), UUID(timeline_id), current_lsn)
 
     # run checkpoint manually to be sure that data landed in remote storage
     with closing(env.pageserver.connect()) as psconn:
         with psconn.cursor() as pscur:
             pscur.execute(f"do_gc {tenant_id} {timeline_id}")
```
SomeoneToIgnore, line 67: > do_gc

Let's replace it with `checkpoint` here and maybe in the replication test too? Seems like a one word change that can save us a lot of trouble.

Totally fine with a separate follow-up task for this.

SomeoneToIgnore (approved):
I like where it moves things generally even with all the corner cases discovered.

IMO, the V1-V2 APIs separation should not stay in the codebase long and get removed as soon as the console integration is added.

pageserver/src/remote_storage/storage_sync/download.rs
SomeoneToIgnore: Yes, I don't like it too, might be worth straightening up a bit, but won't put too much effort: this code has to go, since based on `s3 list` and etcd is around the corner.

control_plane/src/storage.rs
```diff
@@ -86,7 +85,10 @@ impl PageServerNode {
                 &env.pageserver.listen_pg_addr,
             ),
             env: env.clone(),
-            http_client: Client::new(),
+            http_client: Client::builder()
+                .timeout(Duration::from_secs(240))
```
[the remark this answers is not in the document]
LizardWizzard: Oh, this is the debugging code slipped into the final version. It was needed when I've added extensive logging around locks to investigate deadlock issue and that significantly slowed down everything so increased timeout was needed. By the way sometimes I need bigger timeouts for some of the our tests to pass on my laptop.

Thanks for spotting it :)

pageserver/src/remote_storage/storage_sync/download.rs
LizardWizzard: s3 list is orthogonal to etcd introduction as I can see, so this is still relevant

pageserver/src/timelines.rs
```diff
@@ -12,135 +13,126 @@ use std::{
...
-    match timeline.get_current_logical_size_non_incremental(timeline.get_last_record_lsn()) {
-        Ok(size) => Some(size),
-        Err(e) => {
-            error!("Failed to get non-incremental logical size: {:?}", e);
-            None
-        }
-    }
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct RemoteTimelineInfo {
+    pub remote_consistent_lsn: Option<Lsn>,
```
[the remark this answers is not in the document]
LizardWizzard: realistically console shouldn't operate with these lsns at all, they are a detail of storage implementation and used only in storage tests. If these are changed to strings then tests would need to parse them back so I'm not sure about that change

pageserver/src/layered_repository.rs
```diff
@@ -324,158 +327,144 @@ impl Repository for LayeredRepository {
...
+    // Differences from the public:
+    //  * interface in that the caller must already hold the mutex on the 'timelines' hashmap.
+    //  * for internal consumers of this function allows to handle non existing timelines
+    fn get_timeline_load_internal(
         &self,
         timelineid: ZTimelineId,
         timelines: &mut HashMap<ZTimelineId, LayeredTimelineEntry>,
-    ) -> Result<LayeredTimelineEntry> {
+    ) -> anyhow::Result<Option<Arc<LayeredTimeline>>> {
+        // TODO (current) unit test for that
```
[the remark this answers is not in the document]
LizardWizzard: Added a test

pageserver/src/layered_repository.rs
```diff
@@ -324,158 +327,144 @@ impl Repository for LayeredRepository {
...
+                return Ok(None);
             }
-        }
+        };
+        let timeline = self.load_local_timeline(timelineid, timelines)?;
+        let was_loaded = timelines.insert(
+            timelineid,
+            LayeredTimelineEntry::Loaded(Arc::clone(&timeline)),
+        );
+        assert!(
```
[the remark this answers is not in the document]
LizardWizzard: changed to ensure here and in one other place

pageserver/src/layered_repository.rs
LizardWizzard: Cleaned this up in all places

pageserver/src/tenant_mgr.rs
LizardWizzard: Removed

test_runner/batch_others/test_remote_storage.py
LizardWizzard: For some reason I had to increase the margin percent in checks to 0.03 after that change, previously with gc there was no lag at all

pageserver/src/remote_storage/storage_sync/download.rs
SomeoneToIgnore: Sorry, late night comments are hard sometimes for me.

A bit more of an elaboration:

If I've got your ideas about mitigating `s3 list` right, we would access the timeline contents via reading a certain `timeline_dir/index.json` file (naming is arbitrary), thus "download full timeline index" operation could be omitted entirely, if we store the index in the json file.
Alternatively, there's still a way to go with storing the whole remote index entirely in some file and sync those.

etc enforces the switch from `s3 list` towards either of the sync models, since we're finally able to communicate between everything and also limit the number of timelines the pageserver have to care about (due to assigning).

Also Lsm approach at its current form mainly uploads a single file to S3 (unless the PR had changed again, will need to recheck the final form). In general, the whole Lsm approach was supposed to make low level compression work better, ideally making local fs files synced with S3 directly, without compression.

If I remember correctly, alternative approaches from Konstantin planned to upload to S3 incremental db snapshots(?), which is not that many files either.

----

All above are plans for the relatively near future, so I think very soon we most likely won't have any S3 compression or indices anymore, so I still think this code

> worth straightening up a bit, but won't put too much effort

if worth straightening at all.

pageserver/src/timelines.rs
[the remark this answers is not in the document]
SomeoneToIgnore: Yeah, Lsns are not anyhow used in the console now, at least not in the parts related to timeline and tenant creation.

But still worth to note that all other http API except this V2 method return Lsn hex strings, not numbers (which technically won't be fitting in some language's ranges for integers, maybe that's the reason).

I'm fine with leaving it as is: nobody uses the V2 endpoint now, IMO V1 has to go next time we adjust the console code,   so we'll change pageserver anyway and can deal with those `u64` vs `String` at that time too.

test_runner/batch_others/test_remote_storage.py
SomeoneToIgnore: That's odd, but thanks for looking into that.

If I've got it right, Lsm PR might change the way we approach the whole GC approach, so the timing to change this is great anyway.

SomeoneToIgnore (approved):

pageserver/src/remote_storage/storage_sync/download.rs
LizardWizzard: Got your point, I agree with that so I'll remove the comment and we'll see how it goes with new format. Didnt look into details there yet

PR #1165, opened 2022-01-22
Use zimg as builders

cicdteam:
use own `zimg/rust:1.56` in CircleCI flow

- avoid installing deps in ci
- use one image for project build and py tests running
- more reasonable executors naming


.circleci/config.yml
```diff
@@ -38,24 +37,14 @@ jobs:
         # Note this works even though the submodule hasn't been checkout out yet.
       - run:
           name: Get postgres cache key
-          command: |
-            git rev-parse HEAD:vendor/postgres > /tmp/cache-key-postgres
+          command: git rev-parse HEAD:vendor/postgres > /tmp/cache-key-postgres
```
funbringer, line 40: Do you believe this (and similar) change is for the better? Personally, I find this a bit harder to read (those pipes "|" make the next line of code stand out). Also, this causes `git diff` to be slightly bigger than necessary if somebody appends a new line.

funbringer (approved):
Great! I've left a minor styling comment.

.circleci/config.yml
```diff
@@ -1,29 +1,28 @@
...
 
 executors:
-  zenith-build-executor:
+  zenith-xlarge-executor:
     resource_class: xlarge
     docker:
       # NB: when changed, do not forget to update rust image tag in all Dockerfiles
-      - image: cimg/rust:1.56.1
-  zenith-python-executor:
+      - image: zimg/rust:1.56
```
LizardWizzard, line 8: I wonder is it important which version of python we use? Probably not so much, especially with poetry

.circleci/config.yml
cicdteam: it's just about a habit. Usually `|` used for multi-line scripts but when you need execute just one command I see no any advantages this notation.

.circleci/config.yml
cicdteam: I have used 3.7.12 as @funbringer suggested, but we can setup another at corresponding [Dockerfiles](https://github.com/zenithdb/docker-images/blob/main/rust/1.56.1/Dockerfile#L56)

PR #1173, opened 2022-01-24
store performance test results in our staging cluster

### LINKED ISSUE #1129: build dashboard with graphing CI test results between runs
this is about the output of this test: https://app.circleci.com/pipelines/github/zenithdb/zenith/4049/workflows/25af709b-ca88-4c53-9279-4fc2bb5aba70/jobs/36408/steps

implementation ideas:
add a step to push the results somewhere
it's likely that prometheus isn't a best fit for this data, so ztoring in postgres, in zenith is a good idea!

LizardWizzard:
to be able to visualize them in grafana

resolves #1129

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -297,7 +297,7 @@ jobs:
             - PLATFORM: zenith-local-ci
           command: |
             PERF_REPORT_DIR="$(realpath test_runner/perf-report-local)"
-            rm -rf $PERF_REPORT_DIR || true
+            rm -rf $PERF_REPORT_DIR
 
             TEST_SELECTION="test_runner/<< parameters.test_selection >>"
             EXTRA_PARAMS="<< parameters.extra_params >>"
@@ -342,7 +342,6 @@ jobs:
 
             if << parameters.save_perf_report >>; then
               if [[ $CIRCLE_BRANCH == "main" ]]; then
-                # TODO: reuse scripts/git-upload
                 export REPORT_FROM="$PERF_REPORT_DIR"
                 export REPORT_TO=local
                 scripts/generate_and_push_perf_report.sh
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -297,6 +297,7 @@ jobs:
             - PLATFORM: zenith-local-ci
           command: |
             PERF_REPORT_DIR="$(realpath test_runner/perf-report-local)"
+            rm -rf $PERF_REPORT_DIR || true
 
             TEST_SELECTION="test_runner/<< parameters.test_selection >>"
             EXTRA_PARAMS="<< parameters.extra_params >>"
--- a/.github/workflows/benchmarking.yml
+++ b/.github/workflows/benchmarking.yml
@@ -91,7 +91,7 @@ jobs:
       run: |
         # just to be sure that no data was cached on self hosted runner
         # since it might generate duplicates when calling ingest_perf_test_result.py
-        rm -rf perf-report-staging || true
+        rm -rf perf-report-staging
         mkdir -p perf-report-staging
         ./scripts/pytest test_runner/performance/ -v -m "remote_cluster" --skip-interfering-proc-check --out-dir perf-report-staging
 
--- a/.github/workflows/benchmarking.yml
+++ b/.github/workflows/benchmarking.yml
@@ -2,8 +2,8 @@ name: benchmarking
 
 on:
   # uncomment to run on push for debugging your PR
-  # push:
-  #   branches: [ mybranch ]
+  push:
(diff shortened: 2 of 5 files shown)


stepashka (approved):
I suggest to explicitly state the units in panels in grafana. This could be done by setting the Unit in the Standard options in the panel settings, of can be added to the Panel title.

Couple of ideas about showing revisions.
I don't think it's possible to add revision tooltips to each point on a Time Series panel. But here are the other alternatives.

- Build the annotation query with [region events](https://grafana.com/docs/grafana/latest/dashboards/annotations/#adding-regions-events), with start and end of the test run for each revision.

- Make use of [Table panel](https://grafana.com/docs/grafana/latest/visualizations/table/). Have a table panel showing the the list of revisions with their times on the graph, maybe include the the last commit message too. You could then use the [Data Links](https://grafana.com/docs/grafana/latest/linking/data-links/) to allow selecting a specific revision in the table and inspecting only the results related to that revision.

- Make more use of [Table panel](https://grafana.com/docs/grafana/latest/visualizations/table/), and transform some (but not all) of the graph panels into a table. In a table use formatting or transform the data to make it more easily readable. E.g. showing the percentage difference between revisions or using conditional formatting to indicate the trend.

- The last idea is more of a hack, not sure it would work. In Dashboard settings -> General -> Panel options (in the bottom) select the Shared Tooltip. Now all the series names and values on top the the Time Series panels will show up in the tooltips. Maybe this helps get the data you want into the tooltip, not sure how to deal with excessive data that you don't need, you may have to get rid of the legend. :)

I think what we have already is a great result.

LizardWizzard:
> I suggest to explicitly state the units in panels in grafana. This could be done by setting the Unit in the Standard options in the panel settings, of can be added to the Panel title.

Different metrics have different units, unit is stored in the table with results too, but as I mentioned for revision I didnt find a way to bind it to the value. The way to solve it might be to use separate panel for each series (but then for each series I have to extract the unit somehow), and group panels for one test in one row but that might occupy too much space. What do you think?

I'll experiment with your suggestions, thanks!

cicdteam:
sorry, can't help with it now, just approving

.circleci/config.yml
```diff
@@ -292,6 +292,7 @@ jobs:
             - PLATFORM: zenith-local-ci
           command: |
             PERF_REPORT_DIR="$(realpath test_runner/perf-report-local)"
+            rm -rf $PERF_REPORT_DIR || true
```
funbringer, line 295: Why is `|| true` needed here?

.github/workflows/benchmarking.yml
```diff
@@ -89,11 +89,15 @@ jobs:
         BENCHMARK_CONNSTR: "${{ secrets.BENCHMARK_STAGING_CONNSTR }}"
         REMOTE_ENV: "1" # indicate to test harness that we do not have zenith binaries locally
       run: |
+        # just to be sure that no data was cached on self hosted runner
+        # since it might generate duplicates when calling ingest_perf_test_result.py
+        rm -rf perf-report-staging || true
```
funbringer, line 94: Ditto

scripts/generate_and_push_perf_report.sh
```diff
@@ -6,11 +6,14 @@ set -eux -o pipefail
...
-git clone https://$VIP_VAP_ACCESS_TOKEN@github.com/zenithdb/zenith-perf-data.git
+git clone --single-branch --branch testing https://$VIP_VAP_ACCESS_TOKEN@github.com/zenithdb/zenith-perf-data.git  # TODO (PRIOR MERGE) replace back before merge
 cd zenith-perf-data
 mkdir -p reports/
 mkdir -p data/$REPORT_TO
 
+# ingest per test results data into zenith backed postgres running in staging to build grafana reports on that data
+DATABASE_URL=$PERF_TEST_RESULT_CONNSTR poetry run python $SCRIPT_DIR/ingest_perf_test_result.py --ingest $REPORT_FROM
+
 cp $REPORT_FROM/* data/$REPORT_TO
```
funbringer, line 17: Friendly reminder: we could reuse `scripts/git-upload` here, but I won't mind leaving this as-is.

LizardWizzard: Git upload needs some modifications then, html report is generated over all data in the repo, so in that case I have to inject report generation command inside git upload. Something like `git-upload copy ... --run-cmd='<update report>'`. Is it ok to have that in git upload?

funbringer: > Is it ok to have that in git upload?

I won't mind.

funbringer (approved):

PR #1177, opened 2022-01-25
use 2021 rust edition

LizardWizzard:
since we target 1.56+ lets use new edition


funbringer (approved):
Thanks!

SomeoneToIgnore:
The new edition [employs](https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html) the new cargo feature resolver, so presumably `workspace_hack` gets more obsolete now? Should we get rid of it?

LizardWizzard:
Thanks for bringing this up! I think so, I'll check how it feels without it

PR #1181, opened 2022-01-25
Improve coverage tools

funbringer:
This patch optimizes coverage data collection which leads to measurable reduction of CI times (~ -5 mins). However, I was disappointed to see that CircleCI doesn't handle incremental workspace updates very well (a couple of MiBs of coverage data still results in full-blown 1.6 GiB re-uploads; what a misfortune). Maybe @cicdteam knows something.

[Old](https://app.circleci.com/pipelines/github/zenithdb/zenith/4222/workflows/9aca6ccf-cdcb-4025-bcf3-8b6603f2c2d4/jobs/38586) vs [New](https://app.circleci.com/pipelines/github/zenithdb/zenith/4225/workflows/7391a2fb-d311-4548-b7f6-f3ab4764ed74/jobs/38623) (note the last two steps).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -194,6 +194,14 @@ jobs:
           command: |
             cp -a tmp_install /tmp/zenith/pg_install
 
+      - run:
+          name: Merge coverage data
+          command: |
+            # This will speed up workspace uploads
+            if [[ $BUILD_TYPE == "debug" ]]; then
+              scripts/coverage "--profraw-prefix=$CIRCLE_JOB" --dir=/tmp/zenith/coverage merge
+            fi
+
         # Save the rust binaries and coverage data for other jobs in this workflow.
       - persist_to_workspace:
           root: /tmp/zenith
@@ -354,6 +362,13 @@ jobs:
       # The store_test_results step tells CircleCI where to find the junit.xml file.
       - store_test_results:
           path: /tmp/test_output
+      - run:
+          name: Merge coverage data
+          command: |
+            # This will speed up workspace uploads
+            if [[ $BUILD_TYPE == "debug" ]]; then
+              scripts/coverage "--profraw-prefix=$CIRCLE_JOB" --dir=/tmp/zenith/coverage merge
+            fi
       # Save coverage data (if any)
       - persist_to_workspace:
           root: /tmp/zenith
--- a/scripts/coverage
+++ b/scripts/coverage
@@ -14,17 +14,30 @@ from dataclasses import dataclass
 from pathlib import Path
 from tempfile import TemporaryDirectory
 from textwrap import dedent
-from typing import Any, Iterable, List, Optional
+from typing import Any, Dict, Iterator, Iterable, List, Optional
 
 import argparse
+import hashlib
 import json
 import os
 import shutil
+import socket
 import subprocess
 import sys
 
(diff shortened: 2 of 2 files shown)


scripts/coverage
```diff
@@ -205,55 +200,107 @@ class LLVM:
...
+# TODO: implement
+@dataclass
+class ProfDir:
+    cwd: Path
+    llvm: LLVM
+
+    def __post_init__(self) -> None:
+        self.cwd.mkdir(parents=True, exist_ok=True)
+
+    @property
```
LizardWizzard, line 212: if this field accessed several times it might be be good to utilize cached_property

scripts/coverage
```diff
@@ -205,55 +200,107 @@ class LLVM:
         self._cov(subcommand='show', *extras, **kwargs)
 
 
+# TODO: implement
```
LizardWizzard, line 203: Done I guess? :)

scripts/coverage
```diff
@@ -205,55 +200,107 @@ class LLVM:
...
+        if files_mtime >= profdata_mtime:
+            self.llvm.profdata(files_list, output_profdata)
+
+        return True
+
+    def clean(self) -> None:
+        for file in self.cwd.iterdir():
+            os.remove(file)
+
+    def __truediv__(self, other):
```
LizardWizzard, line 244: 👍🏻

scripts/coverage
```diff
@@ -266,9 +313,9 @@ class HtmlReport(Report):
 @dataclass
 class GithubPagesReport(HtmlReport):
     output_dir: Path
-    commit_url: str
+    commit_url: str = 'https://local/deadbeef'
```
LizardWizzard, line 316: Probably unintentional

.circleci/config.yml
```diff
@@ -340,6 +348,13 @@ jobs:
       # The store_test_results step tells CircleCI where to find the junit.xml file.
       - store_test_results:
           path: /tmp/test_output
+      - run:
+          name: Merge coverage data
+          command: |
+            # This will speed up workspace uploads
+            if [[ $BUILD_TYPE == "debug" ]]; then
+              scripts/coverage "--profraw-prefix=$CIRCLE_JOB" --dir=/tmp/zenith/coverage merge
```
LizardWizzard, line 356: $CIRCLE_JOB contains something that needs to be escaped?

LizardWizzard (approved):
Nice! Looks good to me, left few minor nitpicks

scripts/coverage
funbringer: No, in fact, I added this to provide a fallback for local rendering. This dummy url will make it to the report.

scripts/coverage
funbringer: No, I don't think that's a good idea. Sure, it's somewhat costly, but if used carefully, it won't do any harm at all. Just think of the subtle bugs that could happen with such a change.

LizardWizzard: You mean that file list can change on the fly? I think if this is the case many things will get broken anyway. Another option to better describe the intention might be to use get_files or similar name instead. For me `property` always means  something cheap to recreate every time it is accessed. It is not that important though, I suppose there wouldn't be too much files anyway

PR #1184, opened 2022-01-25
cache python deps in circle ci

LizardWizzard:
Previously it took ~30 seconds to install python dependencies, and they were not cached. We invoke dependency installation in 6 jobs so the time is multiplied by 6. 
Now it takes ~10 seconds, 5 for cache restore and 5 to verify that all dependencies are up to date

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -204,9 +204,16 @@ jobs:
     executor: zenith-executor
     steps:
       - checkout
+      - restore_cache:
+          keys:
+            - v1-python-deps-{{ checksum "poetry.lock" }}
       - run:
           name: Install deps
           command: ./scripts/pysync
+      - save_cache:
+          key: v1-python-deps-{{ checksum "poetry.lock" }}
+          paths:
+            - /home/circleci/.cache/pypoetry/virtualenvs
       - run:
           name: Run yapf to ensure code format
           when: always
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -256,9 +263,16 @@ jobs:
...
+      - restore_cache:
+          keys:
+            - v1-python-deps-{{ checksum "poetry.lock" }}
       - run:
           name: Install deps
           command: ./scripts/pysync
+      - save_cache:
+          key: v1-python-deps-{{ checksum "poetry.lock" }}
+          paths:
+            - /home/circleci/.cache/pypoetry/virtualenvs
```
funbringer, line 275: What do you think about using `poetry config virtualenvs.create false --local` instead?

funbringer (approved):
Great!

[It works](https://app.circleci.com/pipelines/github/zenithdb/zenith/4234/workflows/2531ae4c-28a2-439d-980a-52d613d46eca/jobs/38754/parallel-runs/0/steps/0-104).

.circleci/config.yml
LizardWizzard: Then only cache dir changes to pip's default I guess (or to some other dir inside poetry's .cache) Either way is ok to me, what's your preference? Personally I manage virtualenvs via pyenv so I have this setting locally

Another option might be to cache just *.whl files, not the whole venv, but current approach works well and wheel handling requires some boilerplate, so I would rather stick to current approach

funbringer: > Either way is ok to me, what's your preference?

I don't have a strong preference either, just had a thought that the origin of this path might not be as obvious as the local per-project env. I guess we'll change it if it ever becomes a problem.

PR #1187, opened 2022-01-26
Centralize suspending/resuming timeline activity on safekeepers.

arssher:
Timeline is active whenever there is at least 1 connection from compute or
pageserver is not caught up. Currently 'active' means callmemaybes are being
sent.

Fixes race: now suspend condition checking and callmemaybe unsubscribe happen
under the same lock.


lubennikovaav:
Looks good to me. 
Do we have any test that covers this race condition issue?

arssher:
currently nope, we'd need concurrent computes for that

PR #1228, opened 2022-02-06
Generate safekeeper unique IDs.

arssher:
In preparation for storage node messaging.

### DIFF
--- a/walkeeper/src/bin/safekeeper.rs
+++ b/walkeeper/src/bin/safekeeper.rs
@@ -1,17 +1,19 @@
 //
 // Main entry point for the safekeeper executable
 //
-use anyhow::{Context, Result};
+use anyhow::{bail, Context, Result};
 use clap::{App, Arg};
 use const_format::formatcp;
 use daemonize::Daemonize;
 use fs2::FileExt;
-use std::fs::File;
+use std::fs::{self, File};
+use std::io::{ErrorKind, Write};
 use std::path::{Path, PathBuf};
 use std::thread;
 use tracing::*;
 use walkeeper::timeline::{CreateControlFile, FileStorage};
 use zenith_utils::http::endpoint;
+use zenith_utils::zid::ZNodeId;
 use zenith_utils::{logging, tcp_listener, GIT_VERSION};
 
 use tokio::sync::mpsc;
@@ -25,6 +27,7 @@ use zenith_utils::shutdown::exit_now;
 use zenith_utils::signals;
 
 const LOCK_FILE_NAME: &str = "safekeeper.lock";
+const ID_FILE_NAME: &str = "safekeeper.id";
 
 fn main() -> Result<()> {
     zenith_metrics::set_common_metrics_prefix("safekeeper");
@@ -93,6 +96,9 @@ fn main() -> Result<()> {
                 .takes_value(true)
                 .help("Dump control file at path specifed by this argument and exit"),
         )
+        .arg(
+            Arg::new("id").long("id").takes_value(true).help("safekeeper node id: integer")
+        )
         .get_matches();
 
     if let Some(addr) = arg_matches.value_of("dump-control-file") {
@@ -133,10 +139,19 @@ fn main() -> Result<()> {
         conf.recall_period = humantime::parse_duration(recall)?;
     }
 
-    start_safekeeper(conf)
+    let mut given_id = None;
+    if let Some(given_id_str) = arg_matches.value_of("id") {
+        given_id = Some(ZNodeId(
+            given_id_str
+                .parse()
+                .context("failed to parse safekeeper id")?,
+        ));
+    }
+
+    start_safekeeper(conf, given_id)
 }
 
-fn start_safekeeper(conf: SafeKeeperConf) -> Result<()> {
+fn start_safekeeper(mut conf: SafeKeeperConf, given_id: Option<ZNodeId>) -> Result<()> {
     let log_file = logging::init("safekeeper.log", conf.daemonize)?;
 
     info!("version: {}", GIT_VERSION);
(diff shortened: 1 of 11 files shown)


zenith_utils/src/zid.rs
```diff
@@ -4,168 +4,119 @@ use hex::FromHex;
...
-        f.write_str(&hex::encode(self.0))
-    }
-}
-
 macro_rules! zid_newtype {
-    ($t:ident) => {
-        impl $t {
-            pub fn get_from_buf(buf: &mut dyn bytes::Buf) -> $t {
-                $t(ZId::get_from_buf(buf))
+    // $attr meta argument is for attaching the doc string (which is in fact an
```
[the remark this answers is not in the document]
funbringer: I think that using a macro here is fine, since there's no support for const-generic-based arrays in serde, and fixing this with a boilerplate doesn't look much better (if at all). Instead, I'd propose to make ZNodeIds 128-bit to make this change unnecessary. Sure, that'd be somewhat more wasteful, but does it matter?

walkeeper/src/bin/safekeeper.rs
```diff
@@ -257,3 +272,45 @@ fn start_safekeeper(conf: SafeKeeperConf) -> Result<()> {
...
+                }
+            }
+            info!("safekeeper ID {}", my_id);
+        }
+        Err(error) => match error.kind() {
+            ErrorKind::NotFound => {
+                my_id = if let Some(given_id) = given_id {
+                    given_id
+                } else {
+                    ZNodeId::generate()
```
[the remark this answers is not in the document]
LizardWizzard: Parameters to safekeeper are provided each time it launches, so I do not see a difference there between explicit passing of e.g listener address and node id

zenith_utils/src/zid.rs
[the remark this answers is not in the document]
arssher: Interesting idea, though I also think macro here is better. However I just decided to switch to sequential int8:
https://github.com/zenithdb/zenith/pull/1257#issuecomment-1035265127

The refactoring of zid.rs still makes sense to me though, removing basically empty wrapper type, so I'm going to extract that as separate commit.

arssher:
Updated patch makes ids monotonic with decimal representation. @hlinnaka note that this id completely replaces textual 'name' currently known by `zenith` CLI/fixtures. We can add it back as some optional human readable name, but I doubt it makes much sense now.

LizardWizzard (commented):
Still do not get it why we need different flavours of ids. Reposting here for visibility.

If we need to distinguish between entities that ids belong to, lets do that explicitly. We can reserve some of these id bits to some tag value and use it in Debug/Display impls. So id will look something like that: node/abcd12354 or tenant/1234abcd. IMO it is better than some heuristics. Also if our ids are too long, lets shrink them to minimal possible length. e g to have 8 bytes instead of 16.

zenith/src/main.rs
```diff
@@ -18,13 +18,13 @@ use walkeeper::defaults::{
...
 use zenith_utils::postgres_backend::AuthType;
-use zenith_utils::zid::{ZTenantId, ZTimelineId};
+use zenith_utils::zid::{ZNodeId, ZTenantId, ZTimelineId};
 use zenith_utils::GIT_VERSION;
 
 use pageserver::branches::BranchInfo;
 
 // Default name of a safekeeper node, if not specified on the command line.
-const DEFAULT_SAFEKEEPER_NAME: &str = "single";
+const DEFAULT_SAFEKEEPER_ID: ZNodeId = 1;
```
LizardWizzard, line 27: IMO it is pretty scary to have a default for it. If we miss I’d in cli  by accident we will have different nodes with the same id

arssher:
As said in another PR, I'm for that variant mostly due to shortness/readability/being easy to remember. Right now tenant id + timeline id are taking about half of space of log line; imagine adding to that global installation id + shard id once we get to sharding + storage node id, perhaps several (e.g. pageserver and safekeepers). Most of our logs would be ids %).

If you want another example of such "mixed" ids, consider CockroachDB: they have uuid for global cluster id and sequential decimal ids for node ids. For the same reasons, I would expect.

Random 64 bit ids combine both disadvantages: they are not long enough to regard them unique without additional checks (unlike uuid) and still all values are quite long.

Still, I don't care very much here, I just want to proceed. I wouldn't extremely mind the alternative, and it is not very hard to change later.

LizardWizzard (commented):
If one uuid has enough uniqueness can we shrink tenant/timeline ids in half? So together they form this cluster uuid.

I'm not strongly opposed to sequential ids. My concern is having too many concepts to deal with. If one type of id is enough,  then why bother and invent new ones? We can have one id with some tag in the first byte which describes what kind of id it is without guessing 

Regarding logs I think it is inevitable that more and more variables are being added to message context. Modern logging follows machine readable path, e.g json. With that we can filter out the noise.

Feel free to commit this if you're in a hurry.

zenith_utils/src/zid.rs
```diff
@@ -221,6 +221,10 @@ impl fmt::Display for ZTenantTimelineId {
     }
 }
 
+// Unique ID of a storage node (safekeeper or pageserver). Supposed to be
+// sequentially issued by the console.
+pub type ZNodeId = u64;
```
LizardWizzard, line 226: Lets make it an actual newtype, because this way we can safely assign any u64 to a variable of type ZNodeId

arssher: You mean you want the assignment look like `node_id.0 = x` instead of `node_id = x` to prevent accidental assignment?

LizardWizzard: I mean e g when function parameter or a struct variable expects type that is an alias it accepts the original type that alias points to. See the difference here: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e64599c52d153d3b627d95ca7653ab24

And thats why current ZId derivatives are newtypes instead of aliases, that avoids possible confusion with plain ZId type. So you cannot pass ZId to function that expects ZTenantId

arssher: Ah, ok, thanks.

arssher:
Merged through https://github.com/zenithdb/zenith/pull/1310

PR #1251, opened 2022-02-10
[proxy] WIP Async cplane_api, waiters, auth, and routers

bojanserafimov:
This is a pull request into https://github.com/zenithdb/zenith/tree/funbringer-proxy-async so that @funbringer can get a chance to comment before I merge into his branch.

Changes included:
1. Switch the waiter and the cplane api to async
2. Implement 4 kinds of routers https://github.com/zenithdb/console/issues/502 (also relevant: https://github.com/zenithdb/zenith/issues/920)
3. Implement 3 kinds of authentication
4. Add CLI option for router, while preserving backwards compatibility (the way we currently deploy proxy will still work)
5. Add mock cli to speed up development.

Current state:
All 4 routers seem to work. I tested using the cplane mock. Example commands:
- Run static router: `./proxy --router static`
- Run mixed router with mock cplane: `./proxy --mock-cplane localhost:3000`

TODO:
- [ ] Tidy up
- [ ] Write unit or python test

### DIFF
--- a/proxy/src/waiters.rs
+++ b/proxy/src/waiters.rs
@@ -1,6 +1,8 @@
 use anyhow::Context;
-use std::collections::HashMap;
-use std::sync::{mpsc, Mutex};
+use futures::Future;
+use std::{collections::HashMap, pin::Pin, task};
+use tokio::sync::mpsc;
+use parking_lot::Mutex;
 
 pub struct Waiters<T>(pub(self) Mutex<HashMap<String, mpsc::Sender<T>>>);
 
@@ -12,10 +14,10 @@ impl<T> Default for Waiters<T> {
 
 impl<T> Waiters<T> {
     pub fn register(&self, key: String) -> Waiter<T> {
-        let (tx, rx) = mpsc::channel();
+        let (tx, rx) = mpsc::channel(3);
 
         // TODO: use `try_insert` (unstable)
-        let prev = self.0.lock().unwrap().insert(key.clone(), tx);
+        let prev = self.0.lock().insert(key.clone(), tx);
         assert!(matches!(prev, None)); // assert_matches! is nightly-only
 
         Waiter {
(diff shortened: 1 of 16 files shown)


proxy/src/waiters.rs
```diff
@@ -12,10 +14,10 @@ impl<T> Default for Waiters<T> {
 
 impl<T> Waiters<T> {
     pub fn register(&self, key: String) -> Waiter<T> {
-        let (tx, rx) = mpsc::channel();
+        let (tx, rx) = mpsc::channel(3);
```
funbringer, line 17: Wouldn't [oneshot](https://docs.rs/tokio/latest/tokio/sync/oneshot/index.html) work better here? The rebased version of my branch now uses it.

proxy/src/waiters.rs
```diff
@@ -45,14 +46,16 @@ pub struct Waiter<'a, T> {
...
+        self.receiver.poll_recv(cx).map(|t| t.expect("sender dropped"))
     }
 }
 
 impl<T> Drop for Waiter<'_, T> {
     fn drop(&mut self) {
-        self.registry.0.lock().unwrap().remove(&self.key);
+        self.registry.0.lock().remove(&self.key);
     }
 }
```
funbringer, line 61: Yeah, you nailed it! I incorporated this change.

funbringer (commented):
Thanks for the patch! Unfortunately, I wasn't aware that you had went this far, so this is blocked on various conflicts now. Would you mind postponing this till https://github.com/zenithdb/zenith/pull/1270 is merged? I'd like to deploy async version asap.

proxy/src/waiters.rs
bojanserafimov: Yes, I just didn't want to do it in this PR, to simplify review

bojanserafimov:
Superseded by cleaner code from other PRs that accomplishes the same goal.

PR #1254, opened 2022-02-10
docs/sourcetree.md: add command to set Python version inside poetry's virtualenv

yeputons:


hlinnaka:
No objections, but I wonder why you'd want to specify a different python version?

yeputons:
To have the exact same environment we have in CI. Python 3.8 is the default
on Ubuntu 20.04, while we use Python 3.7 in CI. There were some issues with
package updates with pipenv: different versions were resolved when using
different Python versions, which resulted in huge useless diffs. Hopefully
it's not issue with poetry anymore, though.

cicdteam:
3.7 in CI was suggested by @funbringer  but we can easily change it to 3.8 (or any other) if need.

yeputons:
@LizardWizzard, I think the note is still worth it for completeness: we already warn users that newer Pythons may do something different than our CI.

LizardWizzard:
@yeputons CI now uses 3.9.2 is that still a problem? See https://github.com/neondatabase/docker-images/blob/main/rust/Dockerfile#L59

I'm ok with the change, if it's still needed let's fix the version number in the doc and merge.

P.S. Highly recommend using [pyenv](https://github.com/pyenv/pyenv). It allows you to easily install any python version and even configure it on a per-directory basis. Personally, I have venv for zenith managed by pyenv and it is smoothly picked up by poetry. Maybe for that you'll need something like that in your `poetry.toml`:
```
[virtualenvs]
create = false
```

yeputons:
@LizardWizzard ,

> CI now uses 3.9.2 is that still a problem? 

Hmm, how about we bump the minimal supported version to 3.9 then? There are [some incompatibilities](https://docs.python.org/3/whatsnew/3.9.html) between Python 3.9 and currently claimed minimal version 3.7, which we may accidentally use. `poetry` is smart enough to pick up the Python specified in `pyproject.toml` and err if it's not available, so updating it should be ok.

I'd say using a newer version still may cause problems (e.g., the default on Ubuntu 22.04 LTS is [backward-incompatible](https://www.python.org/downloads/release/python-3100/) Python 3.10), so I'd keep the few lines which tell the user how to pinpoint the version they want.

I intend to merge the new three commits as is without squashing.

LizardWizzard (approved):

PR #1257, opened 2022-02-10
support variable length ZId

LizardWizzard:
Involves some workarounds for serde, which is currently does not support
arrays with length expressed as a const generic parameter. This
workaround might be useful in other places if or when we'll want to use
arrays with const generics.

### DIFF
--- a/control_plane/src/local_env.rs
+++ b/control_plane/src/local_env.rs
@@ -12,7 +12,7 @@ use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
 use zenith_utils::auth::{encode_from_key_file, Claims, Scope};
 use zenith_utils::postgres_backend::AuthType;
-use zenith_utils::zid::{opt_display_serde, ZTenantId};
+use zenith_utils::zid::{serde_opt_display, ZTenantId};
 
 //
 // This data structures represents zenith CLI config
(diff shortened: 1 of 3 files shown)


LizardWizzard:
@hlinnaka see https://github.com/zenithdb/zenith/pull/1228#discussion_r803663579

LizardWizzard:
Actually I think 128bit ids are not that big, since we use one per whole machine (https://github.com/zenithdb/zenith/pull/1228#discussion_r803681711) 

Regarding int8 I think it is too small. But anyway, IMO it is more convenient to use one style for all of our ids (timeline, tenant). They are meant to be managed externally by console, so there may be not much sense in random values for all of them then

LizardWizzard:
@arssher are we switching for timeline id and tenant id for sequential ones too? Do we really need two flavours of ids? isn't int8 too small for that? If we want to serve thousands of tenants on a single cluster consisting of multiple pageservers/safekeepers that seems to be not enough.

arssher:
No, only node ids. There is actually some sense in having ids of different flavours -- you can guess the type by looking at it. I'm for that variant mostly due to shortness/readability/being easy to remember. int8 means 8 bytes, not bits, that's postgres terminology, so surely enough.

LizardWizzard:
@arssher If we need to distinguish between entities that ids belong to, lets do that explicitly. We can reserve some of these id bits to some tag value and use it in Debug/Display impls. So id will look something like that: `node/abcd12354` or `tenant/1234abcd`. IMO it is better than some heuristics. Also if our ids are too long, lets shrink them to minimal possible length. e g to have 8 bytes instead of 16 

> that's postgres terminology,

Good to know, thanks!

PR #1265, opened 2022-02-11
merge directories in git-upload instead of removing existing files for perf test result uploads

LizardWizzard:


funbringer (approved):
It seems to work properly this time :)

PR #1270, opened 2022-02-14
[proxy] Migrate to async

funbringer:
This makes most of the proxy async (except for mgmt, but we expect to drop it soon). The auth logic is left almost untouched, since this patch is big enough, and it would be good to merge it and unlock further changes.

### DIFF
--- a/proxy/src/auth.rs
+++ b/proxy/src/auth.rs
@@ -1,0 +1,127 @@
+use crate::compute::DatabaseInfo;
+use crate::config::ProxyConfig;
+use crate::cplane_api::{self, CPlaneApi};
+use crate::stream::PqStream;
+use anyhow::{anyhow, bail, Context};
+use std::collections::HashMap;
+use tokio::io::{AsyncRead, AsyncWrite};
+use zenith_utils::pq_proto::{BeMessage as Be, BeParameterStatusMessage, FeMessage as Fe};
+
+/// Various client credentials which we use for authentication.
+#[derive(Debug, PartialEq, Eq)]
+pub struct ClientCredentials {
+    pub user: String,
+    pub dbname: String,
+}
+
+impl TryFrom<HashMap<String, String>> for ClientCredentials {
+    type Error = anyhow::Error;
+
+    fn try_from(mut value: HashMap<String, String>) -> Result<Self, Self::Error> {
+        let mut get_param = |key| {
+            value
+                .remove(key)
+                .with_context(|| format!("{} is missing in startup packet", key))
+        };
+
+        let user = get_param("user")?;
+        let db = get_param("database")?;
+
+        Ok(Self { user, dbname: db })
+    }
+}
+
+impl ClientCredentials {
+    /// Use credentials to authenticate the user.
+    pub async fn authenticate(
+        self,
+        config: &ProxyConfig,
+        client: &mut PqStream<impl AsyncRead + AsyncWrite + Unpin>,
+    ) -> anyhow::Result<DatabaseInfo> {
+        let db_info = if self.user.ends_with("@zenith") {
+            handle_existing_user(config, client, self).await
+        } else {
+            handle_new_user(config, client).await
+        };
+
+        db_info.context("failed to authenticate client")
+    }
+}
+
+fn new_psql_session_id() -> String {
+    hex::encode(rand::random::<[u8; 8]>())
+}
+
+async fn handle_existing_user(
+    config: &ProxyConfig,
+    client: &mut PqStream<impl AsyncRead + AsyncWrite + Unpin>,
+    creds: ClientCredentials,
+) -> anyhow::Result<DatabaseInfo> {
+    let psql_session_id = new_psql_session_id();
+    let md5_salt = rand::random();
+
+    client
+        .write_message(&Be::AuthenticationMD5Password(&md5_salt))
+        .await?;
+
+    // Read client's password hash
+    let msg = match client.read_message().await? {
+        Fe::PasswordMessage(msg) => msg,
+        bad => bail!("unexpected message type: {:?}", bad),
+    };
+
+    let (_trailing_null, md5_response) = msg
+        .split_last()
+        .ok_or_else(|| anyhow!("unexpected password message"))?;
+
+    let cplane = CPlaneApi::new(&config.auth_endpoint);
+    let db_info = cplane
+        .authenticate_proxy_request(creds, md5_response, &md5_salt, &psql_session_id)
+        .await?;
+
+    client
+        .write_message_noflush(&Be::AuthenticationOk)?
+        .write_message_noflush(&BeParameterStatusMessage::encoding())?;
+
+    Ok(db_info)
+}
+
+async fn handle_new_user(
+    config: &ProxyConfig,
+    client: &mut PqStream<impl AsyncRead + AsyncWrite + Unpin>,
+) -> anyhow::Result<DatabaseInfo> {
+    let psql_session_id = new_psql_session_id();
+    let greeting = hello_message(&config.redirect_uri, &psql_session_id);
+
+    let db_info = cplane_api::with_waiter(psql_session_id, |waiter| async {
+        // Give user a URL to spawn a new database
+        client
+            .write_message_noflush(&Be::AuthenticationOk)?
+            .write_message_noflush(&BeParameterStatusMessage::encoding())?
+            .write_message(&Be::NoticeResponse(greeting))
+            .await?;
+
+        // Wait for web console response
+        waiter.await?.map_err(|e| anyhow!(e))
+    })
+    .await?;
+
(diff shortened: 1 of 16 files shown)


funbringer:
I've just realized that I forgot about `NUM_BYTES_PROXIED_COUNTER`. While it's quite easy to restore this logic, I'm a bit hesitant, since the current approach means that all runtime threads will compete for atomic var's cacheline. I might be overthinking, because there's an awful lot more happening inside tokio's runtime. Thoughts?

bojanserafimov:
> I've just realized that I forgot about `NUM_BYTES_PROXIED_COUNTER`. While it's quite easy to restore this logic, I'm a bit hesitant, since the current approach means that all runtime threads will compete for atomic var's cacheline. I might be overthinking, because there's an awful lot more happening inside tokio's runtime. Thoughts?

There is some cache contention but we only update when we flush, so I wouldn't expect visible performance impact. If there is any, we can just buffer the metric updates even more aggressively.

bojanserafimov (approved):

proxy/src/cancellation.rs
```diff
@@ -0,0 +1,90 @@
...
+    R: std::future::Future<Output = anyhow::Result<V>>,
+{
+    // HACK: We'd rather get the real backend_pid but tokio_postgres doesn't
+    // expose it and we don't want to do another roundtrip to query
+    // for it. The client will be able to notice that this is not the
+    // actual backend_pid, but backend_pid is not used for anything
+    // so it doesn't matter.
+    let key = rand::random();
+
+    // The birthday problem is unlikely to happen here, but it's still possible
```
LizardWizzard, line 76: Let's expand this comment, it is not that obvious what is the problem here

proxy/src/cplane_api.rs
```diff
@@ -1,106 +1,85 @@
...
-
-#[derive(Serialize, Deserialize, Debug, Default)]
-pub struct DatabaseInfo {
-    pub host: String,
-    pub port: u16,
-    pub dbname: String,
-    pub user: String,
-    pub password: Option<String>,
+lazy_static! {
+    static ref CPLANE_WAITERS: Waiters<Result<DatabaseInfo, String>> = Default::default();
```
LizardWizzard, line 9: I wonder is there any noticeable difference with using wakers directly instead of oneshots? With oneshots it is just less amount of code, but wakers could be cheaper. Thoughts?

proxy/src/cplane_api.rs
```diff
@@ -1,106 +1,85 @@
...
-        let waiter = self.waiters.register(psql_session_id.to_owned());
-
-        println!("cplane request: {}", url);
-        let resp = reqwest::blocking::get(url)?;
-        if !resp.status().is_success() {
-            bail!("Auth failed: {}", resp.status())
-        }
+        with_waiter(psql_session_id, |waiter| async {
+            println!("cplane request: {}", url);
+            let resp = reqwest::get(url).await?;
```
LizardWizzard, line 55: Nit: We can reuse connections with [reqwest::Client](https://docs.rs/reqwest/latest/reqwest/struct.Client.html) ?

proxy/src/cplane_api.rs
```diff
@@ -1,106 +1,85 @@
...
+                Ready { conn_info } => Ok(conn_info),
+                Error { error } => bail!(error),
+                NotReady { .. } => waiter.await?.map_err(|e| anyhow!(e)),
+            }
+        })
+        .await
     }
 }
 
+// NOTE: the order of constructors is important.
```
LizardWizzard, line 74: Lets expand this to mention why so

proxy/src/http.rs
```diff
@@ -1,15 +1,30 @@
...
+
+pub async fn thread_main(http_listener: TcpListener) -> anyhow::Result<()> {
+    scopeguard::defer! {
+        println!("http has shut down");
+    }
+
+    let service = || RouterService::new(make_router().build()?);
+
+    hyper::Server::from_tcp(http_listener)?
+        .serve(service().map_err(|e| anyhow!(e))?)
```
LizardWizzard, line 26: Is there an issue for proxy graceful shutdown? Lets create and mention one if needed

proxy/src/proxy.rs
```diff
@@ -1,389 +1,322 @@
...
-            }
-        };
+async fn handle_client(
+    config: &ProxyConfig,
+    stream: impl AsyncRead + AsyncWrite + Unpin,
+) -> anyhow::Result<()> {
+    // The `closed` counter will increase when this future is destroyed.
+    NUM_CONNECTIONS_ACCEPTED_COUNTER.inc();
+    scopeguard::defer! {
+        NUM_CONNECTIONS_CLOSED_COUNTER.inc();
```
LizardWizzard, line 71: What are the benefits of having these two counters instead of one gauge with number of active connections?

LizardWizzard (approved):
Nice! Great Work! Having this covered by unittests is a huge plus.

What do you think about hypothetical Proxy struct that can hold reqwest::Client for console interaction and a cancellation keys map? This way we can also avoid global variables. May also simplify some future tests that manipulate this state.

I've left just a few minor nitpicks

proxy/src/proxy.rs
funbringer: Frankly, I don't know. I tried to stay as close to the source as possible. Maybe @bojanserafimov could elaborate.

proxy/src/http.rs
funbringer: I don't think anything needs to be done here at the moment. I can see how shutdown affects pageserver et al, but there's nothing to persist here, especially in the web server.

proxy/src/cplane_api.rs
funbringer: Why? A curious reader could examine the docs for serde.

proxy/src/cplane_api.rs
funbringer: Do we have to this right now?

proxy/src/cplane_api.rs
funbringer: I don't think this change is necessary here. I'll reconsider this if it ever becomes a bottleneck.

proxy/src/auth.rs
```diff
@@ -0,0 +1,127 @@
...
+use zenith_utils::pq_proto::{BeMessage as Be, BeParameterStatusMessage, FeMessage as Fe};
+
+/// Various client credentials which we use for authentication.
+#[derive(Debug, PartialEq, Eq)]
+pub struct ClientCredentials {
+    pub user: String,
+    pub dbname: String,
+}
+
+impl TryFrom<HashMap<String, String>> for ClientCredentials {
```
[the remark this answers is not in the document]
funbringer: > Am I correct that here parameters are consumed?

Yes.

> Do we forward parameters to compute? I didn't find this so I wonder is it supported and do we need that if not

I'm sorry, I don't follow. This piece of code just takes care of extracting everything needed for authentication process, nothing more and nothing less.

funbringer:
@LizardWizzard 

> What do you think about hypothetical Proxy struct that can hold reqwest::Client for console interaction and a cancellation keys map?

Could you elaborate? I don't see why they should live in a single map.

> This way we can also avoid global variables.

How so? You never said anything about this map's lifetime. Besides, I've just re-introduced some private global maps exactly for the sake of simplicity.

> May also simplify some future tests that manipulate this state.

Not necessarily.

proxy/src/cancellation.rs
funbringer: I've reworded the comment.

proxy/src/cplane_api.rs
funbringer: However, I agree that giving the reader a link wouldn't do any harm :) Updated the comment.

proxy/src/auth.rs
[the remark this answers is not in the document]
funbringer: Ah, I see! You were talking about passing other connection params to compute. This is out of scope of this patch, but we definitely should discuss that.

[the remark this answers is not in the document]
LizardWizzard: I mean that we do not forward parameters to compute, e g application name. This is not introduced in this PR so I created separate issue https://github.com/zenithdb/zenith/issues/1287

proxy/src/http.rs
LizardWizzard: But what about graceful connection termination? Can we send a shutdown message or something? Maybe a warning with some friendly message about maintenance/needed reconnection.

funbringer: I still think this is off-topic. Please, stay the course.

proxy/src/proxy.rs
LizardWizzard: Gotcha, gauge needs to be reset to 0 on startup then. But it is ok either way, lets keep it as is

PR #1279, opened 2022-02-15
remove several obsolete management api commands from pageserver's libpq api

LizardWizzard:
these commands are now available via http api

I think gc/checkpoint commands could be also moved to http (and maybe even basebackup). Thoughts? @hlinnaka 

As far as I'm aware console uses http api, so it should be safe to remove these. Am not sure are we testing this in e2e tests so FYI @ololobus


ololobus:
> As far as I'm aware console uses http api, so it should be safe to remove these. Am not sure are we testing this in e2e tests so FYI 

Yeah, the only libpq command we still use in the cloud is basebackup, executed on the compute side. Console uses a generated http client to create tenant and list timelines, so since e2e tests create a new cluster both endpoints should be tested

SomeoneToIgnore (approved):
Good to know that I don't have to support timeline management commands here.

kelvich (approved):

PR #1280, opened 2022-02-15
Create timeline on safekeepers explicitly

arssher:
based on https://github.com/zenithdb/zenith/pull/1228

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -424,6 +424,7 @@ dependencies = [
  "thiserror",
  "toml",
  "url",
+ "walkeeper",
  "workspace_hack",
  "zenith_utils",
 ]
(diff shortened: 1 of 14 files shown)


petuhovskiy (commented):
Looks good, but explicit timeline creation will probably make things more difficult in general.

I'm not sure if http query for timeline creation should return error if timeline is already exists, because usually such operation should be retriable, and retried until timeline is successfully created.

arssher:
I'll merge this, but noticed that `test_restarts_under_load` sometimes fails at the moment, regardless of this PR. Investigating it now, probably related to batch flushing. FYI @petuhovskiy

arssher:
There is also e2e test failure -- test_cluster_db_delete couldn't reach the started cluster
https://github.com/zenithdb/console/runs/5434205638?check_suite_focus=true
but again I don't think it is related and it is hard to investigate without having compute logs:
https://github.com/zenithdb/console/issues/744

petuhovskiy:
> I'll merge this, but noticed that `test_restarts_under_load` sometimes fails at the moment, regardless of this PR. Investigating it now, probably related to batch flushing. FYI @petuhovskiy

Yeah, I also seen this in my local branch and it seems like it isn't due to a bug, just something has started to work much slower, because test passes if you increase the time limits in it.

PR #1281, opened 2022-02-16
Fix safekeeper metric tags

dhammika:

Use separate tags in sk storage file histo (#1234)


arssher:
There are some more in `src/safekeeper.rs`.

arssher (approved):

PR #1285, opened 2022-02-16
Add more typisation into Python tests

SomeoneToIgnore:
* adds a small wrapper around `zenith` CLI to make it a bit easier to change its API later
* uses `uuid.UUID` instead of `str` for tenant ids where possible

### DIFF
--- a/test_runner/fixtures/zenith_fixtures.py
+++ b/test_runner/fixtures/zenith_fixtures.py
@@ -525,11 +525,11 @@ class ZenithEnv:
 
         # generate initial tenant ID here instead of letting 'zenith init' generate it,
         # so that we don't need to dig it out of the config file afterwards.
-        self.initial_tenant = uuid.uuid4().hex
+        self.initial_tenant = uuid.uuid4()
 
         # Create a config file corresponding to the options
         toml = textwrap.dedent(f"""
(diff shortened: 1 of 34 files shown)


test_runner/fixtures/zenith_fixtures.py
```diff
@@ -839,6 +775,186 @@ class S3Storage:
...
+        self.raw_cli(['tenant', 'create', tenant_id.hex])
+        return tenant_id
+
+    def list_tenants(self) -> 'subprocess.CompletedProcess[str]':
+        return self.raw_cli(['tenant', 'list'])
+
+    def create_branch(self,
+                      branch_name: str,
+                      starting_point: str,
+                      tenant_id: Optional[uuid.UUID] = None) -> 'subprocess.CompletedProcess[str]':
```
SomeoneToIgnore, line 799: Part of the reason I'd like to do so is to get a way to parse the CLI output like

https://github.com/zenithdb/zenith/compare/main...kb/remove-branch-files#diff-41430bb43fbde22288ef4d5c9088e2986f94cc43d4752947c260064daed482e9R791-R804

to get an entity that was created during the operation.

LizardWizzard (approved):
Great! Thanks! Was thinking about something like that too

PR #1286, opened 2022-02-16
Remove timeline names from pageserver

SomeoneToIgnore:
Part of https://github.com/zenithdb/console/issues/693 and will require console adjustments.

Removes branch name, branch tag, related files and related code from pageserver's API, adjusting zenith and tests.
"Branch" concept is gone and now pageserver operates "timeline" only, yet the concept of "branching" is not removed itself: timelines still have ancestors and can branch off one from each other at given Lsn or from the end, as before.
Also, during API adjustment, "point in time" concept got removed, split into explicit "Lsn" and "ZTimelineId" parts.

One implicit advantage of the name removal is a more explicit separation between "postgres node name" and "branch name" that was used as the same entity in [our tests](https://github.com/zenithdb/zenith/pull/1286#discussion_r807832063) when branching timelines.

This PR has a few tests failing, but most of them pass already, so it's time to consider, whether the approach is ok and if we want to go with it in general.


zenith/src/main.rs
```diff
@@ -456,24 +479,42 @@ fn handle_tenant(tenant_match: &ArgMatches, env: &local_env::LocalEnv) -> Result
...
-        // No arguments, list branches for tenant
-        let branches = pageserver.branch_list(&tenantid)?;
-        print_branches_tree(branches)?;
+    match timeline_match.subcommand() {
+        Some(("list", list_match)) => {
+            let tenant_id = get_tenant_id(list_match, env)?;
+            let timelines = pageserver.timeline_list(&tenant_id)?;
+            print_timelines_tree(timelines)?;
+        }
+        Some(("create", create_match)) => {
```
SomeoneToIgnore, line 491: Note that now external API users will be able to create ancestor-less timelines (before these changes, only `init` operation could implicitly create such timelines).

I would also imagine, that we can change the API so that there's `list`, `create` and `branch` parts in it, to add more explicitness? If so, should the same separation be done in the http API?

control_plane/src/local_env.rs
```diff
@@ -57,6 +58,11 @@ pub struct LocalEnv {
 
     #[serde(default)]
     pub safekeepers: Vec<SafekeeperConf>,
+
+    /// Every tenant has a first timeline created for it, currently the only one ancestor-less for this tenant.
+    /// It is used as a default timeline for branching, if no ancestor timeline is specified.
```
hlinnaka, line 63: The CLI `zenith branch' command requires you to specify the parent branch. Is there any scenario where the ancestor isn't given explicitly?

test_runner/fixtures/zenith_fixtures.py
```diff
@@ -1166,13 +1185,11 @@ def create(
         if not config_lines:
             config_lines = []
 
-        if branch is None:
-            branch = node_name
```
SomeoneToIgnore, line 1170: This was the most unusual thing with the current approach, somewhat making the node (pg) <-> branch (pageserver) connection implicit.

So when we do 

```python
env.zenith_cli(["branch", "test_branch_segment_boundary", "test_branch_behind@0/3000000"])
pg = env.postgres.create_start("test_branch_segment_boundary")
```

`test_branch_segment_boundary` are different things, that are mapped together due to the assignment.

zenith/src/main.rs
```diff
@@ -67,21 +68,12 @@ struct BranchTreeEl {
...
-        .help("Node name")
-        .required(false);
+    let pg_node_arg = Arg::new("node").index(1).help("Node name").required(true);
+
+    let safekeeper_node_arg = Arg::new("node").index(1).help("Node name").required(false);
 
-    let timeline_arg = Arg::new("timeline")
-        .index(2)
-        .help("Branch name or a point-in time specification")
+    let timeline_id_arg = Arg::new("timeline")
```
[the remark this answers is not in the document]
hlinnaka: Feel free if you think it makes sense.

zenith/src/main.rs
```diff
@@ -488,86 +529,92 @@ fn handle_pg(pg_match: &ArgMatches, env: &local_env::LocalEnv) -> Result<()> {
...
-                    node.timelineid, // FIXME: resolve human-friendly branch name
+                    node.timeline_id,
                     lsn_str,
                     node.status(),
                 );
             }
         }
         "create" => {
             let node_name = sub_args.value_of("node").unwrap_or("main");
-            let timeline_name = sub_args.value_of("timeline").unwrap_or(node_name);
```
SomeoneToIgnore, line 526: This is another place where we're substituting a `timeline` that is already a `branch_name` and a `node_name` for a typical test of ours due to https://github.com/zenithdb/zenith/pull/1286/files#r807832063

control_plane/src/local_env.rs
```diff
@@ -57,6 +58,11 @@ pub struct LocalEnv {
...
+    #[serde(default)]
+    pub initial_timelines: HashMap<ZTenantId, ZTimelineId>,
```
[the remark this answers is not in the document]
SomeoneToIgnore: Yet now we use `initial_timeline` concept, since pg nodes would require something to start on.
Before, `node_name`, `branch_name` was usually matching in our python tests, and we've got the timeline_id  in any of the `zenith pg ..` methods via `parse_point_in_time`  that called pageserver's `branch_get_by_name` method.
(see https://github.com/zenithdb/zenith/pull/1286/files#r807832063 and #TODO kb for more details)

Now we're not guessing branch names anymore, so have to explicitly set the timeline id into pg node.
To mitigate the verboseness in tests, the `initial_timeline` was introduced into Python test.

hlinnaka (commented):
I'm generally in favour of this, +1.

This does make the CLI user experience much worse though. Previously, you would do this:
```
./target/release/zenith init
./target/release/zenith start
./target/release/zenith pg start main
```
That doesn't work anymore, you have to specify a timeline ID to the 'zenith pg start' command:
```
$ ./target/debug/zenith pg start main
command failed: No timeline id, use --timeline to specify one
```
You have to copy-paste the random timeline ID from the output of the 'zenith init' command. And the advise above isn't actually correct: there is no '--timeline' option:
```
$ ./target/debug/zenith pg start main --timeline 2d88ad2ce4591b9fa55223b6d28153b8
error: Found argument '--timeline' which wasn't expected, or isn't valid in this context

	If you tried to supply `--timeline` as a value rather than a flag, use `-- --timeline`

USAGE:
    zenith pg start [OPTIONS] <node> [timeline]

For more information try --help
```

The correct syntax is:
```
./target/debug/zenith pg start main  2d88ad2ce4591b9fa55223b6d28153b8
```

Anyway, I'd like to keep that developer experience nicer. Could we still have branch names, just for the CLI? They don't need to be stored in the page server, just keep them e.g. a json file in the .zenith directory and map from the branch name to the timeline ID in the CLI.

hlinnaka:
Something's wrong with the 'timeline create' or 'list' command:
```
~/git-sandbox/zenith ((c89fd571...))$ ./target/debug/zenith timeline create --ancestor-timeline-id 627fbb511174568713344749430b7a17
Created timeline '78b01dcd8196eef6089212bbc39daf61' at 0/1698CC0 for tenant: 5cbf22b573093a87de3f207014ce750f
~/git-sandbox/zenith ((c89fd571...))$ ./target/debug/zenith timeline list
 627fbb511174568713344749430b7a17
 3ef01db24c8cbc1d58708235c736958d
 2d88ad2ce4591b9fa55223b6d28153b8
```
The newly created timeline isn't visible in the list.

SomeoneToIgnore:
> Anyway, I'd like to keep that developer experience nicer. Could we still have branch names, just for the CLI?

Totally on board with the need to improve the UX, just not sure if branch names at its current form are good to use?
At least that "node_name" == "branch_name" == "timeline_spec" substitution mentioned in the comments above confuses me: if I'm having an alias, I'd like it to be the same alias in all 3 components semantically, not coincidentally.

Given that we already have to manipulate `Lsn`'s in the CLI for cases when we have to start not on the timeline's end, I assumed that having a default timeline to ask for timeline_id input for non-default cases only could be enough for now (given that we implement timeline listing correctly)
More on that: https://github.com/zenithdb/zenith/pull/1286#discussion_r807850788

So, `pg start main` from your message on the new code should start a pg node on the default timeline of a default tenant.

hlinnaka:
> > Anyway, I'd like to keep that developer experience nicer. Could we still have branch names, just for the CLI?
> 
> Totally on board with the need to improve the UX, just not sure if branch names at its current form are good to use? At least that "node_name" == "branch_name" == "timeline_spec" substitution mentioned in the comments above confuses me: if I'm having an alias, I'd like it to be the same alias in all 3 components semantically, not coincidentally.

Fair enough. I agree it's a bit confusing.

> Given that we already have to manipulate `Lsn`'s in the CLI for cases when we have to start not on the timeline's end, I assumed that having a default timeline to ask for timeline_id input for non-default cases only could be enough for now (given that we implement timeline listing correctly) More on that: [#1286 (comment)](https://github.com/zenithdb/zenith/pull/1286#discussion_r807850788)
> 
> So, `pg start main` from your message on the new code should start a pg node on the default timeline of a default tenant.

Ok, I guess that's enough. And we can add more decorations later when we start to do more stuff with branches, if we need to.

SomeoneToIgnore:
I should be fixing all (or close to that) tests now (including the list command), revamped the API, yet only now fully realized, that `zenith` is a CLI tool and does not save its memory between runs 🙂 
So I think I'll fall back to the "names in json config on zenith level" approach and then this PR will be ready for review.

test_runner/fixtures/zenith_fixtures.py
```diff
@@ -572,8 +571,9 @@ def __init__(self, config: ZenithEnvBuilder):
             self.safekeepers.append(safekeeper)
 
         log.info(f"Config: {toml}")
-
-        self.zenith_cli.init(toml)
+        # TODO kb is this a wrong concept? will break for multiple tenant tests
```
LizardWizzard, line 574: If we have a mapping from timeline name to its id then we can use just "main"

test_runner/fixtures/zenith_fixtures.py
```diff
@@ -798,42 +779,95 @@ def __init__(self, env: ZenithEnv) -> None:
...
 
-    def create_branch(self,
-                      branch_name: str,
-                      starting_point: str,
-                      tenant_id: Optional[uuid.UUID] = None) -> 'subprocess.CompletedProcess[str]':
-        args = ['branch']
-        if tenant_id is not None:
-            args.extend(['--tenantid', tenant_id.hex])
-        args.extend([branch_name, starting_point])
+    def branch_timeline(self,
```
LizardWizzard, line 808: As it was discussed, I think this should accept the timeline name

zenith/src/main.rs
LizardWizzard: I think it is pretty natural to have that feature. We can bring this up in a cloud meeting, because it may affect the way how console works now

LizardWizzard (commented):
Great work! I like these changes

As I mentioned in one of the comments we may even want to remove the create tenant flag from init and let the cli make needed api calls. It may be not so convenient for cli because then we have to start pageserver to make these calls as opposed to init which doesnt have to do so

+1 on a name->mapping for timelines, we can embed the table directly in cli's config file

pageserver/src/timelines.rs
```diff
@@ -0,0 +1,406 @@
...
+    // initdb in the background, and it kept running even after the "zenith init" had exited.
+    // In tests, we started the  page server immediately after that, so that initdb was still
+    // running in the background, and we failed to run initdb again in the same directory. This
+    // has been solved for the rapid init+start case now, but the general race condition remains
+    // if you restart the server quickly. The WAL redo manager doesn't use a separate thread
+    // anymore, but I think that could still happen.
+    let dummy_redo_mgr = Arc::new(crate::walredo::DummyRedoManager {});
+
+    if let Some(tenantid) = create_tenant {
+        let tenantid = ZTenantId::from_str(tenantid)?;
```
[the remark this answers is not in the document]
SomeoneToIgnore: I've taken a stab at it and while it removes a bunch of code from pageserver, it brings the chicken and egg problem, as `create_tenant` is an http-based API method, while `init` phase is a CLI one and needs the `default_tenant_id` before writing the config now.

Starting and stopping the pageserver during `init` seems like an overkill IMO, and the CLI command for creating a tenant is kind of present already, so I think for now it can stay as is, while still confusing indeed.

We could add another command as a requirement to run for the user, but that does not seem right; we could implicitly update the `default_tenant` during `pageserver.start()` invocations, but that is quite odd either.
We could remove  the `pageserver init` command entirely and write a config, if it's absent, during `start`, but this is also not that convenient. Or `zenith init` could manage both configs at the same time, with `pageserver` completely lacking `init` again?

I'm fine with working on this in scope of this PR, just not sure which path is good.

test_runner/fixtures/zenith_fixtures.py
SomeoneToIgnore: Initially, I've wanted to get rid of node names entirely (and use branches instead), but after I had to return back to them and this is the most sane approach indeed, implemented the same way.

control_plane/src/compute.rs
```diff
@@ -37,7 +37,7 @@ impl ComputeControlPlane {
     // pgdatadirs
     // |- tenants
     // |  |- <tenant_id>
-    // |  |   |- <branch name>
```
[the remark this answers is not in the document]
SomeoneToIgnore: Yes, it has to be the `node name`, thank you.

pageserver/src/http/routes.rs
```diff
@@ -367,9 +262,7 @@ pub fn make_router(
             "/v1/timeline/:tenant_id/:timeline_id/detach",
             timeline_detach_handler,
         )
-        .get("/v1/branch/:tenant_id", branch_list_handler)
-        .get("/v1/branch/:tenant_id/:branch_name", branch_detail_handler)
-        .post("/v1/branch", branch_create_handler)
+        .post("/v1/timeline", timeline_create_handler)
```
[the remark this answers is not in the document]
SomeoneToIgnore: I've followed the latter pattern, so that all the routes are now unified.

SomeoneToIgnore:
Names had returned to Zenith CLI, all the feedback had been addressed or answered. The tests seem to pass either.

Please have a look and tell me what you think, I'll prepare some digest on the broken APIs to discuss it in more detail.

SomeoneToIgnore:
IT tests and corresponding console fixes: https://github.com/zenithdb/console/pull/782

pageserver/src/timelines.rs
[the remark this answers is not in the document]
SomeoneToIgnore: With init and start unification, there's a slight UX complication: you might want to `init` with `-c arg1=val1 -c..` once and then just run `pageserver start` always generally. But that's all worth a separate PR and discussion indeed.

What I've found useful to remove is the initial timeline creation during tenant creation in pageserver http api: this helps console and generally simpler: now the "initial timeline" concept is left only in cli `init`, actually, making it a bit more special and worth discussing further.

control_plane/src/local_env.rs
```diff
@@ -58,9 +60,16 @@ pub struct LocalEnv {
 
     #[serde(default)]
     pub safekeepers: Vec<SafekeeperConf>,
+
+    /// Keep human-readable aliases in memory (and persist them to config), to hind ZId hex strings from the user.
```
LizardWizzard, line 64: I think this is a typo, hind -> hide

pageserver/src/timelines.rs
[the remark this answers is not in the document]
SomeoneToIgnore: After I've removed timeline creation from the tenant creation, https://github.com/zenithdb/zenith/issues/1294 became pretty simple to resolve by returning 200 on finding a repo in memory already, feel free to discuss that further in the issue.

Otherwise, I consider some improvement is done on that front, now only `init` has a notion of an initial timeline and maybe it's even fine for cli-oriented approach.

SomeoneToIgnore:
CI tests pass now in the corresponding console PR: https://github.com/zenithdb/console/pull/782
I've reworked pageserver's http api further by
* removing timeline creation from the tenant creation: we have a normal way to create an ancestor-less timeline via http now, so the client can decide when to do it)
* returning 200 on tenant recreation: with the timeline logic removed, tenant creation is simple enough to return early on recreate attempts, assuming that the existing repo is fine ([details](https://github.com/zenithdb/zenith/issues/1294))

Console code shrinked too, yet the currently do not have names in the logic at all, compared to our `zenith` cli implementation now.
Otherwise, I do not plan to touch those PRs, besides rebases and feedback addressing.

Currently, I plan to merge this PR due to mostly general feedback before, after its console counterpart will be ready for merging.
Please share your feedback in either of the PRs, if you have any.

pageserver/src/http/routes.rs
```diff
@@ -73,23 +68,45 @@ async fn status_handler(request: Request<Body>) -> Result<Response<Body>, ApiErr
...
+            request_data.ancestor_start_lsn,
         )
     })
     .await
     .map_err(ApiError::from_err)??;
-    Ok(json_response(StatusCode::CREATED, response_data)?)
+
+    Ok(match new_timeline_info {
+        Some(info) => json_response(StatusCode::CREATED, TimelineInfoResponse::from(info))?,
+        None => json_response(StatusCode::NOT_MODIFIED, ())?,
```
LizardWizzard, line 92: I would rather return 400 with timeline already exists message, WDYT?

pageserver/src/http/routes.rs
```diff
@@ -320,13 +222,20 @@ async fn tenant_create_handler(mut request: Request<Body>) -> Result<Response<Bo
...
+            request_data.new_tenant_id.map(ZTenantId::from),
+        )
     })
     .await
     .map_err(ApiError::from_err)??;
-    Ok(json_response(StatusCode::CREATED, ())?)
+
+    Ok(match new_tenant_id {
+        Some(id) => json_response(StatusCode::CREATED, HexZTenantId::from(id))?,
+        None => json_response(StatusCode::NOT_MODIFIED, ())?,
```
LizardWizzard, line 237: The same as with timeline

test_runner/batch_others/test_wal_acceptor.py
```diff
@@ -55,21 +54,26 @@ def test_many_timelines(zenith_env_builder: ZenithEnvBuilder):
 
     n_timelines = 3
 
-    branches = ["test_wal_acceptors_many_timelines_{}".format(tlin) for tlin in range(n_timelines)]
+    branch_names = [
+        "test_wal_acceptors_many_timelines_{}".format(tlin) for tlin in range(n_timelines)
+    ]
+    branch_names_to_timeline_ids = {}
```
LizardWizzard, line 60: nit: maybe worth a comment explaining why we need this mapping here

test_runner/batch_others/test_wal_acceptor.py
```diff
@@ -39,9 +38,9 @@ def test_normal_work(zenith_env_builder: ZenithEnvBuilder):
 
 
 @dataclass
-class BranchMetrics:
-    name: str
-    latest_valid_lsn: int
+class TimelineMetrics:
+    timeline_id: str
+    last_record_lsn: int
```
LizardWizzard, line 43: Is it correct that we use last_record_lsn instead of disk_consistent_lsn here?

test_runner/fixtures/zenith_fixtures.py
```diff
@@ -810,48 +804,120 @@ class ZenithCli:
...
+        res = self.raw_cli(cmd)
+        res.check_returncode()
 
-    def list_branches(self,
-                      tenant_id: Optional[uuid.UUID] = None) -> 'subprocess.CompletedProcess[str]':
-        args = ['branch']
-        if tenant_id is not None:
-            args.extend(['--tenantid', tenant_id.hex])
-        return self.raw_cli(args)
+        create_timeline_id_extractor = re.compile(r"^Created timeline '(?P<timeline_id>[^']+)'",
```
LizardWizzard, line 875: precompiled regex can be a top level const, or we can even introduce some flag to cli for machine readable output. For cases like this it can be `--quiet` flag and cli changed to output only id without other messages when this flag is supplied. Maybe this is redundant complication, but if someone changes wording in the output this will fail

test_runner/fixtures/zenith_fixtures.py
```diff
@@ -810,48 +804,120 @@ class ZenithCli:
...
+        matches = create_timeline_id_extractor.search(res.stdout)
 
-    def init(self, config_toml: str) -> 'subprocess.CompletedProcess[str]':
+        created_timeline_id = None
+        if matches is not None:
+            created_timeline_id = matches.group('timeline_id')
+
+        if created_timeline_id is None:
+            raise Exception('could not find timeline id after `zenith timeline create` invocation')
+        else:
```
LizardWizzard, line 885: nit: else looks redundant 

PS theoretically mypy should complain about that because return type becomes Optional but it seems that mypy was able to prove exhaustiveness of this "match"

LizardWizzard (commented):
Looks nice! 

I've left some minor comments

control_plane/src/storage.rs
```diff
@@ -310,72 +323,112 @@ impl PageServerNode {
...
 
-    pub fn branch_get_by_name(
+    pub fn timeline_create(
         &self,
-        tenantid: &ZTenantId,
-        branch_name: &str,
-    ) -> Result<BranchInfo> {
-        Ok(self
+        tenant_id: ZTenantId,
+        new_timeline_id: Option<ZTimelineId>,
```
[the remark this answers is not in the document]
SomeoneToIgnore: I think, automatic generation is beneficial in cases when the client does not know what that id is or is unable to generate it. 
Also, this is an attempt to make it less heavy to do "create some timeline" call now, when tenant creation does not create create it along.
Not sure how important that is in our context, but console does generate ids and uses them, so that part is fine.

I have no strong preference, so can change it to require the ids if you insist.

pageserver/src/http/models.rs
```diff
@@ -1,20 +1,121 @@
...
 #[derive(Serialize, Deserialize)]
 pub struct TenantCreateRequest {
+    pub new_tenant_id: Option<HexZTenantId>,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct TimelineInfoResponse {
+    pub kind: String,
+    #[serde(with = "hex")]
+    timeline_id: ZTimelineId,
```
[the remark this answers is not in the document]
SomeoneToIgnore: We can use both, I personally find this variant slightly better since no `From` conversion is done in regular code this way.

I guess, it would make more sense to have hex variants everywhere, if we would have a bit more explicit API around "serializeable json objects" in our logic (`models.rs` is a step in that direction, but not yet complete alas).

test_runner/fixtures/zenith_fixtures.py
SomeoneToIgnore: I've went with the global vars for simplicity 🙂

control_plane/src/storage.rs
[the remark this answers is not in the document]
LizardWizzard: I'm not insisting, but it may be a good idea to add note about that to the api doc

pageserver/src/http/models.rs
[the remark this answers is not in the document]
LizardWizzard: I still do not have complete picture in my head regarding this serde related stuff. I think it may not be possible to use hex variants everywhere, if some struct is serialized to both, binary and json we'll have to duplicate the struct with a Hex id variand and use From impl to convert one to another

pageserver/src/http/routes.rs
LizardWizzard: I think it is a client error to create already existing tenant, I'm ok with either 400 or 409. I can imagine using NOT_MODIFIED status when you asked server to update something, but server state already meets the update criteria. Here we're trying to create new resource, rather than modifying already existing one. Usually for 400 response on some e-commerse websites there is not only a message but also a code message. E g `{"code": "invalid_json", "msg": "Invalid json"}`. Where msg can even be translated to multiple languages. Another point is when there will be multiple kinds of errors we probably won't use different 4xx codes for them, or even when there may be a more than one meaning of some particular error code, e g different types of conflicts. So we will have to distinguish between them anyway. 

I think it is simpler to use an umbrella 400 for that, but do not insist and for current simple case any option is fine.

test_runner/batch_others/test_wal_acceptor.py
LizardWizzard: Hm after some archeology I think it is correct, but the naming is old, last_record_lsn is a replacement for latest_valid_lsn, so you can safely rename old occurrences (or leave as is and I'll clean it up)

pageserver/src/http/models.rs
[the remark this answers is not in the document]
SomeoneToIgnore: There are 3 things happening more or less separately:
* `#[derive(Serialize, Deserialize)]` that does conventional (de)serialization
* `#[serde(with = "smth")]` attribute that can be placed on any field (in this context), overriding its (de)serialize entirely, it's not even required for that annotated field to have `(De)Serialize` trait derived
* different newtypes with different `(De)Serialize` impls.

So

```rust
#[derive(Deserialize, Serialize)]
struct Foo {
    one: ZTimeineId,
    #[serde(with = "hex")]
    two: ZTimeineId,
    three: ZHexTimeineId,
}

fn main() {
    dbg!(serde_json::to_string(&Foo {...}));
}
```

will print `one` as an array of bytes dues its serialization rules (`derive` on a newtype over byte array), will take an override for `two` and use `hex` module to serialize it as a hex string, and `three` as a hex string due its custom `(De)Serialize` impls (as a hex string, with visitor)

pageserver/src/http/routes.rs
SomeoneToIgnore: > I think it is a client error to create already existing tenant

I have one general argument against it, motivating me to return non-error status: if it's an error, the cluster has to defend itself against it by either listing clusters/tenants before, to ensure that we're not recreating anything or by calling verification methods after errors.
At its current model, where any failure would be retried, the cluster has to protect itself from these errors by having an extra call.
Now, when we've split tenant and timeline creation, this means 4 requests in this model, vs 2 requests in a "non-error" one, apparently that [might be important](https://github.com/zenithdb/console/pull/782#discussion_r822860234).

----

So, the "error" approach generally requires more defensive request to be made since is not treated as an error really, still hardly see the benefits myself.

But I actually like the idea of `200 OK {"code": "invalid_json", "msg": "Invalid json"}` approach as more flexible indeed, so maybe it's not worth touching the current odd status now: two repos have to be changed for almost nothing.
But sure feel free to adjust the code yourself later 🙂

test_runner/batch_others/test_wal_acceptor.py
SomeoneToIgnore: There was only one `latest_valid_lsn` in the comment, I've replaced it.

pageserver/src/http/routes.rs
LizardWizzard: > Now, when we've split tenant and timeline creation, this means 4 requests in this model, vs 2 requests in a "non-error" one, apparently that https://github.com/zenithdb/console/pull/782#discussion_r822860234.

I do not get it why two requests are needed, client just retries until either 201 or 400 with already exists error code. There is a saying in python community: “it’s easier to ask for forgiveness than permission”, I think it might be applicable here

I've found a [thread](https://stackoverflow.com/q/3825990/6378300) on SO with many possible options and 409 being the most upvoted one.

Note that safekeepers recently gained the ability to create timelines on them via http api, so I think we should come to a solution which then will be used everywhere. 

Another example with an error is our console which uses 400 during sign up if email is already taken

pageserver/src/http/models.rs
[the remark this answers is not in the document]
LizardWizzard: Hm, having both Hex id newtypes and serde(with="hex") attribute looks a bit complex and confusing but I dont know what we can do about that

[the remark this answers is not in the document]
SomeoneToIgnore: Yes, neither solution is perfect alas, I've tried to find a way to make (de)serialization generic but that did not work, converting everything into ZHex* variant forces you to write extra bloat code, mixed annotations and Hex newtypes are confusing but seems like a lesser evil given that person reads the ZHex* doc.

PR #1288, opened 2022-02-16
Remove redundant pytest plugin declarations

bojanserafimov:
Declaring plugins outside the root conftest.py is redundant and [deprecated](https://docs.pytest.org/en/6.2.x/writing_plugins.html#requiring-plugins-in-non-root-conftests)

### DIFF
--- a/test_runner/performance/test_bulk_insert.py
+++ b/test_runner/performance/test_bulk_insert.py
@@ -4,12 +4,6 @@ from fixtures.log_helper import log
 from fixtures.benchmark_fixture import MetricReport, ZenithBenchmarker
 from fixtures.compare_fixtures import PgCompare, VanillaCompare, ZenithCompare
 
-pytest_plugins = (
-    "fixtures.zenith_fixtures",
-    "fixtures.benchmark_fixture",
-    "fixtures.compare_fixtures",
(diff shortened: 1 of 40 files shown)


test_runner/performance/test_bulk_insert.py
```diff
@@ -4,12 +4,6 @@
 from fixtures.benchmark_fixture import MetricReport, ZenithBenchmarker
 from fixtures.compare_fixtures import PgCompare, VanillaCompare, ZenithCompare
 
-pytest_plugins = (
-    "fixtures.zenith_fixtures",
-    "fixtures.benchmark_fixture",
-    "fixtures.compare_fixtures",
-)
```
LizardWizzard, line 11: Conftest mentions only zenith_fixtures, and benchmark_fixture, probably compare_fixtures should be added too, but everything works even without it (as I can see all the checks are green)

LizardWizzard (approved):
Nice! Thanks!

test_runner/performance/test_bulk_insert.py
bojanserafimov: It's there. Maybe you're looking at a stale main branch :)

Issue #1144, opened 2022-01-18
Optimize flushes and AppendResponses in safekeeper


petuhovskiy:
#1141 showed that single safekeeper can have lower performance compared to single walreceiver. We can try to implement optimization from walreceiver and see if it improves performance.

What we're doing now: reading single AppendRequest from socket, writing WAL to disk, flushing to disk, sending AppendResponse.

What needs to be done: process a batch of AppendRequest that we can read from socket without blocking and then make single disk flush, send single AppendResponse.

walreceiver.c reference: https://github.com/zenithdb/postgres/blob/main/src/backend/replication/walreceiver.c#L447-L491

I'm not sure if we have support for non-blocking reads in PgBackend, so this issue is probably blocked until then.

PR #1290, opened 2022-02-17
Refactor WAL utils in safekeeper

petuhovskiy:
`wal_storage.rs` was split up from `timeline.rs`, `safekeeper.rs` and `send_wal.rs`, and now it contains all WAL related code from the safekeeper. Now there are `PhysicalStorage` for persisting WAL to disk and `WalReader` for reading it. This allows optimizing `PhysicalStorage` without affecting too much of other code. The next step after merging this PR will be to optimize `PhysicalStorage` by splitting write and flush (#1144), keeping File in memory to optimize `open()` syscalls and maybe adding more prometheus metrics. I tried not to change original write/read/truncate functions too much, so they mostly stayed as they were written before.

Also there is a separate structure for persisting control file now, located in `control_file.rs`.


SomeoneToIgnore (commented):
I'm bad with this part of the codebase, noticed nothing suspicious, so LGTM.

lubennikovaav (approved):
Looks good to me

PR #1292, opened 2022-02-17
Add perf test: test_random_writes

bojanserafimov:
This test exploits high write amplification to show a 30x performance difference between zenith and vanilla postgres (on my laptop, release build). The difference might expand even more when we include safekeepers in the test fixture, since vanilla is doing more fsyncs than zenith in this case (which should be done as part of https://github.com/zenithdb/zenith/issues/553 ) Testing with fsync disabled on vanilla expands the gap to 2500x (which makes sense, since segment_size / page_size is in that order of magnitude).

This test is similar to the existing test_write_amplification test and this one https://github.com/zenithdb/zenith/pull/1102 , but it shows a bigger performance gap because it only writes random writes, instead of a mixed hot + cold workload. This also makes it a more realistic scenario to practically happen. If we can safely extrapolate from this test, I'd expect to fully see this problem with random writes on a 300 TB table, and partially on smaller multi-gigabyte tables.

I expect performance to improve a lot after https://github.com/zenithdb/zenith/pull/1133 and I expect the problem to be completely addressed with the LSM tree approach, or any variation of it (as long as checkpointing consists of writing one big file of wal records).

### DIFF
--- a/test_runner/performance/test_random_writes.py
+++ b/test_runner/performance/test_random_writes.py
@@ -11,19 +11,32 @@ import time
 from fixtures.utils import print_gc_result
 
 
+# This is a clear-box test that demonstrates the worst case scenario for the
+# "1 segment per layer" implementation of the pageserver. It writes to random
+# rows, while almost never writing to the same segment twice before flushing.
+# A naive pageserver implementation would create a full image layer for each
+# dirty segment, leading to write_amplification = segment_size / page_size,
+# when compared to vanilla postgres. With segment_size = 10MB, that's 1250.
 def test_random_writes(zenith_with_baseline: PgCompare):
     env = zenith_with_baseline
 
-    # Directly impacts the effective checkpoint distance.
-    # Smaller value makes the test run faster. Larger value is more realistic.
-    # Doesn't affect the zenith / vanilla performance ratio.
+    # Number of rows in the test database. 1M rows runs quickly, but implies
+    # a small effective_checkpoint_distance, which makes the test less realistic.
+    # Using a 300 TB database would imply a 250 MB effective_checkpoint_distance,
+    # but it will take a very long time to run. From what I've seen so far,
+    # increasing n_rows doesn't have impact on the (zenith_runtime / vanilla_runtime)
+    # performance ratio.
     n_rows = 1 * 1000 * 1000  # around 36 MB table
 
-    # Relative measure of write density. Doubling the load factor doubles
-    # average writes per page.
+    # Number of writes per 3 segments. A value of 1 should produce a random
+    # workload where we almost never write to the same segment twice. Larger
+    # values of load_factor produce a larger effective_checkpoint_distance,
+    # making the test more realistic, but less effective. If you want a realistic
+    # worst case scenario and you have time to wait you should increase n_rows instead.
     load_factor = 1
 
-    # Not sure why but this matters in a weird way. Looking into it
+    # Not sure why but this matters in a weird way (up to 2x difference in perf).
+    # TODO look into it
     n_iterations = 1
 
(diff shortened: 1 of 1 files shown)


bojanserafimov:
> First of all I wonder how to use this script with vanilla (in order to compare write amplification with vanilla)? Looks like our pythin scripts may only work with Zenith, this is why I prefer to create shell scripts. But may be I am missing something.
> 
> Concerning write amplification coefficient: it seems to be important to notice that it depends on Big table size. Consider default parameters used in this test: 1M rows, 36Mb table. In this case we have just 4 10Mb segments. Then we start random updating if this segments. Default checkpoint distance is 256Mb. It means that we will create four 64Mb delta layers before checkpointer creates 4 10Mb image layers. So on 256Mb of WAL we wrote in storage 256+40 Mb. Write amplification is 15%. Not so impressive, isn't it?
> 
> If we create much bigger table, than certainly write amplification becomes larger. But to make one updates cause creation of image layer we need 256M segments = 2Pb (2560Tb). This is why I prefer to have two tables:big and small (265Mb). At each iteration we insert 256 Mb in small table which triggers checkpoint and do random updates in big table.In this case actually each update of one record in big relation cause writing complete 10Mb segment.

To compare against vanilla, run pytest with flag `-k test_random_writes`. To only run zenith, run `-k test_random_writes[zenith]`. To only run vanulla `-k test_random_writes[vanilla]`.

Yes, we need a pretty big table to see 30x difference. A 213 TB table would be more than enough. In your calculation you assume that rows are 1 byte, but in this case they're about 36 bytes I think. Since 2560TB / 36 = 71 TB, we can see the effects at 71 TB, as long as we write to each segment exactly once. Here I'm doing random writes, so i multiply the size by 3 to avoid too many collisions.

The small test does something different than you describe. Try running it, and you'll see 30x difference, instead of 15%. It tries to approximate what would happen with a big table. In this test the `load_factor` remains fixed, and the `effective_checkpoint_distance` varies. In the case with a 36 MB database, the effective_checkpoint_distance is 36 bytes. We basically just write one row and flush.

It sounds like a useless test (flushing every 36 bytes is ridiculous), but I've seen that the small test with load factors [1, 10, 100, 1000, etc] approximates the same zenith_runtime / vanilla_runtime that we see on larger tests with those load factors. So it's not ideal, but it's fast and useful for now. For example, I expect that your image layer heuristic will have some visible effect on the small test. (also we might want to increase the number of iterations after that PR merges, to trigger some image layer writes)

On the big:small table test: Say we have a small 256MB table and a big 10 GB table. Instead of writing to all segments in the 10 GB table and then writing 256 MB to the small table, consider just writing 256 random bytes to both of them. The write amplification is the same, exactly ((256 MB + 10 GB) / 256 MB). I don't see a benefit from splitting it into big and small.

knizhnik:
By updating 256 random keys in big table, you modify 256 pages (4Mb) while generate only ~15Kb of WAL.
By writing 256  records in small table, size of modified pages is about the same as size of generated WAL.
That is the difference.

knizhnik:
> To compare against vanilla, ...
O, thank you! I really do no notice compare_fixtures.py

bojanserafimov:
> By updating 256 random keys in big table, you modify 256 pages (4Mb) while generate only ~15Kb of WAL. By writing 256 records in small table, size of modified pages is about the same as size of generated WAL. That is the difference.

Sorry, typo, I meant "write 256 MB of random rows to both of them", not "256 rows".

knizhnik:
But the same story: 256M random rows affect 256M random pages = 4Tb.
Writing 256Mb in small table produce 256Mb of WAL and 256Mb of storage.

PR #1298, opened 2022-02-18
[proxy] Don't print passwords

bojanserafimov:
https://github.com/zenithdb/console/issues/608

Maybe it's ok to delete those lines altogether, but we're deleting the entire mgmt endpoint soon so it doesn't matter.


SomeoneToIgnore (approved):

proxy/src/cplane_api.rs
```diff
@@ -59,7 +59,7 @@ impl CPlaneApi<'_> {
             }
 
             let auth_info: ProxyAuthResponse = serde_json::from_str(resp.text().await?.as_str())?;
-            println!("got auth info: #{:?}", auth_info);
+            println!("got auth info [redacted]"); // Content contains password, don't print it
```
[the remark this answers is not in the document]
funbringer: +1, I think custom Debug instance is the way to go. Other than that, the change is obviously good.

proxy/src/mgmt.rs
```diff
@@ -95,7 +95,7 @@ impl postgres_backend::Handler for MgmtHandler {
 }
 
 fn try_process_query(pgb: &mut PostgresBackend, query_string: &str) -> anyhow::Result<()> {
-    println!("Got mgmt query: '{}'", query_string);
+    println!("Got mgmt query [redacted]"); // Content contains password, don't print it
```
funbringer, line 98: Maybe we could use `cfg(debug_assertions)` to print sensitive info only in debug mode. It's not like we'd use debug builds in prod environment, so this seems to be an appropriate measure.

bojanserafimov: Good idea. I don't see the need for this extra complexity today, but if the need arises it's a valid option. Staying on the safe side for now.

PR #1305, opened 2022-02-21
[proxy] Add daemonize option

bojanserafimov:
This should make testing easier, and potentially help with `zenith proxy start` command, if necessary.

### DIFF
--- a/proxy/src/main.rs
+++ b/proxy/src/main.rs
@@ -9,11 +9,11 @@ use anyhow::{bail, Context};
 use clap::{App, Arg};
 use config::ProxyConfig;
 use futures::FutureExt;
-use std::path::{Path, PathBuf};
 use std::future::Future;
+use std::path::{Path, PathBuf};
 use tokio::{net::TcpListener, task::JoinError};
-use zenith_utils::GIT_VERSION;
 use zenith_utils::shutdown::exit_now;
+use zenith_utils::GIT_VERSION;
 
 use crate::config::{ClientAuthMethod, RouterConfig};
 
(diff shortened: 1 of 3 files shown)


proxy/src/main.rs
```diff
@@ -108,6 +123,37 @@ async fn main() -> anyhow::Result<()> {
...
+            .stderr(stderr);
+
+        // XXX: The parent process should exit abruptly right after
+        // it has spawned a child to prevent coverage machinery from
+        // dumping stats into a `profraw` file now owned by the child.
+        // Otherwise, the coverage data will be damaged.
+        match daemonize.exit_action(|| exit_now(0)).start() {
+            Ok(_) => println!("Success, daemonized"),
+            Err(e) => anyhow::bail!("Error, {}", e),
+        }
```
bojanserafimov, line 154: copied from safekeeper. If this PR merges I'll consider making this a zenith util.

funbringer:
I wonder, why do we keep adding this shim to our code if daemonize/systemd/etc could do this for us? Especially considering stuff like [this](https://github.com/mitsuhiko/systemfd).

proxy/src/main.rs
```diff
@@ -108,6 +123,37 @@ async fn main() -> anyhow::Result<()> {
...
+    }
+
+    if arg_matches.is_present("daemonize") {
+        println!("daemonizing...");
+
+        // There should'n be any logging to stdin/stdout. Redirect it to the main log so
+        // that we will see any accidental manual fprintf's or backtraces.
+        let log_file = zenith_utils::logging::init("proxy.log", true)?;
+        let stdout = log_file.try_clone().unwrap();
+        let stderr = log_file;
```
funbringer, line 139: I firmly believe that being able to redirect logs to a file is orthogonal to daemonization. Sure, daemons can't write logs to stdout, but it doesn't necessarily mean that it shouldn't be possible to choose this option whenever we'd like to (this rant mostly refers to pageserver/safekeeper/etc, but this where you took this from).

bojanserafimov:
> I wonder, why do we keep adding this shim to our code if daemonize/systemd/etc could do this for us? Especially considering stuff like [this](https://github.com/mitsuhiko/systemfd). Not to mention that the dep we're using for daemonization is unsound in the presence of threads and whatnots (due to the quirks of fork).

Not sure. Maybe OSX compatibility? I also think the decision to provide a way to daemonize the proxy is orthogonal to the decision of how we daemonize things in zenith. I'm 3 layers of yak shaving deep, and I don't want to enter the 4th layer, so I copied our existing solution (I don't like it either).

bojanserafimov:
converting to draft to explore ways to write proxy python tests without daemonization

PR #1307, opened 2022-02-21
Fix deadcode warnings

dhammika:

### DIFF
--- a/walkeeper/src/control_file.rs
+++ b/walkeeper/src/control_file.rs
@@ -194,31 +194,35 @@ impl Storage for FileStorage {
         })?;
 
         // fsync the file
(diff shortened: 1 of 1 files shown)


petuhovskiy:
It seems to be introduced by #1290. I think we can fix this warning by actually using conf. In persist function, we should skip calls to fsync() when `conf.no_sync` is set.

dhammika:
Thanks, I'll check it. 
Not sure why this would break e2e tests though. I didn't see that crate used elsewhere.

petuhovskiy:
> Not sure why this would break e2e tests though.

It shouldn't, it looks more like flaky tests or some other issue.

petuhovskiy (approved):
LGTM

PR #1310, opened 2022-02-22
add node id to pageserver

LizardWizzard:
This adds node id parameter to pageserver configuration. Also I use a
simple builder to construct pageserver config struct to avoid setting
node id to some temporary invalid value. Some of the changes in test
fixtures are needed to split init and start operations for envrionment.

### DIFF
--- a/pageserver/src/bin/pageserver.rs
+++ b/pageserver/src/bin/pageserver.rs
@@ -61,7 +61,7 @@ fn main() -> Result<()> {
                 .number_of_values(1)
                 .multiple_occurrences(true)
                 .help("Additional configuration overrides of the ones from the toml config file (or new ones to add there).
-                Any option has to be a valid toml document, example: `-c \"foo='hey'\"` `-c \"foo={value=1}\"`"),
+                Any option has to be a valid toml document, example: `-c=\"foo='hey'\"` `-c=\"foo={value=1}\"`"),
         )
         .get_matches();
 
@@ -115,7 +115,14 @@ fn main() -> Result<()> {
                     option_line
                 )
             })?;
+
             for (key, item) in doc.iter() {
+                if key == "id" {
+                    anyhow::ensure!(
+                        init,
+                        "node id can only be set during pageserver init and cannot be overridden"
+                    );
+                }
                 toml.insert(key, item.clone());
             }
         }
--- a/pageserver/src/config.rs
+++ b/pageserver/src/config.rs
@@ -8,7 +8,7 @@ use anyhow::{bail, ensure, Context, Result};
 use toml_edit;
 use toml_edit::{Document, Item};
 use zenith_utils::postgres_backend::AuthType;
-use zenith_utils::zid::{ZTenantId, ZTimelineId};
+use zenith_utils::zid::{ZNodeId, ZTenantId, ZTimelineId};
 
 use std::convert::TryInto;
 use std::env;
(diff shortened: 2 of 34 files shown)


arssher:
Let's also add ` -c "id='10'"` to `docker-entrypoint.sh` for console development/testing.

pageserver/src/bin/pageserver.rs
```diff
@@ -115,7 +115,14 @@ fn main() -> Result<()> {
                     option_line
                 )
             })?;
+
             for (key, item) in doc.iter() {
+                if key == "id" {
+                    anyhow::ensure!(
```
SomeoneToIgnore, line 121: NIT: Should we rather WARN log it and skip it instead of failing?

pageserver/src/config.rs
```diff
@@ -106,6 +110,168 @@ pub struct PageServerConf {
...
+    max_file_descriptors: Option<usize>,
+
+    workdir: Option<PathBuf>,
+
+    pg_distrib_dir: Option<PathBuf>,
+
+    auth_type: Option<AuthType>,
+
+    auth_validation_public_key_path: Option<Option<PathBuf>>,
+    remote_storage_config: Option<Option<RemoteStorageConfig>>,
```
SomeoneToIgnore, line 136: > `Option<Option<`

Out of curiosity, what does this semantic mean? 
Would we be able to achieve it with a single `Option<>` wrapper?

pageserver/src/config.rs
```diff
@@ -106,6 +110,168 @@ pub struct PageServerConf {
...
+            gc_horizon: Some(DEFAULT_GC_HORIZON),
+            gc_period: Some(
+                humantime::parse_duration(DEFAULT_GC_PERIOD)
+                    .expect("cannot parse default gc period"),
+            ),
+            superuser: Some(DEFAULT_SUPERUSER.to_string()),
+            page_cache_size: Some(DEFAULT_PAGE_CACHE_SIZE),
+            max_file_descriptors: Some(DEFAULT_MAX_FILE_DESCRIPTORS),
+            workdir: Default::default(),
+            pg_distrib_dir: Some(PathBuf::new()),
```
SomeoneToIgnore, line 161: > `Some(PathBuf::new())`

That is a bit odd to me, is `PathBuf::new()` ever a valid path for the `pg_distrib_dir`?

pageserver/src/config.rs
```diff
@@ -106,6 +110,168 @@ pub struct PageServerConf {
...
+            auth_type: Some(AuthType::Trust),
+            auth_validation_public_key_path: Some(None),
+            remote_storage_config: Some(None),
+            id: Default::default(),
```
SomeoneToIgnore, line 165: > `Default::default()`

NIT: that is not that readable + adds more work for the compiler to infer this, `ZNodeId::default()` is way more appealing to me personally.

SomeoneToIgnore (approved):

pageserver/src/config.rs
LizardWizzard: Outer option is needed for builder to distinguish between set/not set, the inner one is the desired value in target structure we are building

pageserver/src/bin/pageserver.rs
LizardWizzard: In my opinion node id sort of related to correctness so I would rather fail loudly. It is an error to have one id in config and a pageserver actually running with different id. If you need to forcibly change the value you can edit the config manually.

pageserver/src/config.rs
LizardWizzard: Hmm, I've blindly copied defaults, in the config checking code I see this: https://github.com/zenithdb/zenith/blob/74a0942a773359754e704a662e7d109cd07489a9/pageserver/src/config.rs#L289
so I think I can directly use it as a default without intermediate value

arssher (approved):

PR #1317, opened 2022-02-23
Rebuild postgres when zenith root Makefile changes

antons-antons:
Zenith root makefile is used to control Postgres build.
Include makefile into a the CI build cache key for Postgres, this will ensure that modifications to Makefile (unfortunately even unrelated) will trigger Postgres rebuild


funbringer (approved):

hlinnaka (approved):
LGTM, but the CI thinks otherwise?

PR #1266, opened 2022-02-11
Combine fsync for incoming AppendRequest's in safekeeper

petuhovskiy:
Implement optimization from https://github.com/zenithdb/zenith/issues/1144.

This seems to improve local results from #1141 by 2 times, but still 1.5-2 times lower than vanilla replication:

| Test info  | pgbench -i -s 200 (time) | pgbench -N -T 60 | wal_msg size | pg_wal size |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| NO PATCH safekeeper + wp (fsync=on, no checkpoints)  | 99 s | 1142 TPS | 411 b | 2608-2592 MB
| PATCHED safekeeper + wp (fsync=on, no checkpoints) | 85 s | 2662 TPS | 662 b | 2656-2592 MB
| vanilla repl (fsync=on, no checkpoints) | 141 s | 4183 TPS | 689 b | 2704-2608 MB

I'll post results for the #799 EC2 test later.

Issue #799, opened 2021-10-25
Basic safekeeper performance numbers


arssher:
Let's have a look at safekeeper performance. Like

- Create 3 VMs, deploy and run 3 safekeepers on them.
- Create VM with postgres configured to run with safekeepers but without pageserver (`zenith.page_server_connstring` not set)
- run pgbench -i -s 400  , (creates db about 5GB size), see how it goes -- time to load.
- Run pgbench for a couple of hours. With e.g. -c 32. Log tps somewhere (-P 1 shows progress), maybe create graph with it. Check out CPU, memory usage.

If everything goes ok, it would be good to look at metrics exposed to prometheus.

Compare the same with Postgres synchronous replicas (synchronous_standby_names=any 2(s1, s2, s3)) instead of safekeepers.

petuhovskiy:
Ran pgbench for one hour on AWS instances with this configuration: 3x safekeeper on i3.large, 1x compute on i3.large.

<details><summary>postgresql.conf</summary>

```
max_wal_senders=10
wal_log_hints=on
max_replication_slots=10
hot_standby=on
shared_buffers=8192MB
fsync=off
max_connections=100
wal_sender_timeout=0
wal_level=replica
listen_addresses='127.0.0.1'
port=5432
wal_keep_size=10TB
shared_preload_libraries=zenith
zenith.page_server_connstring=''
zenith.zenith_tenant=c9bc8b6ea42cba805c62ee42c1637761
zenith.zenith_timeline=d03d475996d127df7839a796c3762df0
synchronous_standby_names = 'walproposer'
wal_acceptors = '172.32.29.146:6500,172.32.24.189:6500,172.32.27.7:6500'
```

</details>

Results:
```
pgbench -i -s 400
...
done in 230.75 s (drop tables 0.00 s, create tables 0.01 s, client-side generate 80.00 s, vacuum 100.50 s, primary keys 50.23 s).

pgbench -c 32 -P 1 -T 3600
...
pgbench (PostgreSQL) 14.0
transaction type: <builtin: TPC-B (sort of)>
scaling factor: 400
query mode: simple
number of clients: 32
number of threads: 1
duration: 3600 s
number of transactions actually processed: 1266470
latency average = 90.886 ms
latency stddev = 56.137 ms
initial connection time = 75.879 ms
tps = 351.663208 (without initial connection time)
```

[...]

PR #1318, opened 2022-02-23
Optimize WAL storage in safekeeper

### LINKED ISSUE #1144: Optimize flushes and AppendResponses in safekeeper

petuhovskiy:
When several AppendRequest's can be read from socket without blocking,
they are processed together and fsync() to segment file is only called
once. Segment file is no longer opened for every write request, now
last opened file is cached inside PhysicalStorage. New metric for WAL
flushes was added to the storage, FLUSH_WAL_SECONDS. More errors were
added to storage for non-sequential WAL writes, now write_lsn can be
moved only with calls to truncate_lsn(new_lsn).

New messages have been added to ProposerAcceptorMessage enum. They
can't be deserialized directly and now are used only for optimizing
flushes. Existing protocol wasn't changed and flush will be called for
every AppendRequest, as it was before.

This PR replaces #1266, as a cleaner version of the same optimization. Closes #1144.
I'll post test results here when they're ready.

### DIFF
--- a/walkeeper/src/receive_wal.rs
+++ b/walkeeper/src/receive_wal.rs
@@ -2,15 +2,21 @@
 //! Gets messages from the network, passes them down to consensus module and
 //! sends replies back.
 
-use anyhow::{bail, Context, Result};
-use bytes::Bytes;
+use anyhow::{anyhow, bail, Result};
+
 use bytes::BytesMut;
 use tokio::sync::mpsc::UnboundedSender;
 use tracing::*;
+use zenith_utils::sock_split::ReadStream;
 
 use crate::timeline::Timeline;
+
 use std::net::SocketAddr;
+use std::sync::mpsc::channel;
+use std::sync::mpsc::Receiver;
+
 use std::sync::Arc;
+use std::thread;
 
 use crate::safekeeper::AcceptorProposerMessage;
 use crate::safekeeper::ProposerAcceptorMessage;
@@ -46,21 +52,6 @@ impl<'pg> ReceiveWalConn<'pg> {
         }
     }
 
-    // Read and extract the bytes of a `CopyData` message from the postgres instance
-    fn read_msg_bytes(&mut self) -> Result<Bytes> {
-        match self.pg_backend.read_message()? {
-            Some(FeMessage::CopyData(bytes)) => Ok(bytes),
-            Some(msg) => bail!("expected `CopyData` message, found {:?}", msg),
-            None => bail!("connection closed unexpectedly"),
-        }
-    }
-
-    // Read and parse message sent from the postgres instance
-    fn read_msg(&mut self) -> Result<ProposerAcceptorMessage> {
-        let data = self.read_msg_bytes()?;
-        ProposerAcceptorMessage::parse(data)
-    }
-
     // Send message to the postgres
     fn write_msg(&mut self, msg: &AcceptorProposerMessage) -> Result<()> {
         let mut buf = BytesMut::with_capacity(128);
@@ -77,18 +68,22 @@ impl<'pg> ReceiveWalConn<'pg> {
         self.pg_backend
             .write_message(&BeMessage::CopyBothResponse)?;
 
+        let r = self
+            .pg_backend
+            .take_stream_in()
+            .ok_or_else(|| anyhow!("failed to take read stream from pgbackend"))?;
+        let mut poll_reader = ProposerPollStream::new(r)?;
+
         // Receive information about server
-        let mut msg = self
-            .read_msg()
-            .context("failed to receive proposer greeting")?;
-        match msg {
+        let next_msg = poll_reader.recv_msg()?;
+        match next_msg {
             ProposerAcceptorMessage::Greeting(ref greeting) => {
                 info!(
                     "start handshake with wal proposer {} sysid {} timeline {}",
                     self.peer_addr, greeting.system_id, greeting.tli,
                 );
             }
-            _ => bail!("unexpected message {:?} instead of greeting", msg),
+            _ => bail!("unexpected message {:?} instead of greeting", next_msg),
         }
 
         // Register the connection and defer unregister.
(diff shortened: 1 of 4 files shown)


walkeeper/src/receive_wal.rs
```diff
@@ -100,16 +95,81 @@ impl<'pg> ReceiveWalConn<'pg> {
...
+                    msg_tx.send(msg)?;
+                }
+                // msg_rx will be dropped here, this will also close msg_tx
+            })?;
+
+        Ok(Self { msg_rx })
+    }
+
+    fn recv_msg(&mut self) -> Result<ProposerAcceptorMessage> {
+        self.msg_rx.recv().context("failed to receive message")
```
petuhovskiy, line 161: I wanted to replace error here with original read error from thread and tried to use `self.read_thread.join().err().unwrap()` for that, but it didn't work. Does anyone know how to make it work?

walkeeper/src/wal_storage.rs
```diff
@@ -339,69 +374,69 @@ impl Storage for PhysicalStorage {
...
             .wal_seg_size
             .ok_or_else(|| anyhow!("wal_seg_size is not initialized"))?;
 
-        // TODO: cross check divergence point
-
-        // nothing to truncate
-        if self.write_lsn == Lsn(0) {
-            return Ok(());
-        }
+        // TODO: cross check divergence point?
```
petuhovskiy, line 408: @arssher do you remember what this comment was about? It looks like some check should be added here, but I don't remember what exactly should be checked.

arssher: The idea was to repeat divergence point search like in SendProposerElected to ensure we truncate at right position. The check should be done before calling `truncate_wal`.

petuhovskiy:
Here are results of the test from #1144, to compare results with #1266.
Code to test the results of this PR: https://github.com/zenithdb/zenith/tree/test-perf-pr-1318
Same test backported to main: https://github.com/zenithdb/zenith/tree/test-perf-backport-1318

These results are for local run on my machine, which has relatively slow fsync (2364 ops/sec, 423 usecs/op) and not very powerful 8 core CPU. I've added "per fsync" metrics, which are useful for disk usage comparison.

 | Test info  | pgbench -i -s 200 (time) | pgbench -N -T 60 | init wal_bytes, per fsync | bench wal_bytes, per fsync | bench txes, per fsync |
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
| NO PATCH safekeeper + wp (fsync=on, no checkpoints)  | 108 s | 1177 TPS | 85kb | 436 b | 1 tx / fsync |
| PATCHED safekeeper + wp (fsync=on, no checkpoints)  | 87 s | 2574 TPS | 417kb | 6314 b | 15.4 tx / fsync |
| vanilla repl (fsync=on, no checkpoints) | 146 s | 4215 TPS | 256kb | 5552 b | 14 tx / fsync |

These results look the same as in #1266.

walkeeper/src/wal_storage.rs
```diff
@@ -135,128 +159,136 @@ impl PhysicalStorage {
...
-    fn wal_file_paths(&self, segno: XLogSegNo) -> Result<(PathBuf, PathBuf)> {
-        let wal_seg_size = self
-            .wal_seg_size
-            .ok_or_else(|| anyhow!("wal_seg_size is not initialized"))?;
+    /// Call fsync if config requires so.
+    fn flush_file(&self, file: &mut File) -> Result<()> {
+        if !self.conf.no_sync {
+            self.metrics
+                .flush_wal_seconds
+                .observe_closure_duration(|| file.sync_all())?;
```
petuhovskiy, line 177: sync_data can be used here

petuhovskiy:
In a more usual 1wp+3sk EC2 test (#799) results are also similar to previous PR #1266, safekeepers are faster than synchronous replication:

| Code branch | Test info  | pgbench -i -s 400 (time) | pgbench TPS | init wal_bytes, per fsync | bench wal_bytes, per fsync | bench txes, per fsync | link to gist |
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
| main | vanilla 1primary(fsync=off)+3replica(fsync=on) | 130 s | 14237 TPS | 800 kb | 1948 b | 6.7 tx / fsync | https://gist.github.com/petuhovskiy/35134419ccb0f287046a6b1c9a6150b8 |
| this PR | 1wp(fsync=off)+3sk(fsync=on) | 92 s | 20400 TPS | 209kb | 1309 b | 4.5 tx / fsync | https://gist.github.com/petuhovskiy/c16668bdfea05acf39a8154b81ac756b |

Interesting that fsync calls count is very different between postgres synchronous replicas when pgbench -i is done, that is visible in gist report.

walkeeper/src/receive_wal.rs
arssher: I think because you need to go one level deeper: `join()` returns `Error` only if thread panicked. In our case it would be typically `Ok<Result<()>>` with inner result containing actual error.

walkeeper/src/receive_wal.rs
```diff
@@ -100,16 +95,80 @@ impl<'pg> ReceiveWalConn<'pg> {
...
+                    let copy_data = match FeMessage::read(&mut r)? {
+                        Some(FeMessage::CopyData(bytes)) => bytes,
+                        Some(msg) => bail!("expected `CopyData` message, found {:?}", msg),
+                        None => bail!("connection closed unexpectedly"),
+                    };
+
+                    let msg = ProposerAcceptorMessage::parse(copy_data)?;
+                    msg_tx.send(msg)?;
+                }
+                // msg_rx will be dropped here, this will also close msg_tx
```
arssher, line 156: rx and tx should be swapped

walkeeper/src/receive_wal.rs
```diff
@@ -100,16 +95,80 @@ impl<'pg> ReceiveWalConn<'pg> {
...
         loop {
-            let reply = spg
-                .timeline
-                .get()
-                .process_msg(&msg)
-                .context("failed to process ProposerAcceptorMessage")?;
-            if let Some(reply) = reply {
-                self.write_msg(&reply)?;
+            if matches!(next_msg, Some(ProposerAcceptorMessage::AppendRequest(_))) {
+                // poll messages without blocking
```
arssher, line 102: Let's extend the comment to 'write data without flushing while it is readily available'.

walkeeper/src/wal_storage.rs
```diff
@@ -339,69 +383,67 @@ impl Storage for PhysicalStorage {
...
+    }
+
+    fn flush_wal(&mut self) -> Result<()> {
+        if self.flush_record_lsn == self.write_record_lsn {
+            // no need to do extra flush
+            return Ok(());
+        }
+
+        if let Some(mut unflushed_file) = self.file.take() {
+            self.fdatasync_file(&mut unflushed_file)?;
```
arssher, line 396: Option can be matched with `ref` keyword to its contents to avoid taking/returning it around.

walkeeper/src/wal_storage.rs
```diff
@@ -294,29 +336,31 @@ impl Storage for PhysicalStorage {
...
 
         {
             let _timer = self.metrics.write_wal_seconds.start_timer();
-            self.write_and_flush(startpos, buf)?;
+            self.write_exact(startpos, buf)?;
         }
 
-        // WAL is written and flushed, updating lsns
-        self.write_lsn = startpos + buf.len() as u64;
+        // WAL is written and flushed, updating LSNs
```
arssher, line 363: it is not flushed yet

walkeeper/src/wal_storage.rs
```diff
@@ -135,128 +159,146 @@ impl PhysicalStorage {
...
         let wal_seg_size = self
             .wal_seg_size
             .ok_or_else(|| anyhow!("wal_seg_size is not initialized"))?;
 
-        let mut bytes_left: usize = buf.len();
-        let mut bytes_written: usize = 0;
-        let mut partial;
-        let mut start_pos = startpos;
-        const ZERO_BLOCK: &[u8] = &[0u8; XLOG_BLCKSZ];
+        if self.write_lsn != pos {
```
arssher, line 263: How this can be true?

arssher (approved):

walkeeper/src/wal_storage.rs
petuhovskiy: I need to call `self.fdatasync_file` in this if, so with `ref` borrow I get an error:
```
cannot borrow `*self` as immutable because it is also borrowed as mutable
immutable borrow occurs hererustc[E0502](https://doc.rust-lang.org/error-index.html#E0502)
```

walkeeper/src/wal_storage.rs
petuhovskiy: Shouldn't be, possible only if someone if using private API directly, functions like `write_exact`.

petuhovskiy:
> These results are for local run on my machine, which has relatively slow fsync (2364 ops/sec, 423 usecs/op) and not very powerful 8 core CPU. I've added "per fsync" metrics, which are useful for disk usage comparison.

I've updated PR to use fdatasync in most places, and ran this test again. It seems that my machine has also much slower fsync, compared to fdatasync, and now results are almost the same:

 | Test info  | pgbench -i -s 200 (time) | pgbench -N -T 60 | init wal_bytes, per fsync | bench wal_bytes, per fsync | bench txes, per fsync |
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
| PATCHED safekeeper + wp (fsync=on, no checkpoints), using fdatasync  | 99 s | 4041 TPS | 301kb | 5439 b | 13.8 tx / fsync |
| vanilla repl (fsync=on, no checkpoints) | 149 s | 4170 TPS | 257kb | 5674 b | 14.5 tx / fsync |

PR #1321, opened 2022-02-23
Expose more config options

lubennikovaav:
Add 'wait_lsn_timeout' and 'wal_redo_timeout' pageserver config options instead of hardcoded defaults

 Add --pageserver-config-override to ZenithEnvBuilder to tune checkpointer and GC in tests.
    Usage example:
    zenith_env_builder.pageserver_config_override = "checkpoint_period = '100 s'; checkpoint_distance = 1073741824"

inspired by #1043 and flaky tests that need some fine-tuning

### DIFF
--- a/pageserver/src/walredo.rs
+++ b/pageserver/src/walredo.rs
@@ -102,8 +102,6 @@ impl crate::walredo::WalRedoManager for DummyRedoManager {
     }
 }
 
-static TIMEOUT: Duration = Duration::from_secs(20);
-
 // Metrics collected on WAL redo operations
 //
 // We collect the time spent in actual WAL redo ('redo'), and time waiting
@@ -221,7 +219,14 @@ impl WalRedoManager for PostgresRedoManager {
                 let result = if batch_zenith {
                     self.apply_batch_zenith(rel, blknum, lsn, img, &records[batch_start..i])
                 } else {
-                    self.apply_batch_postgres(rel, blknum, lsn, img, &records[batch_start..i])
+                    self.apply_batch_postgres(
+                        rel,
+                        blknum,
+                        lsn,
+                        img,
+                        &records[batch_start..i],
+                        self.conf.wal_redo_timeout,
+                    )
                 };
                 img = Some(result?);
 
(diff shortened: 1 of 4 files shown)


pageserver/src/walredo.rs
```diff
@@ -653,7 +667,7 @@ impl PostgresRedoProcess {
             // If we have more data to write, wake up if 'stdin' becomes writeable or
             // we have data to read. Otherwise only wake up if there's data to read.
             let nfds = if nwrite < writebuf.len() { 3 } else { 2 };
-            let n = nix::poll::poll(&mut pollfds[0..nfds], TIMEOUT.as_millis() as i32)?;
+            let n = nix::poll::poll(&mut pollfds[0..nfds], wal_redo_timeout.as_millis() as i32)?;
```
SomeoneToIgnore, line 670: Just a note (it's not the issue of the current PR, I just wanted to highlight the behaviour): `as_millis` returns `u128` and casting it to `i32` will _silently_ truncate the value.

```rust
let i32_max = i32::MAX as u64;
dbg!(Duration::from_millis(i32_max).as_millis() as i32);
dbg!(Duration::from_millis(i32_max + 1).as_millis() as i32);
```

outputs 
```
[src/main.rs:6] Duration::from_millis(i32_max).as_millis() as i32 = 2147483647
[src/main.rs:7] Duration::from_millis(i32_max + 1).as_millis() as i32 = -2147483648
```

Not sure that it's what we want :)
We could use https://doc.rust-lang.org/std/convert/trait.TryFrom.html to explicitly produce such conversion errors.

On the other hand, `i32::MAX` is roughly 596 hours, so not sure that's so critical to fix in this PR, feels like no sane person would want to use such timeouts.

test_runner/fixtures/zenith_fixtures.py
```diff
@@ -1051,6 +1063,12 @@ def append_pageserver_param_overrides(params_to_update: List[str],
             f'--pageserver-config-override={o.strip()}' for o in env_overrides.split(';')
         ]
 
+    if pageserver_config_override is not None:
```
SomeoneToIgnore, line 1066: The order of overrides matters, in any params repeat, the last values will be taken.
I propose to move this block upwards, before `env_overrides = os.getenv('ZENITH_PAGESERVER_OVERRIDES')` since IMO the env vars has to to be the ultimate overrides.

Or we could remove the env override functionality entirely?

SomeoneToIgnore (approved):
LGTM in general, yet I would personally prefer the `ZENITH_PAGESERVER_OVERRIDES` comment to be dealt with before merging this.

test_runner/fixtures/zenith_fixtures.py
lubennikovaav: I think that precedence order where explicit CLI params override ENV is more standard.
check for example AWS cli https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html

For the remote_storage use case, it makes sense to override everything using env to enable it for all tests at once.
But if it is specifically overridden in the particular test, we should use this value.

hlinnaka (approved):

Issue #1320, opened 2022-02-23
Epic: Support time-based PITR in pageserver


stepashka:
start with an RFC

- [ ] GC interval is configured in LSN (bytes), not time
- [ ] GC interval is configured per pageserver, not per tenant

knizhnik:
I wonder if time and LSN criteria for PITR should be ORed or ANDed?
Right now in my implementation I use AND: i.e. record is removed only if it is out of specified time interval and out of LSN range. It seems to be the safest option. But it makes it not possible to say: I want to keep changes for one month but only of them do not exceed 100Gb. Make it possible for user to specify OR/AND? It seems to be overkill... Always use OR?

knizhnik:
I have not changed console code to make it possible to specify per-tenant parameters.
But it is not possible through CLI:
```
zenith tenant create -c "gc_period:10 s" -c "gc_horizon:100000"
```

knizhnik:
Before discussion concerning PITR support scheduled for tomorrow I want to share my thought 
about it:
1. Nobody will want to rollback too far in the past (i.e one week), because it actually means loosing updates for the whole week. Users may want to perform "time travel" and check state of the database a wee ago. But it is actually slightly different story: it is not a "recovery".
2. In case of time travel far in the fast, we do not need to specify precise time, i.e. 7 days, 1 hours, 13 minutes and 33.069 seconds ago. So usually it is enough to save just one snapshot for the particular day. If size of free-tier database is limited by 10Gb, then storing 7 snapshots will require only 70Gb, which is smaller than 100Gb (suggested limit for PITR interval).
3. Users know nothing or little about LSNs and WAL. So saying that PITR interval is limited by 100Gb says nothing to them. Except that despite of requested one week PITR interval, they still may be not able to see state of the database one week ago.

So my suggestions are:
1. Do not public LSN limit for PITR interval. User s will specify only time range for PITR. And it should be limited let's say by 10 days. In this case we can enforce that size of users data will not exceed 100Gb.
2. Assume that WAL (delta layers) are stored only within one day. It can be also configured if needed. Older delta layers are removed and we store only snapshots (image layers). One per day (also can be configured).
3. Image layers older than 10 days can be removed from S3

antons-antons:
[...]

PR #1332, opened 2022-02-25
Define PiTR interval for GC and make it possible to specify per-tenan…

knizhnik:
…t configuration parameters

refer #1320

### DIFF
--- a/control_plane/src/storage.rs
+++ b/control_plane/src/storage.rs
@@ -1,3 +1,4 @@
+use std::collections::HashMap;
 use std::io::Write;
 use std::net::TcpStream;
 use std::path::PathBuf;
@@ -342,10 +343,22 @@ impl PageServerNode {
     pub fn tenant_create(
         &self,
         new_tenant_id: Option<ZTenantId>,
+		settings: HashMap<&str, &str>
     ) -> anyhow::Result<Option<ZTenantId>> {
         let tenant_id_string = self
             .http_request(Method::POST, format!("{}/tenant", self.http_base_url))
-            .json(&TenantCreateRequest::new(tenantid))
+            .json(&TenantCreateRequest {
+                new_tenant_id,
+                checkpoint_distance: settings
+                    .get("checkpoint_distance")
+                    .map(|x| x.parse::<u64>().unwrap()),
+                compaction_period: settings.get("compaction_period").map(|x| x.to_string()),
+                gc_horizon: settings
+                    .get("gc_horizon")
+                    .map(|x| x.parse::<u64>().unwrap()),
+                gc_period: settings.get("gc_period").map(|x| x.to_string()),
+                pitr_interval: settings.get("pitr_interval").map(|x| x.to_string()),
+            })
             .send()?
             .error_from_body()?
             .json::<Option<String>>()?;
--- a/control_plane/src/storage.rs
+++ b/control_plane/src/storage.rs
@@ -343,7 +343,7 @@ impl PageServerNode {
     pub fn tenant_create(
         &self,
         new_tenant_id: Option<ZTenantId>,
-		settings: HashMap<&str, &str>
+        settings: HashMap<&str, &str>,
     ) -> anyhow::Result<Option<ZTenantId>> {
         let tenant_id_string = self
             .http_request(Method::POST, format!("{}/tenant", self.http_base_url))
--- a/control_plane/src/storage.rs
+++ b/control_plane/src/storage.rs
@@ -352,6 +352,9 @@ impl PageServerNode {
                 checkpoint_distance: settings
                     .get("checkpoint_distance")
                     .map(|x| x.parse::<u64>().unwrap()),
+                compaction_target_size: settings
+                    .get("compaction_target_size")
+                    .map(|x| x.parse::<u64>().unwrap()),
                 compaction_period: settings.get("compaction_period").map(|x| x.to_string()),
                 gc_horizon: settings
                     .get("gc_horizon")
--- a/control_plane/src/storage.rs
+++ b/control_plane/src/storage.rs
@@ -345,7 +345,7 @@ impl PageServerNode {
     ) -> anyhow::Result<Option<ZTenantId>> {
         let tenant_id_string = self
             .http_request(Method::POST, format!("{}/tenant", self.http_base_url))
-            .json(&TenantCreateRequest { new_tenant_id })
+            .json(&TenantCreateRequest::new(tenantid))
             .send()?
             .error_from_body()?
             .json::<Option<String>>()?;
--- a/pageserver/src/config.rs
+++ b/pageserver/src/config.rs
@@ -150,6 +150,7 @@ pub const TENANT_CONFIG_NAME: &str = "config";
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
 pub struct TenantConf {
     pub checkpoint_distance: u64,
+    pub compaction_target_size: u64,
     pub compaction_period: Duration,
     pub gc_horizon: u64,
     pub gc_period: Duration,
@@ -164,6 +165,7 @@ impl TenantConf {
             pitr_interval: conf.pitr_interval,
             checkpoint_distance: conf.checkpoint_distance,
             compaction_period: conf.compaction_period,
+            compaction_target_size: conf.compaction_target_size,
         }
     }
     pub fn save(&self, conf: &'static PageServerConf, tenantid: ZTenantId) -> Result<()> {
--- a/pageserver/src/config.rs
+++ b/pageserver/src/config.rs
@@ -164,25 +164,8 @@ impl TenantConf {
             pitr_interval: conf.pitr_interval,
             checkpoint_distance: conf.checkpoint_distance,
             compaction_period: conf.compaction_period,
-		}
-	}
-}
-
-// use dedicated enum for builder to better indicate the intention
-// and avoid possible confusion with nested options
-pub enum BuilderValue<T> {
-    Set(T),
-    NotSet,
-}
-
-impl<T> BuilderValue<T> {
-    pub fn ok_or<E>(self, err: E) -> Result<T, E> {
-        match self {
-            Self::Set(v) => Ok(v),
-            Self::NotSet => Err(err),
         }
     }
-
     pub fn save(&self, conf: &'static PageServerConf, tenantid: ZTenantId) -> Result<()> {
         let _enter = info_span!("saving tenant config").entered();
         let path = conf.tenant_path(&tenantid).join(TENANT_CONFIG_NAME);
@@ -210,6 +193,22 @@ impl<T> BuilderValue<T> {
             Err(err) if err.kind() == std::io::ErrorKind::NotFound => Ok(Self::from(conf)),
             Ok(config_bytes) => Ok(TenantConf::des(&config_bytes)?),
             Err(err) => bail!(err),
+        }
+    }
+}
+
+// use dedicated enum for builder to better indicate the intention
+// and avoid possible confusion with nested options
+pub enum BuilderValue<T> {
+    Set(T),
+    NotSet,
+}
+
+impl<T> BuilderValue<T> {
+    pub fn ok_or<E>(self, err: E) -> Result<T, E> {
+        match self {
+            Self::Set(v) => Ok(v),
+            Self::NotSet => Err(err),
         }
     }
 }
@@ -314,7 +313,7 @@ impl PageServerConfigBuilder {
         self.gc_period = BuilderValue::Set(gc_period)
     }
 
-    pub fn pitr_interval(&mut self, gc_period: Duration) {
+    pub fn pitr_interval(&mut self, pitr_interval: Duration) {
         self.pitr_interval = BuilderValue::Set(pitr_interval)
     }
 
@@ -386,7 +385,9 @@ impl PageServerConfigBuilder {
                 .gc_horizon
                 .ok_or(anyhow::anyhow!("missing gc_horizon"))?,
             gc_period: self.gc_period.ok_or(anyhow::anyhow!("missing gc_period"))?,
-            pitr_interval: self.pitr_interval.ok_or(anyhow::anyhow!("missing pitr_interval"))?,
+            pitr_interval: self
+                .pitr_interval
+                .ok_or(anyhow::anyhow!("missing pitr_interval"))?,
             wait_lsn_timeout: self
                 .wait_lsn_timeout
                 .ok_or(anyhow::anyhow!("missing wait_lsn_timeout"))?,
--- a/pageserver/src/config.rs
+++ b/pageserver/src/config.rs
@@ -15,7 +15,7 @@ use zenith_utils::zid::{ZNodeId, ZTenantId, ZTimelineId};
 
 use std::convert::TryInto;
 use std::env;
-use std::fs::OpenOptions;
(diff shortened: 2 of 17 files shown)


control_plane/src/storage.rs
```diff
@@ -321,11 +322,20 @@ impl PageServerNode {
...
-                tenant_id: tenantid,
+                tenant_id,
+                checkpoint_distance: settings
+                    .get("checkpoint_distance")
+                    .map(|x| x.parse::<u64>().unwrap()),
+                checkpoint_period: settings.get("checkpoint_period").map(|x| x.to_string()),
+                gc_horizon: settings
+                    .get("gc_horizon")
+                    .map(|x| x.parse::<u64>().unwrap()),
+                gc_period: settings.get("gc_period").map(|x| x.to_string()),
```
LizardWizzard, line 337: see [Option::copied](https://doc.rust-lang.org/nightly/std/option/enum.Option.html#method.cloned)

pageserver/src/config.rs
```diff
@@ -118,6 +127,53 @@ pub struct PageServerConf {
...
+    pub fn save(&self, conf: &'static PageServerConf, tenantid: ZTenantId) -> Result<()> {
+        let _enter = info_span!("saving tenant config").entered();
+        let path = conf.tenant_path(&tenantid).join(TENANT_CONFIG_NAME);
+        let mut file =
+            VirtualFile::open_with_options(&path, OpenOptions::new().write(true).create_new(true))?;
+        let config_bytes = self.ser()?;
+        if file.write(&config_bytes)? != config_bytes.len() {
+            bail!("Could not write all the metadata bytes in a single call");
+        }
+        file.sync_all()?;
```
LizardWizzard, line 161: Should we do proper dance with fsyncing the directory on first save etc? for reference you can see how metadata file is written to disk

pageserver/src/layered_repository.rs
```diff
@@ -1743,8 +1758,33 @@ impl LayeredTimeline {
...
-            // 2. Is it needed by a child branch?
+            // 2. It is newer than PiTR interval?
+            // We use modification time of layer file to estimate update time.
+            // This estimation is not quite precise but maintaining LSN->timestamp map seems to be overkill.
+            // It is not expected that users will need high precision here. And this estimation
+            // is conservative: modification time of file is always newer than actual time of version
+            // creation. So it is safe for users.
+            //
+            if let Ok(metadata) = fs::metadata(&l.filename()) {
+                let last_modified = metadata.modified()?;
```
LizardWizzard, line 1769: Also, what do you think about writing layer file creation time inside it? fs metadata can become out of sync with the real file creation time because of s3 upload/download for example. We won't loose anything, but this way we will keep potentially a lot of outdated files. 

Another idea, we can use some index file with timestamps for all layers so this way it is possible to avoid some extra  fs operations to make gc a bit cheaper.

LizardWizzard (commented):
Thanks for the patch! I've left a few comments. 

Would be good to add tenant config specific unit/integration tests

pageserver/src/http/routes.rs
```diff
@@ -318,9 +318,29 @@ async fn tenant_create_handler(mut request: Request<Body>) -> Result<Response<Bo
 
     let request_data: TenantCreateRequest = json_request(&mut request).await?;
 
+    let conf = get_config(&request);
+    let mut tenant_conf = TenantConf::from(conf);
+    if let Some(gc_period) = request_data.gc_period {
```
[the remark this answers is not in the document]
knizhnik: Sorry, 

> Thanks for the patch! I've left a few comments.
> 
> Would be good to add tenant config specific unit/integration tests

Yes, tests definitely are needed. The main problem is that there is no way now to inspect values of this parameters. I wonder if we should add it to HTTP API of pageserver. Or may be there is  some other way to inspect them.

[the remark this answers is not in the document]
knizhnik: > I wonder can it be better with a custom from impl? Or something like `TenantConf::update`

Sorry, do not completely understand.Can you clarify please your suggestion? `TenantConf::from` now just initializing per-tenant config from global config. And it is not clear to mewhat can be passed to  `TenantConf::update` method as parameters? All fields TenantConf as options? Seems to be neither saf, neither convenient...

pageserver/src/layered_repository.rs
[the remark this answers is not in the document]
knizhnik: I am subtracting system times at pageserver. It is not somehow related with time at client side and time at S3. So even if time at pageserver is not correct, it should not cause problems, Certainly there can be cases when system time is adjusted. Either explicitly by admin, either implicitly by NTP protocol. In this case such side effects are really possible. But I do not expect that PiTR interval wil be specified with high resolution (seconds or even hours). Most probably it will be days or even months. In this case such time correction will not be critical.

pageserver/src/tenant_mgr.rs
```diff
@@ -78,15 +78,21 @@ pub fn set_timeline_states(
...
+                        Arc::new(walredo_mgr),
+                        tenant_id,
+                        conf.remote_storage_config.is_some(),
+                    ));
+                    Tenant {
+                        state: TenantState::Idle,
+                        repo,
+                    }
+                }
+                Err(e) => panic!("Failed to load tenant state: {:?}", e),
```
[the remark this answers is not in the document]
knizhnik: Yes. error! is more appreciate here.

pageserver/src/config.rs
knizhnik: I thought that config is less critical than metadata so loosing it is not fatal.
But at the same time it is rarely updated, so not a problem ti fsync parent dir as well.

pageserver/src/layered_repository.rs
knizhnik: Modification time is greater or equal than creation time. As far as we are using it for cut-off, it is more conservative.
The semantic of PiTR interval is that user want to return back to the point which is not older in time than specified value. So using modification data (which correponds to last recent update) is safer.

control_plane/src/storage.rs
LizardWizzard: Oh, it's shame that `cloned` doesn't cover this case. You can use `.map(String::from)` but this has almost the same length

pageserver/src/layered_repository.rs
LizardWizzard: Sorry, I meant slightly different thing, I meant that both creation and modification times may not be preserved through full cycle of upload/download (I confirmed that with a local run of disabled on main test_tenant_relocation). e g file creation/modification time will be the time of the download, not the time of a file creation on the original pageserver it came from.

Also as far as we are using append only style creation vs modification does not make sense.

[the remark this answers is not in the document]
LizardWizzard: You're right, I think I thought more about TimelineMetadata here, but it is unrelated to this code, remaining concerns are covered here https://github.com/zenithdb/zenith/pull/1332#discussion_r823006105

pageserver/src/tenant_mgr.rs
```diff
@@ -78,21 +78,23 @@ pub fn set_timeline_states(
...
-                    ));
-                    Tenant {
-                        state: TenantState::Idle,
-                        repo,
-                    }
+            let tenant_conf = match TenantConf::load(conf, tenant_id) {
+                Ok(tenant_conf) => tenant_conf,
+                Err(e) => {
+                    error!("Failed to load tenant state: {:?}", e);
+                    TenantConf::from(conf)
```
LizardWizzard, line 85: Am I right that we will use default config if we failed to load the tenant specific one? If tenant has PITR of one year and we failed to load the config and loaded default one with PITR of e g one month GC will delete the data we actually need. 

I think we cannot use the default config here and we need to completely skip initialization of this tenant without affecting others. Plus ideally we should report this error somehow, maybe through timeline detail rest api response. But this code is undergoing through massive changes so this can be done separately to avoid conflicts

pageserver/src/layered_repository.rs
knizhnik: First of all eviction procedure for pageserver data is not currently implemented (o I missed something).
So pageserver now contains all data and never request it from S3.
In theory it will be possible that we have independent GC (and PiTR intervals) for pageserver and S3.Butright now it is not the case.

But even if some tenant/timeline was just loaded to pageserver from S3, then current approach is safe: it will not remove data which should not be removed according to PiTR interval. Yes, it may not remove too old layer just because it was just uploaded from S3. And if this layer will be evicted and then once again uploaded to this pageserver, then in theory this layer may never be garbage collected. But IMHO it is very strange and unrealistic scenario.

pageserver/src/tenant_mgr.rs
knizhnik: Not sure if preventing access to some tenant data (because of config file corruption) is better solution than just using default config with risk of deleting data which should not be deleted.

But practically in this peace of code it is not so easy to implement what you have suggested (if not replace  `or_insert_with` method with something else).

pageserver/src/layered_repository.rs
LizardWizzard: > First of all eviction procedure for pageserver data is not currently implemented (o I missed something).

Every checkpoint is uploaded to s3 now. It is not proper eviction but files are transferred and IMO it is enough to cause trouble (bloat)

> So pageserver now contains all data and never request it from S3.

New pageserver can request data from s3 e g during tenant migration.

> Yes, it may not remove too old layer just because it was just uploaded from S3. And if this layer will be evicted and then once again uploaded to this pageserver, then in theory this layer may never be garbage collected. But IMHO it is very strange and unrealistic scenario.

Thats exactly what I'm talking about. It is safe but we potentially may keep a lot of extra data causing bloat.
Eviction/on-demand download are planned to be implemented so this case is definitely realistic to me

pageserver/src/tenant_mgr.rs
LizardWizzard: > Not sure if preventing access to some tenant data (because of config file corruption) is better solution than just using default config with risk of deleting data which should not be deleted.

I think it is better to be on the safe side (IMO data loss never should be considered as a viable option), other options:
* make this timeline readonly
* disable gc for it (can cause bloat without readonly option)
* retry whole operation (e g download)

Personally I'm inclined to retry the operation and deal with that in the same way as with failed download of a timeline. We probably need to guard this file with the checksum too.

@hlinnaka @SomeoneToIgnore thoughts?

> it is not so easy to implement what you have suggested

Feel free to leave a TODO assigned to me, because this code is being heavily reworked in #1163 so I can handle it there or in the follow up PR

LizardWizzard: @knizhnik #1163 was merged, so this is no longer a problem. You can rebase your PR on top. Please reach out too me if you need help with that

pageserver/src/http/routes.rs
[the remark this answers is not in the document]
LizardWizzard: > I wonder if we should add it to HTTP API of pageserver.

+1 on that, don't forget to update API spec

lubennikovaav:
I have rebased this PR on top of the main and moved `compaction_target_size` to per-tenant config.
Now I'm continuing review and going to fix the issue with possibly corrupted tenant config file mentioned above.

lubennikovaav:
So, I've added checksum to tenant config file. 
Now this file is saved when tenant is created and all other load functions expect the file to be there.
@LizardWizzard , could you take another look at the code and help me to combine that with tenant migration?

hlinnaka:
How about storing the config in a json or toml format? That would make it human-readable and editable.

lubennikovaav:
> How about storing the config in a json or toml format? That would make it human-readable and editable.

IMHO, it would be tricky to keep the consistency if we make it human-editable. When should we reload such config? 
How can we make sure that it is not corrupted during tenant relocation (now it is protected with checksum)?

I would rather introduce `zenith tenant config` command  to update the config.

knizhnik:
> > How about storing the config in a json or toml format? That would make it human-readable and editable.
> 
> IMHO, it would be tricky to keep the consistency if we make it human-editable. When should we reload such config? How can we make sure that it is not corrupted during tenant relocation (now it is protected with checksum)?
> 
> I would rather introduce `zenith tenant config` command to update the config.

+1 
It is definitely easier to inspect config in text format. But it also instigate person to edit this file. And in this case we have inconsistency with control panel. Also I am not sure who and why will inspect this file. Certainly it may be useful during debugging. But in production we definitely need some other mechanism (GUI/CLI/...) to inspect configuration files. Especially if there will be thousands of clients.

knizhnik:
We have a discussion with @hlinnaka but still do not come to the consensus:
I think that this PR is unrelated with #1386 

#1386 provides on demand mapping of timestamp to LSN. It requires traversal of all XLOG_XACT_COMMIT records. Number of committed transaction can be very large, especially if cae of larger PITR interval.

I still thinking that users very rarely need to rollback to some particular point in time. To some particular snapshot or database state - yes. But time is not a precise instrument for rollback. It can be useful to perform some kind of time travel or "as of" queries (which results were returned by query yesterday or what plan was chosen for this query one hour ago). But for rollback we need more precise pointer.

This is why efficiency of such time->LSN mapping is not so important.
But if GC will have to performing this mapping before each GC iteration, it can just cause waste of resources (CPU  and disk).

The argument that layer file time may be not correct because of downloading layer from S3 seems to be not correct:
right now we do not have swapping to S3, so layers, may not be uploaded to S3, then evicted from page server and then reloaded on demand.

If we really consider S3 as main storage of data and pageserver - just as cache for most recently/frequently accessed data, then many things in page server have to be rewritten.
For example, GC should not try to drop any layers from local storage. It should only detect layers which can be deleted and then remove them from S3. In this case it can look at file creation date at S3, which should not be changed. And unloading layer should be triggered by completely different criteria than currently used in GC. We can unload layer at any moment when when there are more recent image layers and there is lack of free space at pageserver. And if evicted layer will needed,
it can be downloaded from S3.

LizardWizzard:
@lubennikovaav 

> help me to combine that with tenant migration

Migration currently works on a per timeline basis, heavily relying on the s3, so it is a bit tricky. What will be the source of truth for these configs? Probably it will be the console. So I see two options:
1. Console uses API to fill correct config on pageserver. Existing `tenant_create` to load the initial config and a new endpoint (something like `tenant_update`) to change the config of the existing tenant.
2. As we discussed in the recent meeting about the synchronization, there was a desire to use etcd and subscriptions to changes instead of API requests from console to storage to utilize etcd for that.

IMO either of the options can be implemented separately. The second one needs some preparations that are not ready yet

lubennikovaav:
> The argument that layer file time may be not correct because of downloading layer from S3 seems to be not correct: right now we do not have swapping to S3, so layers, may not be uploaded to S3, then evicted from page server and then reloaded on demand.
> 
And what about tenant migration between pageservers via S3?

knizhnik:
> > The argument that layer file time may be not correct because of downloading layer from S3 seems to be not correct: right now we do not have swapping to S3, so layers, may not be uploaded to S3, then evicted from page server and then reloaded on demand.
> 
> What about tenant migration between pageservers via S3?

I wrote my thoughts about it in PITR Epic ticket: https://github.com/zenithdb/zenith/issues/1320#issuecomment-1091819948
Briefly: if we support swapping of layers to S3, then GC has to be also done at S3 and we can use timestamps of S3 files (which can not be changed because of migration or whatever else other reasons). And we do not need to perform GC at pageserver, we should only implement some eviction policy for layers.

PR #1335, opened 2022-03-01
Fix 1.59 rustc clippy warnings

SomeoneToIgnore:


compute_tools/src/pg_helpers.rs
```diff
@@ -171,7 +171,7 @@ impl PgQuote for PgIdent {
     /// always quotes provided string with `""` and escapes every `"`. Not idempotent,
     /// i.e. if string is already escaped it will be escaped again.
     fn quote(&self) -> String {
-        let result = format!("\"{}\"", self.replace("\"", "\"\""));
+        let result = format!("\"{}\"", self.replace('\"', "\"\""));
```
dhammika, line 174: Escape in `'\"' ` match pattern is redundant right?

PR #1336, opened 2022-03-02
Replace zenith/build build image with zimg/* ones

SomeoneToIgnore:


SomeoneToIgnore:
Also see https://github.com/zenithdb/docker-images/pull/2 : if that PR is correct and works, the unification brings us one step closer to automatic arm docker builds.

funbringer (approved):

PR #1345, opened 2022-03-08
Docker improvement

### LINKED ISSUE #1113: Docker image build time on main is huge
It took 25 minutes to build the docker image. The reason is that it builds everything from scratch without caching

https://app.circleci.com/pipelines/github/zenithdb/zenith/3967/workflows/af369227-1d25-48ad-bd58-57721d042abf/jobs/35386

@cicdteam Is it known issue?

cicdteam:
- docker image size (`zenithdb/zenith`) reduced in ~0.5 Gb
- caching added for rust compiler dependencies
- docker build time reduced in twice: from ~30 min to ~15 min (`cargo build --release` takes 6 min instead of 20 min now)

closes #1113

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -427,7 +427,7 @@ jobs:
   # Build zenithdb/zenith:latest image and push it to Docker hub
   docker-image:
     docker:
-      - image: zimg/rust:1.56
+      - image: cimg/base:2021.04
     steps:
       - checkout
       - setup_remote_docker:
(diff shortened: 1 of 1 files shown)


SomeoneToIgnore (approved):
Feels like there has to be one more PR into https://github.com/zenithdb/docker-images to install `cachepot` on them?

Otherwise decrease looks very promising, let's see how big that cache grows.
I think, we can invalidate every time with rustc bump (less than 10 per year), or simply periodically.

cicdteam:
> Feels like there has to be one more PR into https://github.com/zenithdb/docker-images to install `cachepot` on them?
> 
> Otherwise decrease looks very promising, let's see how big that cache grows. I think, we can invalidate every time with rustc bump (less than 10 per year), or simply periodically.

yes, cachepot was added to zimg images, next step to  reduce build time (not only dockers other steps too) will be moving to GitHub Actions and using ephemeral but powerful own runners

PR #1347, opened 2022-03-08
revert docker build to debian:buster based rust

cicdteam:
`zimg/rust:1.56` has too recent libc6 version (as image based on Debian 11) and not compatible with current zenith storage infrastructure (based on Debian 10)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 # Build Postgres
 #
-FROM zimg/rust:1.56 AS pg-build
+#FROM zimg/rust:1.56 AS pg-build
+FROM zenithdb/build:buster-20220309 AS pg-build
 WORKDIR /pg
 
 USER root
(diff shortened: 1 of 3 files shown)


SomeoneToIgnore (approved):

SomeoneToIgnore:
Out of curiosity, what do we need to migrate storage infra towards new libc?
We still would need zimg to build zenith eventually.

Dockerfile
```diff
@@ -1,6 +1,7 @@
 # Build Postgres
 #
-FROM zimg/rust:1.56 AS pg-build
+#FROM zimg/rust:1.56 AS pg-build
+FROM zenithdb/build:buster-20220309 AS pg-build
```
SomeoneToIgnore, line 4: I've removed the corresponding entry in README, so now would be good to have a comment somewhere about this image and where does it come from.

Unless we plan to move onto zimg soon, then it does not make much sense.

PR #1365, opened 2022-03-15
Compile and run unit tests on macOs

SomeoneToIgnore:
Closes https://github.com/zenithdb/zenith/issues/556

Uses _**non-arm**_ macOs runner from https://github.com/actions/virtual-environments
Currently no arm runners are supported in GitHub.

Note that the GitHub workflow I've edited runs only on `main` branch, which is quite surprising: the job is a basic debug build with unit test run.
Should we rather run it on every PR/push to check postgres macOs miscompilations eagerly?

CI runs for hacked version, showing that Ubuntu and macOs build pass:
* [Ubuntu](https://github.com/zenithdb/zenith/runs/5550732113?check_suite_focus=true)
* [macOs](https://github.com/zenithdb/zenith/runs/5550732224?check_suite_focus=true)

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -13,7 +13,7 @@ jobs:
         # If we want to duplicate this job for different
         # Rust toolchains (e.g. nightly or 1.37.0), add them here.
         rust_toolchain: [stable]
-        os: [ubuntu-latest]
+        os: [ubuntu-latest, macos-latest]
     timeout-minutes: 30
     name: run regression test suite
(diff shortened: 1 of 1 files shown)


SomeoneToIgnore:
Talked to @kelvich more and decided to work towards migrating basic test suite from CircleCI to the GitHub workflow changed in the PR.

As a first step, the workflow is now `on: push`, running on every push anywhere, not only on `main` PRs and pushes, as it was before.

LizardWizzard (approved):
LGTM

PR #1367, opened 2022-03-15
Refactor commit_lsn updates in safekeeper

petuhovskiy:
Fix issue that commit_lsn doesn't update in handle_flush, introduced by #1318

PR #1377, opened 2022-03-17
Add safekeeper information exchange through etcd.

arssher:
Add safekeeper information exchange through etcd.
    
    Safekeers now publish to and pull from etcd per-timeline data. Immediate goal is
    WAL truncation, for which every safekeeper must know remote_consistent_lsn; the
    next would be callmemaybe replacement.
    
    Adds corresponding '--broker' argument to safekeeper and ability to run etcd in
    tests.

### DIFF
--- a/control_plane/src/local_env.rs
+++ b/control_plane/src/local_env.rs
@@ -57,6 +57,10 @@ pub struct LocalEnv {
     #[serde(default)]
     pub private_key_path: PathBuf,
 
+    // A comma separated broker (etcd) endpoints for storage nodes coordination, e.g. 'http://127.0.0.1:2379'.
+    #[serde(default)]
+    pub broker_endpoints: Option<String>,
+
     pub pageserver: PageServerConf,
 
     #[serde(default)]
--- a/test_runner/fixtures/zenith_fixtures.py
+++ b/test_runner/fixtures/zenith_fixtures.py
@@ -33,7 +33,7 @@ from typing_extensions import Literal
 import requests
 import backoff  # type: ignore
 
-from .utils import (get_self_dir, lsn_from_hex, mkdir_if_needed, subprocess_capture)
+from .utils import (etcd_path, get_self_dir, mkdir_if_needed, subprocess_capture, lsn_from_hex)
 from fixtures.log_helper import log
 """
 This file contains pytest fixtures. A fixture is a test resource that can be
@@ -425,7 +425,8 @@ class ZenithEnvBuilder:
                  num_safekeepers: int = 0,
                  pageserver_auth_enabled: bool = False,
                  rust_log_override: Optional[str] = None,
-                 default_branch_name=DEFAULT_BRANCH_NAME):
+                 default_branch_name=DEFAULT_BRANCH_NAME,
+                 broker: bool = False):
         self.repo_dir = repo_dir
         self.rust_log_override = rust_log_override
         self.port_distributor = port_distributor
@@ -434,6 +435,7 @@ class ZenithEnvBuilder:
         self.num_safekeepers = num_safekeepers
         self.pageserver_auth_enabled = pageserver_auth_enabled
         self.default_branch_name = default_branch_name
+        self.broker = broker
         self.env: Optional[ZenithEnv] = None
 
         self.s3_mock_server: Optional[MockS3Server] = None
@@ -509,6 +511,8 @@ class ZenithEnvBuilder:
             self.env.pageserver.stop(immediate=True)
             if self.s3_mock_server:
                 self.s3_mock_server.kill()
+            if self.env.broker is not None:
+                self.env.broker.stop()
 
 
 class ZenithEnv:
@@ -561,6 +565,16 @@ class ZenithEnv:
             default_tenant_id = '{self.initial_tenant.hex}'
         """)
 
+        self.broker = None
+        if config.broker:
+            # keep etcd datadir inside 'repo'
+            self.broker = Etcd(datadir=os.path.join(self.repo_dir, "etcd"),
+                               port=self.port_distributor.get_port(),
+                               peer_port=self.port_distributor.get_port())
+            toml += textwrap.dedent(f"""
+            broker_endpoints = 'http://127.0.0.1:{self.broker.port}'
+        """)
+
         # Create config for pageserver
         pageserver_port = PageserverPort(
             pg=self.port_distributor.get_port(),
@@ -603,11 +617,14 @@ class ZenithEnv:
         self.zenith_cli.init(toml)
 
     def start(self):
-        # Start up the page server and all the safekeepers
+        # Start up the page server, all the safekeepers and the broker
         self.pageserver.start()
 
         for safekeeper in self.safekeepers:
             safekeeper.start()
+
+        if self.broker is not None:
+            self.broker.start()
 
     def get_safekeeper_connstrs(self) -> str:
         """ Get list of safekeeper endpoints suitable for wal_acceptors GUC  """
@@ -1666,6 +1683,7 @@ class Safekeeper:
 class SafekeeperTimelineStatus:
     acceptor_epoch: int
     flush_lsn: str
+    remote_consistent_lsn: str
 
 
 @dataclass
@@ -1689,7 +1707,8 @@ class SafekeeperHttpClient(requests.Session):
         res.raise_for_status()
         resj = res.json()
         return SafekeeperTimelineStatus(acceptor_epoch=resj['acceptor_state']['epoch'],
-                                        flush_lsn=resj['flush_lsn'])
+                                        flush_lsn=resj['flush_lsn'],
+                                        remote_consistent_lsn=resj['remote_consistent_lsn'])
 
     def get_metrics(self) -> SafekeeperMetrics:
         request_result = self.get(f"http://localhost:{self.port}/metrics")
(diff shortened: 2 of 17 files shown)


arssher:
TODO: add etcd to CI images and perhaps assign it random ports.

walkeeper/src/broker.rs
```diff
@@ -0,0 +1,201 @@
...
+        // revive the lease
+        keeper
+            .keep_alive()
+            .await
+            .context("failed to send LeaseKeepAliveRequest")?;
+        ka_stream
+            .message()
+            .await
+            .context("failed to receive LeaseKeepAliveResponse")?;
+        sleep(Duration::from_millis(PUSH_INTERVAL_MSEC)).await;
```
SomeoneToIgnore, line 112: NIT: `Duration` is Copy hence could be easily initialised outside of the loop, thus avoiding division (via `Duration::from_millis`) in a loop.

walkeeper/src/broker.rs
```diff
@@ -0,0 +1,201 @@
...
+async fn push_loop(conf: SafeKeeperConf) -> Result<()> {
+    let mut client = Client::connect(conf.broker_endpoints.as_ref().unwrap(), None).await?;
+
+    // Get and maintain lease to automatically delete obsolete data
+    let lease = client.lease_grant(LEASE_TTL, None).await?;
+    let (mut keeper, mut ka_stream) = client.lease_keep_alive(lease.id()).await?;
+
+    loop {
+        let active_tlis = GlobalTimelines::get_active_timelines();
+        for zttid in &active_tlis {
```
SomeoneToIgnore, line 89: `GlobalTimelines` methods used here and below call `lock()` on an `std::sync::Mutex` and we run all this machinery on a `new_current_thread` runtime, so these locks would block the entire runtime every time.

Another similar note is that `.await` here is done in a `for` loop: so if any of the `.await` would yield and the rest of the futures will just wait before their for loop iteration, without any progress.
There's a way to switch to running other future via `FururesUnordered`, here's an example in our code base:
https://github.com/zenithdb/zenith/blob/7738254f83c86e46795b34db834d18af97197d8d/pageserver/src/remote_storage/storage_sync.rs#L511-L535


Not sure how both are important at this stage, but feel worth of a TODO comment at least?

walkeeper/src/broker.rs
```diff
@@ -0,0 +1,201 @@
...
+    lazy_static! {
+        static ref TIMELINE_SAFEKEEPER_RE: Regex =
+            Regex::new(r"^zenith/([[:xdigit:]]+)/([[:xdigit:]]+)/safekeeper/([[:digit:]])$")
+                .unwrap();
+    }
+    let mut client = Client::connect(conf.broker_endpoints.as_ref().unwrap(), None).await?;
+    loop {
+        let wo = WatchOptions::new().with_prefix();
+        // TODO: subscribe only to my timelines
+        let (_, mut stream) = client.watch("zenith", Some(wo)).await?;
```
SomeoneToIgnore, line 127: > "zenith"

Out of curiosity: why does not push code have any such key? 
If it should have, this feels worthy of being a constant to me.

SomeoneToIgnore (commented):
Did I get you right that this is the PR I'm checking against to move callmemaybe over etcd?

Feels like we won't need a separate client library for the whole thing indeed, thanks for showing the way 🙂

arssher:
> Did I get you right that this is the PR I'm checking against to move callmemaybe over etcd?

yeah

petuhovskiy (commented):
Still looking at this PR, will probably add more comments later. For now I'm thinking about SafekeeperMemState, it looks nice to have, but will conflict with #1367. Maybe it's better to merge it in #1367 separately from this PR?

walkeeper/src/broker.rs
petuhovskiy: It's a prefix of https://github.com/zenithdb/zenith/blob/cd75d0706aed4e52e2b8a14463a1bac5b34eec99/walkeeper/src/broker.rs#L71 but probably still worth a constant or param in config, to be able to run several zenith installations on a single etcd.

walkeeper/src/broker.rs
```diff
@@ -0,0 +1,201 @@
...
+use zenith_utils::{
+    lsn::Lsn,
+    zid::{ZNodeId, ZTenantTimelineId},
+};
+
+use crate::{safekeeper::Term, timeline::GlobalTimelines, SafeKeeperConf};
+
+const RETRY_INTERVAL_MSEC: u64 = 1000;
+const PUSH_INTERVAL_MSEC: u64 = 5000;
+const LEASE_TTL: i64 = 60;
```
petuhovskiy, line 31:
```suggestion
const LEASE_TTL_SEC: i64 = 60;
```

walkeeper/src/broker.rs
```diff
@@ -0,0 +1,201 @@
...
+use zenith_utils::zid::ZTimelineId;
+use zenith_utils::{
+    lsn::Lsn,
+    zid::{ZNodeId, ZTenantTimelineId},
+};
+
+use crate::{safekeeper::Term, timeline::GlobalTimelines, SafeKeeperConf};
+
+const RETRY_INTERVAL_MSEC: u64 = 1000;
+const PUSH_INTERVAL_MSEC: u64 = 5000;
```
petuhovskiy, line 30: May be useful to move these durations to the config, to adjust precisely. Also I'm not sure if values were specially chosen already, 60s TTL seems large to me, I'd prefer 5s TTL with regular keepalives (once a second for example).

petuhovskiy (commented):
Finished reviewing this PR, looks good for initial etcd support.

control_plane/src/local_env.rs
```diff
@@ -56,6 +56,10 @@ pub struct LocalEnv {
     #[serde(default)]
     pub private_key_path: PathBuf,
 
+    // A comma separated broker (etcd) endpoints for storage nodes coordination, e.g. 'http://127.0.0.1:2379'.
+    #[serde(default)]
+    pub broker: Option<String>,
```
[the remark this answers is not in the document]
arssher: Renamed.

As for `Url`, That would require custom deser into `Vec<Url>` here, which is not that useful here to me. I converted to `Url` in safekeeper code though.

walkeeper/src/broker.rs
arssher: I thought about that, but it seems better to have such zenith installation id globally (known by console, pageserver & safekeepers). I'd do that separately.

Introduced common `ZENITH_PREFIX` though.

walkeeper/src/send_wal.rs
```diff
@@ -163,7 +163,7 @@ impl ReplicationConn {
                             let buf = Bytes::copy_from_slice(&m[9..]);
                             let reply = ZenithFeedback::parse(buf);
 
-                            trace!("ZenithFeedback is {:?}", reply);
```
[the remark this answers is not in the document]
arssher: thanks, accidently committed after debugging

test_runner/fixtures/zenith_fixtures.py
```diff
@@ -1719,6 +1736,47 @@ def get_metrics(self) -> SafekeeperMetrics:
...
+    handle: subprocess.Popen = None  # handle of running daemon
+
+    def check_status(self):
+        s = requests.Session()
+        s.mount('http://', requests.adapters.HTTPAdapter(max_retries=1))  # do not retry
+        s.get(f"http://localhost:{self.port}/health").raise_for_status()
+
+    def start(self):
+        pathlib.Path(self.datadir).mkdir(exist_ok=True)
+        etcd_full_path = shutil.which("etcd")
```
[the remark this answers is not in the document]
arssher: Yeah, added that to README and opened https://github.com/zenithdb/docker-images/pull/4

walkeeper/src/timeline.rs
```diff
@@ -410,8 +417,40 @@ impl Timeline {
...
+
+    /// Update timeline state with peer safekeeper data.
+    pub fn record_safekeeper_info(&self, sk_info: &SafekeeperInfo, _sk_id: ZNodeId) {
+        let mut shared_state = self.mutex.lock().unwrap();
+        if let Some(commit_lsn) = sk_info.commit_lsn {
+            shared_state.sk.inmem.s3_wal_lsn = max(commit_lsn, shared_state.sk.inmem.s3_wal_lsn);
+            shared_state.sk.inmem.remote_consistent_lsn =
+                max(commit_lsn, shared_state.sk.inmem.remote_consistent_lsn);
+            shared_state.sk.inmem.peer_horizon_lsn =
+                max(commit_lsn, shared_state.sk.inmem.peer_horizon_lsn);
```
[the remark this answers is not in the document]
arssher: because I'm sloppy, thanks

walkeeper/src/broker.rs
arssher: Added note on that. This code is not performance sensitive at all and we don't await while holding the locks, so there is no deadlock danger -- yes, I wouldn't bother at this stage.

walkeeper/src/broker.rs
arssher: Made them smaller, can do configurable later if we want.

walkeeper/src/timeline.rs
```diff
@@ -410,8 +417,57 @@ impl Timeline {
...
+    pub fn record_safekeeper_info(&self, sk_info: &SafekeeperInfo, _sk_id: ZNodeId) {
+        let mut shared_state = self.mutex.lock().unwrap();
+        // Note: the check is too restrictive, generally we can update local
+        // commit_lsn if our history matches (is part of) history of advanced
+        // commit_lsn provider.
+        if let (Some(commit_lsn), Some(last_log_term)) = (sk_info.commit_lsn, sk_info.last_log_term)
+        {
+            if last_log_term == shared_state.sk.get_epoch() {
+                let local_commit_lsn = min(commit_lsn, shared_state.sk.wal_store.flush_lsn());
+                shared_state.sk.inmem.commit_lsn =
```
petuhovskiy, line 453: I'd update only `global_commit_lsn` here, and waited until it propagates to `commit_lsn` by `SafeKeeper::update_commit_lsn`. Also note that we are calling `notify_wal_senders(commit_lsn)` only inside `process_msg`.

I'm also not really sure if we need to update local commit_lsn from other peers, because it updates just fine when compute is alive. Is there anything we need to do with commit_lsn when compute is not alive?

petuhovskiy (approved):

walkeeper/src/timeline.rs
arssher: > I'd update only global_commit_lsn here,

Makes sense.

> I'm also not really sure if we need to update local commit_lsn from other peers

Yes, currently we don't have to. In perspective, it would be useful as sync-safekeepers replacement though.

PR #1381, opened 2022-03-19
Use serde_with to (de)serialize ZId and Lsn to hex

SomeoneToIgnore:
Improves https://github.com/zenithdb/zenith/pull/1330 with a librarified generic approach to hex (de)serialization in pageserver http api.

Now all http responses return
* `ZTimelineId` and `ZTenantId` as hex strings: `ad50847381e248feaac9876cc71ae418 ` (string) vs `[173,80,132,115,129,226,72,254,170,201,135,108,199,26,228,24]` (array of integers)
* `Lsn` as hex strings: `FFFFFFFF/FFFFFFFF` (string) vs `18446744073709551615` (integer)

when (de)serializing both structs in the existing code. 
This is achieved with specific `serde_with::serde_as` attributes, boiling down to [`DisplayFromStr`](https://docs.rs/serde_with/latest/serde_with/struct.DisplayFromStr.html) using `Display` for serialization and `FromStr` for deserialization of the annotated types.
Both type default (de)serialization approaches stay the same (`[u8;16]` and `u64`).


LizardWizzard (approved):
Looks nice, thanks!

Thats more ergonomic than having two types and convert between them back and forth every time you need a different flavour

PR #1385, opened 2022-03-21
Release main 2022-03-21

ololobus:
Any known stoppers for releasing zenith `main` now?


ololobus:
@cicdteam github says that there some conflicts preventing merging the `main` into `release`. I see that you pushed some hotfix into `release`. It seems that it should be pushed into `main` as well, isn't it?

PR #1395, opened 2022-03-23
some follow ups after s3 integration was enabled on staging

LizardWizzard:
* do not error out when upload file list is empty
* ignore ephemeral files during sync initialization

### DIFF
--- a/pageserver/src/remote_storage.rs
+++ b/pageserver/src/remote_storage.rs
@@ -94,12 +94,13 @@ use std::{
 
 use anyhow::{bail, Context};
 use tokio::{io, sync::RwLock};
-use tracing::{error, info};
+use tracing::{debug, error, info};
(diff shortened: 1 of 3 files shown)


pageserver/src/remote_storage.rs
```diff
@@ -280,6 +283,9 @@ fn collect_timeline_files(
         if entry_path.is_file() {
             if entry_path.file_name().and_then(ffi::OsStr::to_str) == Some(METADATA_FILE_NAME) {
                 timeline_metadata_path = Some(entry_path);
+            } else if is_ephemeral_file(entry_path.file_name().unwrap().to_str().unwrap()) {
```
SomeoneToIgnore, line 286: > `.to_str().unwrap()`

NIT: could be replaced with `&entry_path.file_name().unwrap().to_string_lossy()` to have one less possible panic

SomeoneToIgnore (approved):

PR #1399, opened 2022-03-24
New storage format deploy update

cicdteam:


ololobus:
FYI, I've rebased this branch on top of the #1350 and force-pushed. So it only contains the deploy changes.

PR #1404, opened 2022-03-24
wait a bit longer for timeline availability on new pageserver

LizardWizzard:
sometimes in CI it can take more time


SomeoneToIgnore (approved):

LizardWizzard:
It seems that test does not fail anymore so the change is not needed

PR #1412, opened 2022-03-25
temporary disable s3 integration on staging until LSM storge rewrite lands

LizardWizzard:
There were some problems which were caused by our benchmarking cluster, but there is no sense in resolving them because lsm storage rewrite should not have the problem at all and it is expected to land soon

### DIFF
--- a/.circleci/ansible/deploy.yaml
+++ b/.circleci/ansible/deploy.yaml
@@ -91,19 +91,20 @@
       tags:
       - pageserver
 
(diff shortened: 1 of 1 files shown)


cicdteam:
@LizardWizzard I see you just commented part of ansible playbook. It's ok, but need manually comment/erase that config portion on staging pageserver as well.

cicdteam (approved):

PR #1413, opened 2022-03-25
Hide remote timeline index access details

SomeoneToIgnore:
A follow-up of https://github.com/zenithdb/zenith/pull/1163#discussion_r829517039


dhammika (approved):
+1 I'm not sure if `Accessor` adds any additional info, may be just call it RemoteIndex.

Issue #1414, opened 2022-03-28
Epic: Transfer zenith CI from CircleCI to GitHub Actions


SomeoneToIgnore:
Currently, we have both CI systems in use, but the latter seems to be more spread across the projects and favoured generally.
Additionally, GitHub Actions would allow us to use better build agents, speeding our Rust builds further.

GitHub actions have a set of plugins that could substitute at least part of our functionality (coverage, Docker image builds and tags, etc.), all such transitions could wait and as the first step, the same commands should be moved over to GitHub Actions.

LizardWizzard:
- [ ] set up k8s cluster for self-hosted runners @cicdteam

SergeyMelnikov:
In current setup we run default actions-runner docker image (https://github.com/actions-runner-controller/actions-runner-controller/tree/master/runner), and inside we use zimg (CircleCI based image with rust and everything else).
I'm keeping it for now to make builds in both CI's identical, but we could and should use just one docker (early prototype is in https://github.com/neondatabase/docker-images/pull/10)

SergeyMelnikov:
I use upload-artifact/download-artifact actions to share results between jobs and it is a bottleneck.
Uploading neon debug build neon to github and downloading it back takes up to 20 minutes (and eats space in github org).
We need to find another way of sharing data between jobs.

cicdteam:
> I use upload-artifact/download-artifact actions to share results between jobs and it is a bottleneck.
> Uploading neon debug build neon to github and downloading it back takes up to 20 minutes (and eats space in github org).
> We need to find another way of sharing data between jobs.

what that artifact used for ? can we just cache necessary fils/folders via S3 bucket ? Or just review content of artifact ?

SomeoneToIgnore:
Current status: merged main part of the CircleCI scripts as new GitHub job.

Before removing CircleCI, need more features reimplemented in GH Actions: 
https://github.com/neondatabase/neon/pull/1647#issuecomment-1162016373

Also, need to tune the k8s build agents' availability:
<img width="1721" alt="no_k8s_workers" src="https://user-images.githubusercontent.com/2690773/175008836-c7d75572-3aef-41d5-affc-668e42b31c9c.png">

SomeoneToIgnore:
Seems that benchmark job reliably fails now, see recent `main` runs and another example from the PR: https://github.com/neondatabase/neon/runs/7003132686?check_suite_focus=true

SomeoneToIgnore:
[...]

PR #1416, opened 2022-03-28
use cargo-hakari to manage workspace_hack crate

LizardWizzard:
workspace_hack is needed to avoid recompilation when different crates
inside the workspace depend on the same packages but with different
features being enabled. Problem occurs when you build crates separately
one by one. So this is irrelevant to our CI setup because there we build
all binaries at once, but it may be relevant for local development.

More info about the problem: https://docs.rs/cargo-hakari/latest/cargo_hakari/about/index.html

TODO:
- [ ]  add cargo hakari checks to CI, makes sense to do that after #1414


SomeoneToIgnore (approved):

funbringer (approved):
Great, that's one less thing to think about.

PR #1421, opened 2022-03-28
Bump vendor/postgres to fix commit_lsn going backwards.

arssher:


petuhovskiy (approved):
Tests are passing, so that was a redundant (and incorrect) line 🤔

PR #1422, opened 2022-03-28
Make shared_buffers large in test_pageserver_catchup.

arssher:
We intentionally write while pageserver is down, so we shouldn't query it.

Noticed by @petuhovskiy at
https://github.com/zenithdb/postgres/pull/141#issuecomment-1080261700


petuhovskiy (approved):
Looks fine for a temporal fix, because compute still may want to go to pageserver, but that should be less likely with this fix.

petuhovskiy:
P.S. check-codestyle-python complains on yapf, run `poetry run yapf -ri .`

petuhovskiy:
The issue is probably still there, because I've seen a similar failure in CI after the fix: https://app.circleci.com/pipelines/github/zenithdb/zenith/5362/workflows/ef9e41fd-9a24-4466-bbad-f5ea23ffa34e/jobs/51283/tests

Issue #1405, opened 2022-03-24
CI tests are failing with "Too long with no output" timeout


petuhovskiy:
Here's an example https://app.circleci.com/pipelines/github/zenithdb/zenith/5064/workflows/f673abce-026a-4df0-9fae-03c13bcad41c/jobs/48761

I see such failures pretty often, they're very unpleasant to debug and work with. It can become better after we switch to GitHub Actions, but I think we still need to add timeouts to python tests itself, or at least capture python tests output somehow.

PR #1423, opened 2022-03-28
Add default statement_timeout for tests

petuhovskiy:
This PR sets `statement_timeout=120s` for all tests that are using `PgProtocol.connect()`. It helps with #1405 because timeout in CI is most common while running postgres query

### DIFF
--- a/test_runner/fixtures/zenith_fixtures.py
+++ b/test_runner/fixtures/zenith_fixtures.py
@@ -257,7 +257,8 @@ class PgProtocol:
                 dbname: Optional[str] = None,
                 schema: Optional[str] = None,
                 username: Optional[str] = None,
-                password: Optional[str] = None) -> str:
+                password: Optional[str] = None,
+                statement_timeout_ms: Optional[int] = None) -> str:
         """
         Build a libpq connection string for the Postgres instance.
         """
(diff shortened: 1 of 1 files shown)


test_runner/fixtures/zenith_fixtures.py
```diff
@@ -277,29 +278,36 @@ def connstr(self,
...
+    def connect(
+        self,
+        *,
+        autocommit=True,
+        dbname: Optional[str] = None,
+        schema: Optional[str] = None,
+        username: Optional[str] = None,
+        password: Optional[str] = None,
+        # individual statement timeout in seconds, 2 minutes should be enough for our tests
+        statement_timeout: Optional[int] = 120
```
SomeoneToIgnore, line 296: NIT: we can add `_s` or _`seconds` suffix here too/

SomeoneToIgnore (approved):

test_runner/fixtures/zenith_fixtures.py
petuhovskiy: I thought of it, but `statement_timeout_s` doesn't look pretty IMO and `statement_timeout_seconds` is quite long. Decided to keep it that way because python use seconds as default time units in many places.

arssher (approved):

PR #1424, opened 2022-03-28
Enable S3 for pageserver on staging

ololobus:
Follow-up for #1417. Previously we had a problem uploading to S3
due to huge ammount of existing not yet uploaded data. Now we have a
fresh pageserver with LSM storage on staging, so we can try enabling it
once again.

I think that I mistakenly kept it commented during rebase of #1399 and we can turn it back again.

### DIFF
--- a/.circleci/ansible/deploy.yaml
+++ b/.circleci/ansible/deploy.yaml
@@ -63,20 +63,19 @@
       tags:
       - pageserver
 
-    # Temporary disabled until LSM storage rewrite lands
-    # - name: update config
-    #   when: current_version > remote_version or force_deploy
-    #   lineinfile:
-    #     path: /storage/pageserver/data/pageserver.toml
-    #     line: "{{ item }}"
-    #   loop:
-    #     - "[remote_storage]"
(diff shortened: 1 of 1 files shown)


LizardWizzard (approved):

SomeoneToIgnore:
I can see that `main` CI runs got one more failure starting from now, the error message seems related: 
https://app.circleci.com/pipelines/github/zenithdb/zenith/5157/workflows/4289458c-48e4-46a8-a325-1053620f3af7/jobs/49949

> fatal: [zenith-us-stage-ps-2]: FAILED! => {"msg": "The conditional check 'current_version > remote_version or force_deploy' failed. The error was: error while evaluating conditional (current_version > remote_version or force_deploy): 'remote_version' is undefined\n\nThe error appears to be in '/home/circleci/project/.circleci/ansible/deploy.yaml': line 66, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n    - name: update remote storage (s3) config\n      ^ here\n"}

PR #1427, opened 2022-03-28
Fix tone down compact log chatter

dhammika:

### DIFF
--- a/pageserver/src/layered_repository.rs
+++ b/pageserver/src/layered_repository.rs
@@ -1628,6 +1628,9 @@ impl LayeredTimeline {
                 };
 
                 let num_deltas = layers.count_deltas(&img_range, &(img_lsn..lsn))?;
+                if num_deltas == 0 {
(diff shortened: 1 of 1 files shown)


SomeoneToIgnore:
> Any objections to logging only partitions with deltas?

https://github.com/zenithdb/zenith/pull/1428

SomeoneToIgnore (approved):

SomeoneToIgnore:
image

WDYT, should we make the whole log `debug` instead and remove that `return`?

dhammika:
Isn't `return` to skip scanning once we've found a layer that satisfies delta threshold? No point scanning further if we already know that. 
It can be still quite chatty if the delta layers don't meet threshold criteria.

SomeoneToIgnore:
My bad, I've meant `continue`, related to this logging removal.
But looks like it's already in the works: https://github.com/zenithdb/zenith/pull/1435

PR #1428, opened 2022-03-28
Show more S3 logs and less verbove WAL logs

SomeoneToIgnore:


SomeoneToIgnore:
During https://zenith-esr3609.slack.com/archives/C033A2WE6BZ/p1648492609014039 log explorations, I've stumbled onto a few verbose places I'd decided to suppress:

image

After filtering out the previous message:

image

Also, even without both of the logs, no S3 logging was found despite it was one, so add some basic logging there.

dhammika (approved):
:+1: Assume `DEFAULT_REMOTE_STORAGE_MAX_CONCURRENT_SYNC` change is intentional.

PR #1435, opened 2022-03-29
decrease the log level to debug because it is too noisy

LizardWizzard:
~20k messages per minute on staging


pageserver/src/layered_repository.rs
```diff
@@ -1649,7 +1649,7 @@ impl LayeredTimeline {
                     continue;
                 }
 
-                info!(
```
SomeoneToIgnore, line 1652: I suggest removing the `continue` above, since it was added due to the same reason: https://github.com/zenithdb/zenith/pull/1427

SomeoneToIgnore (approved):

PR #1437, opened 2022-03-30
pair basebackup start log message with corresponding end one of the same severity

LizardWizzard:
Currently, there is only a message about basebackup start, so by looking at logs, it is not clear whether it is successfully finished or not

### DIFF
--- a/pageserver/src/page_service.rs
+++ b/pageserver/src/page_service.rs
@@ -514,6 +514,7 @@ impl PageServerHandler {
     ) -> anyhow::Result<()> {
         let span = info_span!("basebackup", timeline = %timelineid, tenant = %tenantid, lsn = field::Empty);
         let _enter = span.enter();
+        info!("starting");
 
         // check that the timeline exists
         let timeline = tenant_mgr::get_timeline_for_tenant_load(tenantid, timelineid)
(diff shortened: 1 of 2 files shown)


pageserver/src/page_service.rs
```diff
@@ -536,7 +536,7 @@ impl PageServerHandler {
             basebackup.send_tarball()?;
         }
         pgb.write_message(&BeMessage::CopyDone)?;
-        debug!("CopyDone sent!");
+        info!("basebackup CopyDone sent!");
```
SomeoneToIgnore, line 539: > basebackup

I would imagine, 

> info_span!("basebackup", timeline = %timelineid, tenant = %tenantid, lsn = field::Empty);

should already append this info?

LizardWizzard:
@SomeoneToIgnore, another log message happen here https://github.com/zenithdb/zenith/blob/main/pageserver/src/basebackup.rs#L86, and before that, we may not know the exact lsn to take basebackup from. Maybe it is better to add two messages, one at the very beginning without the lsn, and then goes this message, and a finish one after that

SomeoneToIgnore:
> one at the very beginning without the lsn, and then goes this message, and a finish one after that

Yeah, or another one into `send_tarball` (if we stick to my initial "move"  proposal), but I like your way more.
Just feels weird a bit to have a `CopyDone sent!` in the logs (even with the span data), since it does not make much sense compared to the `taking basebackup lsn={}, prev_lsn={}` one 🙂

LizardWizzard:
@SomeoneToIgnore see the updated version :)

SomeoneToIgnore (approved):

PR #1448, opened 2022-03-31
[proxy] Enable keepalives for all tcp connections

funbringer:
This is a reimplementation of @knizhnik's patch https://github.com/zenithdb/zenith/pull/1271. I suggest we merge this and see if it helps with pathological cases like this one https://github.com/zenithdb/zenith/issues/1304#issuecomment-1084807814.


LizardWizzard (approved):
LGTM

Lets deploy this and check how it works on staging

PR #1452, opened 2022-04-01
Fix control file updates in safekeeper

### LINKED ISSUE #1382: Handle updates to control file correctly in case of failures
If control file is failed to persist to disk, SafeKeeperState will be modified in memory, but not on disk. This can cause correctness problems in the future. To fix this we shouldn't update in-memory state if persist to disk has failed.

cc @arssher

petuhovskiy:
Fixes #1382


petuhovskiy:
Waiting for review to merge

arssher (approved):

PR #1460, opened 2022-04-04
feat compute_tools: check write ability handler

Daniel-ef:
Resolve: https://github.com/zenithdb/console/issues/941

### DIFF
--- a/compute_tools/src/checker.rs
+++ b/compute_tools/src/checker.rs
@@ -1,0 +1,46 @@
+use std::sync::{Arc, RwLock};
+
+use anyhow::{anyhow, Result};
+use log::error;
+use postgres::Client;
+use tokio_postgres::NoTls;
+
+use crate::zenith::ComputeState;
+
+pub fn create_writablity_check_data(client: &mut Client) -> Result<()> {
+    let query = "
+    CREATE TABLE IF NOT EXISTS health_check (
+        id serial primary key,
+        updated_at timestamptz default now()
+    );
+    INSERT INTO health_check VALUES (1, now())
+        ON CONFLICT (id) DO UPDATE
+         SET updated_at = now();";
+    let result = client.simple_query(query)?;
+    if result.len() < 2 {
+        return Err(anyhow::format_err!("executed  {} queries", result.len()));
+    }
+    Ok(())
+}
+
+pub async fn check_writability(state: &Arc<RwLock<ComputeState>>) -> Result<()> {
+    let connstr = state.read().unwrap().connstr.clone();
+    let (client, connection) = tokio_postgres::connect(&connstr, NoTls).await?;
+    if client.is_closed() {
+        return Err(anyhow!("connection to postgres closed"));
+    }
+    tokio::spawn(async move {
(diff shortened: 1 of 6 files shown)


compute_tools/src/checker.rs
```diff
@@ -0,0 +1,48 @@
...
+        }
+    });
+
+    let result = client
+        .simple_query(
+            "
+        CREATE TABLE IF NOT EXISTS health_check (
+             id serial primary key,
+            updated_at timestamptz default now()
+        );
```
ololobus, line 27: I thought about adding this (both table create and test row insert) as a step into compute startup routine, so we only did it once on startup. I don't think that `CREATE TABLE IF NOT EXISTS` is expensive, but 1) it just looks cleaner and 2) here we can do an update by id, so if for some reason we will loose table, test row, or btree got damaged we will notice that. Otherwise, we may tolerate a lot of errors.

Not 100% sure actually that what I propose is better, just some thoughts

ololobus (commented):
Otherwise, LGTM. I'd wait for zenithdb/console#963 to pass all tests first

compute_tools/src/checker.rs
Daniel-ef: @ololobus What is the best place to insert this creation?

compute_tools/src/spec.rs
```diff
@@ -244,3 +244,21 @@ pub fn handle_databases(spec: &ClusterSpec, client: &mut Client) -> Result<()> {
 
     Ok(())
 }
+
+pub fn create_writablity_check_data(client: &mut Client) -> Result<()> {
+    let query = "
+    CREATE TABLE IF NOT EXISTS health_check (
+        id serial primary key,
+        updated_at timestamptz default now()
+    );
```
Daniel-ef, line 253: Here we still should use `if not exists` because we can't distinguish creation and start as I understand.

ololobus: Yeah, it seems so

compute_tools/src/spec.rs
```diff
@@ -244,3 +244,21 @@ pub fn handle_databases(spec: &ClusterSpec, client: &mut Client) -> Result<()> {
...
+    INSERT INTO health_check VALUES (1, now())
+        ON CONFLICT (id) DO UPDATE
+         SET updated_at = now();";
+    let result = client.simple_query(query)?;
+    if result.len() != 3 {
+        return Err(anyhow::format_err!(
+            "Failed to create writability check data",
```
ololobus, line 260: Let's be consistent and use lowercase errors everywhere

compute_tools/src/spec.rs
```diff
@@ -244,3 +244,21 @@ pub fn handle_databases(spec: &ClusterSpec, client: &mut Client) -> Result<()> {
 
     Ok(())
 }
+
+pub fn create_writablity_check_data(client: &mut Client) -> Result<()> {
```
ololobus, line 248: This doesn't look as something related to the spec json processing, so I'd put it into the same checker file you created

ololobus (approved):

PR #1461, opened 2022-04-04
Small refactoring after ec3bc741653d.

arssher:
Move record_safekeeper_info inside safekeeper.rs, fix commit_lsn update, sync
control file.

### DIFF
--- a/walkeeper/src/safekeeper.rs
+++ b/walkeeper/src/safekeeper.rs
@@ -6,6 +6,7 @@ use bytes::{Buf, BufMut, Bytes, BytesMut};
 
 use postgres_ffi::xlog_utils::TimeLineID;
 use serde::{Deserialize, Serialize};
+use std::cmp::max;
 use std::cmp::min;
 use std::fmt;
 use std::io::Read;
@@ -15,6 +16,7 @@ use zenith_utils::zid::ZTenantTimelineId;
 
 use lazy_static::lazy_static;
 
+use crate::broker::SafekeeperInfo;
 use crate::control_file;
 use crate::send_wal::HotStandbyFeedback;
 use crate::wal_storage;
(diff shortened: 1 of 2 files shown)


petuhovskiy:
We should fix assert with `self.global_commit_lsn` before merging

petuhovskiy:
We should fix assert with `self.global_commit_lsn` before merging

walkeeper/src/safekeeper.rs
```diff
@@ -825,6 +829,45 @@ where
...
+            self.inmem.remote_consistent_lsn = new_remote_consistent_lsn;
+        }
+        if let Some(peer_horizon_lsn) = sk_info.peer_horizon_lsn {
+            let new_peer_horizon_lsn = max(peer_horizon_lsn, self.inmem.peer_horizon_lsn);
+            sync_control_file |= self.inmem.peer_horizon_lsn + (self.s.server.wal_seg_size as u64)
+                < new_peer_horizon_lsn;
+            self.inmem.peer_horizon_lsn = new_peer_horizon_lsn;
+        }
+        if sync_control_file {
+            self.persist_control_file()?;
```
petuhovskiy, line 867: `persist_control_file` persists only `commit_lsn` and `peer_horizon_lsn`. Should we add `remote_consistent_lsn` and `s3_wal_lsn` there?

petuhovskiy (changes requested):
I see `self.s` somewhere in code, that made me realize that you should rebase this branch because https://github.com/neondatabase/neon/commit/cdf04b6a9fb2d5d225d12a2a74fae6c6eec26da6 was committed to master.

walkeeper/src/safekeeper.rs
```diff
@@ -825,6 +829,45 @@ where
...
+            // commit_lsn if our history matches (is part of) history of advanced
+            // commit_lsn provider.
+            if last_log_term == self.get_epoch() {
+                self.global_commit_lsn = max(commit_lsn, self.global_commit_lsn);
+                self.update_commit_lsn()?;
+            }
+        }
+        if let Some(s3_wal_lsn) = sk_info.s3_wal_lsn {
+            let new_s3_wal_lsn = max(s3_wal_lsn, self.inmem.s3_wal_lsn);
+            sync_control_file |=
```
petuhovskiy, line 848: This check can always be false, because it checks only in-memory delta which is updated right after. The same goes for the next wal_seg_size check.

petuhovskiy (approved):

PR #1469, opened 2022-04-05
Improve random_writes test

bojanserafimov:
If you want to test with a 3GB database by tweaking some constants you'll hit a query timeout. I fix that by batching the inserts.


bojanserafimov:
> By query timeout, do you mean statement_timeout that was recently added in #1423? If it is, you can try just to change statement_timeout for this particular test, you need to pass parameter to `pg.connect()` for that.

Thanks, that's good to know. Though long-running queries are annoying for multiple reasons, and the CI example here is just one instance. Another example is that psycopg2 needs to be specifically configured to watch for Ctrl-C to cancel long-running queries.

PR #1476, opened 2022-04-07
Fix scripts to deploy sk4 on staging

petuhovskiy:
Create `init_safekeeper.sh` inspired by `init_pageserver.sh` and adjust ansible inventory for the new instance.

Safekeeper will be initialized after this PR will be merged to the main.

ref https://github.com/zenithdb/console/pull/970

### DIFF
--- a/.circleci/ansible/scripts/init_safekeeper.sh
+++ b/.circleci/ansible/scripts/init_safekeeper.sh
@@ -1,0 +1,30 @@
+#!/bin/sh
+
+# get instance id from meta-data service
+INSTANCE_ID=$(curl -s http://169.254.169.254/latest/meta-data/instance-id)
+
+# store fqdn hostname in var
+HOST=$(hostname -f)
+
+
+cat <<EOF | tee /tmp/payload
+{
+  "version": 1,
+  "host": "${HOST}",
+  "port": 6500,
+  "region_id": {{ console_region_id }},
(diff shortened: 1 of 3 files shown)


.circleci/ansible/scripts/init_safekeeper.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
+
+# get instance id from meta-data service
+INSTANCE_ID=$(curl -s http://169.254.169.254/latest/meta-data/instance-id)
```
ololobus, line 4: I wonder why we don't use some domain name here as well

ololobus (approved):
LGTM, but maybe @cicdteam has some comments

.circleci/ansible/scripts/init_safekeeper.sh
petuhovskiy: I guess we use instance_id as something that is fixed and unique, this way it's better than domain names.

ololobus: Yeah, but I mean this IP 169.254.169.254

petuhovskiy: It's a special AWS IP for obtaining metadata, I don't see that it has domain name https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html

PR #1477, opened 2022-04-07
allow the users to create trusted extensions: GRANT CREATE ON DATABASE

lubennikovaav:
GRANT CREATE ON DATABASE postgres TO usr ;

from PostgreSQL docs:
CREATE privilege for databases, allows new schemas and publications to be created within the database, and allows trusted extensions to be installed within the database.

To check if user has this privilege, run:
```
select datname, datacl from pg_database ;
  datname  |                                         datacl                                         
-----------+----------------------------------------------------------------------------------------
 template1 | {=c/anastasia,anastasia=CTc/anastasia}
 template0 | {=c/anastasia,anastasia=CTc/anastasia}
 postgres  | {=Tc/anastasia,anastasia=CTc/anastasia,usr=C/anastasia}
 ```
 C is for Create
 
 The only way to check if extension is trusted is to grep its *.control file.

### DIFF
--- a/compute_tools/src/bin/zenith_ctl.rs
+++ b/compute_tools/src/bin/zenith_ctl.rs
@@ -129,6 +129,7 @@ fn run_compute(state: &Arc<RwLock<ComputeState>>) -> Result<ExitStatus> {
 
     handle_roles(&read_state.spec, &mut client)?;
     handle_databases(&read_state.spec, &mut client)?;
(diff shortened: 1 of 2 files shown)


ololobus (approved):

ololobus:
It'd be nice to have an e2e test for it, though. E.g. create some trusted extension and call some func out of it

PR #1478, opened 2022-04-07
Add s3 cost estimate to tests

bojanserafimov:
I add two metrics that estimate our s3 costs in tests. Example results:

Before LSM tree:
```
test_pgbench[zenith].data_uploaded: 227 MB
test_pgbench[zenith].num_files_uploaded: 160
```

After LSM tree:
```
test_pgbench[zenith].data_uploaded: 73 MB
test_pgbench[zenith].num_files_uploaded: 2
```

The numbers will change as we play with layer sizes, image heuristics, compression, etc. I assume it will be nice to have this test handy in the future.


pageserver/src/layered_repository.rs
```diff
@@ -1523,6 +1538,15 @@ impl LayeredTimeline {
                 &metadata,
                 false,
             )?;
+
+            // Update upload metrics
+            for path in vec![new_delta_path.clone()] {
```
SomeoneToIgnore, line 1543: This all should go below `self.upload_layers.load(atomic::Ordering::Relaxed)`, the upload is not enabled by default and not happening.

Also, the whole `vec![new_delta_path.clone()]` seems to be redundant, since `File` is capable of opening `&Path` too?

SomeoneToIgnore (commented):
Nice idea.

IMO, with the current way of reporting the new values, we should either add "estimate" to the metric labels and descriptions, or move the reporting below the `if`.

pageserver/src/layered_repository.rs
bojanserafimov: Ah sorry the vec! stuff slipped. I'll clean it up.

I want a way to count without actually uploading, which is great for tests. I guess there are better ways to write that code to make sure we don't forget to count. I'll see if there's an easy fix.

SomeoneToIgnore: Another way to achieve this would be adding a tag to actually display, if this data was sent to the remote storage and add more logic to determine that tag, something like: 

```
static ref BYTES_SCHEDULED_FOR_UPLOAD: zenith_metrics::IntCounterVec =
    zenith_metrics::register_int_counter_vec!(
        "pageserver_bytes_scheduled_for_upload",
        "Total bytes estimated to be uploaded to the cloud storage, tagged by an actual upload event [uploaded|skipped]",
        &["uploaded", "skipped"]
    )
    .expect("failed to define a bytes scheduled for upload metric");
```

bojanserafimov: I dialed back the language so that the names don't suggest files were actually uploaded. I want to avoid adding test-only metrics to prod.

SomeoneToIgnore (approved):
I think we would eventually need those tags anyway, but this is a good start.

bojanserafimov:
Is there a way to restart the `zenith-remote-ci` test without pushing? It seems more stuck than pending.

pageserver/src/layered_repository.rs
```diff
@@ -1523,6 +1539,12 @@ impl LayeredTimeline {
                 &metadata,
                 false,
             )?;
+
+            // Update upload metrics
+            let new_delta_size = File::open(&new_delta_path)?.metadata()?.len();
```
LizardWizzard, line 1544: There is no need to open a file in order to get a metadata for it. `.metadata` is available on [PathBuf](https://doc.rust-lang.org/stable/std/path/struct.PathBuf.html#method.metadata)

LizardWizzard (approved):
Title become a bit misleading but I like the patch :)

To measure true s3 metrics we can instrument RemoteStorage impl directly. We can even try to convert bytes/operation counts to $

bojanserafimov:
@stepashka FYI here's the new metric https://observer.zenith.tech/d/fWLvMUJ7z/annotations-experiments?orgId=1&from=now-2d&to=now&viewPanel=16

PR #1482, opened 2022-04-08
Zentih user and database management RFC

lubennikovaav:
In this RFC I summed up our understanding of CREATE USER and CREATE DATABASE management.
Please, review and link relevant issues.

### DIFF
--- a/docs/rfcs/016-user-and-database-management.md
+++ b/docs/rfcs/016-user-and-database-management.md
@@ -3,7 +3,8 @@
 ### CREATE USER
 
 CREATE USER is only allowed from console, 
-because it requires generation of the access token and other complex logic.
+because console is involved in the auth process, when users connect via proxy.
+It's much handier to have a single source of truth.
 
 Client has to chose which databases will be accessible by the new user.
 Along with database access, we explicitly grant CREATE ON DATABASE privilege to the user.
@@ -14,8 +15,9 @@ All new users are granted with CREATEDB privilege by default.
 `CREATE ROLE .. WITH CREATEDB`
 Later we may add an option to manage it.
 
-*Q: Are we going to allow user to create new users and NOLIGIN roles to manage privileges?
-*`CREATE ROLE .. WITH CREATEROLE`
+For now we do not allow clients to create new users and NOLIGIN roles to manage privileges.
+i.e. `CREATE ROLE .. WITH CREATEROLE`
+Let's postpone this until we got N feature requests about more complex roles management.
 
 
 ### CREATE DATABASE
@@ -23,7 +25,7 @@ Later we may add an option to manage it.
 Client can CREATE DATABASE via SQL as well as via console.
 
 When database is created from SQL (not by superuser)
-we intercept this action using DDL hook and send it to the console to synchronize the information about databases.
+we intercept this action using DDL hook and send request to the console to synchronize the information about databases.
 
 ### CREATE EXTENSION
 
@@ -39,7 +41,7 @@ Nothing special. PostgreSQL behavior.
 ### DROP DATABASE
 
 From PostgreSQL docs:
-The right to drop an object, or to alter its definition in any way, is not treated as a grantable privilege; it is inherent in the owner, and cannot be granted or revoked. (However, a similar effect can be obtained by granting or revoking membership in the role that owns the object) 
+> The right to drop an object, or to alter its definition in any way, is not treated as a grantable privilege; it is inherent in the owner, and cannot be granted or revoked. (However, a similar effect can be obtained by granting or revoking membership in the role that owns the object) 
 
 Note that ownership can be changed using `REASSIGN OWNED BY .. TO ..` 
 
--- a/docs/rfcs/016-user-and-database-management.md
+++ b/docs/rfcs/016-user-and-database-management.md
@@ -1,0 +1,68 @@
+## Zentih user and database management RFC
+
+### CREATE USER
+
+CREATE USER is only allowed from console, 
+because it requires generation of the access token and other complex logic.
+
+Client has to chose which databases will be accessible by the new user.
+Along with database access, we explicitly grant CREATE ON DATABASE privilege to the user.
+`GRANT CREATE ON DATABASE .. TO .. ;`
+to let the user to create extensions.
+
+All new users are granted with CREATEDB privilege by default.
+`CREATE ROLE .. WITH CREATEDB`
+Later we may add an option to manage it.
+
+*Q: Are we going to allow user to create new users and NOLIGIN roles to manage privileges?
+*`CREATE ROLE .. WITH CREATEROLE`
+
+
+### CREATE DATABASE
+
+Client can CREATE DATABASE via SQL as well as via console.
+
+When database is created from SQL (not by superuser)
+we intercept this action using DDL hook and send it to the console to synchronize the information about databases.
+
+### CREATE EXTENSION
+
+To create trusted extensions in the database, user must have an access privileges on the database.
+
+We explicitly grant CREATE ON DATABASE privilege to the new user along with database access.
+If user creates a database, he becomes the owner and thus has CREATE ON DATABASE privilege by default.
+
+### DROP EXTENSION
+
+Nothing special. PostgreSQL behavior.
+
+### DROP DATABASE
+
+From PostgreSQL docs:
+The right to drop an object, or to alter its definition in any way, is not treated as a grantable privilege; it is inherent in the owner, and cannot be granted or revoked. (However, a similar effect can be obtained by granting or revoking membership in the role that owns the object) 
+
+Note that ownership can be changed using `REASSIGN OWNED BY .. TO ..` 
+
+### DROP USER
+
+DROP USER is only allowed from console.
+
+In case some objects depend on the user, PostgreSQL throws a detailed error.
+
+- ERROR:  role "p" cannot be dropped because some objects depend on it
+DETAIL:  owner of database pdb
+1 object in database pdb
+
+### Competitiors
+
+Competing databases have more or less standart approach to user management.
+There is an admin/master user that can manage other roles and databases.
+
+https://docs.microsoft.com/en-us/azure/postgresql/howto-create-users
+the server admin user is granted these privileges: 
+LOGIN, NOSUPERUSER, INHERIT, CREATEDB, CREATEROLE, REPLICATION
+
+https://aws.amazon.com/blogs/database/managing-postgresql-users-and-roles/
+The master user that is created during Amazon RDS and Aurora PostgreSQL instance creation should be used only for database administration tasks like creating other users, roles, and databases. 
+
+https://cloud.google.com/sql/docs/postgres/users


docs/rfcs/016-user-and-database-management.md
```diff
@@ -0,0 +1,68 @@
+## Zentih user and database management RFC
+
+### CREATE USER
+
+CREATE USER is only allowed from console, 
+because it requires generation of the access token and other complex logic.
```
ololobus, line 6: I'd say more exact. Because console is involved in the auth process, when users connect via proxy. It's much handier to have a single source of truth.

docs/rfcs/016-user-and-database-management.md
```diff
@@ -0,0 +1,68 @@
...
+Along with database access, we explicitly grant CREATE ON DATABASE privilege to the user.
+`GRANT CREATE ON DATABASE .. TO .. ;`
+to let the user to create extensions.
+
+All new users are granted with CREATEDB privilege by default.
+`CREATE ROLE .. WITH CREATEDB`
+Later we may add an option to manage it.
+
+*Q: Are we going to allow user to create new users and NOLIGIN roles to manage privileges?
+*`CREATE ROLE .. WITH CREATEROLE`
```
ololobus, line 18: I think no, at least for now. NOLOGIN roles are not involved in the auth process, but this may complicate things. So I'd postpone this until we got N feature requests about more complex roles management

docs/rfcs/016-user-and-database-management.md
```diff
@@ -0,0 +1,68 @@
...
+
+
+### CREATE DATABASE
+
+Client can CREATE DATABASE via SQL as well as via console.
+
+When database is created from SQL (not by superuser)
+we intercept this action using DDL hook and send it to the console to synchronize the information about databases.
```
ololobus, line 26: NB: we don't send the exact data about created database. We only send some kind of a signal "I'm pretty-cluster-123, please refresh my databases info"

docs/rfcs/016-user-and-database-management.md
```diff
@@ -0,0 +1,68 @@
...
+
+### CREATE EXTENSION
+
+To create trusted extensions in the database, user must have an access privileges on the database.
+
+We explicitly grant CREATE ON DATABASE privilege to the new user along with database access.
+If user creates a database, he becomes the owner and thus has CREATE ON DATABASE privilege by default.
```
ololobus, line 33: Is it really true? We already set default user as an owner of default database `main`, but it cannot create extensions

docs/rfcs/016-user-and-database-management.md
```diff
@@ -0,0 +1,68 @@
...
+
+### DROP EXTENSION
+
+Nothing special. PostgreSQL behavior.
+
+### DROP DATABASE
+
+From PostgreSQL docs:
+The right to drop an object, or to alter its definition in any way, is not treated as a grantable privilege; it is inherent in the owner, and cannot be granted or revoked. (However, a similar effect can be obtained by granting or revoking membership in the role that owns the object)
```
ololobus, line 42: I'd use quote syntax `> smth` here :)

docs/rfcs/016-user-and-database-management.md
```diff
@@ -0,0 +1,68 @@
...
+
+### DROP USER
+
+DROP USER is only allowed from console.
+
+In case some objects depend on the user, PostgreSQL throws a detailed error.
+
+- ERROR:  role "p" cannot be dropped because some objects depend on it
+DETAIL:  owner of database pdb
+1 object in database pdb
```
ololobus, line 54: So the most simple rule now will be that with role deletion we just do `DROP OWNED` **and** notify user in UI that all related objects will be deleted as well. What do you think?

docs/rfcs/016-user-and-database-management.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Competitiors
+
+Competing databases have more or less standart approach to user management.
```
ololobus, line 58: Have you tried how they actually behave? Are there a list of dbs / users in UI? What happens when you create user in shell? Is it immediately updated in UI?

I've played a bit with CockroachDB serverless cloud. They are weird :)

1) You can create both dbs and roles via shell and UI
2) If you create a role via psql you can +- immediately connect using it
3) **But** it never (in some reasonable time) appears in the UI, so it seems like never synced. If you try to create role with the same name in UI it shows "already exists" error, but it's still not listed
4) Yet, dbs created via psql are listed in UI. Considering the load time it refreshes them rather frequently

I'm going to try Planetscale as well

ololobus (commented):
Now, I'm looking on it and starting thinking about why not to allow creating roles as well. Yes, it will bring some delay / sort of eventual consistency. I.e. roles are not immediately synced with console, so could be a small gap when you cannot auth with newly created role, but I don't think it's critical.

At the same time we can allow both role and dbs creation via UI too. Even if there will be a collision (you previously created a role with the same name in UI) at worst we just change role password OR change db owner, because console operation will have a higher priority.

What do you think? @lubennikovaav @kelvich

docs/rfcs/016-user-and-database-management.md
lubennikovaav: It will become true after https://github.com/zenithdb/zenith/pull/1477

docs/rfcs/016-user-and-database-management.md
lubennikovaav: I wouldn't drop any client's data unless it was an explicit request.
Let's show the PotsgreSQL error and make clients do it by themselves.

docs/rfcs/016-user-and-database-management.md
ololobus: Hm, Planetscale seems like a completely different thing %) First of all it's MySQL (yeah, I'm a slowpoke). Second, they don't allow any roles management (as far as I discovered) with a very strange error:
```sql
mysql> create user 'test';
ERROR 1105 (HY000): syntax error at position 12 near 'user'
mysql> create table t(i int);
ERROR 1105 (HY000): direct DDL is disabled
mysql> create role 'test';
ERROR 1105 (HY000): syntax error at position 12 near 'role'
```
According to the docs syntax is correct
https://dev.mysql.com/doc/refman/8.0/en/create-role.html
https://dev.mysql.com/doc/refman/8.0/en/create-user.html

ololobus:
> Could you please give an example of such situation?

You create a role 'test' with password 'pwd1' via psql. It's not yet synced with console. You go to the console and create role 'test' with password 'pwd2'. It creates and operation which goes to the compute and tries to create role 'test' with 'pwd2'. It sees that role exists, but password is different from 'pwd2'. Current logic implies that if password is different, then we change it. That's it. It's not kind of a higher priority maybe, but the last wins

> I think if the role/db exists, we should throw an error. why should we change anything?

As I mentioned above, all of these is implemented via operations, which are heavily async and retried infinitely now. Sending a meaningful feedback to console is cumbersome

docs/rfcs/016-user-and-database-management.md
lubennikovaav: YandexCloud PostgreSQL only allows to create users and databases via their Console (CLI).
There are two "admin" roles but they have very restricted privileges too: https://cloud.yandex.ru/docs/managed-postgresql/concepts/roles

To install extensions, the user should choose them on the Database page in Console.

ololobus:
> I see. Is there any reason it was designed like that? Why don't we check Postgres response?

Because there could be thousands of simultaneous operations and it doesn't look as a good idea to have a blocking waiting on each of them. So the idea is that we form and validate some spec and pass it to k8s, which is a durable orchestration system by design. After that we just wait until compute became ready, as k8s and operations redo job will ensure that any operations will be retried eventually.

And I don't think that this part of design is bad, as it should scale pretty well. Yes, having an asynchronous back channel to the console, so compute was able to report errors would be nice. I think it's possible, but a bit complex thing :) And it won't be an immediate feedback to the user

docs/rfcs/016-user-and-database-management.md
```diff
@@ -0,0 +1,70 @@
...
+i.e. `CREATE ROLE .. WITH CREATEROLE`
+Let's postpone this until we got N feature requests about more complex roles management.
+
+
+### CREATE DATABASE
+
+Client can CREATE DATABASE via SQL as well as via console.
+
+When database is created from SQL (not by superuser)
+we intercept this action using DDL hook and send request to the console to synchronize the information about databases.
```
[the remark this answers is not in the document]
ololobus: We have a UI to query the compute, and you can select some particular user / db pair from existing ones there. It's just much better UX, when you don't need to type in user and db names, and can use GUI selector instead

docs/rfcs/016-user-and-database-management.md
```diff
@@ -0,0 +1,70 @@
+## Zentih user and database management RFC
+
+### CREATE USER
+
+CREATE USER is only allowed from console, 
+because console is involved in the auth process, when users connect via proxy.
```
MMeent, line 6: I think our proxy should not remain the main point of authentication for clusters. Once a cluster has started, the connections should go directly to the cluster whenever possible to reduce latency and overhead.

Console should not be involved in the authentication process. There is 0 reason for that to happen; it only complicates things: Why would we even _want_ to do a bespoke user authentication for PostgreSQL -- it only adds specific behaviors that clients would need to be able to handle and that we would need to document and maintain.

It would be amazing to keep Proxy as _only_ a transparent proxy for those connections that support it, and it would save a lot of work on all sides.

docs/rfcs/016-user-and-database-management.md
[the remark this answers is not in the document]
MMeent: I, too, do not understand why Console needs to manage databases.

docs/rfcs/016-user-and-database-management.md
MMeent: > Planetscale

That's not PostgreSQL.

> YandexCloud

Is that competing? Either way, if we're going to provide "PostgreSQL as a service", I don't expect some bastardized variant that requires a non-PostgreSQL API for doing tasks generally done through the PostgreSQL client, but a fully featured variant that provides all possible PostgreSQL features through the PostgreSQL connection.

docs/rfcs/016-user-and-database-management.md
ololobus: > Console should not be involved in the authentication process. There is 0 reason for that to happen

There is at least 1 reason for that: if we want 'serverless', i.e. spin-up computes when they are needed and tear down them when they are not, and 'needed' means someone wants to connect. Then we have to accept this connection attempts, wake up compute and route traffic to it

MMeent: I do not consider "waking up compute nodes" part of the authentication process but part of setting up the connection (a small but significant difference). Console does not need to know what authentication details are being used to connect to compute node Y; it only needs to know that compute node Y needs to wake up.

docs/rfcs/016-user-and-database-management.md
ololobus: > That's not PostgreSQL

Do I write somewhere that it **is** PostgreSQL? I don't see any reason to don't try some other relational cloud DBaaS.

docs/rfcs/016-user-and-database-management.md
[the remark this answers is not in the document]
ololobus: Ain't that an answer above? Manage --- likely not, at least not a single entry point for creating them. Know about them --- definitely yes, because of the reason above.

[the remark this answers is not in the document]
MMeent: Sorry, my tab hadn't yet synchronized your reply when I opened the issue, and the tab also didn't receive your reply until I sent my comment.

Either way, can't we query the catalogs of the clusters directly? That might be a bad idea, but at the same time, it shouldn't be that hard -- the database's OID is part of each segment, and thus we should be able to extract that information quite easily without specifically involving the console. Console might need to call into PageServer for this information, but that's already part of its job.

docs/rfcs/016-user-and-database-management.md
ololobus: > Console does not need to know what authentication details are being used to connect to compute node Y; it only needs to know that compute node Y needs to wake up.

Then I can spam connection attempts to your cluster, prevent it from being suspended and burn your money. THat's just an example.

And I'm not sure about technical part as well. You propose to don't do auth on proxy or console, then it won't behave like MITM. But when you accept a libpq connection, you basically start the auth flow, so I'm not sure you can say 'wait a second, I've just woken up your Postgres, can you please start all the same auth sequence, but with another server?'

Or you propose that our proxy shouldn't be a libpq proxy? TCP?

MMeent: > Then I can spam connection attempts to your cluster, prevent it from being suspended and burn your money.

Assuming there is a public endpoint to your database and your database access rules allow access to your cluster, then yes.

> Or you propose that our proxy shouldn't be a libpq proxy? TCP?

Yep, basically not much more than a libpq-aware TCP or TLS proxy. I.e., when we implement SNI like "psql -h cluster-name-123456.dbs.zenith.tech", it would just wake up cluster-name-123456 and forward the connection, IP ranges permitting.

Maybe we could add some extra goodies for authentication, such as not waking up the whole PG cluster but only the authentication part, but that's the extent of what I think proxy needs. And for those authentication goodies, there's no real need for Console to authenticate the client or maintain and synchronize a database of users for PostgreSQL Cluster Y.

kelvich: > Maybe we could add some extra goodies for authentication, such as not waking up the whole PG cluster but only the authentication part, but that's the extent of what I think proxy needs.

And exactly for that, you need to decipher traffic and have auth info somewhere not in the destination postgres. Isn't it?

MMeent: ... only if the target cluster is currently offline, in which case you'd already need to contact our control plane. But in the happy case (cluster is running), we shouldn't need to contact Console IMO.

kelvich: > ... only if the target cluster is currently offline, in which case you'd already need to contact our control plane. But in the happy case (cluster is running), we shouldn't need to contact Console IMO.

And this means that if the cluster is ever can be offline we need to always have a synchronous copy of roles info in the console. If the cluster is online we can go directly to the postgres, but that is a tradeoff between 'guard postgres from exhausting max_connection with bruteforce' vs 'have a smaller connection latency'.

antonyc (commented):
The problem with leaving proxy as the primary source of authentication is that when we have an outage and the proxy is down the clients cannot establish new connections to running clusters.

docs/rfcs/016-user-and-database-management.md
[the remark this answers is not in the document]
ololobus: > If we use the DDL hook here then what happens if one creates a database during a console outage? Is it possible for DDL not to fail the database creation during console outage and silently finish?
> And we should probably have a background routine scanning databases anyway, because the console may be offline during creation.

Yes, that was my understanding to. We don't do any blocking operations in the DDL hook. It's just an optional notify to speed up syncing. Console should anyway scan / sync scheme from time to time. E.g. we can do it before suspend and/or during availability checks for active clusters (not implemented yet).

docs/rfcs/016-user-and-database-management.md
ololobus: It could be, but we can minimize any possible downtime later by separating HTTP API service from the current monolithic console. After that it will be a stateless web service, which should be rather durable and less error-prone considering all the k8s automation for high availability

PR #1501, opened 2022-04-13
reenable s3 on staging pagservers by default

LizardWizzard:
After deadlock fix in https://github.com/neondatabase/neon/pull/1496 s3
seems to work normally. There is one more discovered issue but it is not
a blocker so can be fixed separately.


SomeoneToIgnore (approved):
Now we need to remember clearing S3 on any breaking changes for local and remote storage formats.

ololobus (approved):

PR #1504, opened 2022-04-14
Per tenant config + pitr

lubennikovaav:
New version of @knizhnik PR #1332 

This PR implements per-tenant config and also adds new `pitr_interval` parameter.
Maybe it's worth to split pitr part into a separate commit, as there were some questions about the proper implementation of pitr.

TenantConf is still stored as a part of the global PageserverConf, but one can also pass per-tenant conf parameters when tenant is created or using newAPI or  `zenith tenant config` CLI command.
This custom TenantConfOpt is stored in LayeredRepository struct.  
To get the parameter value, check TenantConfOpt, if it is not set, use value from global config.

When updated, tenant conf is saved to the tenant's directory, to recover it on pageserver restart.

To support tenant migraiton, this config must be stored in the console along with other tenant information. @ololobus, FYI

### DIFF
--- a/control_plane/src/storage.rs
+++ b/control_plane/src/storage.rs
@@ -1,3 +1,4 @@
+use std::collections::HashMap;
 use std::io::Write;
 use std::net::TcpStream;
 use std::path::PathBuf;
@@ -9,7 +10,7 @@ use anyhow::{bail, Context};
 use nix::errno::Errno;
 use nix::sys::signal::{kill, Signal};
 use nix::unistd::Pid;
-use pageserver::http::models::{TenantCreateRequest, TimelineCreateRequest};
+use pageserver::http::models::{TenantConfigRequest, TenantCreateRequest, TimelineCreateRequest};
 use pageserver::timelines::TimelineInfo;
 use postgres::{Config, NoTls};
 use reqwest::blocking::{Client, RequestBuilder, Response};
@@ -344,10 +345,32 @@ impl PageServerNode {
     pub fn tenant_create(
         &self,
         new_tenant_id: Option<ZTenantId>,
+        settings: HashMap<&str, &str>,
     ) -> anyhow::Result<Option<ZTenantId>> {
         let tenant_id_string = self
             .http_request(Method::POST, format!("{}/tenant", self.http_base_url))
-            .json(&TenantCreateRequest { new_tenant_id })
+            .json(&TenantCreateRequest {
+                new_tenant_id,
+                checkpoint_distance: settings
+                    .get("checkpoint_distance")
+                    .map(|x| x.parse::<u64>())
+                    .transpose()?,
+                compaction_target_size: settings
+                    .get("compaction_target_size")
+                    .map(|x| x.parse::<u64>())
+                    .transpose()?,
+                compaction_period: settings.get("compaction_period").map(|x| x.to_string()),
+                compaction_threshold: settings
+                    .get("compaction_threshold")
+                    .map(|x| x.parse::<usize>())
+                    .transpose()?,
+                gc_horizon: settings
+                    .get("gc_horizon")
+                    .map(|x| x.parse::<u64>())
+                    .transpose()?,
+                gc_period: settings.get("gc_period").map(|x| x.to_string()),
+                pitr_interval: settings.get("pitr_interval").map(|x| x.to_string()),
+            })
             .send()?
             .error_from_body()?
             .json::<Option<String>>()?;
@@ -362,6 +385,32 @@ impl PageServerNode {
                 })
             })
             .transpose()
+    }
+
+    pub fn tenant_config(&self, tenant_id: ZTenantId, settings: HashMap<&str, &str>) -> Result<()> {
+        self.http_request(Method::PUT, format!("{}/tenant/config", self.http_base_url))
+            .json(&TenantConfigRequest {
+                tenant_id,
+                checkpoint_distance: settings
+                    .get("checkpoint_distance")
+                    .map(|x| x.parse::<u64>().unwrap()),
+                compaction_target_size: settings
+                    .get("compaction_target_size")
+                    .map(|x| x.parse::<u64>().unwrap()),
+                compaction_period: settings.get("compaction_period").map(|x| x.to_string()),
+                compaction_threshold: settings
+                    .get("compaction_threshold")
+                    .map(|x| x.parse::<usize>().unwrap()),
+                gc_horizon: settings
+                    .get("gc_horizon")
+                    .map(|x| x.parse::<u64>().unwrap()),
+                gc_period: settings.get("gc_period").map(|x| x.to_string()),
+                pitr_interval: settings.get("pitr_interval").map(|x| x.to_string()),
+            })
+            .send()?
+            .error_from_body()?;
+
+        Ok(())
     }
 
     pub fn timeline_list(&self, tenant_id: &ZTenantId) -> anyhow::Result<Vec<TimelineInfo>> {
--- a/pageserver/src/bin/pageserver.rs
+++ b/pageserver/src/bin/pageserver.rs
@@ -246,11 +246,12 @@ fn start_pageserver(conf: &'static PageServerConf, daemonize: bool) -> Result<()
 
     for (tenant_id, local_timeline_init_statuses) in local_timeline_init_statuses {
         // initialize local tenant
-        let repo = tenant_mgr::load_local_repo(conf, tenant_id, &remote_index);
+        let repo = tenant_mgr::load_local_repo(conf, tenant_id, &remote_index)
+            .with_context(|| format!("Failed to load repo for tenant {}", tenant_id))?;
         for (timeline_id, init_status) in local_timeline_init_statuses {
             match init_status {
                 remote_storage::LocalTimelineInitStatus::LocallyComplete => {
-                    debug!("timeline {} for tenant {} is locally complete, registering it in repository", tenant_id, timeline_id);
+                    debug!("timeline {} for tenant {} is locally complete, registering it in repository", timeline_id, tenant_id);
                     // Lets fail here loudly to be on the safe side.
                     // XXX: It may be a better api to actually distinguish between repository startup
                     //   and processing of newly downloaded timelines.
--- a/pageserver/src/config.rs
+++ b/pageserver/src/config.rs
@@ -5,6 +5,12 @@
 //! See also `settings.md` for better description on every parameter.
 
 use anyhow::{bail, ensure, Context, Result};
+use std::convert::TryInto;
+use std::env;
+use std::num::{NonZeroU32, NonZeroUsize};
+use std::path::{Path, PathBuf};
+use std::str::FromStr;
+use std::time::Duration;
 use toml_edit;
 use toml_edit::{Document, Item};
 use utils::{
@@ -12,37 +18,17 @@ use utils::{
     zid::{ZNodeId, ZTenantId, ZTimelineId},
 };
 
-use std::convert::TryInto;
-use std::env;
-use std::num::{NonZeroU32, NonZeroUsize};
-use std::path::{Path, PathBuf};
-use std::str::FromStr;
-use std::time::Duration;
-
 use crate::layered_repository::TIMELINES_SEGMENT_NAME;
+use crate::tenant_config::{TenantConf, TenantConfOpt};
 
 pub mod defaults {
+    use crate::tenant_config::defaults::*;
     use const_format::formatcp;
 
     pub const DEFAULT_PG_LISTEN_PORT: u16 = 64000;
     pub const DEFAULT_PG_LISTEN_ADDR: &str = formatcp!("127.0.0.1:{DEFAULT_PG_LISTEN_PORT}");
     pub const DEFAULT_HTTP_LISTEN_PORT: u16 = 9898;
     pub const DEFAULT_HTTP_LISTEN_ADDR: &str = formatcp!("127.0.0.1:{DEFAULT_HTTP_LISTEN_PORT}");
-
-    // FIXME: This current value is very low. I would imagine something like 1 GB or 10 GB
-    // would be more appropriate. But a low value forces the code to be exercised more,
-    // which is good for now to trigger bugs.
-    // This parameter actually determines L0 layer file size.
-    pub const DEFAULT_CHECKPOINT_DISTANCE: u64 = 256 * 1024 * 1024;
-
-    // Target file size, when creating image and delta layers.
-    // This parameter determines L1 layer file size.
-    pub const DEFAULT_COMPACTION_TARGET_SIZE: u64 = 128 * 1024 * 1024;
-    pub const DEFAULT_COMPACTION_PERIOD: &str = "1 s";
-    pub const DEFAULT_COMPACTION_THRESHOLD: usize = 10;
-
-    pub const DEFAULT_GC_HORIZON: u64 = 64 * 1024 * 1024;
-    pub const DEFAULT_GC_PERIOD: &str = "100 s";
 
     pub const DEFAULT_WAIT_LSN_TIMEOUT: &str = "60 s";
     pub const DEFAULT_WAL_REDO_TIMEOUT: &str = "60 s";
@@ -64,14 +50,6 @@ pub mod defaults {
 #listen_pg_addr = '{DEFAULT_PG_LISTEN_ADDR}'
 #listen_http_addr = '{DEFAULT_HTTP_LISTEN_ADDR}'
 
-#checkpoint_distance = {DEFAULT_CHECKPOINT_DISTANCE} # in bytes
-#compaction_target_size = {DEFAULT_COMPACTION_TARGET_SIZE} # in bytes
-#compaction_period = '{DEFAULT_COMPACTION_PERIOD}'
-#compaction_threshold = '{DEFAULT_COMPACTION_THRESHOLD}'
-
-#gc_period = '{DEFAULT_GC_PERIOD}'
-#gc_horizon = {DEFAULT_GC_HORIZON}
-
 #wait_lsn_timeout = '{DEFAULT_WAIT_LSN_TIMEOUT}'
 #wal_redo_timeout = '{DEFAULT_WAL_REDO_TIMEOUT}'
 
@@ -79,6 +57,16 @@ pub mod defaults {
 
 # initial superuser role name to use when creating a new tenant
 #initial_superuser_name = '{DEFAULT_SUPERUSER}'
+
+# [tenant_config]
+#checkpoint_distance = {DEFAULT_CHECKPOINT_DISTANCE} # in bytes
+#compaction_target_size = {DEFAULT_COMPACTION_TARGET_SIZE} # in bytes
+#compaction_period = '{DEFAULT_COMPACTION_PERIOD}'
+#compaction_threshold = '{DEFAULT_COMPACTION_THRESHOLD}'
+
+#gc_period = '{DEFAULT_GC_PERIOD}'
+#gc_horizon = {DEFAULT_GC_HORIZON}
+#pitr_interval = '{DEFAULT_PITR_INTERVAL}'
 
 # [remote_storage]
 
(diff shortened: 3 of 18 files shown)


ololobus:
@alubennikova does pageserver have an API endpoint to get / put this per-tenant config? Without it it will will be impossible to mange it from the console, we don't use ssh anymore

Also I wonder should the PITR config be per-timeline instead? It seems to be dependent on amount of WAL we keep and it's stored per-timeline, isn't it?

knizhnik:
> @alubennikova does pageserver have an API endpoint to get / put this per-tenant config? Without it it will be impossible to mange it from the console, we don't use ssh anymore
> 

Right now tenant configuration is specified by console when tenant is created.
It is possible to add API to change it on the fly but it has not yet be done.
Also in this  case we have to propagate this changes to all active threads (GC, compactor,...)

> Also I wonder should the PITR config be per-timeline instead? It seems to be dependent on amount of WAL we keep and it's stored per-timeline, isn't it?

No, it is stored per tenant.
IMHO there is no sense to have it per timeline, but I may be wrong.
Why do we need per-tenant config?
Because we expect that we will have have some VIP customers for we we can adjust page server parameters.
But I do not see reasons to do it per timeline.
Also please notice that GC and compactor threads are handling all tenants timeline.

lubennikovaav:
> > @alubennikova does pageserver have an API endpoint to get / put this per-tenant config? Without it it will be impossible to mange it from the console, we don't use ssh anymore
> 
> Right now tenant configuration is specified by console when tenant is created. It is possible to add API to change it on the fly but it has not yet be done.

@ololobus 
I've implemented put API: `put("/v1/tenant", tenant_config_handler)`

Now I think that is probably should be
`put("/v1/tenant/:tenant_id/config", tenant_config_handler)
get("/v1/tenant/:tenant_id/config", tenant_config_handler)`

Please, review `openapi_spec.yml` and `routes.rs` changes

>  Also in this case we have to propagate this changes to all active threads (GC, compactor,...)
@knizhnik 
timeline's TenantConf is stored as an Arc to the repo's TenantConf. So when it is changed, all threads should see the change.
If I wrote it right, it should work) Please review the code.

ololobus (commented):
> Now I think that is probably should be
>
> put("/v1/tenant/:tenant_id/config", tenant_config_handler)
> get("/v1/tenant/:tenant_id/config", tenant_config_get_handler)

I think that how you did is fine. IMO, it's a bit inconvenient to pass key (tenant_id) inside body instead of path argument, but I don't see any problems with it.

About get, I don't think we really need it now. It seems that we only need to pass some default config on creation and maybe update it later. So post + put should be enough.

knizhnik (commented):
Originally TenantConf was copied and requires no synchronization on access.
Now it is changed to Arc<RwLock<...>>.
It allows to update configuration online, without restarting threads.
But there are some drawbacks:
1. Synchronization overhead
2. If parameter value is storedin some local variable, then changing configuration will nottake effect
3. Possibility of hangout and even deadlocks.

So may be online update of configuration is really important.
But in this case this three items has to be carefully checked.
I have not noticed places where config is accessed so frequently that is can become a bottleneck. Also I have not noticed placed where config value is cached in some local variable.
But concerning 3, there are placed like
```
std::thread::sleep(tenant_conf.compaction_period);
```
Please notice that here we sleep with RwLock locked. Ansd any attempt to update confiuration will be blocked on this lock.

lubennikovaav:
> But concerning 3, there are placed like
> 
> ```
> std::thread::sleep(tenant_conf.compaction_period);
> ```
> 
> Please notice that here we sleep with RwLock locked. Ansd any attempt to update confiuration will be blocked on this lock.

I don't think, that we sleep with RwLock, it is released as soon, as we exit `get_tenant_conf()`.
I've tested configuration update and it works fine (set compaction_period=100, update it to some smaller value while the thread sleeps, check the config file - the new value is there).
The change itself will be applied on the next wake of the loop. I think it is expected behavior.

lubennikovaav:
@LizardWizzard, @SomeoneToIgnore 
please, take a look at this PR. I need some help to connect it with tenant relocation.
AFAIK, you are refactoring it right now, so I disabled tenant config loading in the latest commit to avoid `test_tenant_relocation` failures. It means that tenant-specific config now doesn't survive relocation. 

I'd like to merge this PR anyway and fix it later, when we will have full support of the tenant relocation elsewhere.
What do you think?

pageserver/src/http/openapi_spec.yml
```diff
@@ -328,19 +328,10 @@ paths:
...
       responses:
-        "201":
-          description: New tenant created successfully
-          content:
-            application/json:
-              schema:
-                type: string
-                format: hex
+        "200":
+          description: Config applied
```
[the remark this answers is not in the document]
lubennikovaav: Thank you, it was a copy-paste mistake. Fixed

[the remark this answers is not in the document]
ololobus: Also we should put this updated spec into https://github.com/neondatabase/cloud/blob/main/spec/pageserver.yml, run `make generate-all` and open a PR to check that everything works well after all these changes

knizhnik:
> 

I see, it is copied...
Sorry for false alarm.

knizhnik (approved):

pageserver/src/layered_repository.rs
```diff
@@ -572,6 +597,70 @@ impl LayeredRepository {
...
+        Ok(tenant_conf)
+    }
+
+    pub fn persist_tenant_config(
+        conf: &'static PageServerConf,
+        tenantid: ZTenantId,
+        tenant_conf: TenantConfOpt,
+    ) -> anyhow::Result<()> {
+        let _enter = info_span!("saving tenantconf").entered();
+        let target_config_path = TenantConf::tenantconf_path(conf, tenantid);
```
LizardWizzard, line 643: nit: probably `tenantconf_path` can be just `path(...)` for simplicity if called like that

pageserver/src/layered_repository.rs
```diff
@@ -1894,8 +2019,27 @@ impl LayeredTimeline {
...
-
-            // 2. Is it needed by a child branch?
+            // 2. It is newer than PiTR interval?
+            // We use modification time of layer file to estimate update time.
+            // This estimation is not quite precise but maintaining LSN->timestamp map seems to be overkill.
+            // It is not expected that users will need high precision here. And this estimation
+            // is conservative: modification time of file is always newer than actual time of version
+            // creation. So it is safe for users.
+            //
+            if let Ok(metadata) = fs::metadata(&l.filename()) {
```
LizardWizzard, line 2029: A possible "bloat" issue still persists here. It can arise from modification time change because of layer upload/download. If we are not solving it now let's mention it in a comment with reasoning or create separate issue for a needed fix.

pageserver/src/layered_repository.rs
```diff
@@ -2232,7 +2376,8 @@ pub mod tests {
             }
 
             let cutoff = tline.get_last_record_lsn();
-            tline.update_gc_info(Vec::new(), cutoff);
+
+            tline.update_gc_info(Vec::new(), cutoff, Duration::ZERO);
```
LizardWizzard, line 2380: `Vec::new()` -> `vec![]`

pageserver/src/page_service.rs
```diff
@@ -672,6 +673,39 @@ impl postgres_backend::Handler for PageServerHandler {
                 }
             }
             pgb.write_message_noflush(&BeMessage::CommandComplete(b"SELECT 1"))?;
+        } else if query_string.starts_with("show ") {
```
LizardWizzard, line 676: Do we really need it? What is the use case? It is already available through http api

pageserver/src/layered_repository.rs
```diff
@@ -572,6 +597,70 @@ impl LayeredRepository {
...
+    pub fn persist_tenant_config(
+        conf: &'static PageServerConf,
+        tenantid: ZTenantId,
+        tenant_conf: TenantConfOpt,
+    ) -> anyhow::Result<()> {
+        let _enter = info_span!("saving tenantconf").entered();
+        let target_config_path = TenantConf::tenantconf_path(conf, tenantid);
+        info!("save tenantconf to {}", target_config_path.display());
+
+        let mut conf_content = r#"# This file conatais a specific per-tenant's config.
```
SomeoneToIgnore, line 646: > 
```suggestion
        let mut conf_content = r#"# This file contains a specific per-tenant's config.
```

pageserver/src/tenant_config.rs
```diff
@@ -0,0 +1,174 @@
...
+    pub compaction_target_size: Option<u64>,
+    pub compaction_period: Option<Duration>,
+    pub compaction_threshold: Option<usize>,
+    pub gc_horizon: Option<u64>,
+    pub gc_period: Option<Duration>,
+    pub pitr_interval: Option<Duration>,
+}
+
+impl TenantConfOpt {
+    pub fn default() -> Self {
```
LizardWizzard, line 69: Please use `derive(Default)` or `impl Default for TenantConfOpt`

pageserver/src/layered_repository.rs
```diff
@@ -465,6 +472,20 @@ impl From<LayeredTimelineEntry> for RepositoryTimeline<LayeredTimeline> {
 
 /// Private functions
 impl LayeredRepository {
+    pub fn get_tenant_conf(&self) -> TenantConf {
```
LizardWizzard, line 475: IMO using getters instead of merging configs every time we need them is a bit cleaner. E g in the compaction thread we're using only `compaction_period`

pageserver/src/layered_repository.rs
```diff
@@ -968,12 +1064,34 @@ impl Timeline for LayeredTimeline {
...
+        tenant_conf
+            .checkpoint_distance
+            .unwrap_or(self.conf.default_tenant_conf.checkpoint_distance)
+    }
+
+    fn get_compaction_target_size(&self) -> u64 {
+        let tenant_conf = self.tenant_conf.read().unwrap();
+        tenant_conf
+            .compaction_target_size
+            .unwrap_or(self.conf.default_tenant_conf.compaction_target_size)
```
SomeoneToIgnore, line 1078: > .unwrap_or

I see that `TenantConfOpt` is used in the layered repository only, and then its values are either used (if present), or defaults are used instead.
That all feels a bit odd, cannot we merge those values once during the init? 

We sure still need the `TenantConfOpt` to parse things from requests/configs, but why do we store it, not the `TenantConf`?

pageserver/src/tenant_config.rs
```diff
@@ -0,0 +1,174 @@
...
+        }
+        if let Some(gc_period) = other.gc_period {
+            self.gc_period = Some(gc_period);
+        }
+        if let Some(pitr_interval) = other.pitr_interval {
+            self.pitr_interval = Some(pitr_interval);
+        }
+    }
+
+    pub fn from_tenant_conf(tenant_conf: TenantConf) -> Self {
```
LizardWizzard, line 125: can be `impl From`. Also I see that it is used only in tests, so maybe makes sense to move it to tests module as a separate helper

pageserver/src/tenant_mgr.rs
```diff
@@ -75,9 +76,13 @@ pub fn load_local_repo(
         // Set up a WAL redo manager, for applying WAL records.
         let walredo_mgr = PostgresRedoManager::new(conf, tenant_id);
 
+        // Load config file
+        let tenant_conf = LayeredRepository::load_tenant_config(conf, tenant_id).unwrap();
```
LizardWizzard, line 80: Lets use a result here instead of a panic. Because this function is used in `apply_timeline_sync_status_updates` so if it panics this will affect other tenants

pageserver/src/tenant_config.rs
```diff
@@ -0,0 +1,174 @@
...
+    pub compaction_period: Duration,
+    // Level0 delta layer threshold for compaction.
+    pub compaction_threshold: usize,
+    pub gc_horizon: u64,
+    pub gc_period: Duration,
+    pub pitr_interval: Duration,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub struct TenantConfOpt {
```
LizardWizzard, line 58: nit: naming suggestion: `TenantConfOverrides` this way it'll indicate that here we override global config

test_runner/batch_others/test_tenant_conf.py
```diff
@@ -0,0 +1,32 @@
+from contextlib import closing
+
+import pytest
+
+from fixtures.zenith_fixtures import ZenithEnvBuilder
+
+
+def test_tenant_config(zenith_env_builder: ZenithEnvBuilder):
```
SomeoneToIgnore, line 8: I propose to test the config update here too.
Would be amazing to update it to something ridiculous and then try to run some operation to see if the config change had influenced that, but getting the value is also ok, I guess.

control_plane/src/storage.rs
```diff
@@ -341,11 +342,29 @@ impl PageServerNode {
 
     pub fn tenant_create(
         &self,
-        new_tenant_id: Option<ZTenantId>,
+        tenant_id: Option<ZTenantId>,
+        settings: HashMap<&str, &str>,
```
SomeoneToIgnore, line 346: Up to you, but I'd rather had some method to parse the strings into `TenantConfOpt` (maybe put the toml parsing into that module too and add some unit tests for both?) and pass it around, not some odd map.

SomeoneToIgnore (commented):
I don't have a solid perspective on tenant relocation, so it looks ok to me to do what you've proposed, but maybe @LizardWizzard has better ideas.

Might it be simpler to use a default config for every tenant by default, if nothing provided?

LizardWizzard (commented):
+1 for fixing tenant relocation tests separately. I have an idea on how to do that without too much code changes

SomeoneToIgnore:
Also feels like we could use more unit tests on parsing at least.

pageserver/src/layered_repository.rs
LizardWizzard: This is my subjective opinion :) Feel free to leave as is

pageserver/src/layered_repository.rs
lubennikovaav: I agree that it looks strange, but this implementation allows us to preserve information about what tenant-specific settings are set. `persist_tenant_conf()` only saves overridden parameters for the same reason.
This is necessary allow global config updates.
I.e. we want to update `gc_horizon` in PageServerConf for all tenants that do not use custom values.
Not sure, if this scenario will be used very often, because it requires a pageserver restart, but technically it is possible.

pageserver/src/config.rs
```diff
@@ -355,6 +269,8 @@ impl PageServerConfigBuilder {
                 .remote_storage_config
                 .ok_or(anyhow::anyhow!("missing remote_storage_config"))?,
             id: self.id.ok_or(anyhow::anyhow!("missing id"))?,
+            // TenantConf is handled separately
```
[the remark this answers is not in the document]
lubennikovaav: TBH, I was trying to avoid using this Builder machinery. 
All TenantConf options have default values, so it's totally fine if some of them are absent.

pageserver/src/layered_repository.rs
```diff
@@ -572,6 +597,70 @@ impl LayeredRepository {
...
+            return Ok(TenantConfOpt::default());
+        }
+
+        // load and parse file
+        let config = fs::read_to_string(target_config_path)?;
+
+        let toml = config.parse::<toml_edit::Document>()?;
+
+        let mut tenant_conf = TenantConfOpt::default();
+        for (key, item) in toml.iter() {
```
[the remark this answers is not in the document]
lubennikovaav: I did it this way to unify this code with global config parsing.
Now we have `[tenant_config] ` table in both files.

pageserver/src/bin/pageserver.rs
```diff
@@ -246,7 +246,7 @@ fn start_pageserver(conf: &'static PageServerConf, daemonize: bool) -> Result<()
         for (timeline_id, init_status) in local_timeline_init_statuses {
             match init_status {
                 remote_storage::LocalTimelineInitStatus::LocallyComplete => {
-                    debug!("timeline {} for tenant {} is locally complete, registering it in repository", tenant_id, timeline_id);
+                    debug!("timeline {} for tenant {} is locally complete, registering it in repository", timeline_id, tenant_id);
```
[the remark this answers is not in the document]
lubennikovaav: Oh, cool. I didn't know that. I'll leave it as-is if you don't mind.
It's unrelated to the PR anyway, I just noticed a typo in the log.

ololobus:
Pageserver from this branch responds with 500 on create timeline request https://github.com/neondatabase/cloud/runs/6093559286?check_suite_focus=true#step:28:137

I think this is because I asked to rename `new_tenant_id` to just `tenant_id` %) Maybe keep both for backward compatibility for some time? Otherwise, all such breaking changes require to deploy both storage and console, but we are not ready to deploy new storage on prod, IIRC. So we can block releases

pageserver/src/config.rs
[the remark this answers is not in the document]
LizardWizzard: I'm ok with avoiding builder machinery for tenant conf, but l'm more about the tenant_conf field in pageserver config. Builder already used there so IMO better option is to handle this field the same way as the others

pageserver/src/http/routes.rs
```diff
@@ -306,6 +329,45 @@ async fn tenant_create_handler(mut request: Request<Body>) -> Result<Response<Bo
     })
 }
 
+async fn tenant_config_handler(mut request: Request<Body>) -> Result<Response<Body>, ApiError> {
+    // check for management permission
+    check_permission(&request, None)?;
```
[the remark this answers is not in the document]
lubennikovaav: Fixed

pageserver/src/config.rs
```diff
@@ -515,9 +426,38 @@ impl PageServerConf {
...
 
+        conf.default_tenant_conf = t_conf.merge(TenantConf::default());
+
         Ok(conf)
     }
 
+    // subroutine of parse_and_validate to parse `[tenant_conf]` section
+
+    pub fn parse_toml_tenant_conf(item: &toml_edit::Item) -> Result<TenantConfOpt> {
+        let mut t_conf = TenantConfOpt::default();
```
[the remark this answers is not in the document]
lubennikovaav: I think we need to keep TenantConfOpt, please see the discussion above:
https://github.com/neondatabase/neon/pull/1504#discussion_r854214360

I'll add some comment.

pageserver/src/layered_repository.rs
lubennikovaav: Agree. I'll implement getters.
Though, we still need get_tenant_conf() to implement show.

pageserver/src/layered_repository.rs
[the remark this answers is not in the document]
LizardWizzard: I'm was not familiar with toml crate api, I thought that types for table entries and plain values should be the same, but unfortunately, it is not true

pageserver/src/config.rs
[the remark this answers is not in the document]
LizardWizzard: Got it, makes sense

pageserver/src/page_service.rs
LizardWizzard: Hm, for some reason I thought that there is already a `get` endpoint for config. Personally, I'm leaning towards using http api as much as possible without duplicating functionality in postgres protocol. But if people find these helpers useful let's keep them

pageserver/src/layered_repository.rs
```diff
@@ -572,6 +597,70 @@ impl LayeredRepository {
...
+    /// Locate and load config
+    pub fn load_tenant_config(
+        conf: &'static PageServerConf,
+        tenantid: ZTenantId,
+    ) -> anyhow::Result<TenantConfOpt> {
+        let target_config_path = TenantConf::tenantconf_path(conf, tenantid);
+
+        info!("load tenantconf from {}", target_config_path.display());
+
+        // XXX If the config file is not found, assume that we're attaching
```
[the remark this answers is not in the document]
lubennikovaav: https://github.com/neondatabase/neon/issues/1555

lubennikovaav:
> Pageserver from this branch responds with 500 on create timeline request https://github.com/neondatabase/cloud/runs/6093559286?check_suite_focus=true#step:28:137
> 
> I think this is because I asked to rename `new_tenant_id` to just `tenant_id` %) Maybe keep both for backward compatibility for some time? Otherwise, all such breaking changes require to deploy both storage and console, but we are not ready to deploy new storage on prod, IIRC. So we can block releases

I just left old name in CreateTenant and introduced new endpoint /t1/tenant/config.
There's some code duplication, but it doesn't hurt. We can clean it up later.

pageserver/src/http/openapi_spec.yml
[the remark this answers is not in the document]
lubennikovaav: https://github.com/neondatabase/cloud/pull/1070

PR #1507, opened 2022-04-14
Set wal_keep_size to zero

petuhovskiy:
- Set wal_keep_size=0 as a default in control_plane
- Test that compute is not keeping all generated WAL on disk
- Test work when safekeepers lag and restart

resolves https://github.com/neondatabase/neon/issues/767

### DIFF
--- a/control_plane/src/compute.rs
+++ b/control_plane/src/compute.rs
@@ -272,9 +272,6 @@ impl PostgresNode {
         conf.append("wal_sender_timeout", "5s");
         conf.append("listen_addresses", &self.address.ip().to_string());
         conf.append("port", &self.address.port().to_string());
-
-        // TODO: Do we have a problem with replication slot when streaming
-        // to pageserver? https://github.com/neondatabase/neon/issues/767
         conf.append("wal_keep_size", "0");
 
         // Configure the node to fetch pages from pageserver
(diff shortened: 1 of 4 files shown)


petuhovskiy:
CI tests passed, so it looks like we don't have any problems with `wal_keep_size=0`.

lubennikovaav (commented):
Hmm. I am not sure that this test properly illustrates the issue.

To illustrate the problem with `wal_keep_size=0` we need to set a lagging safekeeper. 
If safekeeper is able to keep up with postgres WAL, everything is fine,

If safekeeper falls behind by more than `wal_keep_size` megabytes, the sending server might remove a WAL segment still needed by the safekeeper and they will be lost.

petuhovskiy:
We already have a test with lagging safeekepers in test_restarts_under_load. There one of the safekeepers is stopped every 10 seconds and it simulates lag by preventing truncateLsn to move for 10 seconds.

We can add another test with lagging safekeepers and `max_wal_size=32MB` (to trigger segments recycling faster), but I'm not sure about that.

`test_wal_keep_size` is useful to check that compute can work without a large disk usage.

petuhovskiy:
I added new test with lagging safekeepers and fixed description for the previous test

arssher (approved):

Issue #1491, opened 2022-04-12
Safekeepers sync is slow when EC2 safekeeper instance is down


petuhovskiy:
Cluster `bold-pond-326774` from staging had a failed operation `start_compute` at Apr 12, 2022 12:35 PM (GMT+3). Seems that operation failed because safekeepers syncing took more than 2 minutes. [logs](https://observer.zenith.tech/d/IJSLHBOnk/zenith-logs-compute-nodes?orgId=1&var-environment=staging&var-instance=All&var-search=bold-pond-326774&var-exclude=assign%20requested&from=1649756018244&to=1649757002651)

My theory is that we have a problem somewhere near `--sync-safekeepers` exit criteria:
https://github.com/neondatabase/postgres/blob/61afbf978b17764134ab6f1650bbdcadac147e71/src/backend/replication/walproposer.c#L2016-L2018

More specifically, we're waiting for safekeeper in connecting state until TCP timeout happens. This is not reproduced in tests because localhost is always up.

arssher:
Since walproposer uses libpq in async way, we should manually implement `connect_timeout` behaviour of libpq. Or just wait for pageserver to choose safekeeper to fetch WAL from properly.

For the record, currently this occurs on old clusters in staging because they try to reach dead safekeeper as their config is not updated.

PR #1515, opened 2022-04-15
Bump vendor/postgres to add safekeeper connection timeout.

arssher:


petuhovskiy (approved):
LGTM. Will help with https://github.com/neondatabase/neon/issues/1491, sync shouldn't take more than 5 seconds now. But I think issue should be still open until we remove check for SS_OFFLINE state, so sync will not take more than several RTT to finish.

SomeoneToIgnore:
Not sure if related, but the one of the debug jobs on `main` had failed recently with some new (to me) timeout error: https://app.circleci.com/pipelines/github/neondatabase/neon/5575/workflows/ae913a10-d10a-458f-8b3a-aca6c9482de9/jobs/53909

on unrelated changes.

Looks flacky, since previous PRs today did not have such errors.

arssher:
Definitely not related. This PR adds connection timeout on conn attempt to safekeeper, which is logged and that never occured in failure above -- there massive insert times out. `conn_options` sets 2m timeout on any query in our tests, and on my machine that insert takes 70s, so it looks like we need either make insert less massive or increase timeout.

What I don't immediately understand is why that test on my machine always fails with `psycopg2.errors.UndefinedTable: relation "t" does not exist`...

PR #1516, opened 2022-04-15
Use the same version of `rust-postgres` everywhere.

funbringer:
Turns out we still had a stale dep in `compute_tools`.


SomeoneToIgnore (approved):
Nice, had see those double compilations for quite a while, thanks for taking the time to check on that.

PR #1526, opened 2022-04-16
Revert libc upgrade

SomeoneToIgnore:
Despite manual updates to staring libc, I see that there are no logs from all instances approx. after the initial libc PR bump merge:

image

Reverting those changes to see if that helps.


SomeoneToIgnore:
Apparently, the logs missing were caused by something else but malfunctioning libc, physically the logs are still present on envs:

image

PR #1528, opened 2022-04-16
Use mold in Docker image builds

SomeoneToIgnore:
Since now we're able to use `zimg` images for building, we can use a more performant linker that is installed on them.


SomeoneToIgnore:
Old build times:
old


New build times:
new

The results are a bit odd, I would expect both times to go down, yet compute tools build build time had increased.
Yet the biggest time was reduced, so the PR brings some benefit after all.

cicdteam (approved):

PR #1533, opened 2022-04-18
temporary disable bench uploads

LizardWizzard:
The cluster that we upload perf results to has some problems. Insert triggers this error:

```
psycopg2.errors.IoError: could not read block 0 in rel 1663/16385/24577.2 from page server at lsn 0/0247E400
DETAIL:  page server returned error: Failed to get reconstruct data 1663/16385/24577_vm.0 "rel_1663_16385_24577_2_0_00000000019EBD78_0000000001C66381" 0 0/247EA88
```

The cluster is in production env, id `dry-bar-041103`.

Currently, production runs on old version of storage, so I think we can disable the upload for now. Because new storage significantly changed related code, so IMO there is little sense in debugging. Thoughts?

Results are still collected and added to zenith-perf-results repo


LizardWizzard:
I reinserted data into the bench cluster and it seems that it helped

PR #1537, opened 2022-04-19
Add materialized page cache invalidation mechanism to avoid delta lookups

knizhnik:
Right now page cache is used for storing materialized pages.
But if LSN of materialized page is smaller than LSN requested in `get_page_at_lsn`, then page server will have to check if there are some wal records in this range which is not so fast operation.

My proposal is to use cache invalidation mechanism: cached image layers is marked as _latest_.
If some wal record is received for this pages then `latest` flag is cleared.
It allows to use cached page without any extra checks is  `latest`  flag is set.

It gives at my compute speed advance at select-only `pgbench -S -c 10` from 16k to 22k TPS.

### DIFF
--- a/pageserver/src/layered_repository.rs
+++ b/pageserver/src/layered_repository.rs
@@ -881,32 +881,12 @@ impl Timeline for LayeredTimeline {
     /// Look up the value with the given a key
     fn get(&self, key: Key, lsn: Lsn) -> Result<Bytes> {
         debug_assert!(lsn <= self.get_last_record_lsn());
+        self.get_at(key, lsn, false)
+    }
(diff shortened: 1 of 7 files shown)


MMeent:
Agreed in principle, but shouldn't we use a range of LSN in which the materialized page is valid instead of 'latest' and invalidation? I.e., a page is inserted into the cache and marked with [TheLastRecordEndLSN, requestLSN) or [TheLastRecordEndLSN, ThisPage'sNextWALRecordEndLSN] as valid range.

LizardWizzard:
As discussed moving to draft.

Issue #1403, opened 2022-03-24
trim WAL on safekeepers


stepashka:
DOD:
Safekeepers that are not in touch with the pageserver can coordinate and delete WAL after one them confirms that WAL is stored on pageserver and S3 (the trimming itself was implemented in: #944)

We need to keep enough WAL for

- safekeper p2p recovery
- direct S3 offloading from safekeepers
- the last checkpoint offloaded to s3 (remote consistent LSN)

- [ ] #944 
- [ ]

arssher:
Now on staging WAL is removed without waiting for s3 offloading. FYI @stepashka 
https://github.com/neondatabase/neon/pull/1662

arssher:
s3 offloading is enabled on staging, WAL removal happens after it.

PR #1550, opened 2022-04-20
Remove old WAL on safekeepers.

arssher:
Remove when it is consumed by all of 1) pageserver (remote_consistent_lsn) 2)
safekeeper peers 3) s3 WAL offloading.

In test s3 offloading for now is mocked by directly bumping s3_wal_lsn.

ref #1403


safekeeper/src/http/routes.rs
```diff
@@ -122,6 +124,20 @@ async fn timeline_create_handler(mut request: Request<Body>) -> Result<Response<
     json_response(StatusCode::CREATED, ())
 }
 
+/// Used only in tests to hand craft required data.
+async fn timeline_info_handler(mut request: Request<Body>) -> Result<Response<Body>, ApiError> {
```
petuhovskiy, line 128: `timeline_info` is a bit confusing name, my first thought was that it's the same as `timeline_status`. Maybe it's better to add something about record_safekeeper_info/test_only in the name.

safekeeper/src/timeline.rs
```diff
@@ -87,6 +88,7 @@ struct SharedState {
     active: bool,
     num_computes: u32,
     pageserver_connstr: Option<String>,
+    last_removed_segno: XLogSegNo,
```
petuhovskiy, line 91: Not directly related to this PR, but I'd like also to have `disk_start_lsn` stored in the control file, for the minimal LSN available locally.

test_runner/batch_others/test_wal_acceptor.py
```diff
@@ -370,6 +370,52 @@ def test_broker(zenith_env_builder: ZenithEnvBuilder):
...
+            cur.execute("INSERT INTO t SELECT generate_series(1,100000), 'payload'")
+
+    tenant_id = pg.safe_psql("show zenith.zenith_tenant")[0][0]
+    timeline_id = pg.safe_psql("show zenith.zenith_timeline")[0][0]
+
+    # force checkpoint to advance remote_consistent_lsn
+    with closing(env.pageserver.connect()) as psconn:
+        with psconn.cursor() as pscur:
+            pscur.execute(f"checkpoint {tenant_id} {timeline_id}")
+
```
petuhovskiy, line 399: We can also check that 000000010000000000000001 segment exists here (before updating s3_wal_lsn).

safekeeper/src/wal_storage.rs
```diff
@@ -467,6 +473,39 @@ impl Storage for PhysicalStorage {
...
+                continue;
+            }
+            let (segno, _) = XLogFromFileName(fname_str, wal_seg_size);
+            if segno <= segno_up_to {
+                remove_file(entry_path)?;
+            }
+        }
+    }
+    info!(
+        "removed WAL segments up to {}",
```
petuhovskiy, line 505: We can also track number of files removed in the for loop, as well as min-max removed segments and print it the logs, something like this:
```
removed 3 WAL segments from 000000010000000000000001 to 000000010000000000000003
```

petuhovskiy (approved):
LGTM. Left some comments on minor things

PR #1551, opened 2022-04-20
Restructure and rename common libraries

SomeoneToIgnore:
Prepares the common neon libraries for further librarification:
* `zenith_utils` got renamed into `neon_utils` and now it's just a directory that keeps all libraries inside it
* Real `zenith_utils` code is now located in `neon_utils/common_utils` The naming does not shine, so any other proposals are good
* Move and rename `zenith_metrics` into `neon_utils/neon_metrics`

In the future, all other libraries go inside the `neon_utils/` directory.
I plan to add `etcd_neon` if able to properly librarify it, other plans were to split `common_utils` into multiple libraries.


.config/hakari.toml
```diff
@@ -10,6 +10,7 @@ dep-format-version = "2"
 # Hakari works much better with the new feature resolver.
 # For more about the new feature resolver, see:
 # https://blog.rust-lang.org/2021/03/25/Rust-1.51.0.html#cargos-new-feature-resolver
+# Have to keep the resolver still here, despite it's now the default for 2021 edition.
```
hlinnaka, line 13: Why do we need to keep this if it's the default?

hlinnaka (approved):

hlinnaka:
I'm not a fan of pushing `common_utils` and `neon_metrics` deeper, into the new subdirectory. I'd prefer a straight renaming:
```
zenith_utils -> neon_utils
zenith_metrics -> neon_metrics
```
or just
```
zenith_utils -> utils
zenith_metrics -> metrics
```
If we find that we have too many top-level directories in the future, and want to add another directory-level, we can do that later.

(I don't feel very strongly either way though...)

.config/hakari.toml
SomeoneToIgnore: Hakari requires this field to be present, so there has to be something:

```
Work/neon/neon kb/neon-utils:origin/kb/neon-utils ❯ cargo hakari generate          
Error: 
   0: error deserializing Hakari config at /Users/someonetoignore/Work/neon/neon/.config/hakari.toml
   1: missing field `resolver` at line 1 column 1
```

SomeoneToIgnore:
I like this prefix-less approach for the libs, actually

```
zenith_utils -> utils
zenith_metrics -> metrics
```

As for the nested directory, I'd better have this personally, since I expect more libs to appear later:
* there had been talks of splitting the `zenith_utils` into lesser crates, I would split away the `logging` and `http` things at least
* we expect to add more libs later, say, `etcd` with our common prefixes or configs

If we do so, there would be a big mixture of "binary" crates such as pageserver, safekeeper, control_plane, etc. and those util libs in the same directory. Personally not a fan of this at all, I feel that libs should go into some dedicated corner and live there.

------

What if we do this:

```
neon_utils/
                   utils/
                   metrics/
                   _whatever_else_/
```

?

petuhovskiy:
I can suggest:
```
commons/
       utils/
       metrics/
       postgres_ffi/
       whatever_else/
```

IMO neon_ prefix inside the neon repo is redundant.

LizardWizzard:
> commons/

Or just `lib/`

.config/hakari.toml
LizardWizzard: I think It is the default in cargo, not in hakari

SomeoneToIgnore:
image

Decided not to move the `workspace_hack` inside, since not really a lib.

.config/hakari.toml
SomeoneToIgnore: Phrased this as 

```
# Have to keep the resolver still here since hakari requires this field,
# despite it's now the default for 2021 edition & cargo.
```

PR #1567, opened 2022-04-22
Limit concurrent S3 and IAM interactions

SomeoneToIgnore:
Based on https://neondb.slack.com/archives/C034ED3MM7A/p1650624886422879

Number 50 for concurrency limit is chosen arbitrarily, but looks like we're rather limited by IAM server here?

```rust
/// Currently, sync happens with AWS S3, that has two limits on requests per second:
/// ~200 RPS for IAM services
/// https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/UsingWithRDS.IAMDBAuth.html
/// ~3500 PUT/COPY/POST/DELETE or 5500 GET/HEAD S3 requests
/// https://aws.amazon.com/premiumsupport/knowledge-center/s3-request-limit-avoid-throttling/
```

### DIFF
--- a/pageserver/src/remote_storage/storage_sync.rs
+++ b/pageserver/src/remote_storage/storage_sync.rs
@@ -62,7 +62,7 @@ pub mod index;
 mod upload;
 
 use std::{
-    collections::{hash_map, HashMap, HashSet, VecDeque},
+    collections::{HashMap, HashSet, VecDeque},
     fmt::Debug,
     num::{NonZeroU32, NonZeroUsize},
     ops::ControlFlow,
@@ -132,7 +132,9 @@ lazy_static! {
 /// mpsc approach was picked to allow blocking the sync loop if no tasks are present, to avoid meaningless spinning.
 mod sync_queue {
     use std::{
-        collections::{hash_map, HashMap},
+        collections::{hash_map, HashMap, HashSet},
+        num::NonZeroUsize,
+        ops::ControlFlow,
         sync::atomic::{AtomicUsize, Ordering},
     };
 
@@ -179,7 +181,7 @@ mod sync_queue {
     /// Polls a new task from the queue, using its receiver counterpart.
     /// Does not block if the queue is empty, returning [`None`] instead.
     /// Needed to correctly track the queue length.
-    pub async fn next_task(
+    async fn next_task(
         receiver: &mut UnboundedReceiver<(ZTenantTimelineId, SyncTask)>,
     ) -> Option<(ZTenantTimelineId, SyncTask)> {
         let task = receiver.recv().await;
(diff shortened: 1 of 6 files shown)


ololobus:
Interesting, are S3 credentials cached somehow after EC2 instance startup / first use? Or we do IAM request + S3 request on each S3 operation?

SomeoneToIgnore:
Yes, from what I see in `rusoto` code for downloads, for instance:
* We use `client.get_object()` https://github.com/neondatabase/neon/blob/3e6087a12f26ebefe6b91ea78be5d927c72b2a48/pageserver/src/remote_storage/s3_bucket.rs#L203-L210
* Rusoto constructs a request and sends it 
```rust
let mut response = self
            .sign_and_dispatch(request, GetObjectError::from_response)
            .await?; 
```
(https://docs.rs/rusoto_s3/latest/src/rusoto_s3/generated.rs.html#21041)
* Which calls the request prodiver inside:
```rust
let credentials = if let Some(to) = timeout {
    time::timeout(to, provider.credentials())
        .await
        .map_err(|_| CredentialsError {
            message: "Timeout getting credentials".to_owned(),
        })
        .and_then(std::convert::identity)
} else {
    provider.credentials().await
}
.map_err(SignAndDispatchError::Credentials)?;
```
https://github.com/rusoto/rusoto/blob/a58a4ce9a4d4fe05a88db6c0c26cc3d16ae1b88d/rusoto/core/src/client.rs#L154-L162

* and for our case, this is the IAM fetch code: https://github.com/rusoto/rusoto/blob/master/rusoto/credential/src/instance_metadata.rs
pure requests, no caching.

SomeoneToIgnore:
I see that official AWS approach does not differ much: https://github.com/awslabs/aws-sdk-rust/blob/main/sdk/aws-config/src/ecs.rs

If I've got the docs right, there's a way to set a `AWS_CONTAINER_AUTHORIZATION_TOKEN` token, but not sure this is ok for our case. Otherwise — pure requests.

LizardWizzard (approved):
Looks good to me

Maybe makes sense to keep limiter inside remote storage impl
Pros: 
* there is no need to froward limiter along with storage 
* it is not possible to use remove storage without limiter
Cons?

SomeoneToIgnore:
Thanks, I've reworked the PR as you've proposed, now there's one limit on the amount of timelines to process (for the sync_queue) and the other one for S3 bucket, with `Semaphore` used only in the `s3_bucket` module.

pageserver/src/remote_storage/storage_sync.rs
```diff
@@ -195,15 +196,33 @@ mod sync_queue {
     /// or two (download and upload, if both were found in the queue during batch construction).
     pub async fn next_task_batch(
         receiver: &mut UnboundedReceiver<(ZTenantTimelineId, SyncTask)>,
-        mut max_batch_size: usize,
-    ) -> HashMap<ZTenantTimelineId, SyncTask> {
-        if max_batch_size == 0 {
-            return HashMap::new();
+        max_timelines_to_sync: usize,
```
LizardWizzard, line 199: Maybe it is not a good idea, but NonZeroUsize should be able to eliminate zero case handled below

LizardWizzard (approved):
Looks good to me, thanks!

pageserver/src/remote_storage/storage_sync.rs
SomeoneToIgnore: I think this is a great idea actually, fixed.

PR #1571, opened 2022-04-25
Use fsync+rename for atomic downloads from remote storage

LizardWizzard:
The problem I have is that we need to build pageserver with the failpoint feature enabled. Probably we do not need that for production builds, do we? I plan to enable it for debug builds only so it can be used in tests properly

I guess existing failpoint invocation in test_backpressure is inactive

### DIFF
--- a/pageserver/Cargo.toml
+++ b/pageserver/Cargo.toml
@@ -4,8 +4,12 @@ version = "0.1.0"
 edition = "2021"
 
 [features]
-default = []
+# It is simpler infra-wise to have failpoints enabled by default
+# It shouldnt affect perf in any way because failpoints 
+# are not placed in hot code paths
+default = ["failpoints"]
 profiling = ["pprof"]
+failpoints = ["fail/failpoints"]
 
 [dependencies]
 chrono = "0.4.19"
--- a/pageserver/src/remote_storage/storage_sync.rs
+++ b/pageserver/src/remote_storage/storage_sync.rs
@@ -62,7 +62,9 @@ pub mod index;
 mod upload;
 
 use std::{
+    borrow::Cow,
     collections::{HashMap, HashSet, VecDeque},
+    ffi::OsStr,
     fmt::Debug,
     num::{NonZeroU32, NonZeroUsize},
     ops::ControlFlow,
@@ -89,7 +91,10 @@ use self::{
 use super::{LocalTimelineInitStatus, LocalTimelineInitStatuses, RemoteStorage, SyncStartupData};
 use crate::{
     config::PageServerConf,
-    layered_repository::metadata::{metadata_path, TimelineMetadata},
+    layered_repository::{
+        metadata::{metadata_path, TimelineMetadata},
+        LayeredRepository,
+    },
     repository::TimelineSyncStatusUpdate,
     tenant_mgr::apply_timeline_sync_status_updates,
     thread_mgr,
@@ -103,6 +108,7 @@ use metrics::{
 use utils::zid::{ZTenantId, ZTenantTimelineId, ZTimelineId};
 
 pub use self::download::download_index_part;
+pub use self::download::TEMP_DOWNLOAD_EXTENSION;
 
 lazy_static! {
     static ref REMAINING_SYNC_ITEMS: IntGauge = register_int_gauge!(
@@ -782,8 +788,14 @@ where
     P: Debug + Send + Sync + 'static,
     S: RemoteStorage<StoragePath = P> + Send + Sync + 'static,
 {
-    match download_timeline_layers(storage, current_remote_timeline, sync_id, new_download_data)
-        .await
+    match download_timeline_layers(
+        conf,
+        storage,
+        current_remote_timeline,
+        sync_id,
+        new_download_data,
+    )
+    .await
     {
         DownloadedTimeline::Abort => {
             register_sync_status(sync_start, task_name, None);
@@ -852,18 +864,28 @@ async fn update_local_metadata(
 
     if local_lsn < Some(remote_lsn) {
         info!("Updating local timeline metadata from remote timeline: local disk_consistent_lsn={local_lsn:?}, remote disk_consistent_lsn={remote_lsn}");
-
-        let remote_metadata_bytes = remote_metadata
-            .to_bytes()
-            .context("Failed to serialize remote metadata to bytes")?;
-        fs::write(&local_metadata_path, &remote_metadata_bytes)
-            .await
-            .with_context(|| {
-                format!(
-                    "Failed to write remote metadata bytes locally to path '{}'",
-                    local_metadata_path.display()
-                )
-            })?;
+        // clone because spawn_blocking requires static lifetime
+        let cloned_metadata = remote_metadata.to_owned();
+        let ZTenantTimelineId {
+            tenant_id,
+            timeline_id,
+        } = sync_id;
+        tokio::task::spawn_blocking(move || {
+            LayeredRepository::save_metadata(conf, timeline_id, tenant_id, &cloned_metadata, true)
+        })
+        .await
+        .with_context(|| {
+            format!(
+                "failed to join save_metadata task for {}",
+                local_metadata_path.display()
+            )
+        })?
+        .with_context(|| {
+            format!(
+                "Failed to write remote metadata bytes locally to path '{}'",
+                local_metadata_path.display()
+            )
+        })?;
     } else {
         info!("Local metadata at path '{}' has later disk consistent Lsn ({local_lsn:?}) than the remote one ({remote_lsn}), skipping the update", local_metadata_path.display());
     }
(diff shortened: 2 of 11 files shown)


pageserver/src/remote_storage/storage_sync.rs
```diff
@@ -1192,6 +1194,22 @@ fn register_sync_status(sync_start: Instant, sync_name: &str, sync_status: Optio
...
+    );
+    let new_extension = match original_path
+        .as_ref()
+        .extension()
+        .map(OsStr::to_string_lossy)
+    {
+        Some(extension) => format!("{extension}.{suffix}"),
+        None => suffix.to_owned(),
+    };
+    original_path.as_ref().with_extension(new_extension)
```
SomeoneToIgnore, line 1210: NITs:

* you can use `Cow` and avoid cloning with `to_owned`: 
```rust
{
        Some(extension) => Cow::Owned(format!("{extension}.{suffix}")),
        None => Cow::Borrowed(suffix),
    };
    original_path.as_ref().with_extension(new_extension.as_ref())
```

* not super fond of this panic, might it be ok to be lenient and "merge" such dots (could be multiple dots though..)? Or even ignore them, let there be multiple dots in the name?

* might be good to add a negative test in case you want to fight dots

pageserver/src/remote_storage/storage_sync/download.rs
```diff
@@ -149,6 +158,19 @@ where
...
+                let parent_path = temp_file_path
+                    .parent()
+                    .expect("no parent dir for layer file");
+                let parent_dir = fs::File::open(parent_path).await?;
+                parent_dir.sync_all().await?;
+
+                fail::fail_point!("remote-storage-download-pre-rename", |_| {
+                    bail!("failpoint triggered")
+                });
+                fs::rename(temp_file_path, &layer_desination_path).await?;
```
SomeoneToIgnore, line 173: I wonder, should we try fsyncing those in parallel after all download sync tasks are done, with par_fsync that we have?

Also, this `fs::rename`, should it be also somehow fsynced? 

Last but not least, we do sync `metadata` lately, one level above, in `update_local_metadata`.
Should we do both/one of/neither `fsync` and `*.tmp` move for it too?

pageserver/src/remote_storage/storage_sync/download.rs
```diff
@@ -149,6 +158,19 @@ where
...
+
+                // fsync parent dir
+                let parent_path = temp_file_path
+                    .parent()
+                    .expect("no parent dir for layer file");
+                let parent_dir = fs::File::open(parent_path).await?;
+                parent_dir.sync_all().await?;
+
+                fail::fail_point!("remote-storage-download-pre-rename", |_| {
+                    bail!("failpoint triggered")
```
SomeoneToIgnore, line 171: NIT: could be more specific on what point exactly was triggered, so somebody could grep it easily.

SomeoneToIgnore (commented):
That looks simpler than I've anticipated, great.
Not sure how to help with failpoints detection, I think it's ok for now.

pageserver/src/remote_storage/storage_sync/download.rs
```diff
@@ -131,10 +133,17 @@ where
...
-
-                let mut destination_file = fs::File::create(&layer_desination_path)
-                    .await
-                    .with_context(|| {
+                // we perform a rename inspired by durarble_rename from file_utils.c
+                // the differnce is that we sync parent dir once before rename.
+                // For us it is not critical if rename will be rolled back. In such case
+                // file will be deleted on startup and redownloaded.
+                // For more context about durable_rename check this email from postgres mailing list:
+                // https://www.postgresql.org/message-id/56583BDD.9060302@2ndquadrant.com
```
hlinnaka, line 141:
```suggestion
                // Perform a rename inspired by durable_rename from file_utils.c.
                // The difference is that we sync parent dir once, before the rename.
                // For us it is not critical if the rename is rolled back. In such case,
                // the file will be deleted on startup and re-downloaded.
                // For more context about durable_rename, see this thread on pgsql-hackers:
                // https://www.postgresql.org/message-id/56583BDD.9060302@2ndquadrant.com
```

hlinnaka: If it's OK to lose the file, why do we bother fsyncing the directory at all? It seems pretty pointless to fsync it before the rename.

hlinnaka:
>     * We can add a new config variable and wrap each failpoint in a conditional around it so even if binary is compiled with failpoint support it will check for config first and then dive into failpoint crate internals.

Yeah. I'm surprised 'fail-rs' doesn't do that itself.

> Personally, I'm for the simplest option :) We do not use failpoint in hot places, and we'll see it in flame graphs if it'll become noticeable. Thoughts?

+1

pageserver/src/remote_storage/storage_sync/download.rs
LizardWizzard: Hmmm, you're right. Given the need for metadata, sync @SomeoneToIgnore pointed to above, I think proper sync after rename is needed here. For some time, I felt that we could avoid fsync for the parent directory because we can redownload the file, but if we don't fsync the dir, rename may rollback, but metadata update will be performed to match that layer. I shouldn't cause any harm because we'll still redownload the file and overwrite metadata after that again. But it is a bit scary to have metadata with data that does not match files available on disk. What do you think?

I had a quick chat with @antons-antons touching that so would be glad to get some feedback

hlinnaka: No objections to fsyncing the directory too. We can revisit later if it becomes a performance problem, but better to err on the safe side.

So it would be:
1. fsync the file with old name
2. rename
3. fsync the file with new name
4. fsync parent directory

Right?

LizardWizzard: From the description [here](https://lwn.net/Articles/457667/) the sequence is that: 
1. create a new temp file (on the same file system!)
2. write data to the temp file
3. fsync() the temp file
4. rename the temp file to the appropriate name
5. fsync() the containing directory

So the only difference with your proposal is the absence of fsync on a file after rename. Any downsides to that approach?

hlinnaka: The [thread on pgsql-hackers](https://www.postgresql.org/message-id/flat/56583BDD.9060302%402ndquadrant.com) concluded with the fsync() on the file, too. I'm not sure why, I don't see that point being mentioned explicitly, everyone seems to just assume that it's needed.

Based on the LWN article I think you're right that it's not needed. But why take the risk? An extra fsync() is very cheap when the file has already been flushed to disk. Better safe than sorry.

LizardWizzard: Agreed, I've changed the code to match. Thanks!

pageserver/src/remote_storage/storage_sync.rs
```diff
@@ -852,18 +864,28 @@ async fn update_local_metadata(
...
-                    local_metadata_path.display()
-                )
-            })?;
+        // clone because spawn_blocking requires static lifetime
+        let cloned_metadata = remote_metadata.to_owned();
+        let ZTenantTimelineId {
+            tenant_id,
+            timeline_id,
+        } = sync_id;
+        tokio::task::spawn_blocking(move || {
```
LizardWizzard, line 873: @SomeoneToIgnore what do you think about moving metadata update to layered repo's apply_sync_status_update?

SomeoneToIgnore (approved):
Spotted a few unnecessary nits, but looks nice now.

pageserver/src/remote_storage/storage_sync.rs
```diff
@@ -1062,7 +1084,7 @@ where
                 debug!("Successfully fetched index part for {id}");
                 index_parts.insert(id, index_part);
             }
-            Err(e) => warn!("Failed to fetch index part for {id}: {e:?}"),
```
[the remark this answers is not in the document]
LizardWizzard: If the timeline has no checkpoints it is ok to have an error there (in case of nothing was uploaded yet). While debugging I spent some time figuring out the reason for this scary backtrace and it was just that :) Since there is no checkpoint for initial timeline every time pageserver starts after init there will be a backtrace

pageserver/Cargo.toml
```diff
@@ -4,8 +4,12 @@ version = "0.1.0"
 edition = "2021"
 
 [features]
-default = []
+# It is simpler infra-wise to have failpoints enabled by default
+# It shouldnt affect perf in any way because failpoints 
+# are not placed in hot code paths
+default = ["failpoints"]
 profiling = ["pprof"]
+failpoints = ["fail/failpoints"]
```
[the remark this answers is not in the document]
LizardWizzard: It is in default features. I think it should be included by default. How can I check that?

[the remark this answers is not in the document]
SomeoneToIgnore: Sorry, have missed the default features. Never mind me then.

pageserver/src/remote_storage/storage_sync.rs
LizardWizzard: It started to take longer than I expected, I'll do that separately

Issue #932, opened 2021-11-22
safekeeper panics in find_end_of_wal_segment


petuhovskiy:
Happened in test_restarts_under_load running in [CI job 28426](https://app.circleci.com/pipelines/github/zenithdb/zenith/3386/workflows/087b05fe-8d7a-42c9-8e4c-e3d289afcdec/jobs/28426/tests#failed-test-0). One of the safekeepers panicked with `'attempt to subtract with overflow'`, postgres_ffi/src/xlog_utils.rs:198:23.

<details>
<summary>Full stacktrace from sk2/safekeeper.log</summary>

```
2021-11-19T14:38:34.624712Z  INFO creating timeline dir 19e0a56a95188c69230e13f68ea0a509, create is True
2021-11-19T14:38:34.624835Z  INFO loading control file ./19e0a56a95188c69230e13f68ea0a509/safekeeper.control, create=True lock file "./19e0a56a95188c69230e13f68ea0a509/safekeeper.lock"
thread 'WAL service thread' panicked at 'attempt to subtract with overflow', postgres_ffi/src/xlog_utils.rs:198:23
stack backtrace:
   0: rust_begin_unwind
             at /rustc/c8dfcfe046a7680554bf4eb612bad840e7631c4b/library/std/src/panicking.rs:515:5
   1: core::panicking::panic_fmt
             at /rustc/c8dfcfe046a7680554bf4eb612bad840e7631c4b/library/core/src/panicking.rs:92:14
   2: core::panicking::panic
             at /rustc/c8dfcfe046a7680554bf4eb612bad840e7631c4b/library/core/src/panicking.rs:50:5
   3: postgres_ffi::xlog_utils::find_end_of_wal_segment
             at /home/circleci/project/postgres_ffi/src/xlog_utils.rs:198:23
   4: postgres_ffi::xlog_utils::find_end_of_wal
             at /home/circleci/project/postgres_ffi/src/xlog_utils.rs:306:25
   5: walkeeper::timeline::SharedState::create_restore
             at /home/circleci/project/walkeeper/src/timeline.rs:141:13
   6: walkeeper::timeline::GlobalTimelines::get
             at /home/circleci/project/walkeeper/src/timeline.rs:426:36
   7: <core::option::Option<alloc::sync::Arc<walkeeper::timeline::Timeline>> as walkeeper::timeline::TimelineTools>::set
             at /home/circleci/project/walkeeper/src/timeline.rs:389:22
[...]

PR #1574, opened 2022-04-25
Test `find_end_of_wal_segment` and fix its contrecord handling

yeputons:
Related to #544: fixes some troubles with handling WAL records that cross a segment boundary.

This PR starts fixing them:

* Add a separate library/binary for carefully crafting weird WALs using a Postgres server. Right now it can craft three kinds of WAL: simple, one ending in a record that crosses segment boundary, and a similar one with a small message appended.
  * The latter two were not handled correctly, and prior to #1280 (or somewhere around), it was even possible to crash Safekeeper.  It is harder now because it starts `find_end_of_wal` from commit lsn.
* Add lots of tracing to `find_end_of_wal_segment`. It's called on Safekeeper startup only, so should not yield a big performance penalty.
* Fix `find_end_of_wal_segment`'s bug: it worked incorrectly when the first record in the last segment is a contrecord. The old logic was to crash, new logic is to skip this record and find at least one full valid record in the segment.

I think it's better to get the barebones in first, but there are some todos left:

* Handle corner cases in `find_end_of_wal_segment` marked as `TODO` in this PR. I believe they never happen in our tests.
* Backtrack if there is no full valid record in the last segment file. This test is currently `#[ignore]`d.


libs/postgres_ffi/src/xlog_utils.rs
```diff
@@ -218,42 +259,96 @@ fn find_end_of_wal_segment(
...
+                trace!(
+                    "  reading rec_hdr[{}..{}] <-- [0x{:x}; 0x{:x})",
+                    rec_offs,
+                    rec_offs + len,
+                    page_offs,
+                    page_offs + len
+                );
                 rec_hdr[rec_offs..rec_offs + len].copy_from_slice(&buf[page_offs..page_offs + len]);
             }
+            // TODO: is it possible that the header is split between pages?
```
knizhnik, line 286: Certainly it is possible, then header is accumulated in `rec_hdr` buffer

libs/postgres_ffi/src/xlog_utils.rs
```diff
@@ -218,42 +259,96 @@ fn find_end_of_wal_segment(
...
             if rec_offs <= XLOG_RECORD_CRC_OFFS && rec_offs + n >= XLOG_SIZE_OF_XLOG_RECORD {
+                // TODO: what if CRC field is on the previous page? Read from rec_hdr, probably.
```
knizhnik, line 288: CRC can not be split between pages because WAL records is 8 byte aligned. So it is either at one page or at another. In any cae we calculate CRC when proceeding this page and CRC is in the available rage.

knizhnik (approved):

libs/postgres_ffi/src/xlog_utils.rs
```diff
@@ -154,9 +155,11 @@ fn find_end_of_wal_segment(
     file.seek(SeekFrom::Start(offs as u64))?;
     let mut rec_hdr = [0u8; XLOG_RECORD_CRC_OFFS];
 
+    trace!("find_end_of_wal_segment(data_dir={}, segno={}, tli={}, wal_seg_size={}, start_offset=0x{:x})", data_dir.display(), segno, tli, wal_seg_size, start_offset);
```
antons-antons, line 158: Should we validate here that 
1. start_offset is < wal_seg_size?
2. wal_seg_size is a power of 2 and between 1Mb and 1Gb?

libs/postgres_ffi/src/xlog_utils.rs
```diff
@@ -170,30 +173,53 @@ fn find_end_of_wal_segment(
...
+                            warn!("trying to find_end_of_wal with start_offset ({}) in the middle of the first contrecord (xlp_rem_len={})", start_offset, xlp_rem_len);
+                            contlen -= start_offset - offs;
+                        }
+                    }
+                } else {
+                    trace!("  first record is not contrecord");
                 }
             } else {
                 offs += XLOG_SIZE_OF_XLOG_SHORT_PHD;
             }
```
antons-antons, line 214: we're handling the first page in the segment with XLP_FIRST_IS_CONTRECORD while it's possible that a page in the middle of the segment with this flag, should we handle it?

libs/postgres_ffi/src/xlog_utils.rs
```diff
@@ -170,30 +173,53 @@ fn find_end_of_wal_segment(
...
             // ... and step forward again if asked
+            trace!("  skipped header to 0x{:x}", offs);
             offs = max(offs, start_offset);
-
         // beginning of the next record
         } else if contlen == 0 {
             let page_offs = offs % XLOG_BLCKSZ;
```
antons-antons, line 220: I find this function has unnecessary nesting, do you think we can simplify this?

libs/postgres_ffi/src/xlog_utils.rs
```diff
@@ -170,30 +173,53 @@ fn find_end_of_wal_segment(
...
             let xl_tot_len = LittleEndian::read_u32(&buf[page_offs..page_offs + 4]) as usize;
+            trace!("offs=0x{:x}: new record, xl_tot_len={}", offs, xl_tot_len);
             if xl_tot_len == 0 {
```
antons-antons, line 223: I think we should handle the case of xl_tot_len < XLOG_SIZE_OF_XLOG_RECORD and at least warn about this.

libs/postgres_ffi/src/xlog_utils.rs
```diff
@@ -218,42 +259,96 @@ fn find_end_of_wal_segment(
...
+                trace!(
+                    "  padded offs to 0x{:x}, crc is {:x}, expected crc is {:x}",
+                    offs,
+                    crc,
+                    xl_crc
+                );
+                if skipping_first_contrecord {
+                    // do nothing, the flag will go down on next iteration
+                    trace!("  first conrecord has been just completed");
+                } else if crc == xl_crc {
```
antons-antons, line 332: In addition to the crc validation we can check if xl_prev points to last_valid_rec_pos (with the only exception of the first log record received by SK)

antons-antons (commented):
I was looking at https://github.com/neondatabase/neon/issues/932 and had a similar change to yours, but you have beaten me to the CR and have tests, wanna link the issue with this PR?

libs/postgres_ffi/wal_generate/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+    ensure!(
+        message_lsn < after_message_lsn,
+        "No record found after the emitted message"
+    );
+
+    ensure!(
+        after_message_lsn == client.pg_current_wal_flush_lsn()?,
+        "WAL is either not flushed or is extended with something unknown"
+    );
+    Ok(after_message_lsn)
```
arssher, line 215: And where is small message following crossing record one?

libs/postgres_ffi/wal_generate/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+    // First few created tables take more WAL bytes than later.
+    info!("LSN initial = {}", client.pg_current_wal_insert_lsn()?);
+    client.execute("create table t_base_1 (x int)", &[])?;
+    client.execute("create table t_base_2 (x int)", &[])?;
+
+    // We want to create ~400 tables: ~4300 bytes in WAL per each, ~353 bytes in the COMMIT message per each.
+    // Commit message will take approx 137 KiB and will start ~24 KiB before segment's end.
+    let start_creating_at_lsn = PgLsn::from(0x0200_0000 - 8192 * 3 - 4300 * 400);
+    let stop_creating_at_lsn = PgLsn::from(0x0200_0000 - 8192 * 3);
+    let expect_commit_end_after_lsn = PgLsn::from(0x0200_0000);
```
arssher, line 230: Why it is so complicated, what is the problem with single large record crossing the segment?

arssher (changes requested):

yeputons:
Removing complex WAL generation logic from 891d0523613750181a7cc916da6fae9a828daf8e in favor of a simpler strategy: emit necessary logical message, flush it to the disk via `neon_xlogflush` from https://github.com/neondatabase/postgres/pull/160, kill the server. No `COMMIT` record.

Disabling autovacuum just in case on Anton's suggestion. Maybe some other stuff needs to be disabled as well.

libs/postgres_ffi/wal_generate/src/lib.rs
yeputons: It's the implicit COMMIT record.

libs/postgres_ffi/wal_generate/src/lib.rs
yeputons: The original problem was that there was no way to ensure WAL is flushed to the disk. No longer a problem with `XLogFlush` exposed as a SQL function

yeputons:
Will rebase 8e5562bf2c2f419482bd75cfda995f03568823db onto `main` and squash all commits, so it's ready to merge after review.

libs/utils/src/lib.rs
```diff
@@ -95,3 +95,11 @@ macro_rules! project_git_version {
...
     };
 }
+
+/// Same as `assert!`, but evaluated during compilation and gets optimized out in runtime.
+#[macro_export]
+macro_rules! const_assert {
+    ($($args:tt)*) => {
+        const _: () = assert!($($args)*);
+    };
+}
```
yeputons, line 105: There is a popular `static_assertions` crate, but it [wasn't updated for a while](https://github.com/nvzqz/static-assertions-rs/issues/43).

libs/postgres_ffi/src/xlog_utils.rs
```diff
@@ -224,42 +264,103 @@ fn find_end_of_wal_segment(
...
+                trace!(
+                    "  reading rec_hdr[{}..{}] <-- [0x{:x}; 0x{:x})",
+                    rec_offs,
+                    rec_offs + len,
+                    page_offs,
+                    page_offs + len
+                );
                 rec_hdr[rec_offs..rec_offs + len].copy_from_slice(&buf[page_offs..page_offs + len]);
             }
+            // TODO: handle `rec_offs + n < XLOG_SIZE_OF_XLOG_RECORD`: the header is not completely read.
```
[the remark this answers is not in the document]
knizhnik: Sorry, I do not understand the problem.
Yes, record head can be cut at any 8-byte word. Only 8-byte alignment is enforced.
So it can happen that record header starts at one page, but CRC is at another page.

But this accumulation of page header in  `rec_hdr` is used to address this case. CRC field can not be splitted between two different pages, because of 8-byte alignment. So the following check `rec_offs <= XLOG_RECORD_CRC_OFFS && rec_offs + n >= XLOG_SIZE_OF_XLOG_RECORD` should catch moment when we have CRC field in the buffer.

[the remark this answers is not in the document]
yeputons: It will be accumulated in `rec_hdr`, but it will also update `crc` with a piece of header in the `else`-branch below.

Ah, I see, it does not matter as we reset `crc` back to zero in the `if` branch once we have read the header.

yeputons:
Wow, it's a real flaky failure introduced (or reproduced?) by this PR! Investigating. Does not reproduce locally.

```
2022-05-20T17:11:33.180580Z ERROR {tid=8}: query handler for 'START_WAL_PUSH postgresql://no_user:@localhost:18106' failed: failed to restore shared state

Caused by:
    Condition failed: `rec_offs + n <= XLOG_RECORD_CRC_OFFS` (23 vs 20)

Stack backtrace:
   0: anyhow::ensure::render
             at /home/circleci/.cargo/registry/src/github.com-1ecc6299db9ec823/anyhow-1.0.53/src/ensure.rs:97:20
   1: <(A,B) as anyhow::ensure::BothDebug>::__dispatch_ensure
             at /home/circleci/.cargo/registry/src/github.com-1ecc6299db9ec823/anyhow-1.0.53/src/ensure.rs:20:9
      postgres_ffi::xlog_utils::find_end_of_wal_segment
             at /home/circleci/project/libs/postgres_ffi/src/xlog_utils.rs:331:17
      postgres_ffi::xlog_utils::find_end_of_wal
             at /home/circleci/project/libs/postgres_ffi/src/xlog_utils.rs:439:25
   2: <safekeeper::wal_storage::PhysicalStorage as safekeeper::wal_storage::Storage>::init_storage
             at /home/circleci/project/safekeeper/src/wal_storage.rs:329:17
   3: safekeeper::safekeeper::SafeKeeper<CTRL,WAL>::new
             at /home/circleci/project/safekeeper/src/safekeeper.rs:554:9
   4: safekeeper::timeline::SharedState::restore
             at /home/circleci/project/safekeeper/src/timeline.rs:130:17
      safekeeper::timeline::GlobalTimelines::get
             at /home/circleci/project/safekeeper/src/timeline.rs:598:21
   5: <core::option::Option<alloc::sync::Arc<safekeeper::timeline::Timeline>> as safekeeper::timeline::TimelineTools>::set
             at /home/circleci/project/safekeeper/src/timeline.rs:513:22
      <safekeeper::handler::SafekeeperPostgresHandler as utils::postgres_backend::Handler>::process_query
             at /home/circleci/project/safekeeper/src/handler.rs:108:13
   6: utils::postgres_backend::PostgresBackend::process_message
             at /home/circleci/project/libs/utils/src/postgres_backend.rs:431:33
   7: utils::postgres_backend::PostgresBackend::run_message_loop
             at /home/circleci/project/libs/utils/src/postgres_backend.rs:283:31
      utils::postgres_backend::PostgresBackend::run
             at /home/circleci/project/libs/utils/src/postgres_backend.rs:265:19
   8: safekeeper::wal_service::handle_socket
             at /home/circleci/project/safekeeper/src/wal_service.rs:55:5
   9: safekeeper::wal_service::thread_main::{{closure}}
             at /home/circleci/project/safekeeper/src/wal_service.rs:26:43
      std::sys_common::backtrace::__rust_begin_short_backtrace
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/std/src/sys_common/backtrace.rs:123:18
  10: std::thread::Builder::spawn_unchecked::{{closure}}::{{closure}}
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/std/src/thread/mod.rs:484:17
      <core::panic::unwind_safe::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/panic/unwind_safe.rs:271:9
      std::panicking::try::do_call
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/std/src/panicking.rs:406:40
      std::panicking::try
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/std/src/panicking.rs:370:19
      std::panic::catch_unwind
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/std/src/panic.rs:133:14
      std::thread::Builder::spawn_unchecked::{{closure}}
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/std/src/thread/mod.rs:483:30
      core::ops::function::FnOnce::call_once{{vtable.shim}}
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/ops/function.rs:227:5
  11: <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/alloc/src/boxed.rs:1694:9
      <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/alloc/src/boxed.rs:1694:9
      std::sys::unix::thread::Thread::new::thread_start
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/std/src/sys/unix/thread.rs:106:17
  12: start_thread
  13: clone
2022-05-20T17:11:33.444770Z  INFO Got SIGQUIT. Terminating in immediate shutdown mode
```

yeputons:
Gosh, I love programming. No sarcasm.

1. Unable to reproduce locally.
2. Re-running the test in CI made it pass, so it's flaky, and I got fortunate it failed the first time. Another test failed, though: `test_replace_safekeeper`, and for the same reason. Re-running once got it everything pass.
3. Compiling with `--release` and running the test multiple times got me the local reproduction.

So, the corner case: if a record gets split between two segments, `find_end_of_wal_segment` does not care and still reads the second half of the record as if it was a correct record with a header, CRC, and stuff, but just does not check it. However, that behavior changed in my last commit: I ensured that the record never stops inside the `xl_crc` field. That would be correct for if we were reading a valid record, but we are reading a tail of a record. Hence the failure.

Nice coincidence, IMHO:

1. A record got split between two pages; and
2. Its second half has a length of either 21, 22, or 23 bytes exactly; and
3. It got caught by a completely unrelated test in or CI.

Fast fix: check less when skipping the first contrecord. A probably cleaner one would to clearly specify the current state of `find_end_of_wal_segment` (e.g. "skipping the first cont record", "reading the header", "reading the record body"), but I'd leave it for a separate PR, see https://github.com/neondatabase/neon/issues/544#issuecomment-1133043210

arssher (approved):
IOW, rec_offs is unknown in case of record split between segments.

PR #1575, opened 2022-04-25
Use item.get for tenant config toml parsing

LizardWizzard:
Previously we've used table interface, but there was no easy way to pass
it as an override to pageserver through cli. Use the same strategy as
for remote storage config parsing


dhammika (approved):
+1, tnx for checking this.

lubennikovaav (approved):
Thank you for fixing!

Issue #1561, opened 2022-04-21
Track lowest locally available LSN in safekeeper


petuhovskiy:
Not sure how this LSN should be called, some options:
- `disk_start_lsn`
- `local_start_lsn`

We can also track first-ever LSN was broadcasted from compute to sk (in other words LSN after initdb), I'd call it `origin_lsn`. Both LSNs should be stored inside control file. `local_start_lsn` should be updated when WAL is trimmed.

The next step after adding `local_start_lsn` will be to add more checks with it, for example don't allow streaming below this LSN.

petuhovskiy:
Resolved in https://github.com/neondatabase/neon/pull/1602

PR #1583, opened 2022-04-27
WIP: SK WAL Backup

antons-antons:
what's working
1. etcd leader election
2. signals from PhysicalStorage to WalBackup

what's not working:
1. control file and etcd metrics maintanance
2. proper integration with segment file names (incl. segment size)
3. pickup of missing segments on timeline restore
4. logging is verbose in some places, needs to be tamed
5. S3 related code needs some TLC (it had gone through a cosmetic changes and with tests removed due to PS assumptions)

Need help with:
1. Proper factory function to created RemoteStorage based on configuration (in cpp words I need a pointer to an interface, not the subclass, how to do it, see remote_storage.rs pub fn create<S, P>)
2. An idea on how to store a refference to a RemoteStorage in WalBackup (something has to do with lifetime, no idea how to code around that) 
4. Ideas for a mechanism on how to communicate local flush lsn (wait_for_lsn ?) without taking a hard dependency on Timeline. maybe channel with peek option would work here?
5. How to test such code in Rust (literally no idea on how unit tests)


safekeeper/src/wal_backup.rs
```diff
@@ -0,0 +1,218 @@
...
+    conf: SafeKeeperConf,
+
+    backup_lsn : Lsn,
+
+    // etcd lease id, used for leader election
+    lease_id : Option<i64>,
+    timeline_id: ZTenantTimelineId, 
+}
+
+async fn detect_task(
```
SomeoneToIgnore, line 53: Can add some `.instrument(info_span!(...` calls around the main sync loop: we can add fields into this span, such as timeline_id to automatically log every entry inside the span with this id.

This way, we can distinguish sync attempts in the logs.

safekeeper/src/timeline.rs
```diff
@@ -101,7 +103,14 @@ impl SharedState {
     ) -> Result<Self> {
         let state = SafeKeeperState::new(zttid, peer_ids);
         let control_store = control_file::FileStorage::create_new(zttid, conf, state)?;
-        let wal_store = wal_storage::PhysicalStorage::new(zttid, conf);
+
+        let segment_complete = watch::channel((Lsn::MAX, Lsn::MAX));
+        let lsn_persisted = watch::channel(Lsn::MAX);
```
SomeoneToIgnore, line 108: Feels like `channel(Lsn(0))` is a more sane default choice for what was persisted?

safekeeper/src/wal_backup.rs
```diff
@@ -0,0 +1,218 @@
...
+
+use tracing::*;
+
+use crate::remote_storage::{LocalFs, RemoteStorage};
+use crate::{broker, SafeKeeperConf};
+
+lazy_static! {
+    static ref BACKUP_RUNTIME: Runtime = {
+        Builder::new_multi_thread()
+            .worker_threads(16)
```
SomeoneToIgnore, line 24: `16` looks magical, esp. considering that we're good with a bunch of separate single threaded runtimes in general: all `tokio::fs` operations spawn separate threads to deal with the i/o concurrently, the rest of the operations seem to be non-blocking, so we might just use a single threaded runtime for experiments for now.

Or add a configurable value. Or a comment at least.

safekeeper/src/wal_backup.rs
```diff
@@ -0,0 +1,218 @@
...
+            None
+        };
+
+        let backup = Self {_remote_storage : &remote_storage, backup_lsn : Lsn::MAX , lease_id: lease, timeline_id: *timeline_id, conf: conf};
+
+        if let Some(lease_id) = lease {
+            // TODO: antons how to do it in this function???
+            // ensure!( conf.broker_endpoints.is_none(););
+
+            BACKUP_RUNTIME.spawn(broker::lease_keep_alive(lease_id, conf.clone()));
```
SomeoneToIgnore, line 109: We would need a cancellation mechanism here, similar to what's in the [similar place](https://github.com/neondatabase/neon/blob/778744d35ca4ff57237a6ef5b4323084797de9bd/pageserver/src/remote_storage/storage_sync.rs#L546) of pageserver.

What's worse is the fact that every `BACKUP_RUNTIME.spawn` returns a join handle, but we don't use that to wait for the task's completion, rather leaving it spinning in the runtime implicitly.

This means that the futures will be abruptly cancelled when the runtime is dropped only, abruptly terminating the upload streaming with that drop.
Should not be a big issue with S3 guarantees, but feels rather odd to see in general.

safekeeper/src/timeline.rs
```diff
@@ -101,7 +103,14 @@ impl SharedState {
...
+        let wal_store = wal_storage::PhysicalStorage::new(zttid, conf, lsn_persisted.0, segment_complete.0);
+
+        let remote_storage = remote_storage::create(conf.remote_storage_config.as_ref(), conf.timeline_dir(zttid))?;
```
SomeoneToIgnore, line 111: Current remote storage limits s3 requests using an internal semaphore, so multiple remote storages would have their own, independent limits.
This is odd and I think there should be only one `Arc<RemoteStorage>` stored somewhere on a top level (safekeper main file?) and shared among the rest of the code.
Every method in the `RemoteStorage` requires `&self` so it's fine to be used within `Arc`s.

Also, for pageserver's LocalStorage at least, we've intended its root to be the directory with the `tenants` folder.

SomeoneToIgnore (changes requested):
Looks quite raw and could take a few more iterations before being submitted as a draft PR IMO (I'd close this and take it all offline), but good to start some discussion, maybe somebody else can help to move this forward.
Please consider asking about Safekeeper approaches from somebody else though, so I can only give specific recommendations on S3 integration in general.

I'm not sure that I've spotted all issues, neither is sure which comments of mine have the most priority here.
I feel that the leader selection is odd and worth a commentary at least, if works as intended + the sync loop handling would need a redesign, hopefully hidden behind some struct with a state and proper future termination/cancellation handling.

---------------------

> Need help with:

> Proper factory function to created RemoteStorage based on configuration (in cpp words I need a pointer to an interface, not the subclass, how to do it, see remote_storage.rs pub fn create<S, P>)

See the init code around the code pointers for the next answer.

> An idea on how to store a refference to a RemoteStorage in WalBackup (something has to do with lifetime, no idea how to code around that)

Two ways:
* `Arc` and generics in methods, same way as it's initialized and passed around in pageserver's `spawn_storage_sync_thread` in `storage_sync.rs`: https://github.com/neondatabase/neon/blob/3e6087a12f26ebefe6b91ea78be5d927c72b2a48/pageserver/src/remote_storage/storage_sync.rs#L472
* Add an enum for a hardcoded static dispatch: https://github.com/neondatabase/neon/blob/3e6087a12f26ebefe6b91ea78be5d927c72b2a48/pageserver/src/http/routes.rs#L40-L43

> Ideas for a mechanism on how to communicate local flush lsn (wait_for_lsn ?) without taking a hard dependency on Timeline. maybe channel with peek option would work here?

Not sure what's the issue here, but I'd start with a proper sync design first, then it might be a bit more obvious. So far there are definitely some RAII holes and maybe too many static, non-struct functions.

> How to test such code in Rust (literally no idea on how unit tests)

I think it's better to create a python tests for that, there's a remote storage test in pageserver already: https://github.com/neondatabase/neon/blob/3e6087a12f26ebefe6b91ea78be5d927c72b2a48/test_runner/batch_others/test_remote_storage.py

You would need to tweak the code to adjust safekeeper init config to enable the remote storage there and think on how to reproduce it via some actions with safekeeper

Otherwise, creating a unit test in rust is a simple as adding somewhere to the bottom of the file something like

```rust
#[cfg(test)]
mod pg_helpers_tests {

    use std::fs::File;

    use compute_tools::pg_helpers::*;
    use compute_tools::zenith::ClusterSpec;

    #[test]
    fn params_serialize() {
        let file = File::open("tests/cluster_spec.json").unwrap();
        let spec: ClusterSpec = serde_json::from_reader(file).unwrap();

        assert_eq!(
            spec.cluster.databases.first().unwrap().to_pg_options(),
            "LC_COLLATE 'C' LC_CTYPE 'C' TEMPLATE template0 OWNER \"alexk\""
        );
        assert_eq!(
            spec.cluster.roles.first().unwrap().to_pg_options(),
            "LOGIN PASSWORD 'md56b1d16b78004bbd51fa06af9eda75972'"
        );
    }
}
```

you have to think on how to init your code for testing yourself, so consider adding some unit tests on the core logic here, but leave the rest for the python test.

petuhovskiy (commented):
Can't understand some of the code, maybe it's unfinished, so I'll try again sometime later. 

In my understanding the whole process should look like this:
1. list directory with segments, get segment names from lowest to highest
2. check if segment is uploaded to S3 one by one, until non-uploaded segment is found (or call List in S3 and compare to local files)
3. repeat the upload until everything is uploaded
4. wait for commit_lsn updates to fully cover new segment and go to previous step

And there can be some optimizations, which are not strictly necessary, but improve overall:
- take an exclusive lock on S3 upload for the current timeline (not necessary, IMO it'll be good for upload to work even without coordination)
- store last uploaded segment in some shared storage (etcd), to optimize fs/s3 calls
- trim old WAL segments on all safekeepers once they're uploaded to S3

safekeeper/src/config.rs
```diff
@@ -0,0 +1,67 @@
...
+    /// The storage connection configuration.
+    pub storage: RemoteStorageKind,
+}
+
+/// A kind of a remote storage to connect to, with its connection configuration.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum RemoteStorageKind {
+    /// Storage based on local file system.
+    /// Specify a root folder to place all stored files into.
+    LocalFs(PathBuf),
```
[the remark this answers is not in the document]
LizardWizzard: Can be useful for testing, that's the same way how it is done for pageserver

safekeeper/src/wal_backup.rs
LizardWizzard: Actually that's a bit scary and is valid for pagesever, too, if for some reason the uploader fails (panics etc.) but lease keepalives are still sent no one will be able to step in and replace the failed uploader

antons-antons:
Addressed some of the comments.

missing:
* correct integration with lease keepalive
* communication of wal_segment size 
* tests

safekeeper/src/wal_storage.rs
```diff
@@ -247,6 +256,10 @@ impl PhysicalStorage {
             let (wal_file_path, wal_file_partial_path) =
                 wal_file_paths(&self.timeline_dir, segno, wal_seg_size)?;
             fs::rename(&wal_file_partial_path, &wal_file_path)?;
+
+            let start_lsn = XLogSegNoOffsetToRecPtr(segno, 0, wal_seg_size);
+            let end_lsn = XLogSegNoOffsetToRecPtr(segno + 1, 0, wal_seg_size);
+            self.notify_segment_complete(Lsn(start_lsn), Lsn(end_lsn))?;
```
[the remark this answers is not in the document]
antons-antons: addressed

safekeeper/src/timeline.rs
antons-antons: re. semaphore: as discussed via Slack, throttling implementation with semaphore is probably not optimal. This is exacerbated by the fact we're throttled by IAM while we access S3 (IAM control plane is in USE1, which poses an availability issue in addition to much lower API request limits as compared to S3).

safekeeper/src/wal_backup.rs
antons-antons: agree, the cancellation story isn't sound, at the same time backup runtime is supposed to live for the whole duration of SK process life (new implementation makes it lazily initialized).
I'd like to address cancellation (incl. timeline drop) and SK replacement handling separately with the addition of time-based upload (in addition to LSN range). This would address keep-alive independence.

safekeeper/src/wal_backup.rs
antons-antons: addressed

safekeeper/src/safekeeper.rs
```diff
@@ -153,7 +155,7 @@ pub struct PeerInfo {
 impl PeerInfo {
     fn new() -> Self {
         Self {
-            s3_wal_lsn: Lsn(0),
+            backup_lsn: Lsn::INVALID,
```
SomeoneToIgnore, line 158: Just interested: are those `Lsn(0)` below valid or also invalid, but just not changed?
Might be a bit confusing to read and even the former is true (Lsn(0) and Lsn::Invalid are different by the meaning), then we'd rather express it with an enum instead?

safekeeper/src/wal_backup.rs
```diff
@@ -0,0 +1,267 @@
...
+                    break;
+                }
+            }
+
+            if commit_lsn < backup_lsn {
+                bail!("Commit lsn {} moved back {}", commit_lsn, backup_lsn);
+            }
+
+            if backup_lsn.segment_number(wal_seg_size) == commit_lsn.segment_number(wal_seg_size) {
+                continue;
```
SomeoneToIgnore, line 74: Curious: should we check that we have something to backup before contending for leadership?

I would also keep a single client all the time, without creating and connecting every time I need to adjust the leadership: I wonder, if dealing with leadership even supposed to used this way with reconnects?

safekeeper/src/wal_backup.rs
```diff
@@ -0,0 +1,267 @@
...
+
+    Ok(res)
+}
+
+async fn backup_single_segment(seg: Segment, wal_seg_size: usize, conf: &SafeKeeperConf, timeline_id: ZTenantTimelineId) -> Result<()> {
+
+    let segment_file_name = seg.file_path(&timeline_id, conf)?;
+    let dest_name = PathBuf::from(format!("{}/{}", timeline_id.tenant_id, timeline_id.timeline_id)).with_file_name(seg.file_name());
+
+    info!("Backup of {:?} requested", segment_file_name);
```
SomeoneToIgnore, line 114: Styling nit:

```suggestion
    info!("Backup of {} requested", segment_file_name.display());
```

A bunch more plaves can have this changed.

safekeeper/src/wal_backup.rs
SomeoneToIgnore: Yes, now that I realise that we have `lease_keep_alive` with another client polling the whole lease liveness, I wonder how this all is supposed to work.

I would create the client here, polling for both the lease, leadership and new events in the same loop, cancelling the loop if no lease is granted, doing nothing if not a leader, etc?

safekeeper/src/wal_backup.rs
SomeoneToIgnore: Should we consider two different cancellation parts?

* "reasonable" bits such as reacting to SIGINT, being able to "shut down a tenant", etc. Not sure how it all is relevant for safekeeper, but we seem to not to have any means to support this in the new code
* sound cancellation with handling all edge cases, panics, etc.

The latter is out of scope for the current early version for sure, but what about regular termination commands?

safekeeper/src/wal_backup.rs
```diff
@@ -0,0 +1,267 @@
...
+    mut wal_seg_size: Receiver<u32>,
+    lease_id : Option<i64>, 
+    mut lsn_durable: Receiver<Lsn>, 
+    lsn_backed_up: Sender<Lsn>,
+    ) -> Result<()> {
+
+    // TODO: antons come up with better naming
+    wal_seg_size.changed().await.expect("Failed to recieve wal segment size");
+    let wal_seg_size = *wal_seg_size.borrow();
+    let mut wal_seg_size = usize::try_from(wal_seg_size).with_context(|| format!("Couldn't convert {} to usize", wal_seg_size))?;
```
SomeoneToIgnore, line 46: > ?

There are a lot of such places in the loop, that just abruptly return `Err(anyhow::Error)`, stopping the loop.
We don't check for that and never restart the loop: we drop the handle of this task spawned. Neither we check if the task spawned had panicked.

IMO, we should log such cases and restart the backup loop for every timeline. Some exponential backoff could be used to avoid odd network issues.

safekeeper/src/wal_backup.rs
```diff
@@ -0,0 +1,267 @@
...
+            backup_lsn = backup_lsn_range(backup_lsn, commit_lsn, wal_seg_size, &conf, timeline_id).await?;
+            lsn_backed_up.send(backup_lsn)?;
+        }
+    }
+}
+
+pub async fn backup_lsn_range(start_lsn: Lsn, end_lsn: Lsn, wal_seg_size: usize, conf: &SafeKeeperConf, timeline_id: ZTenantTimelineId) -> Result<Lsn> {
+
+    let mut res = start_lsn;
+    for s in get_segments(start_lsn, end_lsn, wal_seg_size) {
```
SomeoneToIgnore, line 88: > .await

Not familiar with safekeeper internals, unfortunately, but how many such segments do we expect to receive?
Now we process them "one by one", with one task at a time waiting on `.await` and the rest of the sync tasks just queuing.

We can use [`FuturesUnordered`](https://github.com/neondatabase/neon/blob/81cad6277a2666ca47b97848628ffeafd6bf6aba/pageserver/src/remote_storage/storage_sync/upload.rs#L156) or its ordered counterpart to make it more concurrent.

SomeoneToIgnore (changes requested):
A step forward, I'm glad to see that lease and leadership election gets forward: this is IMO the most important part of this PR.

safekeeper/src/safekeeper.rs
```diff
@@ -511,6 +512,10 @@ pub struct SafeKeeper<CTRL: control_file::Storage, WAL: wal_storage::Storage> {
     pub state: CTRL,               // persistent state storage
 
     pub wal_store: WAL,
+
+    lsn_committed: Sender<Lsn>,
+    lsn_backed_up: Receiver<Lsn>,
+    wal_segment_known: Sender<u32>,
```
petuhovskiy, line 518: I can suggest an alternative to adding channels inside SafeKeeper struct. These channels are used for communication with walbackup, and I think we can move this communication to `timeline.rs`. This will follow the current code structure, where we place everything that affects consensus inside `safekeeper.rs` and gluing components together inside `timeline.rs`. 

For example, `commit_lsn` updates when `timeline.rs` calls `process_msg` and then it gets `AppendResponse`. In that `AppendResponse` we have latest `commit_lsn`, so `timeline.rs` can notify `send_wal.rs` that `commit_lsn` has just updated (that happens inside `notify_wal_senders`). So we can remove `lsn_committed` channel and just modify existing `notify_wal_senders` to notify walbackup.

`lsn_backed_up` can be replaced with just calling some `SafeKeeper` method, or directly persisting update to `safekeeper.state`.

When we creating/restoring timeline, we usually already know `wal_seg_size`, because it's persisted to control file. If not, we can wait for Greeting message in `process_msg` inside `timeline.rs`. Once we know `wal_seg_size`, we can pass it to walbackup (or initialize walbackup at this moment, see below).

One way to initialize walbackup inside `timeline.rs` is to place `walbackup: Option<WalBackup>` inside `SharedState`, and create it only when we know `wal_seg_size`.

safekeeper/src/wal_backup.rs
```diff
@@ -0,0 +1,267 @@
...
+            .enable_all()
+            .build()
+            .expect("Couldn't create wal backup runtime")
+    });
+
+    let lease_id = conf.broker_endpoints
+        .as_ref()
+        .map(|_| {
+            let lease = broker::get_lease(conf);
+            runtime.spawn(broker::lease_keep_alive(lease, conf.clone()));
```
petuhovskiy, line 148: This lease seems not to be released once timeline is deactivated. We can create WalBackup struct which will "hold" the lease, and release it in the destructor.

safekeeper/src/wal_backup.rs
petuhovskiy: It's better to upload segments one by one, so we will have no holes in the uploaded range.

safekeeper/src/safekeeper.rs
petuhovskiy: `Lsn(0)` is always invalid. Enum sounds nice, but it's not much convenient to use in the code.

antons-antons: I would rewrite all of the LSN(0) as Lsn::INVALID _(and/or introduce Lsn::LOWEST_VALID where valid LSN smaller than anything Postgres Could write is semantically appropriate)_ but it will bloat the PR massively.
I'd say a dedicated pr for such change is a better idea.

safekeeper/src/wal_backup.rs
antons-antons: functionally the usage of a new client is okay for leadership validation is fine as the leader is associated with a lease (and once the lease expires a new candidate would become a leader if such a candidate exists). Currently caching a client makes code more complicated as the client needs to be mutable while participating in multiple activities. I would defer optimization of this code path until we get a better idea if we want to have general timeline leaders not backup specific.

_note: agree that flipping lsn check with leadership check is a freebie that would improve preformance_

safekeeper/src/wal_backup.rs
antons-antons: at this one-by-one should be fine, agree that parallel upload will be a desirable functionality once larger-than-free tier, but I would bundle that with time based upload

safekeeper/src/safekeeper.rs
```diff
@@ -746,9 +765,8 @@ where
     /// Persist in-memory state to the disk.
     fn persist_control_file(&mut self) -> Result<()> {
         let mut state = self.state.clone();
-
         state.commit_lsn = self.inmem.commit_lsn;
-        state.s3_wal_lsn = self.inmem.s3_wal_lsn;
+        state.backup_lsn = *self.lsn_backed_up.borrow();
```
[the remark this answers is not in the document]
antons-antons: agree we may upload a segment twice if crashed at the wrong time, will be addressed in https://github.com/neondatabase/neon/issues/1636

safekeeper/src/wal_storage.rs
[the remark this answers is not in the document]
antons-antons: fixed

safekeeper/src/wal_backup.rs
```diff
@@ -0,0 +1,218 @@
...
+            .worker_threads(16)
+            .enable_all()
+            .build()
+            .unwrap()
+    };
+}
+
+const BACKUP_ELECTION_PATH: &str = "WAL_BACKUP";
+
+// TODO: there's need to track last upload time to avoid losing partially filled segments
```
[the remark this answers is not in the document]
antons-antons: implemented handling for the new timelines, should be okay to ignore missing segments when backup_lsn is not initialized yet

safekeeper/src/broker.rs
```diff
@@ -84,6 +91,65 @@ fn timeline_safekeeper_path(zttid: &ZTenantTimelineId, sk_id: ZNodeId) -> String
...
+            .context("failed to receive LeaseKeepAliveResponse")
+            .unwrap();
+
+        sleep(push_interval).await;
+    }
+}
+
+/// create an election for a given timeline and wait to become a leader 
+pub async fn become_leader(lease_id: i64, election_name: String, timeline_id: &ZTenantTimelineId, conf: SafeKeeperConf) -> Result<()> {
+    let mut client = Client::connect(conf.broker_endpoints.as_ref().unwrap(), None).await?;
```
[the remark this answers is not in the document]
antons-antons: while keepalive is supposed to be addressed, leaving the on-demand client creation for now

safekeeper/src/remote_storage.rs
```diff
@@ -0,0 +1,166 @@
+/// This is a copy pasta from Page Server
```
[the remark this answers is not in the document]
antons-antons: yup, will integrate with https://github.com/neondatabase/neon/pull/1634 before finalizing this PR

safekeeper/src/wal_backup.rs
antons-antons: made a lousy attempt at this (at the detect task scheduling) but somehow can't log timeline there (something with macro expansion isn't right)

safekeeper/src/wal_backup.rs
antons-antons: added ability to cancel WAL backup, but haven't wired it to [de]activation.
spilling this functionality over into https://github.com/neondatabase/neon/issues/1644

safekeeper/src/wal_backup.rs
antons-antons: please check if the latest looks any better

safekeeper/src/wal_backup.rs
antons-antons: applied

libs/utils/src/lsn.rs
```diff
@@ -103,6 +106,12 @@ impl Lsn {
     pub fn is_aligned(&self) -> bool {
         *self == self.align()
     }
+
+    /// Return if the LSN is valid
+    /// mimics postgres XLogRecPtrIsInvalid macro
+    pub fn is_valid(self) -> bool {
+        self.0 != Lsn::INVALID.0
```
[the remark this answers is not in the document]
antons-antons: done

libs/remote_storage/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+// Note that with git_version prefix is `git:` and in case of git version from env its `git-env:`.
```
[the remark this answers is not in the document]
antons-antons: assuming this will be fixed by the integration of https://github.com/neondatabase/neon/pull/1634 ?

safekeeper/src/broker.rs
```diff
@@ -84,6 +91,65 @@ fn timeline_safekeeper_path(zttid: &ZTenantTimelineId, sk_id: ZNodeId) -> String
...
+        let mut c = Client::connect(conf.broker_endpoints.as_ref().unwrap(), None).await.unwrap();
+        c.lease_grant(LEASE_TTL_SEC, None).await.unwrap()
+    });
+
+    y.id()
+}
+
+// create keepalive task with default heatbeat interval
+pub async fn lease_keep_alive(lease_id: i64, conf: SafeKeeperConf) {
+    let mut client = Client::connect(conf.broker_endpoints.as_ref().unwrap(), None).await.unwrap();
```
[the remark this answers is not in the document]
antons-antons: supposedly addressed in the later versions, please validate

libs/remote_storage/src/lib.rs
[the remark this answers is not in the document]
SomeoneToIgnore: Yep, I'd have tried to rebase on top already and see if that somehow helps the PR.

control_plane/src/local_env.rs
```diff
@@ -63,6 +63,34 @@ pub struct LocalEnv {
     #[serde(default)]
     pub broker_endpoints: Option<String>,
 
+    // s3 bucket name for SK WAL Backup
+    #[serde(default)]
+    pub wal_backup_s3_bucket: Option<String>,
```
SomeoneToIgnore, line 68: Feels a bit odd that pageserver managed to have its buckets enabled on stating without the control_plane adjustments.
I'd hide those at least into `SafekeeperConf` at least and prepared some common struct just in case pageserver would also need such params here?

SomeoneToIgnore (changes requested):
I don't have an extensive knowledge on safekeeper code, so cannot comment the `control_plane` param adjustments now, but they look a but odd.

The cyclomatic complexity of the `detect_task` method could be reduced drastically: current versions with two loops, two kinds of breaks is a bit dangerous to work with in the future IMO

I wonder, what's left at this point, globally?
* remote_storage extraction and rebase
* a basic python test on the functionality, better to have more that one and unit tests also
* some refactoring
* ??? any pieces of logic missing? Looks like not (besides the active discussions), but hard to say for me at least at this point

I'd rebased this PR on the remote storage extraction PR (unless you disagree with that one fundamentally on something) or even merged that one
Then I'd write the Python test and ensured that those and the rest of the tests pass on the CI.
Some active discussions can be addressed along the way, also the refactorings.

After that, I think we can review it once more in bulk.
I can also review smaller increments, if you prefer so, just go on with what you're doing then.

safekeeper/src/wal_backup.rs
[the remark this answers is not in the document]
petuhovskiy: > should be okay to ignore missing segments when backup_lsn is not initialized yet

WDYM? Now #1561 is implemented and we can check if we have incomplete segment locally.

petuhovskiy (commented):
Looks better! I agree that some refactoring and advanced features can be done later (#1636). For now we want to get a PR with working S3 upload which doesn't break anything critical.

What I think should be definitely done before merge:
- wait until #1634 and #1635 merged to main, to prevent duplicate code
- rebase this PR on top of main
- add a python test for S3 upload
- fix discovered bugs :)

Other important things, that we should keep in mind and somehow test:
- We should ensure all segments are uploaded to S3, and we won't skip any. In other words, we won't delete a segment from safekeeper until it's uploaded to S3
- S3 upload won't stuck if there are errors or non-working safekeepers.
- S3 upload won't consume all available CPU resources. Usually that happens in a fail-retry infinite loop, so we should have sane retries. I'm a bit concerned about creating etcd client for every request, but not sure how much resources it takes and if we can reuse single client for everything.

Also note that #1602 was merged recently, so now you can use local_start_lsn and timeline_start_lsn to detect incomplete segments.

safekeeper/src/safekeeper.rs
antons-antons: I think we should discuss this proposal over a call.
looking at the code structure either I'm confused or the implementation would be more convoluted

safekeeper/src/wal_backup.rs
```diff
@@ -0,0 +1,473 @@
...
+                match result {
+                    Ok(l) => { leader = Some(l);},
+                    Err(e) => {
+                        error!("Error during leader election {:?}", e);
+                        sleep(Duration::from_millis(BROKER_CONNECTION_RETRY_DELAY_MS)).await;
+                        continue;
+                    },
+                }
+            }
+            _ = shutdown.changed() => {}
```
arssher, line 77: As discussed, there can be just single top-level select over shutdown chan and the rest. And we can do cleanup and exit right in the arm handling.

safekeeper/src/wal_backup.rs
```diff
@@ -0,0 +1,473 @@
...
+            select! {
+                durable_changed  = lsn_durable.changed() => {
+                    if let Err(e) = durable_changed {
+                        error!("Channel closed shutting down wal backup {:?}", e);
+                        cancel = true;
+                        break;
+                    }
+                }
+                _ = shutdown.changed() => {}
+                _ = sleep(Duration::from_millis(retry_delay)) => {}
```
arssher, line 115: select! can use conditional arm, see e.g. broker.rs

safekeeper/src/wal_backup.rs
```diff
@@ -0,0 +1,473 @@
...
+            let start_lsn = XLogSegNoOffsetToRecPtr(s, 0, seg_size);
+            let end_lsn = XLogSegNoOffsetToRecPtr(s + 1, 0, seg_size);
+            Segment::new(s, Lsn::from(start_lsn), Lsn::from(end_lsn))
+        })
+        .collect();
+
+    res
+}
+
+static REMOTE_STORAGE: OnceCell<Box<Option<GenericRemoteStorage>>> = OnceCell::new();
```
arssher, line 432: I doubt this should be *sync* OnceCell, and don't see why it is enum instead of trait -- that forces the manual `match` below.

libs/remote_storage/src/lib.rs
```diff
@@ -203,6 +204,91 @@ pub fn path_with_suffix_extension(original_path: impl AsRef<Path>, suffix: &str)
         .with_extension(new_extension.as_ref())
 }
 
+impl RemoteStorageConfig {
+    pub fn from_json_string(s: String) -> anyhow::Result<RemoteStorageConfig> {
```
arssher, line 208: Let's add a comment that this should be gone once safekeeper and pageserver conf is unified.

safekeeper/Cargo.toml
```diff
@@ -30,6 +30,11 @@ const_format = "0.2.21"
 tokio-postgres = { git = "https://github.com/zenithdb/rust-postgres.git", rev="d052ee8b86fff9897c77b0fe89ea9daba0e1fa38" }
 tokio-util = { version = "0.7", features = ["io"] }
 git-version = "0.3.5"
+rusoto_core = "0.47"
```
arssher, line 33: These are not needed as remote storage is separated.

test_runner/batch_others/test_walbackup_tenantreloc.py
```diff
@@ -0,0 +1,256 @@
...
+                    # the result, so sometimes selected value is larger by one than we expect
+                    assert cur.fetchone()[0] - inserted_ctr <= 1
+                    log.info("successfully recovered %s", inserted_ctr)
+                    failed = False
+                    load_ok_event.set()
+    log.info('load thread stopped')
+
+
+@pytest.mark.parametrize('with_load', ['with_load', 'without_load'])
+def test_tenant_relocation(zenith_env_builder: ZenithEnvBuilder,
```
petuhovskiy, line 99: This test looks like it checks tenant relocation on pageservers rather than safekeeper WAL backup, am I missing something?

For safekeepers WAL backup I expected a simple test in which we have:
- setup for some remote storage, local should be enough for this test, but I'd also add a test with S3 compatible storage later
- start usual configuration, most tests use 3safekeepers+1pageserver+1compute
- insert `generate_series(a lot)` to generate at least 2-3 WAL segments (each 16MB)
- run checkpoint on pageserver, so pageserver will tell safekeeper WAL can be offloaded from disk to s3
- check that WAL segments exist in remote storage
- check that segments are removed from all safekeepers disk
- check correct LSNs are stored in etcd (optional)

arssher:
So I
    - Added start/stop of per timeline offloading tasks on relevant events. Separate
      launcher tasks does the job.
    - Added neon_local & tests infrastructure, test itself.
    - Moved glue to from safekeeper.rs to timeline.rs.
    - Addressed some of reviews.
    - Rebased.
 And generally cleaned up a lot of things. I still dislike quite a number of them, notably condition for participating in election, but want to merge this ASAP and address them in follow ups. Currently I consider it mostly complete, what left is

- fix cause of test_restarts failure
- see what went wrong with e2e test
- update safekeeper args in deploy

safekeeper/src/wal_backup.rs
```diff
@@ -0,0 +1,422 @@
...
+    });
+
+    let mut tasks: HashMap<ZTenantTimelineId, WalBackupTaskHandle> = HashMap::new();
+
+    loop {
+        // channel is never expected to get closed
+        let zttid = wal_backup_launcher_rx.recv().await.unwrap();
+        let is_wal_backup_required = is_wal_backup_required(zttid);
+        if conf.remote_storage.is_none() {
+            continue; /* just drain the channel and do nothing */
```
SomeoneToIgnore, line 94: I wonder, if it makes sense to go that deep into the s3 code to implement this.
Might it be simpler to always require the remote storage during this thread init, and give to timelines some `/dev/null`-like channel (or just `Option`) end if there's no such storage given? 

----

Below, there's also another symptom:

```rust
// FIXME remove this
None => {
    panic!("no backup storage configured",);
}
```

safekeeper/src/bin/safekeeper.rs
```diff
@@ -252,7 +260,8 @@ fn start_safekeeper(mut conf: SafeKeeperConf, given_id: Option<ZNodeId>, init: b
     let signals = signals::install_shutdown_handlers()?;
     let mut threads = vec![];
     let (callmemaybe_tx, callmemaybe_rx) = mpsc::unbounded_channel();
-    GlobalTimelines::set_callmemaybe_tx(callmemaybe_tx);
+    let (wal_backup_launcher_tx, wal_backup_launcher_rx) = mpsc::channel(100);
```
SomeoneToIgnore, line 263: > 100

Is that enough for our plans to have ~1000 tenants served by a pageserver?
I would assume, one safekeeper has to know about ~333 tenants at least which might start randomly blocking others during intensive writes?

safekeeper/src/wal_backup.rs
```diff
@@ -0,0 +1,422 @@
...
+                );
+            } else {
+                // need to stop the task
+                info!("stopping wal backup task for {}", zttid);
+
+                let wb_handle = tasks.remove(&zttid).unwrap();
+                // Tell the task to shutdown. Error means task exited earlier, that's ok.
+                let _ = wb_handle.shutdown_tx.send(()).await;
+                // Await the task itself. TODO: restart panicked tasks earlier.
+                // Hm, why I can't await on reference to handle?
```
SomeoneToIgnore, line 138: You can on a `&mut` one.

safekeeper/src/broker.rs
```diff
@@ -44,6 +47,117 @@ fn timeline_safekeeper_path(
...
+        .await
+        .context("failed to create keepalive stream")?;
+
+    loop {
+        let push_interval = Duration::from_millis(PUSH_INTERVAL_MSEC);
+
+        keeper
+            .keep_alive()
+            .await
+            .context("failed to send LeaseKeepAliveRequest")?;
```
SomeoneToIgnore, line 134: Is there a way we are now able to react on these errors?
Presumably, we are not the leader anymore after this happens, but where is the logic to cover the leadership loss?

We never log these events and never really join on this task, that seems quite suspicious.
I wonder, if `check_am_i` and keepalive has to be the same thing, running in `select!` somewhere?

----

Another reason for being suspicious is the `WalBackupTask::run` workflow, where the keepalive task is being spawned and killed constantly in a loop.

safekeeper/src/wal_backup.rs
```diff
@@ -0,0 +1,422 @@
...
+            if let Some(ref l) = self.leader {
+                l.give_up().await;
+            }
+
+            match broker::get_leader(&self.election).await {
+                Ok(l) => {
+                    self.leader = Some(l);
+                }
+                Err(e) => {
+                    error!("error during leader election {:?}", e);
```
SomeoneToIgnore, line 214: Is that really an error or could that be a normal situation, where a node is not a leader, hence not backing up and doing a regular job? If the latter, might want to `info!` or even `debug!` this and replace `{:?}` with something else.

Or we might use better types here, since there seem to be both fatal and expected errors that this function can return.

safekeeper/src/wal_backup.rs
```diff
@@ -0,0 +1,422 @@
...
+        end_lsn,
+        start_lsn,
+    );
+    Ok(res)
+}
+
+async fn backup_single_segment(seg: &Segment, timeline_dir: &Path) -> Result<()> {
+    let segment_file_name = seg.file_path(timeline_dir)?;
+
+    // TODO: antons implement retry logic
```
SomeoneToIgnore, line 337: Looks implemented already, above.

SomeoneToIgnore (commented):
Looks nice in general.
I would have put more love into remote storage & channels init code + into the keepalive tasks.

safekeeper/src/wal_backup.rs
```diff
@@ -0,0 +1,422 @@
...
+    commit_lsn_watch_rx: watch::Receiver<Lsn>,
+    leader: Option<ElectionLeader>,
+    election: Election,
+}
+
+/// Offload single timeline.
+async fn backup_task_main(
+    zttid: ZTenantTimelineId,
+    timeline_dir: PathBuf,
+    mut shutdown_tx: Receiver<()>,
```
petuhovskiy, line 160: `shutdown_rx`

safekeeper/src/wal_backup.rs
```diff
@@ -0,0 +1,422 @@
...
+                    // wait for new WAL to arrive
+                    if let Err(e) = self.commit_lsn_watch_rx.changed().await {
+                        // should never happen, as we hold Arc to timeline.
+                        error!("commit_lsn watch shut down: {:?}", e);
+                        return;
+                    }
+                } else {
+                    // or just sleep if we errored previously
+                    let retry_delay = min(
+                        UPLOAD_FAILURE_RETRY_MIN_MS << retry_attempt,
```
petuhovskiy, line 232: Can we overflow here if retry_attempt will be too big?

safekeeper/src/safekeeper.rs
```diff
@@ -193,9 +194,9 @@ pub struct SafeKeeperState {
     /// Part of WAL acknowledged by quorum and available locally. Always points
     /// to record boundary.
     pub commit_lsn: Lsn,
-    /// First LSN not yet offloaded to s3. Useful to persist to avoid finding
-    /// out offloading progress on boot.
-    pub s3_wal_lsn: Lsn,
+    /// LSN that points to the end of the last backed up segment. Useful to
+    /// persist to avoid finding out offloading progress on boot.
+    pub backup_lsn: Lsn,
```
petuhovskiy, line 199: Should we bump control file version and add migration for it, or this simple rename will work as is?

safekeeper/src/safekeeper.rs
```diff
@@ -930,9 +930,9 @@ where
...
     /// While it is safe to use inmem values for determining horizon,
     /// we use persistent to make possible normal states less surprising.
-    pub fn get_horizon_segno(&self, s3_offload_enabled: bool) -> XLogSegNo {
-        let s3_offload_horizon = if s3_offload_enabled {
-            self.state.s3_wal_lsn
+    pub fn get_horizon_segno(&self, wal_backup_enabled: bool) -> XLogSegNo {
+        let wal_backup_horizon = if wal_backup_enabled {
+            self.state.backup_lsn
         } else {
             Lsn(u64::MAX)
```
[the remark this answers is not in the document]
arssher: Good notice. I decided just to keep --enable-wal-backup option.

safekeeper/src/broker.rs
```diff
@@ -44,6 +47,117 @@ fn timeline_safekeeper_path(
...
+        .await
+    {
+        keep_alive.abort();
+        return Err(e.into());
+    }
+
+    Ok(ElectionLeader { client, keep_alive })
+}
+
+pub async fn lease_keep_alive(mut client: Client, lease_id: i64) -> Result<()> {
```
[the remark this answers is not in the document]
arssher: Made private. But it is not infallible -- lease keep alive can easily fail.

Generally, I would like to rework the whole election thing by moving it into launcher and most probably without using election API at all.

safekeeper/src/timeline.rs
```diff
@@ -129,15 +136,62 @@ impl SharedState {
...
+        self.active = self.is_active();
+        self.is_wal_backup_action_pending()
+    }
+
+    /// Should we run s3 offloading in current state?
+    fn is_wal_backup_required(&self) -> bool {
+        let seg_size = self.get_wal_seg_size();
+        self.num_computes > 0 ||
+        // Currently only the whole segment is offloaded, so compare segment numbers.
+               (self.sk.inmem.commit_lsn.segment_number(seg_size) >
```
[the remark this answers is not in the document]
arssher: No. If e.g. commit_lsn is `0/2696628`, but backup lsn is `0/2000000`, there is nothing to do currently. We could compare LSNs, but truncate commit_lsn to segment beginning beforehand, not sure that would be simpler.

safekeeper/src/safekeeper.rs
```diff
@@ -764,6 +764,14 @@ where
         self.inmem.commit_lsn = commit_lsn;
         self.metrics.commit_lsn.set(self.inmem.commit_lsn.0 as f64);
 
+        // We got our first commit_lsn, which means we should sync
+        // everything to disk, to initialize the state.
+        if self.state.commit_lsn == Lsn(0) && commit_lsn > Lsn(0) {
```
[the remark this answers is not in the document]
arssher: Here we actually could, replaced it.

safekeeper/src/wal_storage.rs
```diff
@@ -598,7 +598,7 @@ fn write_zeroes(file: &mut File, mut count: usize) -> Result<()> {
 }
 
 /// Helper returning full path to WAL segment file and its .partial brother.
-fn wal_file_paths(
+pub fn wal_file_paths(
```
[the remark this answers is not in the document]
arssher: yes, thanks

safekeeper/src/wal_backup.rs
arssher: Currently we sometimes do send there from async and sometimes from sync context, so at this point it wouldn't simplify things. But I removed the 'symptom', thanks for noticing.

safekeeper/src/timeline.rs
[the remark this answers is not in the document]
SomeoneToIgnore: Maybe I don't get something, but due to

```rust
impl Lsn {
// ...
    pub fn segment_number(self, seg_sz: usize) -> u64 {
        self.0 / seg_sz as u64
    }
// ...
}
```

`(self.sk.inmem.commit_lsn.segment_number(seg_size) > self.sk.inmem.backup_lsn.segment_number(seg_size))`
can be pseudo rewritten as 
`(self.sk.inmem.commit_lsn / seg_size) > (self.sk.inmem.backup_lsn / seg_size)`

and since `seg_size` is `usize` and we're sure it's not 0, it's just 

`self.sk.inmem.commit_lsn > self.sk.inmem.backup_lsn`

?

safekeeper/src/bin/safekeeper.rs
arssher: Every safekeeper will push here 1000 times. But that shouldn't take long, launcher just spins/kills the task and ready take the next one from the channel.

safekeeper/src/wal_backup.rs
arssher: Just `&mut` didn't help -- actually awaiting on `JoinHandle` consumes it, which is reasonable. I fixed another place (killing lease task) according to this.

SomeoneToIgnore: Just out of curiosity, can you elaborate a bit on the consuming part?
I've meant something like 
```rust
#[tokio::main]
async fn main() {
    let mut handle = tokio::spawn(async {
        tokio::time::sleep(std::time::Duration::from_secs(3)).await;
        println!("Helloooooo from taaaaaaaaaaaask");     
    });
    
    loop {
        tokio::select! {
            _ = &mut handle => {
                println!("Task finished");
                break;
            }
            
            _ = tokio::time::sleep(std::time::Duration::from_secs(1)) => {
                println!("Waiting for the task");
            }
        }
    }
}
```

that seem to work without (explicitly?) consuming anything and awaiting on the same handle multiple times:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b83a0a089e97197c7cb270fc6cf4deda

safekeeper/src/timeline.rs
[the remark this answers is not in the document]
arssher: Well, in math world, yes. Here, where remainder is thrown away, no.

safekeeper/src/broker.rs
arssher: As soon as lease is lost, next call to `check_am_i` will tell our leadership is lost. Then we jump out of offloading loop into election loop and destroy the lease task (if it hasn't terminated itself at this point). But again, I feel like we could get away without Election API at all later.

safekeeper/src/wal_backup.rs
arssher: Interesting. Why then this doesn't work?
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ffe24232424869e889525f6e05a24262

safekeeper/src/wal_backup.rs
arssher: When node is not a leader, `client.campaign` call in `get_leader` will hang till it errors out or leadership is given (yes, that's not obvious).

safekeeper/src/broker.rs
SomeoneToIgnore: Due to compiler quirks: by default, the most "owning" resolution is used, similar to 

```rust
struct Foo {
    field: String,
}

fn main() {
    let foo = Foo {
        field: "field".to_string(),
    };

    let foo_ref_field_1 = &foo.field;
    let foo_ref_field_2 = &foo.field;

    // moves out the value
    let foo_field_1 = foo.field;
    // error, cannot move twice
    let foo_field_2 = foo.field;
}
```

so if you rewrite `async fn join(&mut self) ` contents as 

```
async fn join(&mut self) {
        let zz = &mut self.h;
        zz.await;
    }
```

it  would work.

safekeeper/src/wal_backup.rs
arssher: That would take about 2.8 * 10^12 years to happen, but I fixed that.

safekeeper/src/wal_backup.rs
arssher: TIL, thanks. One liner also works: `(&mut self.h).await;`.

safekeeper/src/safekeeper.rs
arssher: Tested, it works. Apparently bincode doesn't put field name inside.

safekeeper/src/wal_backup.rs
petuhovskiy: I'd say that it will take 6 minutes and it's not fixed still. When `retry_attempt` will be 64, then `UPLOAD_FAILURE_RETRY_MIN_MS << retry_attempt` will be 0, if I'm not mistaken anywhere.

SomeoneToIgnore (approved):

safekeeper/src/wal_backup.rs
arssher: right, here is one more attempt

PR #1585, opened 2022-04-27
WIP [proxy] Prototype new cloud api methods using a local mock

funbringer:


kelvich:
Thanks!

If we commit this PR in its current form we will break staging and e2e tests for quite a long time -- right now console responds with a plaintext password instead of a hash. And in order to fix that we will need to commit a big patch in the console which touches both the control plane and UI (since now the password is per cluster with a hardcoded username, not per user and we will need to address that too).

What about the following approach:

* leave `authenticate_proxy_client()` as it is
* when the username ends with `@zenith` we use the old flow with a plaintext password. We can model that as `AuthInfo::Password`
* when the username does not end with `@zenith` we can use SCRAM and a new API

or alternatively, we can switch old/new behavior with the CLI flag (by selecting "remote" router, which should be the default). BTW "remote" sounds strange, maybe call it "legacy"?

PR #1605, opened 2022-05-02
Deploy standalone SRAM proxy on staging

kelvich:


kelvich:
Should be merged after https://github.com/neondatabase/neon/pull/1604

cicdteam:
@kelvich I've pushed minor fixes but have one question about new proxy.

Is console still uses `PSQL_PROXY_MGMT_ADDRESS` variable to communicate with new (SCRAM) proxy ? if yes then we need use another variable as current value is:

`PSQL_PROXY_MGMT_ADDRESS = "host=zenith-proxy port=7000 sslmode=disable"` and it points to old proxy only.

kelvich:
Thanks!

Scram proxy do not use mgmt api, it just issues an http calls to the proxy. However now I've split password-based and links-based flows without guessing it by '@zenith' username suffix. So the idea is to have two proxy instances:
* scram one (--auth-backend=console) on 'cloud.neon.tech' which avoids mgmt api and can be scaled up
* link one (--auth-backend=link) on 'start.neon.tech' which is used during the onboarding and uses mgmt api

cicdteam (approved):

.circleci/helm-values/staging.proxy-scram.yaml
```diff
@@ -6,7 +6,7 @@ image:
 
 settings:
   authBackend: "console"
-  authEndpoint: "https://console.stage.neon.tech:9095/management/api/v2"
+  authEndpoint: "http://console-staging.local/management/api/v2/healthz"
```
kelvich: Hm, why `/healthz`? This string would be a base to construct API calls like `/management/api/v2/proxy_get_scram_secret?role=aaa&cluster=bbb`. Also port is 9095 as per `https://github.com/neondatabase/cloud/blob/main/helm/zenith-console/values-stage.yaml#L78-L85`.

PR #1611, opened 2022-05-02
Make remote storage test less flacky

SomeoneToIgnore:


LizardWizzard (approved):

arssher:
I wonder why it wasn't so flaky before? Or this is just my observational bias?

LizardWizzard:
@arssher I added new checks on friday, test passed locally and in my PR but started to fail on main

SomeoneToIgnore:
Also, somewhere around that time I've created a PR that arrived in `main` separately: we now load all layers into pageserver's memory eagerly on startup, hence kicking off the gc and compaction threads.

PR #1613, opened 2022-05-02
fix domain names in proxy deploy

kelvich:


.circleci/helm-values/staging.proxy.yaml
```diff
@@ -20,7 +20,7 @@ exposedService:
     service.beta.kubernetes.io/aws-load-balancer-type: external
     service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: ip
     service.beta.kubernetes.io/aws-load-balancer-scheme: internet-facing
-    external-dns.alpha.kubernetes.io/hostname: connect.stage.neon.tech
+    external-dns.alpha.kubernetes.io/hostname: start.stage.neon.tech
```
cicdteam, line 24: why we back to `start.` there ? it's ok with `start.` of course but I though this part of hostname will be the same in all environments (like `connect.neon.tech` for prod).

PR #1620, opened 2022-05-03
avoid detach with alive branches

LizardWizzard:


pageserver/src/layered_repository.rs
```diff
@@ -393,9 +393,23 @@ impl Repository for LayeredRepository {
 
     fn detach_timeline(&self, timeline_id: ZTimelineId) -> anyhow::Result<()> {
         let mut timelines = self.timelines.lock().unwrap();
+        // check no child timelines, because detach will remove files, which will brake child branches
+        // FIXME this can still be violated because we do not guarantee
+        //   that all ancestors are downloaded/attached to the same pageserver
+        dbg!(&timeline_id);
```
SomeoneToIgnore: `dbg!` (here and below) looks like something worth removing

pageserver/src/layered_repository.rs
```diff
@@ -393,9 +393,22 @@ impl Repository for LayeredRepository {
 
     fn detach_timeline(&self, timeline_id: ZTimelineId) -> anyhow::Result<()> {
         let mut timelines = self.timelines.lock().unwrap();
+        // check no child timelines, because detach will remove files, which will brake child branches
+        // FIXME this can still be violated because we do not guarantee
+        //   that all ancestors are downloaded/attached to the same pageserver
```
SomeoneToIgnore, line 398: > downloaded

We can actually check the remote index too, can't we?
It has `metadata: TimelineMetadata` inside of the corresponding `RemoteTimeline`

Feels worth a test also, but up to you.

pageserver/src/layered_repository.rs
LizardWizzard: Cleaned it up, thanks!

SomeoneToIgnore: That comment did disappear after I hit `Refresh` 🤔

pageserver/src/layered_repository.rs
LizardWizzard: It still misses tenants that are not attached to that pageserver. But then check has to happen in dettach_handler itself. Not sure are there any benefits in checking both places

SomeoneToIgnore: Guess me being paranoid and not trusting the layer map.
Also, we don't remove from the remote index now, so it will be more complete after multiple detaches.

But no strong opinion, it all has to be amended anyway during better s3 <-> local <-> inmem <-?-> etcd transitions design.

SomeoneToIgnore (approved):

pageserver/src/layered_repository.rs
LizardWizzard: Layer map is a different thing :) This is a timelines one

PR #1621, opened 2022-05-03
tone down tenant not found error

LizardWizzard:


SomeoneToIgnore (approved):
Or we can use `{e:#}` that would not print a multiline error, AFAIK

kelvich:
Hm, do we actually need a backtrace there?

kelvich (approved):

kelvich:
Let's merge this as is -- this will help with the investigations now. Later we can think of introducing custom errors and guidelines on when to print backtrace.

PR #1631, opened 2022-05-04
Fix control file update for b9fd8a36ad3b

arssher:
(old clusters on staging are broken without this)


petuhovskiy (approved):
Probably we need a compatibility test for that, similar to the pageserver one we discussed on a meeting.

PR #1635, opened 2022-05-04
Prepare etcd_broker crate for further pageserver usage

SomeoneToIgnore:
etcd librarification part, extracted from https://github.com/neondatabase/neon/pull/1619

### DIFF
--- a/libs/etcd_broker/src/lib.rs
+++ b/libs/etcd_broker/src/lib.rs
@@ -1,0 +1,335 @@
+//! A set of primitives to access a shared data/updates, propagated via etcd broker (not persistent).
+//! Intended to connect services to each other, not to store their data.
+use std::{
+    collections::{hash_map, HashMap},
+    fmt::Display,
+    str::FromStr,
+};
+
+use regex::{Captures, Regex};
+use serde::{Deserialize, Serialize};
+use serde_with::{serde_as, DisplayFromStr};
+
+pub use etcd_client::*;
+
+use tokio::{sync::mpsc, task::JoinHandle};
+use tracing::*;
+use utils::{
+    lsn::Lsn,
+    zid::{ZNodeId, ZTenantId, ZTenantTimelineId},
+};
+
+#[derive(Debug, Deserialize, Serialize)]
+struct SafekeeperTimeline {
+    safekeeper_id: ZNodeId,
+    info: SkTimelineInfo,
+}
+
+/// Published data about safekeeper's timeline. Fields made optional for easy migrations.
+#[serde_as]
+#[derive(Debug, Deserialize, Serialize)]
+pub struct SkTimelineInfo {
+    /// Term of the last entry.
+    pub last_log_term: Option<u64>,
+    /// LSN of the last record.
+    #[serde_as(as = "Option<DisplayFromStr>")]
+    #[serde(default)]
+    pub flush_lsn: Option<Lsn>,
+    /// Up to which LSN safekeeper regards its WAL as committed.
+    #[serde_as(as = "Option<DisplayFromStr>")]
+    #[serde(default)]
+    pub commit_lsn: Option<Lsn>,
+    /// LSN up to which safekeeper offloaded WAL to s3.
+    #[serde_as(as = "Option<DisplayFromStr>")]
+    #[serde(default)]
+    pub s3_wal_lsn: Option<Lsn>,
+    /// LSN of last checkpoint uploaded by pageserver.
+    #[serde_as(as = "Option<DisplayFromStr>")]
+    #[serde(default)]
+    pub remote_consistent_lsn: Option<Lsn>,
+    #[serde_as(as = "Option<DisplayFromStr>")]
+    #[serde(default)]
+    pub peer_horizon_lsn: Option<Lsn>,
+    #[serde(default)]
+    pub wal_stream_connection_string: Option<String>,
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum BrokerError {
+    #[error("Etcd client error: {0}. Context: {1}")]
+    EtcdClient(etcd_client::Error, String),
+    #[error("Error during parsing etcd data: {0}")]
+    ParsingError(String),
+    #[error("Internal error: {0}")]
+    InternalError(String),
+}
+
+/// A way to control the data retrieval from a certain subscription.
+pub struct SkTimelineSubscription {
+    safekeeper_timeline_updates:
+        mpsc::UnboundedReceiver<HashMap<ZTenantTimelineId, HashMap<ZNodeId, SkTimelineInfo>>>,
+    kind: SkTimelineSubscriptionKind,
+    watcher_handle: JoinHandle<Result<(), BrokerError>>,
+    watcher: Watcher,
+}
+
+impl SkTimelineSubscription {
+    /// Asynchronously polls for more data from the subscription, suspending the current future if there's no data sent yet.
+    pub async fn fetch_data(
+        &mut self,
+    ) -> Option<HashMap<ZTenantTimelineId, HashMap<ZNodeId, SkTimelineInfo>>> {
+        self.safekeeper_timeline_updates.recv().await
+    }
+
+    /// Cancels the subscription, stopping the data poller and waiting for it to shut down.
+    pub async fn cancel(mut self) -> Result<(), BrokerError> {
+        self.watcher.cancel().await.map_err(|e| {
+            BrokerError::EtcdClient(
+                e,
+                format!(
+                    "Failed to cancel timeline subscription, kind: {:?}",
+                    self.kind
+                ),
+            )
+        })?;
+        self.watcher_handle.await.map_err(|e| {
+            BrokerError::InternalError(format!(
+                "Failed to join the timeline updates task, kind: {:?}, error: {e}",
+                self.kind
+            ))
+        })?
+    }
+}
+
+/// The subscription kind to the timeline updates from safekeeper.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct SkTimelineSubscriptionKind {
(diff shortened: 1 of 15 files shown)


safekeeper/src/timeline.rs
```diff
@@ -432,11 +435,23 @@ impl Timeline {
                 shared_state.sk.inmem.remote_consistent_lsn,
             )),
             peer_horizon_lsn: Some(shared_state.sk.inmem.peer_horizon_lsn),
-        }
+            wal_stream_connection_string: shared_state
+                .pageserver_connstr
+                .as_deref()
+                .map(|pageserver_connstr| {
```
petuhovskiy, line 441: Why do we need pageserver_connstr here? I thought that it's used only by callmemaybe logic, when sk connects to ps and asks it to start replication.

It looks like you're preparing for etcd integration in the pageserver, and we are going away from using callmemaybe. Then you just need to advertise sk callback address, so any pageserver can connect to this safekeeper when it needs to. Am I missing something?

safekeeper/src/timeline.rs
```diff
@@ -485,6 +500,29 @@ impl Timeline {
     }
 }
 
+// pageserver connstr is needed to be able to distinguish between different pageservers
+// it is required to correctly manage callmemaybe subscriptions when more than one pageserver is involved
+// TODO it is better to use some sort of a unique id instead of connection string, see https://github.com/zenithdb/zenith/issues/1105
+fn callmemaybe_connstr(
```
petuhovskiy, line 506: If it's not related to the old callmemaybe logic, then I'd like to invent a new name for it. Maybe something like `callback_connstr`, or just `safekeeper_connstr`.

libs/etcd_broker/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+    pub remote_consistent_lsn: Option<Lsn>,
+    #[serde_as(as = "Option<DisplayFromStr>")]
+    #[serde(default)]
+    pub peer_horizon_lsn: Option<Lsn>,
+    #[serde(default)]
+    pub wal_stream_connection_string: Option<String>,
+}
+
+// TODO: add global zenith installation ID.
+const ZENITH_BROKER_PREFIX: &str = "zenith";
```
petuhovskiy, line 58: Some time ago we had a discussion that we're using single etcd for several neon installations. So I think we can take this const from `ETCD_BROKER_PREFIX` envvar. It's not urgent, we can create an issue and do it later.

safekeeper/src/timeline.rs
SomeoneToIgnore: I have left this small method as some sort of an easter egg, but I like your sense of naming more.

Yet `callback_connstr` seems too generic for a WAL streaming tasks, so let it be `wal_stream_connection_string`?

safekeeper/src/timeline.rs
SomeoneToIgnore: I don't think you miss anything, I just follow the path of least resistance and reuse the connection string that was propagated by safekeepers to pageservers before: this way, I have to write no new code to connect and stream the wal.

That connection string was constructed from the `pageserver_connstr` and the `callmemaybe_connstr` logic in the old code, I've removed this in the "remove callmemaybe" request, but left it here.

safekeeper/src/timeline.rs
petuhovskiy: `wal_stream_connection_string` sounds fine for me

libs/etcd_broker/src/lib.rs
SomeoneToIgnore: Will adjust, thanks. Also have to change it to `neon` anyway..

SomeoneToIgnore: Have thought about it more and placed that into `SafekeeperConf` without using env vars at all.
Lmk what you think.

petuhovskiy: Yeah, thought of it too, but forgot to mention. We currently have different approaches to configuration in safekeeper and pageserver, but we need this prefix to be the same for both. `SafekeeperConf` is ok, but I guess we should also add it to pageserver config then.

libs/etcd_broker/src/lib.rs
```diff
@@ -0,0 +1,318 @@
...
+                    if let Some(kv) = event.kv() {
+                        match parse_etcd_key_value(subscription_kind, &regex, kv) {
+                            Ok(Some((zttid, timeline))) => {
+                                match timeline_updates
+                                    .entry(zttid)
+                                    .or_default()
+                                    .entry(timeline.safekeeper_id)
+                                {
+                                    hash_map::Entry::Occupied(mut o) => {
+                                        if o.get().flush_lsn < timeline.info.flush_lsn {
```
petuhovskiy, line 230: I'm not sure about this, because in theory `flush_lsn` is not always a correct metric for identifying latest update. More correct metric would be `(last_log_term, commit_lsn, flush_lsn)`, but I'd not use it because it's not very obvious. Can we just send all updates with `Some(zttid, timeline)` to a channel one by one, or there will be performance and other kind of issues?

petuhovskiy (approved):
Looks good to me, but I think other potential users of this crate should take a look too :)

libs/etcd_broker/src/lib.rs
SomeoneToIgnore: We can send those and I'm sure this code is more like a precaution rather then a real situation with etcd: AFAIK, per subscription poll on the key, it returns the "aggregated" data with the last versions per key value (node id in our case).

Feels odd to panic/error in such case though, so maybe just merge it over the triplet?

LizardWizzard (commented):
* Etcd client supports having multiple watches from one client. It is a bit tricky because you start first watch using `client.watch`, and then use returned `Watcher` object to add more watched keys to the same stream. See here: https://github.com/etcdv3/etcd-client/blob/master/examples/watch.rs#L24. This can be done later if we're in a hurry to merge this PR.
* Since this is a new package, are there any cons to using thiserror here and avoid anyhow entirely?

SomeoneToIgnore:
> Added a way to add another subscription into the existing one.

Just looked at watcher more and I don't think this is the way to use the watcher in pageserver: we'd love to cancel the watches for certain timelines on their detach (and stop receiving their etcd values), while the Watcher API seems to only accept new watches without the ability to remove those.

libs/etcd_broker/src/lib.rs
```diff
@@ -0,0 +1,374 @@
...
+    pub fn timeline(broker_prefix: String, timeline: ZTenantTimelineId) -> Self {
+        Self {
+            broker_prefix,
+            kind: SubscriptionKind::Timeline(timeline),
+        }
+    }
+
+    fn watch_regex(&self) -> Regex {
+        match self.kind {
+            SubscriptionKind::All => Regex::new(&format!(
```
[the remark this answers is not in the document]
LizardWizzard: If prefix is specified in config it is still possible to use once_cell то initialize the value with a custom argument. Like it is done in [this](https://docs.rs/once_cell/latest/once_cell/#safe-initialization-of-global-data) example with explicit call to `.set`

LizardWizzard (approved):
Looks good to me. Thanks!

Plus one on polishing multiple subscriptions case later

libs/etcd_broker/src/lib.rs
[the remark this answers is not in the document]
SomeoneToIgnore: But now our constructor accepts an arbitrary string, so we cannot init a regex once?
Maybe I'm thinking it wrong? I can review a PR on the topic :)

safekeeper/src/timeline.rs
arssher: While callmemaybe is still there, we could leave that code in place but publish safekeeper address in SkTimelineInfo already cleanly, without pageserver_connstr appended.

libs/etcd_broker/src/lib.rs
```diff
@@ -0,0 +1,335 @@
...
+                e,
+                format!("Failed to init the watch for subscription {subscription:?}"),
+            )
+        })?;
+
+    let (timeline_updates_sender, safekeeper_timeline_updates) = mpsc::unbounded_channel();
+
+    let subscription_kind = subscription.kind;
+    let regex = subscription.watch_regex();
+    let watcher_handle = tokio::spawn(async move {
```
arssher, line 211: Do we need separate spawn (and communicating channel) here? I suspect subscriber will run in its own task anyway.

arssher (commented):
Somewhat late to the party, but here some comments.

libs/etcd_broker/src/lib.rs
SomeoneToIgnore: We don't need it, but if we remove the spawn and the channel, we'll have to remove the subscription polling in the loop, leaving tiny bits of shared code here and making both safekeeper and pageserver to implement their own polling, cancellation, etc?

Could be done, yet my point was to create an isolated timeline data watcher, that could use etcd or anything else as an impl detail: this way, we don't leak `etcd_client` types much for the subscription logic.

libs/etcd_broker/src/lib.rs
SomeoneToIgnore: https://etcd.io/docs/v3.5/learning/api_guarantees/#isolation-level-and-consistency-of-replicas

> etcd does not ensure linearizability for watch operations. Users are expected to verify the revision of watch responses to ensure correct ordering.

Looks like we have to check versions ourselves, so I'll add some code and comments there.

safekeeper/src/timeline.rs
SomeoneToIgnore: Sure, this way we can add the connstr as a separate field later, looks even better.

SomeoneToIgnore:
https://github.com/neondatabase/neon/pull/1686

PR #1638, opened 2022-05-05
get_binaries script fix

cicdteam:
get_binaries script fixed to retrieve correct docker tags

it fix attempts to deploy wrong versions like this
```console
ok: [zenith-us-stage-ps-2] => {
    "msg": "Version to deploy - test-1535"
}
```

### DIFF
--- a/.circleci/ansible/get_binaries.sh
+++ b/.circleci/ansible/get_binaries.sh
@@ -7,7 +7,7 @@ RELEASE=${RELEASE:-false}
 # look at docker hub for latest tag for neon docker image
 if [ "${RELEASE}" = "true" ]; then
     echo "search latest relase tag"
-    VERSION=$(curl -s https://registry.hub.docker.com/v1/repositories/neondatabase/neon/tags |jq -r -S '.[].name' | grep release | sed 's/release-//g' | grep '^[0-9]*$' | sort -n | tail -1)
+    VERSION=$(curl -s https://registry.hub.docker.com/v1/repositories/neondatabase/neon/tags |jq -r -S '.[].name' | grep release | sed 's/release-//g' | grep -E '^[0-9]+$' | sort -n | tail -1)
     if [ -z "${VERSION}" ]; then
         echo "no any docker tags found, exiting..."
         exit 1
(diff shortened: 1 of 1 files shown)


.circleci/ansible/get_binaries.sh
```diff
@@ -7,7 +7,7 @@ RELEASE=${RELEASE:-false}
 # look at docker hub for latest tag for neon docker image
 if [ "${RELEASE}" = "true" ]; then
     echo "search latest relase tag"
-    VERSION=$(curl -s https://registry.hub.docker.com/v1/repositories/neondatabase/neon/tags |jq -r -S '.[].name' | grep release | sed 's/release-//g' | tail -1)
+    VERSION=$(curl -s https://registry.hub.docker.com/v1/repositories/neondatabase/neon/tags |jq -r -S '.[].name' | grep release | sed 's/release-//g' | grep '^[0-9]*$' | sort -n | tail -1)
```
arssher, line 10: Note that I don't see any images containing 'release' currently.

arssher (approved):

.circleci/ansible/get_binaries.sh
```diff
@@ -16,7 +16,7 @@ if [ "${RELEASE}" = "true" ]; then
     fi
 else
     echo "search latest dev tag"
-    VERSION=$(curl -s https://registry.hub.docker.com/v1/repositories/neondatabase/neon/tags |jq -r -S '.[].name' | grep -v release | tail -1)
+    VERSION=$(curl -s https://registry.hub.docker.com/v1/repositories/neondatabase/neon/tags |jq -r -S '.[].name' | grep -v release | grep '^[0-9]*$' | sort -n | tail -1)
```
kelvich, line 19: nit: grep '^[0-9]+$' is a bit safer
nit: grep -v release is unecessary

.circleci/ansible/get_binaries.sh
kelvich, line 10: JFYI: right now there no images prefixed with 'release'

kelvich (approved):

.circleci/ansible/get_binaries.sh
cicdteam: yes currently there are no neon images with release prefix (they are in  `zenithdb` docker account) but they will appear when we merge to release.

.circleci/ansible/get_binaries.sh
cicdteam: I've tested grep on MacOS and there grep hasn't `-E`, about `+` you are right but it should be escaped probably (in shell it should be `^[0-9]\+` but not sure how it should be inside docker in CircleCI

kelvich: Strange! My mac has `grep -E` but may be I've overwrote standard grep, not sure

cicdteam: yes, seems my fault, updated script

PR #1639, opened 2022-05-05
Feature/lineages as a value

MMeent:
Draft implementation of the Lineage value-merging pipeline


MMeent:
The tests fail on seccomp issues, and I've not been able to reproduce this locally. Apart from that, I think this is ready for review. I will add more comments to the code tomorrow.

lubennikovaav (commented):
@MMeent, please, add more comments to the new structures. 
Also, I think it's worth to describe lineages in some Readme

LizardWizzard:
As discussed moving to draft.

PR #1645, opened 2022-05-06
Bump vendor/postgres to shut down on wrong basebackup.

arssher:


petuhovskiy (approved):
Looks good, but we should have a test for it. How hard is it to create such test now?

arssher:
Shouldn't be hard, but let's do that separately.

PR #1648, opened 2022-05-06
Enable etcd and safekeepers in all tests

SomeoneToIgnore:
Part of https://github.com/neondatabase/neon/pull/1619

This is a PR that should create preliminary breakage before the main breakage is implemented.
* starts one safekeeper for every test 
I'm not sure if it uses that though, I should I also remove [this](https://github.com/neondatabase/neon/blob/ef40e404cf15cc335fbf6a226879e5358aa628eb/test_runner/fixtures/zenith_fixtures.py#L1543) line from fixtures? Anywhere else?
* starts an etcd for every test

Makes both pageserver and safekeeper binaries to require etcd broker endpoints specified for them in configs/startup args.

Without the endpoints, no etcd connection is possible.
Without etcd connection, no WAL streaming from Safekeeper to Pageserver is possible after the WIP PR mentioned above.

### DIFF
--- a/control_plane/simple.conf
+++ b/control_plane/simple.conf
@@ -9,3 +9,6 @@ auth_type = 'Trust'
 id = 1
 pg_port = 5454
 http_port = 7676
+
+[etcd_broker]
+broker_endpoints = ['http://127.0.0.1:2379']
--- a/control_plane/src/local_env.rs
+++ b/control_plane/src/local_env.rs
@@ -97,6 +97,7 @@ pub struct PageServerConf {
 
     // jwt auth token used for communication with pageserver
     pub auth_token: String,
+    pub broker_endpoints: Vec<String>,
 }
 
 impl Default for PageServerConf {
@@ -107,6 +108,7 @@ impl Default for PageServerConf {
             listen_http_addr: String::new(),
             auth_type: AuthType::Trust,
             auth_token: String::new(),
+            broker_endpoints: Vec::new(),
         }
     }
 }
@@ -401,6 +403,7 @@ impl LocalEnv {
 
         self.pageserver.auth_token =
             self.generate_auth_token(&Claims::new(None, Scope::PageServerApi))?;
+        self.pageserver.broker_endpoints = self.broker_endpoints.clone();
 
         fs::create_dir_all(self.pg_data_dirs_path())?;
 
--- a/control_plane/src/local_env.rs
+++ b/control_plane/src/local_env.rs
@@ -60,14 +60,7 @@ pub struct LocalEnv {
     #[serde(default)]
     pub private_key_path: PathBuf,
 
-    // Broker (etcd) endpoints for storage nodes coordination, e.g. 'http://127.0.0.1:2379'.
-    #[serde(default)]
-    #[serde_as(as = "Vec<DisplayFromStr>")]
-    pub broker_endpoints: Vec<Url>,
-
-    /// A prefix to all to any key when pushing/polling etcd from a node.
-    #[serde(default)]
-    pub broker_etcd_prefix: Option<String>,
+    pub etcd_broker: EtcdBroker,
 
     pub pageserver: PageServerConf,
 
@@ -81,6 +74,62 @@ pub struct LocalEnv {
     // https://toml.io/en/v1.0.0 does not contain a concept of "a table inside another table".
     #[serde_as(as = "HashMap<_, Vec<(DisplayFromStr, DisplayFromStr)>>")]
     branch_name_mappings: HashMap<String, Vec<(ZTenantId, ZTimelineId)>>,
+}
+
+/// Etcd broker config for cluster internal communication.
+#[serde_as]
+#[derive(Serialize, Deserialize, PartialEq, Eq, Clone, Debug)]
+pub struct EtcdBroker {
+    /// A prefix to all to any key when pushing/polling etcd from a node.
+    #[serde(default)]
+    pub broker_etcd_prefix: Option<String>,
+
+    /// Broker (etcd) endpoints for storage nodes coordination, e.g. 'http://127.0.0.1:2379'.
+    #[serde(default)]
+    #[serde_as(as = "Vec<DisplayFromStr>")]
+    pub broker_endpoints: Vec<Url>,
+
+    /// Etcd binary path to use.
+    #[serde(default)]
+    pub etcd_binary_path: PathBuf,
+}
+
+impl EtcdBroker {
+    pub fn locate_etcd() -> anyhow::Result<PathBuf> {
+        let which_output = Command::new("which")
+            .arg("etcd")
+            .output()
+            .context("Failed to run 'which etcd' command")?;
+        let stdout = String::from_utf8_lossy(&which_output.stdout);
+        ensure!(
+            which_output.status.success(),
+            "'which etcd' invocation failed. Status: {}, stdout: {stdout}, stderr: {}",
+            which_output.status,
+            String::from_utf8_lossy(&which_output.stderr)
+        );
+
+        let etcd_path = PathBuf::from(stdout.trim());
+        ensure!(
+            etcd_path.is_file(),
+            "'which etcd' invocation was successful, but the path it returned is not a file or does not exist: {}",
+            etcd_path.display()
+        );
+
+        Ok(etcd_path)
+    }
+
+    pub fn comma_separated_endpoints(&self) -> String {
+        self.broker_endpoints.iter().map(Url::as_str).fold(
+            String::new(),
+            |mut comma_separated_urls, url| {
+                if !comma_separated_urls.is_empty() {
+                    comma_separated_urls.push(',');
+                }
+                comma_separated_urls.push_str(url);
+                comma_separated_urls
+            },
+        )
+    }
 }
 
 #[derive(Serialize, Deserialize, PartialEq, Eq, Clone, Debug)]
@@ -97,7 +146,6 @@ pub struct PageServerConf {
 
     // jwt auth token used for communication with pageserver
     pub auth_token: String,
-    pub broker_endpoints: Vec<String>,
 }
 
 impl Default for PageServerConf {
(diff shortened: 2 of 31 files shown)


SomeoneToIgnore:
At least the e2e tests fail expectedly correctly.
I'll adjust those later, but would be glad to know if I'm doing everything right and no direct compute streaming tests are left.

SomeoneToIgnore:
Console PR: https://github.com/neondatabase/cloud/pull/1195

SomeoneToIgnore:
Just realized that I have to adjust neon_base cli now to start etcd on init and stop at least.

Another good topic to discuss is whether to make the broker_endpoints mandatory as it is now: their absence in the new model makes both binary pretty useless, but for local hacking at least, it might be ok to start without the broker?

neon_local/src/main.rs
```diff
@@ -896,20 +896,23 @@ fn handle_safekeeper(sub_match: &ArgMatches, env: &local_env::LocalEnv) -> Resul
     Ok(())
 }
 
-fn handle_start_all(sub_match: &ArgMatches, env: &local_env::LocalEnv) -> Result<()> {
+fn handle_start_all(sub_match: &ArgMatches, env: &local_env::LocalEnv) -> anyhow::Result<()> {
+    etcd::start_etcd_process(env)?;
```
arssher, line 900: In current coding etcd is started by both python code and neon_local, we clearly need one of them. The question is which one. Python code gives slightly more handy manual control, but leaving neon_local untested is not nice. I think we can do everything from neon_local, but make it more functional:

- on stop wait handle, bail out if stop failed;
- wait for start by checking status;
- make the whole etcd managing business optional (it might be running elsewhere) -- e.g. bool flag in conf;
- probably add separate etcd start/stop commands. For example, currently ZenithEnvBuilder on __exit__ doesn't call `neon stop`, it shuts down services individually. etcd start/stop can also be useful for some tests.

Or just sit with python version for some more time.

arssher (changes requested):

arssher:
> Another good topic to discuss is whether to make the broker_endpoints mandatory as it is now: their absence in the new model makes both binary pretty useless, but for local hacking at least, it might be ok to start without the broker?

Yes, I would keep it optional. For example, for benchmarking safekeepers it is useful to have working setup without pageserver at all (we actually have such test).

arssher:
> I'm not sure if it uses that though, I should I also remove [this](https://github.com/neondatabase/neon/blob/ef40e404cf15cc335fbf6a226879e5358aa628eb/test_runner/fixtures/zenith_fixtures.py#L1543) line from fixtures? Anywhere else?

Yes, with this PR all tests use safekeeper. As for this line -- its fate should be the same as `callmemaybe_connstring` configuration in `setup_pg_conf`. If we don't support setup without safekeepers at all, it can be removed. But I'm dubious on that. Such setup is still might be useful for tests/experiments/local deployment. But for this to work, we must either don't drop callmemaybe support from pageserver (which I don't expect to be hard actually) or reverse the flow (initiate replication connection by compute), which is a good idea already discussed, but more involved.

control_plane/src/local_env.rs
```diff
@@ -59,13 +59,7 @@ pub struct LocalEnv {
...
     pub private_key_path: PathBuf,
 
-    // A comma separated broker (etcd) endpoints for storage nodes coordination, e.g. 'http://127.0.0.1:2379'.
-    #[serde(default)]
-    pub broker_endpoints: Option<String>,
-
-    /// A prefix to all to any key when pushing/polling etcd from a node.
-    #[serde(default)]
-    pub broker_etcd_prefix: Option<String>,
+    pub etcd_broker: EtcdBroker,
```
arssher, line 62: Should also add this to `simple.conf`

petuhovskiy (commented):
Looks fine, but I have several questions:
- What's with removing etcd from python fixtures? Can we handle it all in control_plane?
- What's current approach for etcd reuse? Are we launching separate etcd instances for each test? Can we have one for all the tests?

control_plane/src/local_env.rs
```diff
@@ -412,3 +457,19 @@ fn base_path() -> PathBuf {
         None => ".zenith".into(),
     }
 }
+
+// TODO kb split fs checks and parsing methods, to test the conf for completeness
```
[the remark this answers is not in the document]
SomeoneToIgnore: We should, sorry for the amount of pushes, I was fighting with the bash scripting to make our `broker_endpoints` propagated to every e2e test.
Luckily, this was the last one and tests are green now: https://github.com/neondatabase/cloud/actions/runs/2316232005

This means, we've got all possible places `broker_endpoints` covered and I can throw away the code that fails on this config value absence and address the rest of the comments.

To justify the extra pushes here, I've been addressing some comments and this TODO is a result of the `simple.conf` comment: I want to create a test for parsing the file and left a reminder.

SomeoneToIgnore:
> What's with removing etcd from python fixtures? Can we handle it all in control_plane?

We can and we should, but I don't think I want to do this now, since it's quite a big change per se.
I'm into the idea fully though, we can create a follow-up task for that.

> What's current approach for etcd reuse? Are we launching separate etcd instances for each test? Can we have one for all the tests?

So far, it's a single cluster of a single etcd node in e2e tests: https://github.com/neondatabase/cloud/pull/1241/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3
and a separate etcd per test here, with its own logs directory.
Looks like an overkill indeed, I'll move it out same way we have the remote storage now when addressing the rest of the unresolved comments.

control_plane/simple.conf
```diff
@@ -9,3 +9,6 @@ auth_type = 'Trust'
 id = 1
 pg_port = 5454
 http_port = 7676
+
+[etcd_broker]
+broker_endpoints = ['127.0.0.1::2379']
```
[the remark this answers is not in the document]
SomeoneToIgnore: There were no tests for that, I've added those separately: https://github.com/neondatabase/neon/pull/1697

Funny that `['127.0.0.1::2379']` gets parsed as `Url` just fine:

```
broker_endpoints: [
            Url {
                scheme: "localhost",
                cannot_be_a_base: true,
                username: "",
                password: None,
                host: None,
                port: None,
                path: ":2379",
                query: None,
                fragment: None,
            },
        ],
```

Anyway, let's continue there.

test_runner/fixtures/zenith_fixtures.py
```diff
@@ -1891,7 +1896,13 @@ def test_output_dir(request: Any) -> str:
...
 
 
-SKIP_DIRS = frozenset(('pg_wal', 'pg_stat', 'pg_stat_tmp', 'pg_subtrans', 'pg_logical'))
+SKIP_DIRS = frozenset(('pg_wal',
+                       'pg_stat',
+                       'pg_stat_tmp',
+                       'pg_subtrans',
+                       'pg_logical',
+                       'pg_replslot/wal_proposer_slot',
+                       'pg_xact'))
```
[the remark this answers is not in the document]
SomeoneToIgnore: Reverted back, that was wrong, thanks.

neon_local/src/main.rs
SomeoneToIgnore: I've decided to sit with the python for now without extra big changes.
More on the topic: https://github.com/neondatabase/neon/pull/1648#issuecomment-1125948912

control_plane/src/local_env.rs
SomeoneToIgnore: That resulted in https://github.com/neondatabase/neon/pull/1697 eventually

SomeoneToIgnore:
Ok, had addressed all feedback, but please ask if anything's unclear, can be reviewed further.

Few action points: 
* there's now one global etcd and mock s3 servers in our python tests (s3 is still enabled per test, so test logic is still the same)
* a separate simple.conf set of fixes: https://github.com/neondatabase/neon/pull/1697
* all `broker_endpoints` vars were first made mandatory and a cloud PR was created to ensure that e2e tests pass and have endpoints: https://github.com/neondatabase/cloud/pull/1241
Then, the mandatory requirement was reverted, so now pageserver in this PR does nothing about the variables besides parsing, safekeeper issues a warning on empty endpoints

LizardWizzard:
Are we really sure that we want to start new etcd instance on per-test basis? IMO the better option is to introduce installation id in etcd paths as @arssher suggested and use it to namespace different test envs in a single etcd instance. 

Per test etcd set up will be extremely resource demanding for parallel test runs.

petuhovskiy:
I looked at the CI, the only failed job was `pg_regress-tests-debug` that failed with `RuntimeError: timed out waiting 294s for etcd start: HTTPConnectionPool(host='127.0.0.1', port=16001)`. Does it mean that this PR is almost ready now?

IMO we can merge this PR once all tests in CI are stable and fix other things later (better to create an issue for that).

SomeoneToIgnore:
Prefixes is a nice idea actually, had not implemented that, but considered the PR to be ready indeed, wanted to create a separate issue for better python-control_plane separation and had thought that the rest of the feedback got addressed.

I'll run the tests a few more times, to see what's wrong with the startups.

arssher (approved):
Perhaps remove `start_etcd_process` from `handle_start_all` if for now we use python version in tests? Fixtures don't use `start all`, but that's confusing on the first look.

SomeoneToIgnore:
I've managed to overcome the flackines via [somewhat obscure way](https://app.circleci.com/pipelines/github/neondatabase/neon/6406/workflows/62dcba74-8562-4744-b1ea-b2f994304e6b/jobs/64411/artifacts):
image

Apparently, the `@pytest.fixture(scope='session')` dependencies are different for parallel tests, so they start different etcds.
I could try and write even more code to detect an already running etcd, but decided to keep things easier for now and keep this set up.
If a few reruns won't show failures, I'll merge the PR.

SomeoneToIgnore:
Follow-up tasks:

* https://github.com/neondatabase/neon/issues/1716
* https://github.com/neondatabase/neon/issues/1717

I seem to have addressed everything by this point?

PR #1662, opened 2022-05-10
Remove WAL on safekeepers without s3 offloading.

arssher:
Until it is merged.

### DIFF
--- a/.circleci/ansible/systemd/safekeeper.service
+++ b/.circleci/ansible/systemd/safekeeper.service
@@ -6,7 +6,7 @@ After=network.target auditd.service
 Type=simple
 User=safekeeper
 Environment=RUST_BACKTRACE=1 ZENITH_REPO_DIR=/storage/safekeeper/data LD_LIBRARY_PATH=/usr/local/lib
-ExecStart=/usr/local/bin/safekeeper -l {{ inventory_hostname }}.local:6500 --listen-http {{ inventory_hostname }}.local:7676 -p {{ first_pageserver }}:6400 -D /storage/safekeeper/data --broker-endpoints={{ etcd_endpoints }}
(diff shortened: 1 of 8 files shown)


petuhovskiy (commented):
It seems that we wanted to do that only on staging. Should we add some temp configuration to distinguish staging from production?

arssher:
Hm. I hope s3 offloading lands before prod release, but let's do it optional anyway.

petuhovskiy (approved):
Ok, the next prod deployment is likely to be done by you, so you know what to fix before release :)

But option in config to trim without s3 still can be useful for capacity testing.

UPD: I missed the new changes before approval somehow, but I took a look and it's ok.

.circleci/ansible/systemd/safekeeper.service
```diff
@@ -6,7 +6,7 @@ After=network.target auditd.service
 Type=simple
 User=safekeeper
 Environment=RUST_BACKTRACE=1 ZENITH_REPO_DIR=/storage/safekeeper/data LD_LIBRARY_PATH=/usr/local/lib
-ExecStart=/usr/local/bin/safekeeper -l {{ inventory_hostname }}.local:6500 --listen-http {{ inventory_hostname }}.local:7676 -p {{ first_pageserver }}:6400 -D /storage/safekeeper/data --broker-endpoints={{ etcd_endpoints }}
+ExecStart=/usr/local/bin/safekeeper -l {{ inventory_hostname }}.local:6500 --listen-http {{ inventory_hostname }}.local:7676 -p {{ first_pageserver }}:6400 -D /storage/safekeeper/data --broker-endpoints={{ etcd_endpoints }} --enable-s3-offload={{ safekeeper_enable_s3_offload }}
```
petuhovskiy, line 9: What will happen to the old safekeeper when it'll get unknown option in the arguments? Will it ignore it silently?

arssher: No, it will crash and burn :) Uh, I wanted to avoid programming in ansible.

arssher: Hang on, how is it possible to get fresh service file without updating the binary?

arssher:
Got on more test_pageserver_catchup failure, created issue on that
https://github.com/neondatabase/neon/issues/1666

PR #1674, opened 2022-05-11
Make callmemaybe less agressive until we fix it/migrate to bigger machines.

arssher:


petuhovskiy (approved):
I would also give up callmemaybe activity on receiving `ERROR: Cannot load local timeline`

PR #1686, opened 2022-05-12
Correctly operate etcd safekeeper timeline data

SomeoneToIgnore:
Follow up of https://github.com/neondatabase/neon/pull/1635#pullrequestreview-967206057


safekeeper/src/timeline.rs
```diff
@@ -435,18 +435,10 @@ impl Timeline {
...
-                    wal_stream_connection_string(
-                        self.zttid,
-                        &shared_state.listen_pg_addr,
-                        pageserver_connstr,
-                    )
-                })
-                .transpose()
-                .context("Failed to get the pageserver callmemaybe connstr")?,
+            safekeeper_connection_string: format!(
+                "postgresql://no_user@{}/no_db",
```
arssher, line 439: Plain `listen_pg_addr` should work as well. And we can (should) just take it from config and never put to shared_state.

arssher (approved):

PR #1689, opened 2022-05-12
Create an initial timeline on CLI tenant creation

### LINKED ISSUE #1655: Neon CLI does not create an initial timeline on tenant creation
Based on https://neondb.slack.com/archives/C036U0GRMRB/p1650800300576519?thread_ts=1650799566.220579&cid=C036U0GRMRB

Currently, we use [`pageserver init`](https://github.com/neondatabase/neon/blob/d4e155aaa3b818981717e5b1a1ac6fb7af5cc9cd/neon_local/src/main.rs#L492-L507) during `neon_base init` cli command and that leaves us with a tenant and a timeline initialised for them, with `main` name on a timeline stored as branch into inside the `neon_base` config.

Yet when the tenant is [created](https://github.com/neondatabase/neon/blob/d4e155aaa3b818981717e5b1a1ac6fb7af5cc9cd/neon_local/src/main.rs#L529-L543) directly (that's the only other way to create a tenant **and** we don't allow multiple `neon init` commands), no timeline is created and the user needs to guess that he has to create one manually.

`pageserver init` behaviour is a separate question: by default, pageserver does not create any tenant or timeline except for this command for idempotence reasons.

But for the local CLI, it seems reasonable to ease developer's life and without any timelines, the tenant is pretty useless from the DB point of view, so we have to create a default timeline and register it in the `neon_base` mapping.

aome510:
Resolves #1655

### DIFF
--- a/neon_local/src/main.rs
+++ b/neon_local/src/main.rs
@@ -540,6 +540,33 @@ fn handle_tenant(tenant_match: &ArgMatches, env: &mut local_env::LocalEnv) -> an
                 "tenant {} successfully created on the pageserver",
                 new_tenant_id
             );
+
+            // Create an initial timeline for the new tenant
+            let new_timeline_id = parse_timeline_id(create_match)?;
+            let timeline = pageserver
+                .timeline_create(new_tenant_id, new_timeline_id, None, None)?
+                .ok_or_else(|| {
+                    anyhow!(
+                        "Failed to create initial timeline for tenant {}",
+                        new_tenant_id
+                    )
+                })?;
+            let new_timeline_id = timeline.timeline_id;
+            let last_record_lsn = timeline
+                .local
+                .expect("no local timeline info")
+                .last_record_lsn;
+
+            env.register_branch_mapping(
+                DEFAULT_BRANCH_NAME.to_string(),
+                new_tenant_id,
+                new_timeline_id,
+            )?;
+
+            println!(
+                "Created an initial timeline '{}' at Lsn {} for tenant: {}",
+                new_timeline_id, last_record_lsn, new_tenant_id
+            );
         }
         Some(("config", create_match)) => {
             let tenant_id = get_tenant_id(create_match, env)?;
--- a/test_runner/batch_others/test_zenith_cli.py
+++ b/test_runner/batch_others/test_zenith_cli.py
@@ -1,7 +1,7 @@
 import uuid
 import requests
 
-from fixtures.zenith_fixtures import ZenithEnv, ZenithEnvBuilder, ZenithPageserverHttpClient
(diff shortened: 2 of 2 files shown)


neon_local/src/main.rs
```diff
@@ -540,6 +540,33 @@ fn handle_tenant(tenant_match: &ArgMatches, env: &mut local_env::LocalEnv) -> an
                 "tenant {} successfully created on the pageserver",
                 new_tenant_id
             );
+
+            // Create an initial timeline for the new tenant
+            let new_timeline_id = parse_timeline_id(create_match)?;
+            let timeline = pageserver
+                .timeline_create(new_tenant_id, new_timeline_id, None, None)?
+                .ok_or_else(|| {
```
SomeoneToIgnore, line 548: NIT:

1. You can put identifiers (not expressions though) into `{}` brackets now:
` "Failed to create initial timeline for tenant {new_tenant_id}"` note that expressions are still not supported and something like this won't work: ` "Failed to create initial timeline for tenant {new_tenant_id.to_string()}"`

2. You can apply `anyhow::Context` trait methods onto `Option` directly:
`.timeline_create(new_tenant_id, new_timeline_id, None, None)?.with_context(|| format!("Failed to create initial timeline for tenant {new_tenant_id}"))?`

neon_local/src/main.rs
```diff
@@ -540,6 +540,33 @@ fn handle_tenant(tenant_match: &ArgMatches, env: &mut local_env::LocalEnv) -> an
...
+                    anyhow!(
+                        "Failed to create initial timeline for tenant {}",
+                        new_tenant_id
+                    )
+                })?;
+            let new_timeline_id = timeline.timeline_id;
+            let last_record_lsn = timeline
+                .local
+                .expect("no local timeline info")
```
SomeoneToIgnore, line 557: > `.expect`

Let's avoid panics if we can, `.context` or `.with_context` + ids would be safer: panic kills the entire thread and [exits](https://neondb.slack.com/archives/C036U0GRMRB/p1652201272837079?thread_ts=1652185280.790199&cid=C036U0GRMRB) the pageserver, which is bad.

SomeoneToIgnore (approved):
Thank you.

neon_local/src/main.rs
aome510: > NIT:
> 
> 1. You can put identifiers (not expressions though) into `{}` brackets now:
>    ` "Failed to create initial timeline for tenant {new_tenant_id}"` note that expressions are still not supported and something like this won't work: ` "Failed to create initial timeline for tenant {new_tenant_id.to_string()}"`
> 2. You can apply `anyhow::Context` trait methods onto `Option` directly:
>    `.timeline_create(new_tenant_id, new_timeline_id, None, None)?.with_context(|| format!("Failed to create initial timeline for tenant {new_tenant_id}"))?`

TIL about using `anyhow::Context` with `Option` :eyes:. Thanks

Issue #1565, opened 2022-04-22
GIT_VERSION always `local`


cicdteam:
On staging:

```console
$ pageserver --version
Zenith page server git-env:local profiling:false
```

As I see last several builds have the same "version".  CircleCi config is correct and it pass `CIRCLE_SHA1` as build arg in docker image build step, but...

Also from CircleCi logs:

<img width="765" alt="Снимок экрана 2022-04-22 в 18 48 39" src="https://user-images.githubusercontent.com/14281009/164749780-4d5bc7c8-fad5-4bbd-bc78-b2264c6234df.png">

I worry it can affect on cachepot settings as well @SomeoneToIgnore

SomeoneToIgnore:
We can disable it (by setting RUSTC_WRAPPER="") and check if it is to blame.

SomeoneToIgnore:
This is indeed seem to be related to `cachepot` indeed, since with disabled one a version appears.
It's odd that [`utils/build.rs`](https://github.com/neondatabase/neon/blob/main/libs/utils/build.rs) malfunctions though, since it was created for this specific invalidation task: might be a cachepot's bug.

There's a dirty way to fix the situation by using a macro like
```rust
#[macro_export]
macro_rules! project_git_version {
    () => {
        const GIT_VERSION: &str = git_version::git_version!(
            prefix = "git:",
            fallback = concat!(
                "git-env:",
                env!("GIT_VERSION", "Missing GIT_VERSION envvar")
            ),
            args = ["--abbrev=40", "--always", "--dirty=-modified"] // always use full sha
        );
    };
}
```

but that requires adding this macro invocation into every crate (instead of importing the `GIT_VERSION` as today) and adding a `git_version` crate into every `Cargo.toml`. This way no `build.rs` is needed and every binary gets a new version correctly.

I'd like to keep cachepot since it boosts builds [quite significantly](https://github.com/neondatabase/cloud/pull/1033#issuecomment-1100935036), so might use the macro-based approach if won't find a better way to tweak build.rs and cachepot.

SomeoneToIgnore:
```
admin@zenith-us-stage-ps-2:~$ /usr/local/bin/pageserver --version
Zenith page server git-env:cded72a580266d978fee5260be9e0e56abbb42b9 profiling:false failpoints:true
```

Looks like it works.

PR #1694, opened 2022-05-12
Force actual GIT_VERSION via declaration macros

SomeoneToIgnore:
Closes https://github.com/neondatabase/neon/issues/1565

I'm not getting down to the root cause in the near future and I think it's better to have a proper version and build caching.
So it might make sense to fix it at least this way and increment a proper fix on top, when somebody gets more time for that.


dhammika (approved):
:+1: 
Looks like this is a known issue in cachepot,
https://github.com/paritytech/cachepot/issues/137
Sort of scary if it's widespread and could link bad code from prior builds but I didn't find anything along those lines.

funbringer (approved):

PR #1653, opened 2022-05-06
Safekeeper: add timeline/tenant force delete HTTP endpoings

yeputons:
This implements the basic version of #895.

* There is no auth in Safekeeper HTTP at all currently, so simply calling `check_permission` (like in Pageserver) is not enough.
* There are no checks of Safekeeper still working with the data, as "still working" is burry now: a timeline may be "active" while there are no compute nodes and all data is propagated.
* However, there is an attempt to report at least some information about "active" timelines (even if inaccurate) back to the caller. I'm not sure if it's information useful, but it's there just in case. Maybe for some metrics or whatnot.

PR #1696, opened 2022-05-13
Fix test_delete_force from #1653 conflicting with #1692

yeputons:


aome510 (approved):
LGTM! Thanks for the fix

PR #1697, opened 2022-05-13
Test simple.conf and handle broker_endpoints better

SomeoneToIgnore:
A follow-up of https://github.com/neondatabase/neon/pull/1648#discussion_r872232041, moved away separately since touches a bit more than etcd.
I intentionally did not work extensively on reworking the related bits, only on the things that help to be sure that our configs are not forgotten and parsed properly.

* replaces `String` with `Url` in types
* moves binary existence verification logic into the `init` function instead of parsing
* adds tests on parsing that covers the wrong url case above


aome510 (approved):
LGTM. I guess we'll need to wait for https://github.com/neondatabase/neon/pull/1696 to land to see if the tests passed in this PR.

petuhovskiy (approved):

PR #1746, opened 2022-05-18
Safekeeper: avoid holding mutex when deleting a tenant

yeputons:
Following discussion with @arssher after #1653

### DIFF
--- a/safekeeper/src/timeline.rs
+++ b/safekeeper/src/timeline.rs
@@ -679,29 +679,34 @@ impl GlobalTimelines {
 
     /// Deactivates and deletes all timelines for the tenant, see `delete()`.
     /// Returns map of all timelines which the tenant had, `true` if a timeline was active.
+    /// There may be a race if new timelines are created simultaneously.
     pub fn delete_force_all_for_tenant(
         conf: &SafeKeeperConf,
         tenant_id: &ZTenantId,
     ) -> Result<HashMap<ZTenantTimelineId, TimelineDeleteForceResult>> {
         info!("deleting all timelines for tenant {}", tenant_id);
-        let mut state = TIMELINES_STATE.lock().unwrap();
-        let mut deleted = HashMap::new();
-        for (zttid, tli) in &state.timelines {
-            if zttid.tenant_id == *tenant_id {
-                deleted.insert(
-                    *zttid,
-                    GlobalTimelines::delete_force_internal(
-                        conf,
-                        zttid,
-                        tli.deactivate_for_delete()?,
-                    )?,
-                );
+        let mut to_delete = HashMap::new();
+        {
+            let mut state = TIMELINES_STATE.lock().unwrap();
+            for (zttid, tli) in &state.timelines {
+                if zttid.tenant_id == *tenant_id {
+                    to_delete.insert(*zttid, tli.deactivate_for_delete()?);
+                }
             }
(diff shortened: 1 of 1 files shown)


safekeeper/src/timeline.rs
```diff
@@ -679,29 +679,34 @@ impl GlobalTimelines {
...
-        state
-            .timelines
-            .retain(|zttid, _| !deleted.contains_key(zttid));
+        let mut deleted = HashMap::new();
+        for (&zttid, &was_active) in &to_delete {
+            deleted.insert(
+                zttid,
+                GlobalTimelines::delete_force_internal(conf, &zttid, was_active)?,
+            );
+        }
```
yeputons, line 707: I'd be delighted to rewrite this using `map` or something, but it cannot handle `Result<>` errors. Which makes sense: `map` is typically side-effect-free. In this case, it should be fine because we can always issue the deletion request one more time, and it will work.

yeputons: Well, it can actually be written like

```rust
        let deleted = to_delete
            .iter()
            .map(|(&zttid, &was_active)| {
                GlobalTimelines::delete_force_internal(conf, &zttid, was_active)
                    .map(|response| (zttid, response))
            })
            .collect::<Result<HashMap<ZTenantTimelineId, TimelineDeleteForceResult>>>()?;
```

but I prefer the imperative version: it a) makes it clear that the first error is returned; b) actually stops processing on the first error instead of smartly "collecting" all intermediate errors and suppressing everything except the first.

aome510: I just looked into this briefly and came up with a similar solution:
```rust
        let deleted = to_delete
            .into_iter()
            .map(|(zttid, was_active)| {
                Ok((
                    zttid,
                    GlobalTimelines::delete_force_internal(conf, &zttid, was_active)?,
                ))
            })
            .collect::<Result<HashMap<_, _>>>()?;
```

You don't need to explicitly specify the inner types for container in `std::iter::collect`.

That said, I agree that the imperative version is better b/c of the above reasons

aome510: Update: dig into this a bit deeper. `std::iter::collect` actually does stop processing on the first error (reference: [rust doc](https://doc.rust-lang.org/std/result/enum.Result.html#impl-FromIterator%3CResult%3CA%2C%20E%3E%3E)). 

Also, using this map/collect approach may reduce the runtime compared to `n` independent insertions (I'm not so sure about how `collect` works under the hood for `Hashmap`)

yeputons:
I've tried writing in a more functional style in df07426c based on @aome510's suggestions, but I still like ba68ab4 more. Too few operations, so unusual `collect::<Result<....>>` stands out and makes reading harder.

safekeeper/src/timeline.rs
```diff
@@ -687,25 +687,23 @@ impl GlobalTimelines {
...
             }
             // TODO: test that the correct subset of timelines is removed. It's complicated because they are implicitly created currently.
-            state
-                .timelines
-                .retain(|zttid, _| !to_delete.contains_key(zttid));
+            timelines.retain(|zttid, _| !to_delete.contains_key(zttid));
         }
         let mut deleted = HashMap::new();
-        for (&zttid, &was_active) in &to_delete {
+        for (zttid, was_active) in to_delete {
```
yeputons, line 701: This for loop consumes `to_delete`

PR #1781, opened 2022-05-23
Potential fix to #1626. Fixed typo is Makefile.

KlimentSerafimov:
Changed pageserver.rs to bail in addition to printing an error when there is an error thrown when creating a pageserver.
Potential fix to: https://github.com/neondatabase/neon/issues/1626
Issue and determining source of error by Heikki, fixed with help from Stas.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ else ifeq ($(BUILD_TYPE),debug)
 	PG_CONFIGURE_OPTS = --enable-debug --with-openssl --enable-cassert --enable-depend
 	PG_CFLAGS = -O0 -g3 $(CFLAGS)
 else
-$(error Bad build type `$(BUILD_TYPE)', see Makefile for options)
+	$(error Bad build type '$(BUILD_TYPE)', see Makefile for options)
 endif
 
 # macOS with brew-installed openssl requires explicit paths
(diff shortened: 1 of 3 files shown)


SomeoneToIgnore (approved):
Let's also address the [comment](https://github.com/neondatabase/neon/issues/1626#issuecomment-1117037648) and fix the same thing in safekeeper:

https://github.com/neondatabase/neon/blob/134eeeb096de28c44c8fc7de1d771ed5350598c2/safekeeper/src/bin/safekeeper.rs#L248

KlimentSerafimov:
I don't know how to recreate the bug for the safekeeper code.
I tried running 

> ./target/debug/safekeeper --id 2 -D .zenith --daemonize

to start a safekeeper  and then

> ./target/debug/safekeeper --id 3 -D .zenith --daemonize

to start a second one, but it gives error

```
Error: control file ./safekeeper.lock is locked by some other process

Caused by:
    Resource temporarily unavailable (os error 35)

```

When recreating the pageserver bug, we simply changed the listen_http_addr and listen_pg_addr to point to different sockets so there is no error thrown before the bug we want to recreate, but for the safekeeper code I don't know how to bypass the error above. Alternatively I tried

> ./target/debug/safekeeper --id 3 -D .zenith --daemonize --broker-endpoints http://127.0.0.1:2379

After reading the help instructions, but the same error persists.

How do I recreate the bug on for the safekeeper code?

SomeoneToIgnore:
No idea how to repro the case and not sure why it's needed.

According to the [comment](https://github.com/neondatabase/neon/issues/1626#issuecomment-1117036967)
> So, if the Daemonize::start() call fails, we print the error, but continue running. Surely we should exit on failure.

So, that `error` is wrong and we can `bail` instead, no repro is needed, especially now, when the pageserver counterpart is adjusted, it's better to tread daemonisation errors the same way.

kelvich:
+1 to @SomeoneToIgnore point. High chances that with a safekeeper current error handling does not lead to any real problem. However, it still makes sense to fix it because it clearly an error and may resurface with some future refactoring.

PR #1798, opened 2022-05-25
add simple metrics for remote storage operations

LizardWizzard:
track number of operations and number of their failures


SomeoneToIgnore (approved):
Not sure how good it is to put so many functions into an S3 wrapper, but it's better than no observability for sure.

LizardWizzard:
> Not sure how good it is to put so many functions into an S3 wrapper

Since it is used in both pageserver and safekeeper looks like a good common place to have them :)

PR #1811, opened 2022-05-27
s3 WAL offloading staging review.

arssher:
- Uncomment accidently `self.keep_alive.abort()` commented line, due to this
  task never finished, which blocked launcher, which overflowed the channel and blocked everything.
- Mess up with initialization one more time, to fix offloader trying to back up
  segment 0. Now we initialize all required LSNs in handle_elected,
  where we learn start LSN for the first time.
- Fix blind attempt to provide safekeeper service file with remote storage
  params.

### DIFF
--- a/libs/utils/src/zid.rs
+++ b/libs/utils/src/zid.rs
@@ -218,7 +218,7 @@ impl ZTenantTimelineId {
 
 impl fmt::Display for ZTenantTimelineId {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}-{}", self.tenant_id, self.timeline_id)
+        write!(f, "{}/{}", self.tenant_id, self.timeline_id)
     }
 }
 
--- a/safekeeper/src/safekeeper.rs
+++ b/safekeeper/src/safekeeper.rs
@@ -731,24 +731,36 @@ where
         {
             let mut state = self.state.clone();
 
-            // Remeber point where WAL begins globally, if not yet.
+            // Here we learn initial LSN for the first time, set fields
+            // interested in that.
+
             if state.timeline_start_lsn == Lsn(0) {
+                // Remember point where WAL begins globally.
                 state.timeline_start_lsn = msg.timeline_start_lsn;
                 info!(
                     "setting timeline_start_lsn to {:?}",
                     state.timeline_start_lsn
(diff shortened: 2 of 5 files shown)


libs/utils/src/zid.rs
```diff
@@ -218,7 +218,7 @@ impl ZTenantTimelineId {
 
 impl fmt::Display for ZTenantTimelineId {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}-{}", self.tenant_id, self.timeline_id)
+        write!(f, "{}/{}", self.tenant_id, self.timeline_id)
```
petuhovskiy, line 221: What this change is about?

arssher: Ah, it is for easier copypaste of log pieces into various places, for example safekeeper timeline status endpoint accepts {tenant_id}/{timeline_id}, not {tenant_id}-{timeline_id}.

safekeeper/src/safekeeper.rs
```diff
@@ -731,24 +731,36 @@ where
...
             }
+            // Initializing commit_lsn before acking first flushed record is
+            // important to let find_end_of_wal skip the whole in the beginning
+            // of the first segment.
+            //
+            // NB: on new clusters, this happens at the same time as
+            // timeline_start_lsn initialization, it is taken outside to provide
+            // upgrade.
+            self.global_commit_lsn = max(self.global_commit_lsn, state.timeline_start_lsn);
+            self.inmem.commit_lsn = max(self.inmem.commit_lsn, state.timeline_start_lsn);
```
petuhovskiy, line 756: It seems that now we will have `commit_lsn > flush_lsn` when we set commit_lsn for the first time here. flush_lsn is still zero, but we already set commit_lsn.

Moreover, assert in `update_commit_lsn` will probably fail. Also zero `flush_lsn` will be reported to compute, and compute can respond with zero `commit_lsn`?

petuhovskiy (approved):
Tests have passed, so it's better to merge this now to fix staging and fix other problems later.

safekeeper/src/safekeeper.rs
arssher: https://github.com/neondatabase/neon/pull/1812

PR #1812, opened 2022-05-27
Prevent commit_lsn <= flush_lsn violation after a42eba3cd7.

arssher:
Nothing complained about that yet, but we definitely don't hold at least one assert, so let's keep it this way until better version.


petuhovskiy (approved):

arssher:
test_wal_acceptor.py::test_wal_backup[mock_s3] failed for unclear reasons, which is upsettings, but I don't think this PR is related.

PR #1823, opened 2022-05-28
Implement timeline-based metrics in safekeeper

petuhovskiy:
Some of these metrics were implemented before, but now they are reimplemented with this new approach. I can't say new approach is better in every aspect, there are differences.

What's better:
- It's easier to add metrics, no need to insert `.set` in every place near modification
- Metrics now are a consistent snapshot, so should be no issues like https://github.com/neondatabase/neon/issues/1470
- Old metrics for inactive timelines won't show up in `/metrics` response

What's worse:
- We need to take mutex locks, that can affect metrics latency and timelines processing

Some places in timelines collector are not performance-optimized, let's not have premature optimizations until there'll be problems with it.

### DIFF
--- a/safekeeper/src/metrics.rs
+++ b/safekeeper/src/metrics.rs
@@ -1,11 +1,11 @@
 //! This module exports metrics for all active timelines.
 
-use std::time::{SystemTime, Instant};
+use std::time::{Instant, SystemTime};
 
 use metrics::{
     core::{AtomicU64, Collector, Desc, GenericGaugeVec, Opts},
     proto::MetricFamily,
-    IntGaugeVec, Gauge,
+    Gauge, IntGaugeVec,
 };
 use postgres_ffi::xlog_utils::XLogSegNo;
 use utils::{lsn::Lsn, zid::ZTenantTimelineId};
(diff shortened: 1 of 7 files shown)


safekeeper/src/metrics.rs
```diff
@@ -0,0 +1,336 @@
...
+        self.acceptor_term.reset();
+
+        let timelines = GlobalTimelines::active_timelines_metrics();
+
+        for tli in timelines {
+            let tenant_id = tli.zttid.tenant_id.to_string();
+            let timeline_id = tli.zttid.timeline_id.to_string();
+            let labels = &[tenant_id.as_str(), timeline_id.as_str()];
+
+            let mut most_advanced: Option<utils::pq_proto::ZenithFeedback> = None;
```
arssher, line 245: There is `get_replicas_state` which does this. Though it is kinda broken anyway, we should switch to tracking on per-pageserver (e.g. keyed by node id) basis.

safekeeper/src/timeline.rs
```diff
@@ -777,6 +805,16 @@ impl GlobalTimelines {
             .collect()
     }
 
+    /// Return FullTimelineInfo for all active timelines.
+    pub fn active_timelines_metrics() -> Vec<FullTimelineInfo> {
+        let state = TIMELINES_STATE.lock().unwrap();
```
arssher, line 810: Holding both map lock and per-tli lock during the whole scrape isn't cool and it is easy to avoid, but it is ok to see the real problem first.

arssher (approved):

safekeeper/src/timeline.rs
petuhovskiy: Right, I just did the same as in `get_active_timelines`. If scrape will take a substantial time (100ms+) then we'll get back to it.

safekeeper/src/metrics.rs
petuhovskiy: Yeah, but AFAIK `get_replicas_state` is timestamp based. It's not a big difference, I just wanted to track minimal sk->ps lag here.

safekeeper/src/metrics.rs
```diff
@@ -0,0 +1,336 @@
...
+                    .set(feedback.ps_writelsn);
+                if let Ok(unix_time) = feedback.ps_replytime.duration_since(SystemTime::UNIX_EPOCH)
+                {
+                    self.feedback_last_time_seconds
+                        .with_label_values(labels)
+                        .set(unix_time.as_secs());
+                }
+            }
+
+            if tli.last_removed_segno != 0 {
```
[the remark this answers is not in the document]
arssher: I'm ok keeping at as is as, in case of unknown it is not hard to calculate disk usage manually. Or properly init last_removed_segno when timeline is loaded by checking oldest available segment.

petuhovskiy:
Performance stats: https://observer.zenith.tech/goto/l1z5xyr7z?orgId=1

PR #1833, opened 2022-05-30
Use `HOMEBREW_PREFIX` instead of hard-coded path

aome510:


aome510:
Context: on my laptop, `$HOMEBREW_PREFIX`  is set to `/opt/homebrew`. I guess, on others, it is set `/usr/local`

PR #1834, opened 2022-05-31
Close file descriptors for redo process

bojanserafimov:
Fixes https://github.com/neondatabase/neon/issues/1814

Tested manually using `lsof .zenith/pageserver.pid`

### DIFF
--- a/pageserver/src/walredo.rs
+++ b/pageserver/src/walredo.rs
@@ -569,7 +569,19 @@ impl<C: CommandExt> CloseFileDescriptors for C {
     fn close_fds(&mut self) -> &mut Command {
         unsafe {
             self.pre_exec(move || {
-                // close_fds crate is safe to use in this context (see docs)
+                // SAFETY: Code executed inside pre_exec should have async-signal-safety,
+                // which means it should be safe to execute inside a signal handler.
+                // The precise meaning depends on platform. See `man signal-safety`
+                // for the linux definition.
+                //
+                // The set_fds_cloexec_threadsafe function is documented to be
+                // async-signal-safe.
+                //
+                // Aside from this function, the rest of the code is re-entrant and
+                // doesn't make any syscalls. We're just passing constants.
+                //
+                // NOTE: It's easy to indirectly cause a malloc or lock a mutex,
+                // which is not async-signal-safe. Be careful.
                 close_fds::set_fds_cloexec_threadsafe(3, &[]);
                 Ok(())
             })
@@ -644,6 +656,18 @@ impl PostgresRedoProcess {
             .env("LD_LIBRARY_PATH", conf.pg_lib_dir())
             .env("DYLD_LIBRARY_PATH", conf.pg_lib_dir())
             .env("PGDATA", &datadir)
+            // The redo process is not trusted, so it runs in seccomp mode
+            // (see seccomp in zenith_wal_redo.c). We have to make sure it doesn't
+            // inherit any file descriptors from the pageserver that would allow
+            // an attacker to do bad things.
+            //
+            // The Rust standard library makes sure to mark any file descriptors with
+            // as close-on-exec by default, but that's not enough, since we use
+            // libraries that directly call libc open without setting that flag.
+            //
+            // One example is the pidfile of the daemonize library, which doesn't
+            // currently mark file descriptors as close-on-exec. Either way, we
+            // want to be on the safe side and prevent accidental regression.
             .close_fds()
             .spawn()
             .map_err(|e| {
--- a/pageserver/src/walredo.rs
+++ b/pageserver/src/walredo.rs
@@ -556,6 +556,28 @@ impl PostgresRedoManager {
 }
 
 ///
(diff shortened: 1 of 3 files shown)


hlinnaka (approved):
Let's add a comment somewhere to explain why it's important to close the file descriptors:
1. It's a security issue if we don't. The WAL process is a sandbox that's not supposed to be able to access anything in the parent process.
2. The concrete problem with the lock file.

pageserver/src/walredo.rs
```diff
@@ -554,6 +555,28 @@ impl PostgresRedoManager {
...
+    /// Close file descriptors (other than stdin, stdout, stderr) in child process
+    ///
+    fn close_fds(&mut self) -> &mut Command;
+}
+
+impl<C: CommandExt> CloseFileDescriptors for C {
+    fn close_fds(&mut self) -> &mut Command {
+        unsafe {
+            self.pre_exec(move || {
+                // close_fds crate is safe to use in this context (see docs)
```
SomeoneToIgnore, line 572: NIT: Simple folk like me would really thank you for writing a SAFETY paragraph with explicit explanation, why it's safe.
One of the many examples: https://github.com/rust-lang/rust/blob/a28109a7679fbb254ca0962865a8b32b91ddbbd9/library/std/src/sys/sgx/waitqueue/unsafe_list.rs#L41

That would also help to validate the assumptions: "see docs" would require a person to construct his own and guess for you instead, that does not help to validate things simple.

phoenix24 (approved):

SomeoneToIgnore (approved):
Thank you!

hlinnaka (approved):

funbringer:
I believe this solution is suboptimal. If the goal is to fix just #1814, it would suffice to ensure that somebody sets the `FD_CLOEXEC` flag for the pidfile's fd. Since we use https://github.com/knsd/daemonize, this is exactly the entity which should do that for us. Unfortunately, that's not the case for the version [0.4.1](https://github.com/knsd/daemonize/blob/b59073bf3e51870a6a8bb44f954faa1b38b20d2d/src/lib.rs#L423). Compare this to the main branch: the code is [there](https://github.com/knsd/daemonize/blob/f7be28efa1b4a70e43bb37b5f4ff4d664992edca/daemonize/src/lib.rs#L401), but the fix hasn't made it to the crates.io yet (I wonder why).

One could argue that there may be more fds like that. Fortunately, std already uses `cloexec` where possible (see the docs for [exec](https://doc.rust-lang.org/std/os/unix/process/trait.CommandExt.html#tymethod.exec)). This leads me to think that we should do two three things instead:

* Somehow use the latest version of daemonize, or better yet, get rid of it. [That's not how one should daemonize services in this day and age anyway](https://0pointer.de/public/systemd-man/daemon.html#New-Style%20Daemons).
* Audit the whole project for other places where we don't set `cloexec` either due to the broken deps or sloppiness with libs like nix.
* Insert an `assert!` to the `pre_exec` to blow up the fork if we detect any stray fds.

bojanserafimov:
> I believe this solution is suboptimal. If the goal is to fix just #1814, it would suffice to ensure that somebody sets the `FD_CLOEXEC` flag for the pidfile's fd. Since we use https://github.com/knsd/daemonize, this is exactly the entity which should do that for us. Unfortunately, that's not the case for the version [0.4.1](https://github.com/knsd/daemonize/blob/b59073bf3e51870a6a8bb44f954faa1b38b20d2d/src/lib.rs#L423). Compare this to the main branch: the code is [there](https://github.com/knsd/daemonize/blob/f7be28efa1b4a70e43bb37b5f4ff4d664992edca/daemonize/src/lib.rs#L401), but the fix hasn't made it to the crates.io yet (~I wonder why,~ the project seems to be dead: last commit on 31 Jul 2021).
> 
> One could argue that there may be more fds like that. Fortunately, std already uses `cloexec` where possible (see the docs for [exec](https://doc.rust-lang.org/std/os/unix/process/trait.CommandExt.html#tymethod.exec)). This leads me to think that we should do two three things instead:
> 
>     * Somehow use the latest version of daemonize, or better yet, get rid of it. [That's not how one should daemonize services in this day and age anyway](https://0pointer.de/public/systemd-man/daemon.html#New-Style%20Daemons).
> 
>     * Audit the whole project for other places where we don't set `cloexec` either due to the broken deps or sloppiness with libs like nix.
> 
>     * Insert an `assert!` to the `pre_exec` to blow up the fork if we detect any stray fds.

Yes https://github.com/neondatabase/neon/issues/1814#issuecomment-1141365290

Short term: Would you rather deal with unreleased commits from daemonize, or this new close_fds crate? We might need this crate anyway for the pre_exec assertion that you mention, unless there's a simpler way to correctly iterate open fds in a multi-thread program. Or alternatively we can do this check after exec, in the postgres code.

Long term: Moving away from daemonize also fixes https://github.com/neondatabase/neon/issues/1840, among other things. It's a bigger project though, we shouldn't block `a/reliability` fixes on this transition.

PR #1835, opened 2022-05-31
Do not skip page header for basebackup LSN

knizhnik:


arssher:
Looks ok, but why it was there in the first place? Perhaps we should do the reverse, i.e. `normalize_lsn` at safekeeper side?

kelvich:
That kind of change should come with a rigorous explanation of why it won't cause a crash during Postgres launch, and I think it actually will crash -- this normalization was to please Postgres. We now have a decent restriction -- pageserver returns normalized LSN's. And instead of imposing this rule on a safekeeper too, this patch removes restrictions at all.

+1 to coding it in the safekeepers (ideally but not sure how that will play with other code there) or in the wal_proposeser (normalize LSN tight after reading it).

arssher:
I force pushed to run CI.

I also examined `StartupXLOG` to check whether something on compute side requires skipped page header LSN. As far as I see nothing. `XLogBeginRead` is ok with page header.

The only thing I see obviously incorrect is
```
			xlogPageHdr->xlp_info = XLP_FIRST_IS_CONTRECORD; // FIXME
			xlogPageHdr->xlp_rem_len = offs - SizeOfXLogShortPHD;
```
but I doubt anyone ever reads this page header.

kelvich:
> XLogBeginRead is ok with page header.

`XLogRecPtrToBytePos()` asserts `Assert(offset >= SizeOfXLogLongPHD);`
`StartupXLOG()` calls `XLogRecPtrToBytePos()`

arssher (approved):

arssher:
Did it per above. @knizhnik please have a look.

PR #1836, opened 2022-05-31
Bump vendor/postgres to hotfix basebackup LSN comparison.

arssher:


kelvich:
e2e failed with:

```
Compute logs from project 'aged-darkness-798983' and test 'test_creation_stress'

thread 'main' panicked at 'pageserver connstr should be provided', compute_tools/src/bin/compute_ctl.rs:120:10
stack backtrace:
   0: rust_begin_unwind
             at ./rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/std/src/panicking.rs:498:5
   1: core::panicking::panic_fmt
             at ./rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/panicking.rs:107:14
   2: core::panicking::panic_display
             at ./rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/panicking.rs:63:5
   3: core::option::expect_failed
             at ./rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/option.rs:1637:5
   4: core::option::Option<T>::expect
             at ./rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/option.rs:709:21
   5: compute_ctl::main
             at ./home/circleci/project/compute_tools/src/bin/compute_ctl.rs:116:30
   6: core::ops::function::FnOnce::call_once
             at ./rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/ops/function.rs:227:5
```

@lubennikovaav @ololobus is that due to the extension rename?

ololobus:
>  is that due to the extension rename?

Yes, because of `zenith.page_server_connstring -> neon.pageserver_connstring`. Is it tries to push a rebased Postgres branch?

kelvich:
Right. @arssher you took old Postgres branch that lacks two recent commits

arssher:
I took 'release' one. Probably just merge the whole 'main' to 'release' then?

kelvich:
Ah, got it. This PR is intended for production. @ololobus would console pick up new compute if we'll merge this?

ololobus:
> would console pick up new compute if we'll merge this?

No, because storage is older

arssher:
Re-pushed with vendor/postgres forked from latest main, but I'm confused now. Which compute version would console pick up? Obviously that rename is not prod yet, otherwise vendor/postgres `release` would rename it. So this is just e2e tests problem?

ololobus:
> Re-pushed with vendor/postgres forked from latest main, but I'm confused now.

Not sure this is right way to do it. Main uses renamed GUCs, which are not supported by the prod console. So the forking out of release was a right way to do it. I just hadn't noticed that this is a PR to release

> Which compute version would console pick up?

Console picks the minimal version tag across all assigned storage node version to maintain backward compatibility. So if you just push a new compute image it won't pick it up automatically. Once this is merged it will rollout storage partially, I guess. We only deploy safekeepers now

> Obviously that rename is not prod yet, otherwise vendor/postgres release would rename it. So this is just e2e tests problem?

Yes, e2e is running the latest console, because `neon` doesn't have any option to reference some particular console version/tag

ololobus:
> Console picks the minimal version tag across all assigned storage node version to maintain backward compatibility. So if you just push a new compute image it won't pick it up automatically. Once this is merged it will rollout storage partially, I guess. We only deploy safekeepers now

To solve this @cicdteam previously just re-tagged new compute as the old PS version too. But maybe we can just bump old prod PS version in the console db manually? It doesn't look that right, but are there any problems this can cause?

arssher:
> Console picks the minimal version tag across all assigned storage node version to maintain backward compatibility. So if you just push a new compute image it won't pick it up automatically. 

Hm. We live for quite some time on staging with old and new pageservers, how it works then? Surely we don't run old computes?

> Once this is merged it will rollout storage partially, I guess. We only deploy safekeepers now
My reading of production.hosts says we deploy both safekeepers and new pageserver (2).

ololobus:
> how it works then? Surely we don't run old computes?

Because of the hack with re-tagging, I suppose

>  both safekeepers and new pageserver (2).

Ah, we use new pageserver on prod already, didn't know about that. Anyway, to make this properly working we need either bump old PS version in the console db, or do the same re-tag trick as @cicdteam did before

ololobus:
BTW, is this fix already merged to main and tested on staging? What is the corresponding PR / branch in `postgres` repo? I cannot find one

kelvich:
+1 we anyway will need a process for hotfixes. And one with relabeling seems to be okay. Basically, it means that if we have version 123 we can also create version 123.hotfix-1. It will be correct if all the next versions have the same fix in them.

kelvich:
But let's have a checklist for ad-hoc prod DB edits

PR #1838, opened 2022-05-31
Use :local compute-tools tag to build compute-node image

ololobus:
Postgres PR https://github.com/neondatabase/postgres/pull/170
Cloud PR https://github.com/neondatabase/cloud/pull/1446

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -453,9 +453,6 @@ jobs:
       - checkout
       - setup_remote_docker:
           docker_layer_caching: true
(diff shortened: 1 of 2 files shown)


kelvich (approved):
LGTM, thanks?

Do we even need to push `compute-tools` to the dockerhub?

ololobus:
> Do we even need to push compute-tools to the dockerhub?

I don't know, previously I used it from time to time if I needed to build compute-node locally, but didn't want to rebuild everything. Sounds like a very rare case, though. Yet, it shouldn't hurt is we push a couple more images

PR #1839, opened 2022-05-31
improve test normal work to start several computes

LizardWizzard:
This way it reproduces test failure caught by e2e tests in #1619

### DIFF
--- a/test_runner/batch_others/test_normal_work.py
+++ b/test_runner/batch_others/test_normal_work.py
@@ -1,0 +1,47 @@
+from fixtures.log_helper import log
+from fixtures.zenith_fixtures import ZenithEnv, ZenithEnvBuilder, ZenithPageserverHttpClient
+
+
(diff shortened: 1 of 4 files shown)


test_runner/batch_others/test_normal_work.py
```diff
@@ -0,0 +1,38 @@
...
+    assert res[0] == (5000050000, )
+
+    pageserver_http.timeline_detach(tenant_id, timeline_id)
+
+
+def test_normal_work(zenith_env_builder: ZenithEnvBuilder):
+    """
+    Basic test, write something in setup with wal acceptors,
+    ensure that successfully commits.
+    Check several timelines.
```
SomeoneToIgnore, line 31: NIT: Might worth adding that a new tenant and timeline creation & detach happens here.

SomeoneToIgnore (approved):

PR #1848, opened 2022-06-01
GitHub Actions: pin Rust version to 1.58 like on CircleCI

yeputons:


yeputons:
Actually, let's add Rust version to the caching key so they do not collide between branches during update.

yeputons:
There are two pending checks: `run regression test suite (stable, macos-latest)` and `run regression test suite (stable, ubuntu-latest)`. They are superseded by `Build and Test / run regression test suite (1.58, ubuntu-latest)` and `run regression test suite (1.58, macos-latest) (pull_request)`, correspondingly. I'm not sure why the former are still expected by GitHub in this PR, but I think it is not worth a deep investigation.

I'm merging this PR and observing whether the set of checks on the `main` branch is correct.

yeputons:
Ah, I cannot merge until these statuses pass. Trying [a workaround from StackOverflow](https://stackoverflow.com/a/67986789): close and re-open this PR to update list of checks.

yeputons:
Ok, it did not help. Rebasing manually. If it does not help as well, will just push the rebased commit to `main`.

yeputons:
Yes, it did not help, these checks are still pending for whatever reason, even though they're removed in this PR. Pushing straight to `main` and monitoring for breakages.

изображение

yeputons:
I don't have permission to push to `main` (understandably). I suspect that this PR changes the set of workflows (actually, the collection of required checks); hence old (removed) workflows cannot pass as (probably) needed by the repository's settings.

pageserver/src/walreceiver.rs
```diff
@@ -469,12 +469,9 @@ async fn timeline_wal_broker_loop_step(
             updates = broker_subscription.fetch_data() => match updates {
                 Some(mut all_timeline_updates) => {
                     if let Some(subscribed_timeline_updates) = all_timeline_updates.remove(&id) {
-                        match wal_connection_manager.select_connection_candidate(subscribed_timeline_updates) {
-                            Some(candidate) => {
-                                info!("Switching to different safekeeper {} for timeline {id}, reason: {:?}", candidate.safekeeper_id, candidate.reason);
-                                wal_connection_manager.change_connection(candidate.safekeeper_id, candidate.wal_producer_connstr).await;
-                            },
-                            None => {}
+                        if let Some(candidate) = wal_connection_manager.select_connection_candidate(subscribed_timeline_updates) {
```
arssher, line 472: is this accidental?

yeputons: No, it's because Clippy in 1.58 is somewhat stricter than in 1.60, see https://github.com/rust-lang/rust-clippy/issues/8928

PR #1851, opened 2022-06-01
Fix url path printing in etcd CLI

SomeoneToIgnore:
Based on https://neondb.slack.com/archives/C036U0GRMRB/p1654113394396859?thread_ts=1654082043.264319&cid=C036U0GRMRB


SomeoneToIgnore:
The only place that does not like URLs is the etcd CLI, the etcd-client Rust crate does not mind trailing `/`

aome510 (approved):
Work for me now! Thanks for the quick fix

Issue #1855, opened 2022-06-01
Require authentication for safekeeper/pageserver metrics


yeputons:
Currently `endpoint::make_router()` exposes `/metrics` without any checks. However, it may contain relatively sensitive information like tenant/timeline IDs. We should require a managemental token instead.

Careful: not to break Grafana.

Careful: implement after #1856 is merged, otherwise there is not much auth in Safekeeper.

Careful: do not merge until #1856 is correctly configured on staging/prod, requires configuration updates. Even then, coordinate with SREs.

stepashka:
we don't need to auth metrics, as they contain aggregated info and anonymized ids only
it makes much more sense to clean up logs first

Issue #1854, opened 2022-06-01
 Implement the authentication in safekeeper Postgres API


yeputons:
Implement based on #1856, otherwise there is not much auth in Safekeeper.

Careful: do not merge until #1856 is correctly configured on staging/prod, this issue requires configuration updates. Even then, coordinate with SREs.

LizardWizzard:
Related #451

PR #1856, opened 2022-06-01
Safekeeper HTTP API auth (closes #1753)

### LINKED ISSUE #1753: Implement the authentication in safekeeper HTTP API
Probably akin to Pageserver's. Especially interesting after #1653 , we don't want random timeline deletion from safekeepers.

yeputons:
Closes #1753. My understanding is that #1068 is irrelevant here.

Follow-up issues:

* #1854
* #1855

**Careful**: do not merge until consulted with SREs whether this will work on staging and/or break the console. Some extra configuration may be needed to enable it.

### DIFF
--- a/safekeeper/src/http/routes.rs
+++ b/safekeeper/src/http/routes.rs
@@ -222,6 +222,7 @@ pub fn make_router(
             "/v1/timeline/:tenant_id/:timeline_id",
             timeline_status_handler,
         )
+        // Will be used in the future instead of implicit timeline creation
         .post("/v1/timeline", timeline_create_handler)
         .delete(
             "/v1/tenant/:tenant_id/timeline/:timeline_id",
--- a/safekeeper/src/http/routes.rs
+++ b/safekeeper/src/http/routes.rs
@@ -1,8 +1,9 @@
-use hyper::{Body, Request, Response, StatusCode};
+use hyper::{Body, Request, Response, StatusCode, Uri};
 
+use once_cell::sync::Lazy;
 use serde::Serialize;
 use serde::Serializer;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::fmt::Display;
 use std::sync::Arc;
 
@@ -12,8 +13,9 @@ use crate::timeline::{GlobalTimelines, TimelineDeleteForceResult};
 use crate::SafeKeeperConf;
 use etcd_broker::subscription_value::SkTimelineInfo;
 use utils::{
+    auth::JwtAuth,
     http::{
-        endpoint,
+        endpoint::{self, auth_middleware, check_permission},
         error::ApiError,
         json::{json_request, json_response},
         request::{ensure_no_body, parse_request_param},
@@ -32,6 +34,7 @@ struct SafekeeperStatus {
 
 /// Healthcheck handler.
 async fn status_handler(request: Request<Body>) -> Result<Response<Body>, ApiError> {
+    check_permission(&request, None)?;
     let conf = get_conf(&request);
     let status = SafekeeperStatus { id: conf.my_id };
     json_response(StatusCode::OK, status)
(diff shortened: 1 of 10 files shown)


test_runner/batch_others/test_wal_acceptor.py
```diff
@@ -380,8 +386,22 @@ def test_wal_removal(zenith_env_builder: ZenithEnvBuilder):
...
+            auth_token=env.auth_keys.generate_tenant_token(uuid4().hex))
+        http_cli_noauth = env.safekeepers[0].http_client()
+
     # Pretend WAL is offloaded to s3.
+    if auth_enabled:
+        for cli in [http_cli_other, http_cli_noauth]:
+            with pytest.raises(cli.HTTPError, match='Forbidden|Unauthorized'):
+                cli.record_safekeeper_info(tenant_id,
+                                           timeline_id, {'backup_lsn': 'FFFFFFFF/FEFFFFFF'})
+        # TODO: ensure the record_safekeeper_info was not executed
```
yeputons, line 404: Can we do it somehow?

safekeeper/src/http/routes.rs
```diff
@@ -32,6 +34,7 @@ struct SafekeeperStatus {
 
 /// Healthcheck handler.
 async fn status_handler(request: Request<Body>) -> Result<Response<Body>, ApiError> {
+    check_permission(&request, None)?;
```
yeputons, line 37: Probably should remove this line and leave a comment that this route is in `ALLOWLIST`?

safekeeper/src/http/routes.rs
```diff
@@ -125,6 +129,7 @@ async fn timeline_create_handler(mut request: Request<Body>) -> Result<Response<
         tenant_id: request_data.tenant_id,
         timeline_id: request_data.timeline_id,
     };
+    check_permission(&request, Some(zttid.tenant_id))?;
```
yeputons, line 132: This line is untested: `timeline_create_handler` is never called in the tests. Should we remove it, is it used at all?

test_runner/batch_others/test_wal_acceptor.py
yeputons: TODO: yes, we can, `record_safekeeper_info` is mostly for manual testing/fixes, it just changes some variables inside the Safekeeper. So we should be able to read them before/after the change.

safekeeper/src/http/routes.rs
yeputons: Discussed with @ars: it's a stub for a future code which will create timelines explicitly (which is probably better than implicit creation). So let's keep the code for now. We will test it when it's actually used.

yeputons: Discussed with @arssher: it's a stub for a future code which will create timelines explicitly (which is probably better than implicit creation). So let's keep the code for now. We will test it when it's actually used.

hlinnaka (approved):

test_runner/batch_others/test_wal_acceptor.py
```diff
@@ -921,7 +966,14 @@ def test_delete_force(neon_env_builder: NeonEnvBuilder):
...
-    sk_http = sk.http_client()
+    if not auth_enabled:
+        sk_http = sk.http_client()
+        sk_http_other = sk_http
+    else:
+        sk_http = sk.http_client(auth_token=env.auth_keys.generate_tenant_token(tenant_id))
+        sk_http_other = sk.http_client(
+            auth_token=env.auth_keys.generate_tenant_token(tenant_id_other))
+        sk_http_noauth = sk.http_client()
     assert (sk_data_dir / tenant_id / timeline_id_1).is_dir()
```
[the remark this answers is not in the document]
yeputons: It can. Good idea, but I'd do it in a separate PR for both Pageserver and Safekeeper simultaneously.

yeputons:
Actually, I think it's better to enable auth on staging/production _after_ the follow-up issues. Otherwise we won't be able to check (even on staging) whether Grafana is correctly authenticating with Safekeeper, for example, because corresponding routes are not checked. Same for postgres protocol.

Issue #1799, opened 2022-05-26
Replace remote storage logs with metrics


petuhovskiy:
We can replace some very common pageserver logs with metrics. Now we have such logs every second for each active timeline:
```
2022-05-26T07:50:40.913116Z  INFO storage_sync_loop_step:process_sync_task_batch{sync_id=AAA-BBB}:upload_timeline_data: Processed a sync task in 0.03 seconds
2022-05-26T07:50:40.885634Z  INFO storage_sync_loop_step:process_sync_task_batch{sync_id=AAA-BBB}:upload_timeline_data: Uploading remote index for the timeline
2022-05-26T07:50:40.885602Z  INFO storage_sync_loop_step:process_sync_task_batch{sync_id=AAA-BBB}:upload_timeline_data: No layers to upload after filtering, aborting
```

When we have some data to upload:
```
2022-05-26T07:56:22.279100Z  INFO storage_sync_loop_step:process_sync_task_batch{sync_id=AAA-BBB}:upload_timeline_data: Successfully uploaded all layersShow context
2022-05-26T07:56:21.870670Z  INFO storage_sync_loop_step:process_sync_task_batch{sync_id=AAA-BBB}:upload_timeline_data: Uploading 1 timeline layers, new lsn: Some(0/16960E8) created initial timeline BBB timeline.lsn 0/16960E8
```

We can turn these logs to trace level and export this info as metrics. Metric names can be:
- histogram `pageserver_upload_timeline_task_seconds`, to replace Processed a sync task in **0.03 seconds**
- counter `pageserver_upload_timeline_layers_total`, to replace Uploading **1 timeline layers**, new lsn: Some(0/16960E8)

LizardWizzard:
I think the better option would be to add metrics, and tone down logs. They may be useful for debugging. Even if we include ids as metric labels.

PR #1865, opened 2022-06-02
Etcd leftovers

SomeoneToIgnore:
Context: https://neondb.slack.com/archives/C033QLM5P7D/p1654106294534139?thread_ts=1654092724.305229&cid=C033QLM5P7D


pageserver/src/walreceiver.rs
```diff
@@ -468,14 +468,17 @@ async fn timeline_wal_broker_loop_step(
...
+                        Some(subscribed_timeline_updates) => {
+                            match wal_connection_manager.select_connection_candidate(subscribed_timeline_updates) {
+                                Some(candidate) => {
+                                    info!("Switching to different safekeeper {} for timeline {id}, reason: {:?}", candidate.safekeeper_id, candidate.reason);
+                                    wal_connection_manager.change_connection(candidate.safekeeper_id, candidate.wal_producer_connstr).await;
+                                },
+                                None => debug!("No connection candidate was selected for timeline"),
+                            }
                         }
+                        None => warn!("Timeline has an active broker subscription, but got no updates. Other data length: {}", all_timeline_updates.len()),
```
arssher, line 481: How this is possible? If it doesn't, let's add comment/assert about being unreachable.

libs/etcd_broker/src/lib.rs
```diff
@@ -282,54 +290,64 @@ pub async fn subscribe_to_safekeeper_timeline_updates(
...
+            )));
+        }
     };
+    let info = serde_json::from_str(value_str).map_err(|e| {
+        BrokerError::ParsingError(format!(
+            "Failed to parse '{value_str}' as safekeeper timeline info: {e}"
+        ))
+    })?;
 
     let (zttid, safekeeper_id) = match subscription_kind {
```
arssher, line 323: Key is the same in all kinds of subs, it would be simpler to avoid this match. (and generally 3 kinds of regexp I guess)

arssher (approved):

pageserver/src/walreceiver.rs
SomeoneToIgnore: I've just looked at the logs, and looks like we've got a conflict of a kind, that triggers exactly this: the only failed test currently, https://app.circleci.com/pipelines/github/neondatabase/neon/6914/workflows/46edb373-663a-4863-9d09-f460e36e68fc/jobs/70757, sometimes receives `WAL_BACKUP` keys only, since those share the same key prefix, as the one we subscribe to:

```
2022-06-02T11:21:51.798462Z ERROR wal_receiver_main:timeline{id=ddf794d7b424414ba62dbab6afe69829/49ec01927debff5d347a6aaa96420a80}:etcd_subscription:etcd_broker: Failed to parse timeline update: Error during parsing etcd data: KV has unexpected key 'neon/ddf794d7b424414ba62dbab6afe69829/49ec01927debff5d347a6aaa96420a80/safekeeper/WAL_BACKUP/694d81241b4118c7' that does not match required regex ^neon/ddf794d7b424414ba62dbab6afe69829/49ec01927debff5d347a6aaa96420a80/safekeeper/([[:digit:]]+)$
2022-06-02T11:21:51.798737Z ERROR wal_receiver_main:timeline{id=ddf794d7b424414ba62dbab6afe69829/49ec01927debff5d347a6aaa96420a80}:etcd_subscription:etcd_broker: Failed to parse timeline update: Error during parsing etcd data: KV has unexpected key 'neon/ddf794d7b424414ba62dbab6afe69829/49ec01927debff5d347a6aaa96420a80/safekeeper/WAL_BACKUP/694d81241b4118c5' that does not match required regex ^neon/ddf794d7b424414ba62dbab6afe69829/49ec01927debff5d347a6aaa96420a80/safekeeper/([[:digit:]]+)$
2022-06-02T11:21:51.798914Z  WARN wal_receiver_main:timeline{id=ddf794d7b424414ba62dbab6afe69829/49ec01927debff5d347a6aaa96420a80}: Timeline has an active broker subscription, but got no updates. Other data length: 0
```

should we split those keys entirely, to avoid overlap in the same subscription?

SomeoneToIgnore:
One "flacky" test fails due to me incorrectly applying LSN lag calculation to PS <-> SK Lsn values instead of old SK <-> new SK.

Very weird that the tests in the old PR did not fail after all those retries, fixing it here too.

pageserver/src/walreceiver.rs
arssher: Yes, let's skip `safekeeper`, adjusting `get_campaign_name`.

pageserver/src/storage_sync.rs
```diff
@@ -892,7 +892,7 @@ fn storage_sync_loop<P, S>(
 
         REMAINING_SYNC_ITEMS.set(remaining_queue_length as i64);
         if remaining_queue_length > 0 || !batched_tasks.is_empty() {
-            info!("Processing tasks for {} timelines in batch, more tasks left to process: {remaining_queue_length}", batched_tasks.len());
+            debug!("Processing tasks for {} timelines in batch, more tasks left to process: {remaining_queue_length}", batched_tasks.len());
```
petuhovskiy, line 895: Actually wanted to do that for some time :) ref https://github.com/neondatabase/neon/issues/1799

petuhovskiy (approved):

libs/etcd_broker/src/lib.rs
SomeoneToIgnore: That required splitting off the etcd_prefix first, but done.

SomeoneToIgnore:
I think I've addressed all feedback and issues that I've found, I see no jumpiness in the logs too, let's see how the tests behave on CI now.

pageserver/src/walreceiver.rs
```diff
@@ -648,68 +668,78 @@ impl WalConnectionManager {
...
+            (_, None) => None,
+
+            // No updates from the old SK in this batch, but some candidate is available:
+            // check how long time ago did we receive updates from the current SK, switch connections in case it's over the threshold
+            (None, Some((&new_sk_id, _, new_wal_producer_connstr))) => {
+                match self.wal_connection_data.as_ref() {
+                    Some(current_connection) => {
+                        let last_sk_interaction_time =
+                            match current_connection.last_wal_receiver_data.as_ref() {
+                                Some((_, data_submission_time)) => *data_submission_time,
```
arssher, line 685: It is better to compare here not just the latest message timestamp, but timestamp of latest commit_lsn change -- otherwise we might stuck connected to safekeeper which doesn't receive any new WAL but successfully sends keepalives.

pageserver/src/walreceiver.rs
```diff
@@ -648,68 +668,78 @@ impl WalConnectionManager {
...
-    fn reason_to_reconnect(
-        &self,
-        current_connection: &WalConnectionData,
-        new_sk_timeline: &SkTimelineInfo,
-    ) -> Option<ReconnectReason> {
-        let last_sk_interaction_time = match current_connection.last_wal_receiver_data.as_ref() {
-            Some((last_wal_receiver_data, data_submission_time)) => {
-                let new_lsn = new_sk_timeline.commit_lsn?;
-                match new_lsn.0.checked_sub(last_wal_receiver_data.ps_writelsn)
+            // Both current SK got updated via etcd and there's another candidate with suitable Lsn:
```
arssher, line 717: I think this is mostly unreachable code as nothing guarantees candidate sk data and current sk data will be in the same watch batch. Need to check, but in normally I'd expect it deliver single kv in single message.

The way I imagined it is the following: broker maintains map of sk id -> data, and data is timestamped on each new info arrival. Before selecting candidate we remove entries which weren't updated for some time (safekeeper stopped sending data and can be considered dead).

PR #1879, opened 2022-06-03
Silence etcd safekeeper info key parse errors.

arssher:
When we subscribe to everything, it is ok to receive not only safekeeper
timeline updates.

### DIFF
--- a/libs/etcd_broker/src/lib.rs
+++ b/libs/etcd_broker/src/lib.rs
@@ -65,7 +65,9 @@ pub struct SkTimelineInfo {
 pub enum BrokerError {
     #[error("Etcd client error: {0}. Context: {1}")]
     EtcdClient(etcd_client::Error, String),
-    #[error("Error during parsing etcd data: {0}")]
+    #[error("Error during parsing etcd key: {0}")]
+    InvalidKey(String),
+    #[error("Error during parsing etcd value: {0}")]
     ParsingError(String),
     #[error("Internal error: {0}")]
     InternalError(String),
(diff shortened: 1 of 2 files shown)


libs/etcd_broker/src/lib.rs
```diff
@@ -383,7 +388,8 @@ mod tests {
                 &timeline_subscription,
             ] {
                 let (id, _timeline) =
-                    parse_etcd_key_value(subscription, &key_string, value_str)
+                    parse_safekeeper_timeline_kv(subscription, &key_string, value_str)
+                        .unwrap()
```
SomeoneToIgnore, line 392: I don't feel good about that `.unwrap`, given that we cannot control what we receive: let's rewrite it so that there's no way unusual KV input can cause panics here.

SomeoneToIgnore (commented):
I would add some `debug!` to cover the parsing cases, just in case something's not right with the data and we want to investigate it later.

To do so, I would restore the initial state of the code (modulo naming), but made `parse_etcd_key_value/parse_safekeeper_timeline_kv` function to return `Result<.., CustomError>` with 

```rust
enum CustomError {
    WrongKey(...),
    ParsingFailed(...),
}
```
and then `debug!`'ed on the first variant.

arssher:
Not sure I feel much difference in this case, but ok.

SomeoneToIgnore (approved):

LizardWizzard:
@arssher whats the purpose of creating a duplicate test_normal_work? I've kept it but moved it to a separate file here: https://github.com/neondatabase/neon/blob/main/test_runner/batch_others/test_normal_work.py

arssher:
I just like to run small test during development. It runs faster and has 3 safekeepers instead of 1. It is skipped by default (in CI).

LizardWizzard:
Lets add 3 safekeepers to it if that's that's the only problem. It can be parametrized with a number of tenants and have cases with 1 and 3 tenants so you can run short one locally. 

I'm strongly against unnecessary code duplication. Skipped tests begin to rot over time and silently break.

arssher:
makes sense, https://github.com/neondatabase/neon/pull/1881/files

PR #1886, opened 2022-06-04
Delete dead code

bojanserafimov:
Looks like the file was copied to `storage_sync` and the old copy is still in the repo, unlinked.


SomeoneToIgnore (approved):
Thank you.

PR #1892, opened 2022-06-06
Rework etcd broker keys

SomeoneToIgnore:
A preparation PR for https://github.com/neondatabase/neon/pull/1865#discussion_r888609043

* Adds a typed API for constructing the etcd keys: no more key overlap, checked by the compiler
* Restructure the key segments to receive less unrelated events for more generic subscriptions
* Adjust the etcd pull code to send the updates only if there are some relevant updates came from etcd

### DIFF
--- a/libs/etcd_broker/src/lib.rs
+++ b/libs/etcd_broker/src/lib.rs
@@ -1,91 +1,43 @@
 //! A set of primitives to access a shared data/updates, propagated via etcd broker (not persistent).
 //! Intended to connect services to each other, not to store their data.
+
+/// All broker keys, that are used when dealing with etcd.
+pub mod subscription_key;
+/// All broker values, possible to use when dealing with etcd.
+pub mod subscription_value;
+
 use std::{
     collections::{hash_map, HashMap},
-    fmt::Display,
     str::FromStr,
 };
 
-use once_cell::sync::Lazy;
-use regex::{Captures, Regex};
-use serde::{Deserialize, Serialize};
-use serde_with::{serde_as, DisplayFromStr};
+use serde::de::DeserializeOwned;
 
-pub use etcd_client::*;
-
+use subscription_key::SubscriptionKey;
 use tokio::{sync::mpsc, task::JoinHandle};
 use tracing::*;
-use utils::{
-    lsn::Lsn,
-    zid::{NodeId, ZTenantId, ZTenantTimelineId},
-};
+use utils::zid::{NodeId, ZTenantTimelineId};
+
+use crate::subscription_key::SubscriptionFullKey;
+
+pub use etcd_client::*;
 
 /// Default value to use for prefixing to all etcd keys with.
 /// This way allows isolating safekeeper/pageserver groups in the same etcd cluster.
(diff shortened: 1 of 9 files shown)


libs/etcd_broker/src/lib.rs
```diff
@@ -1,209 +1,136 @@
...
+    key: SubscriptionKey,
+) -> Result<BrokerSubscription<V>, BrokerError>
+where
+    V: DeserializeOwned + Send + 'static,
+{
+    subscribe_for_values(client, key, |_, value_str| {
+        match serde_json::from_str::<V>(value_str) {
+            Ok(value) => Some(value),
+            Err(e) => {
+                error!("Failed to parse value str '{value_str}': {e}");
```
LizardWizzard, line 88: Is there a reason to avoid BrokerError::ValueNotParsed here?

LizardWizzard (approved):
Looks good to me

petuhovskiy:
I just looked at the etcd gRPC API and found that single watch stream can have multiple subscriptions: https://github.com/etcd-io/etcd/blob/main/api/etcdserverpb/rpc.proto#L66-L78

What do you think about using this? I'm not sure if it's much better than using separate watch stream for every subscriber, but we have such an option.

libs/etcd_broker/src/subscription_key.rs
```diff
@@ -0,0 +1,313 @@
+//! Etcd broker keys, used in the project and shared between instances.
+//! The keys are split into two categories:
+//!
+//! * [`SubscriptionFullKey`] full key format: `<cluster_prefix>/<node_kind>/<operation>/<tenant>/<timeline>/<node_id>`
```
petuhovskiy, line 4: AFAIR we wanted to place `/<tenant>/<timeline>/` earlier, so we can use a prefix to watch all updates on the selected timeline as a prefix. If `<node_kind>/<operation>` is placed before, than we need to subscribe to multiple prefixes to watch all timeline updates.

SomeoneToIgnore: I will return it back to the beginning, thanks.
Seemed a good solution for both SK and PS subscriptions, but can be improved later.

libs/etcd_broker/src/lib.rs
SomeoneToIgnore: No reason to have this variant now, AFAIK: we allow the user to react on value parsing and do not stop on errors, just log them.

libs/etcd_broker/src/subscription_key.rs
```diff
@@ -0,0 +1,310 @@
+//! Etcd broker keys, used in the project and shared between instances.
+//! The keys are split into two categories:
+//!
+//! * [`SubscriptionFullKey`] full key format: `<cluster_prefix>/<tenant>/<timeline>/<node_kind>/<operation>/<node_id>`
```
arssher, line 4: FWIW I'm not a fan of using 'subscription' word here all around. This is just a schema, and subscribing to changes is only a (one) way to use it.

PR #1921, opened 2022-06-09
Added invariant check for project name.

KlimentSerafimov:
Resolving issue #1805 

### Summary
Added invariant check that throws an error if the inferred project name contains a character outside the set {alphanumeric U '-'}.

#### Note
Used common name from server.crt to determine project name from SNI (the common name is the suffix).

### Behavior

Start the proxy with 

> ./target/debug/proxy -c server.crt -k server.key --auth-backend=console

In another terminal, if you start a postgres instance without the correct name formatting, it throws an error:

> Kliments-MBP:neon klimentserafimov$ psql 'postgres://my_cluster_123.localtest.me:4432/main?sslmode=require'
psql: error: connection to server at "my_cluster_123.localtest.me" (127.0.0.1), port 4432 failed: ERROR:  Project name must contain only alphanumeric characters and hyphens ('-'). Project name: 'my_cluster_123'.

Here, underscores ('_') are not allowed.

Otherwise, it works as expected.

#### Untested behavior:
This functionality requires the common name to be set, otherwise it throws a "Common name is not set." error.

### DIFF
--- a/proxy/Cargo.toml
+++ b/proxy/Cargo.toml
@@ -39,6 +39,9 @@ utils = { path = "../libs/utils" }
 metrics = { path = "../libs/metrics" }
 workspace_hack = { version = "0.1", path = "../workspace_hack" }
 
+x509-parser = "0.13.2"
+substring = "1.4.5"
+
 [dev-dependencies]
 rcgen = "0.8.14"
 rstest = "0.12"
--- a/proxy/Cargo.toml
+++ b/proxy/Cargo.toml
@@ -40,7 +40,6 @@ metrics = { path = "../libs/metrics" }
 workspace_hack = { version = "0.1", path = "../workspace_hack" }
 
 x509-parser = "0.13.2"
-substring = "1.4.5"
 
 [dev-dependencies]
 rcgen = "0.8.14"
--- a/proxy/src/auth/credentials.rs
+++ b/proxy/src/auth/credentials.rs
@@ -27,7 +27,7 @@ pub enum ClientCredsParseError {
     CommonNameNotSet,
 
     #[error("SNI ('{1}') inconsistently formatted with respect to common name ('{0}'). SNI should be formatted as '<project-name>.<common-name>'.")]
-    InconsistentCommonNameAndSNI(&'static str, String),
+    InconsistentCommonNameAndSNI(String, String),
 
     #[error("Project name ('{0}') must contain only alphanumeric characters and hyphens ('-').")]
     ProjectNameContainsIllegalChars(String),
@@ -52,8 +52,8 @@ impl ClientCredentials {
 
     pub fn parse(
         mut options: HashMap<String, String>,
-        sni_data: Option<String>,
-        common_name: Option<&'static str>,
+        sni_data: Option<&str>,
+        common_name: Option<&str>,
     ) -> Result<Self, ClientCredsParseError> {
         let mut get_param = |key| {
             options
@@ -64,7 +64,7 @@ impl ClientCredentials {
         let user = get_param("user")?;
         let dbname = get_param("database")?;
         let project_name = get_param("project").ok();
-        let project_name = get_project_name(&sni_data, &common_name, &project_name)?.to_string();
+        let project_name = get_project_name(sni_data, common_name, project_name.as_deref())?;
 
         Ok(Self {
             user,
@@ -85,20 +85,23 @@ impl ClientCredentials {
 }
 
 /// Inferring project name from sni_data.
-fn project_name_from_sni_data<'sni>(
-    sni_data: &'sni str,
-    common_name: &'static str,
-) -> Result<&'sni str, ClientCredsParseError> {
+fn project_name_from_sni_data(
+    sni_data: &str,
+    common_name: &str,
+) -> Result<String, ClientCredsParseError> {
     let common_name_with_dot = format!(".{common_name}");
     // check that ".{common_name_with_dot}" is the actual suffix in sni_data
     if !sni_data.ends_with(&common_name_with_dot) {
         return Err(ClientCredsParseError::InconsistentCommonNameAndSNI(
-            common_name,
+            common_name.to_string(),
             sni_data.to_string(),
         ));
     }
     // return sni_data without the common name suffix.
-    Ok(sni_data.strip_suffix(&common_name_with_dot).unwrap())
+    Ok(sni_data
+        .strip_suffix(&common_name_with_dot)
+        .unwrap()
+        .to_string())
 }
 
 #[cfg(test)]
@@ -109,10 +112,10 @@ mod tests_for_project_name_from_sni_data {
     fn passing() {
         let target_project_name = "my-project-123";
         let common_name = "localtest.me";
-        let sni_data = target_project_name.to_string() + "." + common_name;
+        let sni_data = format!("{target_project_name}.{common_name}");
         assert_eq!(
             project_name_from_sni_data(sni_data.as_str(), common_name).ok(),
-            Some(target_project_name)
+            Some(target_project_name.to_string())
         );
     }
 
@@ -122,12 +125,12 @@ mod tests_for_project_name_from_sni_data {
         let common_name = "localtest.me";
         let wrong_suffix = "wrongtest.me";
         assert_eq!(common_name.len(), wrong_suffix.len());
-        let wrong_common_name = ".wrong".to_string() + wrong_suffix;
-        let sni_data = target_project_name.to_string() + wrong_common_name.as_str();
+        let wrong_common_name = format!("wrong{wrong_suffix}");
+        let sni_data = format!("{target_project_name}.{wrong_common_name}");
         assert_eq!(
             project_name_from_sni_data(sni_data.as_str(), common_name).err(),
             Some(ClientCredsParseError::InconsistentCommonNameAndSNI(
-                common_name,
+                common_name.to_string(),
                 sni_data
             ))
         );
@@ -135,46 +138,35 @@ mod tests_for_project_name_from_sni_data {
 }
 
 /// Determine project name from SNI or from project_name parameter from options argument.
-fn get_project_name<'ret>(
-    sni_data: &'ret Option<String>,
-    common_name: &Option<&'static str>,
-    project_name: &'ret Option<String>,
-) -> Result<&'ret str, ClientCredsParseError> {
+fn get_project_name(
+    sni_data: Option<&str>,
+    common_name: Option<&str>,
+    project_name: Option<&str>,
+) -> Result<String, ClientCredsParseError> {
     // determine the project name from sni_data if it exists, otherwise from project_name.
     let ret = match sni_data {
-        // if sni_data exists, use it to determine project name
         Some(sni_data) => {
-            // extract common name. If unset, throw a CommonNameNotSet error.
-            let common_name = match &common_name {
-                Some(common_name) => common_name,
-                None => return Err(ClientCredsParseError::CommonNameNotSet),
-            };
-            // extract project name from sni.
+            let common_name = common_name.ok_or(ClientCredsParseError::CommonNameNotSet)?;
             let project_name_from_sni = project_name_from_sni_data(sni_data, common_name)?;
             // check invariant: project name from options and from sni should match
             if let Some(project_name) = &project_name {
                 if !project_name_from_sni.eq(project_name) {
                     return Err(ClientCredsParseError::InconsistentProjectNameAndSNI(
-                        project_name_from_sni.to_string(),
+                        project_name_from_sni,
                         project_name.to_string(),
                     ));
                 }
             }
             project_name_from_sni
         }
-        // otherwise use project_option if it was manually set thought options parameter.
         None => project_name
-            .as_ref()
             .ok_or(ClientCredsParseError::MissingSNIAndProjectName)?
-            .as_str(),
+            .to_string(),
     };
 
-    // checking that formatting invariant holds.
-    // project name must contain only alphanumeric characters and hyphens.
+    // check formatting invariant: project name must contain only alphanumeric characters and hyphens.
     if !ret.chars().all(|x: char| x.is_alphanumeric() || x == '-') {
-        return Err(ClientCredsParseError::ProjectNameContainsIllegalChars(
-            ret.to_string(),
-        ));
+        return Err(ClientCredsParseError::ProjectNameContainsIllegalChars(ret));
     }
 
     Ok(ret)
@@ -188,10 +180,10 @@ mod tests_for_project_name_only {
     fn passing_from_sni_data_only() {
         let target_project_name = "my-project-123";
         let common_name = "localtest.me";
-        let sni_data = target_project_name.to_string() + "." + common_name;
+        let sni_data = format!("{target_project_name}.{common_name}");
         assert_eq!(
-            get_project_name(&Some(sni_data), &Some(common_name), &None).ok(),
-            Some(target_project_name)
+            get_project_name(Some(&sni_data), Some(common_name), None).ok(),
+            Some(target_project_name.to_string())
         );
     }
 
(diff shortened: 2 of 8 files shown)


cicdteam:
> 
> #### Untested behavior:
> This functionality requires the common name to be set, otherwise it throws a "Common name is not set." error.

It retrieves CN from certificate ? Right ? or we need specify it some way ?

KlimentSerafimov:
It retrieves the common name from the certificate ("server.crt"). Nothing additional needs specifying.
If you are interested check proxy/src/main.rs for the code that extracts the common name from server.crt.

proxy/src/main.rs
```diff
@@ -118,11 +119,34 @@ async fn main() -> anyhow::Result<()> {
     let mgmt_address: SocketAddr = arg_matches.value_of("mgmt").unwrap().parse()?;
     let http_address: SocketAddr = arg_matches.value_of("http").unwrap().parse()?;
 
+    // determine common name from tls-cert (-c server.crt param).
+    // used in asserting project name formatting invariant.
+    let common_name = match arg_matches.value_of("tls-cert") {
+        Some(cert_path) => {
+            let file = std::fs::File::open(cert_path).unwrap();
+            let almost_common_name = x509_parser::pem::Pem::read(std::io::BufReader::new(file))
+                .unwrap()
```
kelvich, line 128: `unwrap()` would cause a crash when triggered. It is better to use `?` so that an error message will be printed instead.

kelvich:
Generally LGTM.

Let's also add unit tests for `project_name_from_sni_data` and for common name parsing out of some pre-generated certificates. Look up `mod tests` in `messages.rs` for rust unit test examples.

proxy/src/main.rs
KlimentSerafimov: how about in the next section, there is also an unwrap there as well.
Should I do this for both the unwraps above? (since the second one should always pass as I understand).

kelvich: >  Should I do this for both the unwraps above? (since the second one should always pass as I understand)

Yes, better fix both. It is easy to trigger second one as well -- try adding test on that case :)

funbringer:
@KlimentSerafimov That was fast :) I didn't have a chance to highlight the remaining code style issues in your previous PR, so I went ahead and did a small refactoring https://github.com/neondatabase/neon/pull/1917. It looks like one of us has to take a step back. I think this PR could benefit from similar improvements, thus I see two options:

a. We'll just merge this as soon as CI passes. Then I'll chime in and refactor the code one more time, asking for your review.
b. I'll highlight everything I have on my mind in this PR and then we'll try to fix those.

What do you think?

proxy/src/auth/credentials.rs
```diff
@@ -40,49 +43,319 @@ impl ClientCredentials {
...
-                if !project_name_from_options.eq(project_name_from_sni_data) {
-                    return Err(ProjectNameError::Inconsistent(
-                        project_name_from_sni_data.to_string(),
-                        project_name_from_options.to_string(),
-                    ));
+/// Inferring project name from sni_data.
+/// If common_name is None, throws an CommonNameNotSet error,
+/// otherwise checks if common_name is suffix to sni_data.
+/// if common_name is not suffix to sni_data throws InconsistentCommonNameAndSNI error,
+/// otherwise returns the prefix of sni_data without the common_name suffix.
```
funbringer, line 70: I believe we should not duplicate these details in the docstring comment. It's the code that should give the answer to the question "how exactly does this function do what it's asked to do?", not the comment. Apart from the most general overview, the docstring should tell us "why" or "what are the subtleties" or even "what are the preconditions that can't be enforced by the type system".

TLDR: try making **the code** more compact and clear, because comments like this will rot and misguide the reader.

proxy/src/auth/credentials.rs
```diff
@@ -40,49 +43,319 @@ impl ClientCredentials {
...
+            sni_suffix.to_string(),
+        ));
+    }
+
+    // return sni_data without the common name suffix.
+    Ok(sni_data.substring(0, sni_data.len() - common_name.len()))
+}
+
+#[cfg(test)]
+mod tests_for_project_name_from_sni_data {
```
funbringer, line 97: Why bother with `tests_for_project_name_from_sni_data`?

proxy/src/proxy.rs
```diff
@@ -101,6 +101,7 @@ async fn handshake<S: AsyncRead + AsyncWrite + Unpin>(
     stream: S,
     mut tls: Option<TlsConfig>,
     cancel_map: &CancelMap,
+    common_name: &Option<String>,
```
funbringer, line 104: `Option<&str>` is better than `&Option<String>`. You can't move the `String` using a reference anyway, and `None` is better than `&None`.

proxy/src/proxy.rs
```diff
@@ -153,6 +154,9 @@ async fn handshake<S: AsyncRead + AsyncWrite + Unpin>(
                     creds.sni_data = tls.get_ref().1.sni_hostname().map(|s| s.to_owned());
                 }
 
+                // Set common_name
+                creds.common_name = common_name.clone();
```
funbringer, line 158: I'm not convinced that common name should be cloned. We have only one common name during the runtime, so why should we clone it instead of providing a read only reference to a single string?

Besides, the current implementation changes the `creds` object after it has been constructed, which is a code smell. We should provide a `build` or `parse` function accepting all the prerequisites instead.

proxy/src/main.rs
```diff
@@ -118,11 +119,32 @@ async fn main() -> anyhow::Result<()> {
...
+            let almost_common_name = x509_parser::pem::Pem::read(std::io::BufReader::new(file))?
+                .0
+                .parse_x509()?
+                .tbs_certificate
+                .subject
+                .to_string();
+            let prefix = "CN=*";
+            // making sure formatting is as expected: almost_common_name = "'CN=*'[<common name>]"
+            assert_eq!(almost_common_name.substring(0, prefix.len()), prefix);
+            let common_name = almost_common_name.substring(prefix.len(), almost_common_name.len());
```
funbringer, line 136: This doesn't look natural to me. See https://doc.rust-lang.org/std/primitive.str.html#method.strip_prefix

proxy/Cargo.toml
```diff
@@ -39,6 +39,9 @@ utils = { path = "../libs/utils" }
 metrics = { path = "../libs/metrics" }
 workspace_hack = { version = "0.1", path = "../workspace_hack" }
 
+x509-parser = "0.13.2"
+substring = "1.4.5"
```
funbringer, line 43: I don't think we need this.

proxy/src/auth/credentials.rs
```diff
@@ -40,49 +43,319 @@ impl ClientCredentials {
...
+) -> Result<&'sni str, ProjectNameError> {
+    // extract common name. If unset, throw a CommonNameNotSet error.
+    let common_name = match &common_name {
+        Some(common_name) => common_name,
+        None => return Err(ProjectNameError::CommonNameNotSet),
+    };
+
+    // check that the common name passed from common_name is the actual suffix in sni_data.
+    use substring::Substring;
+    let sni_suffix = sni_data.substring(sni_data.len() - common_name.len(), sni_data.len());
```
funbringer, line 83: See https://doc.rust-lang.org/std/primitive.str.html#method.strip_suffix

KlimentSerafimov:
I just took a look at your changes and I think they are neat :D . I'm still getting accustomed to the best way of writing things in Rust. I'll prefer option b because that will help me think through the style of changes you suggest, and I already have your existing changes to reference when refactoring. Option a is also good. 

Few thoughts: Stas asked me to [*1] refactor and add unit tests for extracting the common name from server.crt, so that's another thing that has to be done in addition to the [*2] refactoring you suggested.

So the options now are: 
O1: commit this PR as it is (after the CI resolves the 'neon-cloud-e2e' test), and make other pull requests for [1] and [2] after.
O2: complete [*2] in this PR, and [*1] in a next one.
O3: complete [*1] in this PR, and [*2] in a next one.
O4: complete [*1] and [*2] in this PR.

I would go with O1 because it's just more step-by-step, also this change is large enough. 
Any thoughts @kelvich @funbringer

KlimentSerafimov:
@funbringer I'm taking a look at your suggestions now.

proxy/src/auth/credentials.rs
```diff
@@ -40,49 +43,319 @@ impl ClientCredentials {
...
+/// otherwise returns the prefix of sni_data without the common_name suffix.
+fn project_name_from_sni_data<'sni>(
+    sni_data: &'sni str,
+    common_name: &Option<String>,
+) -> Result<&'sni str, ProjectNameError> {
+    // extract common name. If unset, throw a CommonNameNotSet error.
+    let common_name = match &common_name {
+        Some(common_name) => common_name,
+        None => return Err(ProjectNameError::CommonNameNotSet),
+    };
```
funbringer, line 79: Sorry, I'm confused. Why do you take `Option` if not having a value is an immediate error?

funbringer (changes requested):
Sorry, after reading this a bit more I think this is undercooked.

proxy/src/proxy.rs
KlimentSerafimov: On the first note: I think I added .clone() bc the compiler was complaining. I'll check again.
On the second note: I agree. I thought it was fine because creds.sni_data was also modified. Will refactor this.

proxy/src/auth/credentials.rs
KlimentSerafimov: Good catch. 
common_name is an option from the start (in main), because it's derived from server.crt, which is also an option. So it could happen that common_name is None, and since we have to handle this at some point, I decided to handle it here.

proxy/Cargo.toml
KlimentSerafimov: I added it bc the compiler complained, I'll check again.

proxy/src/auth/credentials.rs
KlimentSerafimov: sounds good!

proxy/src/auth/credentials.rs
KlimentSerafimov: thanks!

proxy/src/auth/credentials.rs
KlimentSerafimov: Added it as practice. I agree it's not necessary especially since project_name(..) covers the entire behavior.

proxy/src/main.rs
KlimentSerafimov: will refactor!

proxy/src/proxy.rs
KlimentSerafimov: ok, thanks

proxy/src/proxy.rs
KlimentSerafimov: will wait to handle this on monday

proxy/Cargo.toml
KlimentSerafimov: done

KlimentSerafimov:
Since all checks are passing, would it be reasonable to get this committed, and work on the refactoring in a new PR? @kelvich @funbringer

proxy/src/config.rs
```diff
@@ -65,6 +82,26 @@ pub fn configure_tls(key_path: &str, cert_path: &str) -> anyhow::Result<TlsConfi
...
+        let option_common_name = almost_common_name.strip_prefix(expected_prefix);
+        let common_name = match option_common_name {
+            Some(common_name) => common_name,
+            None => panic!("Expected {expected_prefix} prefix before the common name."),
+        };
+        common_name.to_string()
+    };
+    Ok(TlsConfig {
+        tls_config: config.into(),
+        common_name: Some(Box::<str>::leak(common_name.into_boxed_str())),
```
funbringer, line 105: We don't need `Some` (see above)

proxy/src/proxy.rs
```diff
@@ -366,7 +375,11 @@ mod tests {
         let (client, server) = tokio::io::duplex(1024);
 
         let (_, server_config) = generate_tls_config("localhost")?;
-        let proxy = tokio::spawn(dummy_proxy(client, Some(server_config), NoAuth));
+        let proxy = tokio::spawn(dummy_proxy(
+            client,
+            Some(TlsConfig::new(server_config)),
```
funbringer, line 380: I think we should move `TlsConfig::new(...)` into `generate_tls_config`. This would eliminate the need for such changes.

proxy/src/config.rs
```diff
@@ -36,7 +36,24 @@ pub struct ProxyConfig {
     pub auth_link_uri: ApiUrl,
 }
 
-pub type TlsConfig = Arc<rustls::ServerConfig>;
+#[derive(Clone)]
+pub struct TlsConfig {
+    pub tls_config: Arc<rustls::ServerConfig>,
+    pub common_name: Option<&'static str>,
```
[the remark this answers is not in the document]
KlimentSerafimov: I used 'Option<&'static str>' and not '&'static str', because there are some tests that construct this struct using TlsConfig::new, where common_name is set None (because they are isolated tests that don't create a common name).

-----

ok, I'll implement the second suggestion.

proxy/src/config.rs
```diff
@@ -36,7 +36,24 @@ pub struct ProxyConfig {
...
+    #[cfg(test)]
+    pub fn new(tls_config: Arc<rustls::ServerConfig>) -> TlsConfig {
+        TlsConfig {
+            tls_config,
+            common_name: None,
+        }
+    }
+    pub fn get_tls_config(self) -> Arc<rustls::ServerConfig> {
+        self.tls_config
+    }
```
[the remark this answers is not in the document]
KlimentSerafimov: Sounds good, I'll make the change! Although I'm slightly confused because in the previous version of the code, we moved the tls_config, rather than cloning it, which is why this new access method now does the same. Also the compiler is not complaining about it. Let me know what you think about this reasoning! Maybe another solution is to simply rename the method move_tls_config, indicating that it's a move rather than a clone.
Regardless, your proposed changed sounds safer and doesnt make much of a performance difference.

proxy/src/proxy.rs
```diff
@@ -105,6 +105,8 @@ async fn handshake<S: AsyncRead + AsyncWrite + Unpin>(
     // Client may try upgrading to each protocol only once
     let (mut tried_ssl, mut tried_gss) = (false, false);
 
+    let common_name = tls.as_ref().map(|tls| tls.common_name);
```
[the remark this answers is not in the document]
KlimentSerafimov: yes I know xD

proxy/src/proxy.rs
KlimentSerafimov: thanks! I didn't notice this.

proxy/src/auth/credentials.rs
```diff
@@ -120,3 +83,247 @@ impl ClientCredentials {
...
+
+    #[test]
+    fn passing() {
+        let target_project_name = "my-project-123";
+        let common_name = "localtest.me";
+        let sni_data = format!("{target_project_name}.{common_name}");
+        assert_eq!(
+            project_name_from_sni_data(sni_data.as_str(), common_name).ok(),
+            Some(target_project_name.to_string())
+        );
```
funbringer, line 119: nit:

```
assert!(matches!(
    project_name_from_sni_data(&sni_data, common_name),
    Err(target_project_name),
));
```

Other than that, the tests look fine. Maybe I'll make them more terse later (I'll ask for your review in that case).

proxy/src/config.rs
```diff
@@ -49,10 +60,19 @@ pub fn configure_tls(key_path: &str, cert_path: &str) -> anyhow::Result<TlsConfi
         keys.pop().map(rustls::PrivateKey).unwrap()
     };
 
+    let cert_chain_bytes = {
+        let mut cert_file = std::fs::File::open(cert_path)
+            .context(format!("Failed to open TLS cert file at '{cert_path}'."))?;
+        let mut cert_chain_bytes = Vec::new();
+        let _ = cert_file.read_to_end(&mut cert_chain_bytes)?;
+        cert_chain_bytes
+    };
```
funbringer, line 69: This used to be `let cert_chain_bytes = std::fs::read(cert_path).context("TLS cert file")?;`

proxy/src/config.rs
```diff
@@ -49,8 +59,15 @@ pub fn configure_tls(key_path: &str, cert_path: &str) -> anyhow::Result<TlsConfi
         keys.pop().map(rustls::PrivateKey).unwrap()
     };
 
+    // read PEM file at cert_path.
+    let cert_file =
+        std::fs::File::open(cert_path).context("Failed to open TLS cert file at '{cert_path}.'")?;
+    let buffer_reader = std::io::BufReader::new(cert_file);
+    let pem = x509_parser::pem::Pem::read(buffer_reader)
+        .context("Failed to read from buffer reader of cert file at '{cert_path'}")?
+        .0;
```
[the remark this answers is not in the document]
KlimentSerafimov: Do we ever expect there to be more than one pem in the file? In case we don't, let's simply throw an error if there is more than one.

proxy/src/auth/credentials.rs
KlimentSerafimov: I understood what you meant. Changes applied.

funbringer (approved):
Thank you! I think this is good enough.

Issue #1893, opened 2022-06-07
Epic: tenant migration between storage versions


kelvich:
We've talked about using logical dump/restore to move tenants in case of the storage format change. But that requires compute to be enabled and currently we don't have too much manual control over it. Also, a logical dump is somewhat a crutch, and building some features to support it seems like the wrong idea.

With some brainstorming during on-call handovers we've come up with a plan that seems to be better:
- [x] Implement pg_basebackup-compatible endpoint in the pageserver. Whatever storage version pageserver is running it anyway can decode it to the postgres format. So we can leverage this format as "lingua franca" of our storage formats. Also, we don't care if compute is running or not -- we just talk to the pageserver.
https://github.com/neondatabase/neon/pull/1911 - PR for the old pageserver
https://github.com/neondatabase/neon/pull/1923 - PR for main
- [x] validate data after migration on staging
- [ ] Implement something like per-tenant start_backup/stop_backup in the pageserver. To be able to import physical backup we need to be sure that the safekeeper has the necessary WAL. We can delay sending s3_consistent_lsn from the pageserver to achieve that. start_backup starts that delay, stop_backup stops.
- [x] Backpatch this endpoint to the old pageserver on staging.
- [x] Implement a backup receiver endpoint in the pageserver. Code there should be quite close to `import_local_directory`. As for the wire format, it would be cool to use the same basebackup protocol.
https://github.com/neondatabase/neon/pull/1925

- [x] Implement `push_backup` binary that uploads postgres backup into the pageserver. Again, the iterator over files should be close to the one used in `import_local_directory`.
https://github.com/neondatabase/neon/pull/1925 - This pr implements the following API:
https://github.com/neondatabase/neon/blob/43d1531e66e46768963dbcedd56a7b47314afe0b/pageserver/src/page_service.rs#L933-L934

- [x] #1980

- [x] https://github.com/neondatabase/neon/issues/2020

- [x] https://github.com/neondatabase/cloud/issues/897

- [ ] https://github.com/neondatabase/neon/issues/2097

That seems to be enough to handle relocation manually.

Follow ups:

- Import can be used as an alternative to making `initdb` on the pageserver. If one day we will support custom settings in `initdb` we can make it near compute and feed resulting files into pageserver.
[...]

PR #1925, opened 2022-06-12
Import basebackup into pageserver

bojanserafimov:
https://github.com/neondatabase/neon/issues/1893

Lots of TODOs left in the code. Pls comment on what seems important to address right now


neon_local/src/main.rs
```diff
@@ -159,6 +159,20 @@ fn main() -> Result<()> {
...
+                .arg(Arg::new("node-name").long("node-name").takes_value(true)
+                    .help("Name to assign to the imported timeline"))
+                .arg(Arg::new("base-tarfile").long("base-tarfile").takes_value(true)
+                    .help("Basebackup tarfile to import"))
+                .arg(Arg::new("base-lsn").long("base-lsn").takes_value(true)
+                    .help("Lsn the basebackup starts at"))
+                .arg(Arg::new("wal-tarfile").long("wal-tarfile").takes_value(true)
+                    .help("Wal to add after base"))
+                .arg(Arg::new("end-lsn").long("end-lsn").takes_value(true)
+                    .help("Lsn the basebackup ends at")))
```
bojanserafimov, line 175: For the sake of tests I'm using `neon_local` as the client interface, but we will need a binary that talks directly to deployed pageservers. I'll make one after this PR, unless there's some `psql` trick for streaming a file using copy protocol.

pageserver/src/import_datadir.rs
```diff
@@ -403,3 +425,219 @@ fn import_wal<R: Repository>(
...
+    let mut segno = start_lsn.segment_number(pg_constants::WAL_SEGMENT_SIZE);
+    let mut offset = start_lsn.segment_offset(pg_constants::WAL_SEGMENT_SIZE);
+    let mut last_lsn = start_lsn;
+    let mut walingest = WalIngest::new(tline, start_lsn)?;
+
+    // Ingest wal until end_lsn
+    info!("importing wal until {}", end_lsn);
+    let mut pg_wal_tar = tar::Archive::new(reader);
+    let mut pg_wal_entries_iter = pg_wal_tar.entries()?;
+    while last_lsn <= end_lsn {
```
bojanserafimov, line 485: This loop is almost copy-pasted from the `import_wal` function. Working on removing copied code, and cleaning up the function names and interfaces in this module ..

pageserver/src/import_datadir.rs
```diff
@@ -403,3 +425,219 @@ fn import_wal<R: Repository>(
...
+        let bytes = read_all_bytes(reader)?;
+        modification.put_twophase_file(xid, Bytes::copy_from_slice(&bytes[..]))?;
+        info!("imported twophase file");
+    } else if file_path.starts_with("pg_wal") {
+        panic!("found wal file in base section");
+    } else {
+        info!("ignored");
+    }
+
+    // TODO: pg_tblspc ??
```
bojanserafimov, line 634: @hlinnaka no clue what's in `pg_tblspc`, but I copied this comment from `import_timeline_from_postgres_datadir`. Let me know if I should look into it now.

pageserver/src/page_service.rs
```diff
@@ -200,6 +202,96 @@ impl PagestreamBeMessage {
...
+    /// Overflow buffer for bytes sent in CopyData messages
+    /// that the reader (caller of read) hasn't asked for yet.
+    /// TODO use BytesMut?
+    buf: Vec<u8>,
+
+    /// Bytes before `buf_begin` are considered as dropped.
+    /// This allows us to implement O(1) pop_front on Vec<u8>.
+    /// The Vec won't grow large because we only add to it
+    /// when it's empty.
+    buf_begin: usize,
```
bojanserafimov, line 218: I hope there's a simpler solution.

The root cause of the problem is that I'm receiving data from a client that decides how much it wants to send, and giving it to the tar library, which decides how much it wants to read. So I have to buffer the bytes in between.

I tried using `VecDeque` as a buffer but it's missing lots of functionality, and I basically had to copy entries one by one in for loops. Not great.

This `Vec` version is fine, but would be nice if there was a simpler (less bug prone) existing solution, and maybe something that copies less. I was looking into `BytesMut`, I see we use it in `waldecoder`, but I'm not sure if `.advance()` does what I think it does.

pageserver/src/import_datadir.rs
```diff
@@ -173,12 +187,10 @@ fn import_relfile<R: Repository>(
...
     let mut buf: [u8; 8192] = [0u8; 8192];
 
-    let len = file.metadata().unwrap().len();
-    ensure!(len % pg_constants::BLCKSZ as u64 == 0);
-    let nblocks = len / pg_constants::BLCKSZ as u64;
+    ensure!(len % pg_constants::BLCKSZ as usize == 0);
+    let nblocks = len / pg_constants::BLCKSZ as usize;
 
     if segno != 0 {
         todo!();
```
lubennikovaav, line 196: This must be fixed to import multi-segment files. 
I'll send a PR today

pageserver/src/import_datadir.rs
```diff
@@ -403,3 +425,219 @@ fn import_wal<R: Repository>(
...
+            let entry = pg_wal_entries_iter.next().expect("expected more wal")?;
+            let header = entry.header();
+            let file_path = header.path().unwrap().into_owned();
+
+            match header.entry_type() {
+                tar::EntryType::Regular => {
+                    // FIXME: assume postgresql tli 1 for now
+                    let expected_filename = XLogFileName(1, segno, pg_constants::WAL_SEGMENT_SIZE);
+                    let file_name = file_path.file_name().unwrap().to_string_lossy();
+                    ensure!(expected_filename == file_name);
```
bojanserafimov, line 496: Potential issue: we assume the tar was created with segments in sorted order.

pageserver/src/page_service.rs
```diff
@@ -447,6 +539,73 @@ impl PageServerHandler {
...
+
+        // TODO leave clean state on error
+
+        // Import basebackup provided via CopyData
+        info!("importing basebackup");
+        pgb.write_message(&BeMessage::CopyInResponse)?;
+        let reader = CopyInReader::new(pgb);
+        import_basebackup_from_tar(&mut datadir_timeline, reader, base_lsn)?;
+
+        info!("done");
```
lubennikovaav, line 573: Let's reword this message to something more informative, i.e. `imported timeline {} timeline.lsn {}`

Also, we probably need to call ` timeline.tline.checkpoint(CheckpointConfig::Forced)?;` here to flush the data.

pageserver/src/page_service.rs
```diff
@@ -447,6 +539,73 @@ impl PageServerHandler {
...
+        _end_lsn: Lsn,
+    ) -> anyhow::Result<()> {
+        thread_mgr::associate_with(Some(tenant_id), Some(timeline_id));
+        let _enter =
+            info_span!("import basebackup", timeline = %timeline_id, tenant = %tenant_id).entered();
+
+        // Create empty timeline
+        info!("creating new timeline");
+        let repo = tenant_mgr::get_repository_for_tenant(tenant_id)?;
+        let timeline = repo.create_empty_timeline(timeline_id, Lsn(0))?;
```
lubennikovaav, line 557: if the timeline already exists, this call will fail, but not before overwriting the existing entry...

```
    fn create_empty_timeline(
    ...
       let r = timelines.insert(
            timelineid,
            LayeredTimelineEntry::Loaded(Arc::clone(&timeline)),
        );
        ensure!(
            r.is_none(),
            "assertion failure, inserted duplicate timeline"
        );
```
Let's fix this weird behavior of `create_empty_timeline`.

pageserver/src/import_datadir.rs
lubennikovaav: Postgres tablespaces allow user to define datadir where to store the files.
see: https://www.postgresql.org/docs/current/manage-ag-tablespaces.html

It doesn't make sense in Neon "bottomless" setup, as we store everything in pageserver anyway.
We do not allow creating new tablespaces in neon, but imported user database may contain them...

Let's cover it with a test and error out here, until we fix it elsewhere.
I'll send a PR.

neon_local/src/main.rs
lubennikovaav: So this API allows us to upload existing basebackup.tar

Maybe just add another endpoint to `page_service.rs`:
 `"import fullbackup pageserver_connstr tenant_id timeline_id lsn"` 
 that will connect another pageserver directly, issue `"fullbackup tenant_id timeline_id lsn"` request and receive the data via copy protocol?

pageserver/src/page_service.rs
bojanserafimov: We're inside `info_span!`, which provides that information on the same line.

Good point about checkpointing. In the case of import from postgres, we're not done until the data gets to s3.

Off topic: will there be issues with two pageservers writing the same timeline to s3, in the case of "import from other pageserver"?

pageserver/src/import_datadir.rs
lubennikovaav: `pg_basebackup` packs tablespaces into separate tars, so we will need to adjust API to support them.
Let's leave them as TODO, because neon clusters never use tablespaces.

pageserver/src/page_service.rs
SomeoneToIgnore: Two concurrent writes into the same timelines are indeed not desirable at the very least.
Alas, there's no sync mechanism to ensure exclusive S3 write in pageserver (yet it's not that hard to implement one via etcd now), but we can do either:

* turn off S3 upload on the new pageserver during the migration period (just omit the S3 settings, basically)
* reproduce the situation manually and see how pageservers would behave, chances are that layers created by different pageservers won't overlap

bojanserafimov: In addition, the new pageserver should never read layer files written by the old pageserver. They have conflicting histories, since the new pageserver thinks the entire database was just created. Not to mention, they likely have conflicting storage formats.

The easiest solution is to write in s3 under a new prefix. For example, prefix the layer object names with storage version number. That prevents migration across pageservers with the same version though (not necessary, but nice to have).

lubennikovaav: I think that we should rather disable/reroute s3 writes for old pageserver instead, because old S3 layers won't be useful after migration.
 
Something like:
1. change old tenant config to `[s3_path].old`
2. move old tenant layers to `[s3_path].old`
3. move data between pageservers. new tenant will write to normal s3 location.

pageserver/src/page_service.rs
bojanserafimov: good catch

lubennikovaav:
https://github.com/neondatabase/neon/commit/e6976d6a5a09b5eff13755b3b9b81f0cbd30e9f9

I've added the test and immediately found an export-import API mismatch:

vanilla `pg_basebackup` writes WAL into a separate file `pg_wal.tar`.
pageserver's `fullbackup` writes WAL into the base tar as files in `pg_wal` subdir.
(just like pageserver's `basebackup`)
It is just one WAL segment, that we generate to pass checkpoint info.

It is crucial to import it this WAL segment to update `prev_lsn` needed for compute start. 

We can either fix import to parse WAL segment from the base tar, 
or change export so that it returned two tars: base and pg_wal.

I prefer the first option because it requires less backporting.
. @bojanserafimov , WDYT?

bojanserafimov:
@lubennikovaav Seems like `fullbackup` puts the wal at the end of the tar, so first solution will work in this case.

But postgres also creates an empty pg_wal inside `base.tar`, so we have to be careful with the logic. When we get a **_nonempty_** `pg_wal` dir inside the tar, we can assume the rest of the tar is only wal files.

The second solution is also fine, but if we return two files we'll need a bespoke binary to parse the end of the first file, write the two files, etc.

lubennikovaav:
UPD
We import all the data at `base_lsn` which is the lsn at which we got the fullbackup. 
So there won't be any WAL after this `lsn`. It means that we don't need to parse this WAL segment 
Instead, we need to set `prev_lsn` to a correct value. Where do we get it?
We can take it from pageserver using `zenith.signal` just like in `basebackup`.

I fixed that in the commit: https://github.com/neondatabase/neon/pull/1955/commits/b699b271272e829aae4cc5f5d4144e706fd34bee 
Now `test_import_from_pageserver` passes.
@bojanserafimov , please review #1955
It is based on basebackup-import rebased on top of main.

neon_local/src/main.rs
bojanserafimov: let's leave this for another PR. It's an extra feature

pageserver/src/import_datadir.rs
bojanserafimov: Good enough for now. At least it works when there's no wal to import, which is our only use case for now

pageserver/src/page_service.rs
bojanserafimov: Added checkpoint

pageserver/src/import_datadir.rs
bojanserafimov: I removed 80% of the code repetition in the module. Seems premature to refactor further

bojanserafimov:
Lots of TODOs left in the code. Pls comment on what seems important to address right now

pageserver/src/import_datadir.rs
```diff
@@ -403,3 +285,241 @@ fn import_wal<R: Repository>(
...
+
+                if let Some(res) = import_file(&mut modification, file_path.as_ref(), entry, len)? {
+                    // We found the pg_control file.
+                    pg_control = Some(res);
+                }
+            }
+            tar::EntryType::Directory => {
+                info!("directory {:?}", file_path);
+                if file_path.starts_with("pg_wal") {
+                    info!("found pg_wal in base lol");
```
lubennikovaav, line 320: Let's remove this message.
pg_wal directory can be present in basebackup. 
That's fine. We ignore wal files anyway.

pageserver/src/import_datadir.rs
```diff
@@ -403,3 +285,241 @@ fn import_wal<R: Repository>(
...
+    Ok(())
+}
+
+pub fn import_file<R: Repository, Reader: Read>(
+    modification: &mut DatadirModification<R>,
+    file_path: &Path,
+    reader: Reader,
+    len: usize,
+) -> Result<Option<ControlFileData>> {
+    info!("looking at {:?}", file_path);
```
lubennikovaav, line 418: There are too many `info!` messages. Please, them into `debug!` or remove at all.

pageserver/src/page_service.rs
```diff
@@ -200,6 +202,96 @@ impl PagestreamBeMessage {
...
+                Err(e) => {
+                    if !is_socket_read_timed_out(&e) {
+                        return Err(io::Error::new(io::ErrorKind::Other, e));
+                    }
+                }
+            }
+        }
+
+        // Shutting down
+        // TODO is this fine?
```
lubennikovaav, line 290: Huh? What could go wrong here?
Please, add some details or remove the comment.

pageserver/src/page_service.rs
```diff
@@ -447,6 +539,93 @@ impl PageServerHandler {
...
+        // Create empty timeline
+        info!("creating new timeline");
+        let repo = tenant_mgr::get_repository_for_tenant(tenant_id)?;
+        let timeline = repo.create_empty_timeline(timeline_id, Lsn(0))?;
+        let repartition_distance = repo.get_checkpoint_distance();
+        let mut datadir_timeline =
+            DatadirTimeline::<LayeredRepository>::new(timeline, repartition_distance);
+
+        // TODO mark timeline as not ready until it reaches end_lsn?
+        // This would prevent compute node from connecting to it and writing conflicting wal.
```
lubennikovaav, line 563: I suspect that compute node won't be able to connect anyway.
To connect, it needs to receive basebackup at some lsn and get the data, which is invisible until `    modification.commit()?;`

we can test it using sleep failpoint here.

pageserver/src/page_service.rs
```diff
@@ -447,6 +539,93 @@ impl PageServerHandler {
...
+
+        // TODO leave clean state on error
```
lubennikovaav, line 565: Is it still a TODO?
What is a workaround?

pageserver/src/page_service.rs
```diff
@@ -447,6 +539,93 @@ impl PageServerHandler {
...
+
+        // TODO check checksum
+        // Meanwhile you can verify client-side by taking fullbackup
+        // and checking that it matches in size with what was imported.
+        // It wouldn't work if base came from vanilla postgres though,
+        // since we discard some log files.
+
+        // Flush data to disk, then upload to s3
+        info!("flushing layers");
+        datadir_timeline.tline.checkpoint(CheckpointConfig::Flush)?;
```
lubennikovaav, line 581: Why don't we use `CheckpointConfig::Forced` like in `bootstrap_timeline`?
This choice deserves a comment.

pageserver/src/page_service.rs
```diff
@@ -447,6 +539,93 @@ impl PageServerHandler {
...
+        let _enter =
+            info_span!("import wal", timeline = %timeline_id, tenant = %tenant_id).entered();
+
+        let repo = tenant_mgr::get_repository_for_tenant(tenant_id)?;
+        let timeline = repo.get_timeline_load(timeline_id)?;
+        let repartition_distance = repo.get_checkpoint_distance();
+        let mut datadir_timeline =
+            DatadirTimeline::<LayeredRepository>::new(timeline, repartition_distance);
+
+        // TODO ensure start_lsn matches current lsn
```
lubennikovaav, line 608: What is current lsn in this context?  Is it `timeline.last_record_lsn`?
Let's add this sanity check before merging.

pageserver/src/page_service.rs
```diff
@@ -447,6 +539,93 @@ impl PageServerHandler {
...
+        // and checking that it matches in size with what was imported.
+        // It wouldn't work if base came from vanilla postgres though,
+        // since we discard some log files.
+
+        // Flush data to disk, then upload to s3
+        info!("flushing layers");
+        datadir_timeline.tline.checkpoint(CheckpointConfig::Flush)?;
+
+        // TODO Wait for s3 upload to complete
+        // info!("uploading layers");
```
lubennikovaav, line 584: This looks important.

If pageserver crashes right after data loading, we will lose the data.
That's less of a problem for bootstrap, because we can re-run initdb anytime.

@LizardWizzard, is there any way to force s3 uploading for the timeline? 
To wait until completion, we must check `remote_consistent_lsn == last_record_lsn`

pageserver/src/page_service.rs
```diff
@@ -750,6 +929,41 @@ impl postgres_backend::Handler for PageServerHandler {
...
+            // 2. Run:
+            // cat my_backup/base.tar | psql -h $PAGESERVER \
+            //     -c "import basebackup $TENANT $TIMELINE $START_LSN $END_LSN"
+            let (_, params_raw) = query_string.split_at("import basebackup ".len());
+            let params = params_raw.split_whitespace().collect::<Vec<_>>();
+            ensure!(params.len() == 4);
+            let tenant = ZTenantId::from_str(params[0])?;
+            let timeline = ZTimelineId::from_str(params[1])?;
+            let base_lsn = Lsn::from_str(params[2])?;
+            let end_lsn = Lsn::from_str(params[2])?;
```
lubennikovaav, line 947: Do we ever use this `end_lsn`?

test_runner/fixtures/neon_fixtures.py
```diff
@@ -1395,12 +1395,12 @@ def start(self, log_path: Optional[str] = None):
         if log_path is None:
             log_path = os.path.join(self.pgdatadir, "pg.log")
 
-        self.pg_bin.run_capture(['pg_ctl', '-D', self.pgdatadir, '-l', log_path, 'start'])
+        self.pg_bin.run_capture(['pg_ctl', '-w', '-D', self.pgdatadir, '-l', log_path, 'start'])
```
lubennikovaav, line 1398: -w is default for pg_ctl, so it doesn't change anything.
>This is supported for the modes start, stop, restart, promote, and register, and is the default for those modes.

test_runner/batch_others/test_import.py
```diff
@@ -0,0 +1,183 @@
...
+        "-F",
+        "tar",
+        "-d",
+        vanilla_pg.connstr(),
+        "-D",
+        basebackup_dir,
+    ])
+
+    # Make corrupt base tar with missing pg_control
+    unpacked_base = os.path.join(basebackup_dir, "unpacked-base")
```
lubennikovaav, line 43: I see strange artifact in CI run:

[tmp/test_output/test_import_from_vanilla/basebackup/unpacked-base/pg.log](https://output.circle-artifacts.com/output/job/e701e0fe-03c8-4cb5-b5b8-aca404142838/artifacts/0/tmp/test_output/test_import_from_vanilla/basebackup/unpacked-base/pg.log)

What process starts postgres in this directory? I thought it was a temp dir for a tar...

pageserver/src/page_service.rs
bojanserafimov: Possible scenario:
1. We import the base
2. We start importing wal
3. But before that happens, compute connects and writes wal

I'll clarify. No need to fix now, since it doesn't affect pageserver->pageserver migration

pageserver/src/page_service.rs
bojanserafimov: I agree it's important. But afaik it's a general problem, even if we're not importing timelines. Currently GC could eat some layers that are only scheduled for upload, but not uploaded (or something of that sort).

I guess before merging, I'll confirm this is a bigger issue and make a github issue if so.

pageserver/src/page_service.rs
bojanserafimov: If we'd like to keep the timeline closed until all wal is imported, we would mark it closed until this LSN is reached.

pageserver/src/page_service.rs
bojanserafimov: added comment

test_runner/batch_others/test_import.py
bojanserafimov: `pg.log` is unpacked from `base.tar`

pageserver/src/page_service.rs
bojanserafimov: I'm wondering if the reader will keep trying to read, or stop when it reads 0 bytes. If the former is the case, it could get stuck forever.

pageserver/src/page_service.rs
bojanserafimov: clarified with comment

pageserver/src/page_service.rs
bojanserafimov: Seems like the condition "The layer has been uploaded" is missing [here](https://github.com/neondatabase/neon/blob/basebackup-import/pageserver/src/layered_repository.rs#L2248), but maybe it's implied from the other conditions somehow?

lubennikovaav:
Let's maybe reduce table size in `test_import_from_vanilla` ?
We won't test multi-segment scenario, but the test will be more stable.

pageserver/src/page_service.rs
SomeoneToIgnore: > is there any way to force s3 uploading for the timeline?

We can trigger a checkpoint, using psql for instance: https://github.com/neondatabase/neon/blob/a91e0c299d3604086be9186d31209b08b733766e/test_runner/batch_others/test_remote_storage.py#L73

if it works without errors, it will create a layer that'll schedule its upload to the remote storage and upload it eventually, thus updating the remote Lsn.

We don't have any good way to synchronise the sync part of pageserver with async remote storage part in terms of timeline statuses (neither we have a good timeline status transition impl in pageserver generally).
We register remote storage timeline updates once per remote storage sync loop, placing all changed data into the remote index that can be accessed from anywhere, including the HTTP timeline info endpoint, that can be used to check whether we've synced more data to the remote storage for a certain timeline.

bojanserafimov: I clarified that it's the client's responsibility to wait for upload and implemented that client side in the tests. We can improve on this later, but it's good enough I think.

lubennikovaav (approved):

PR #1926, opened 2022-06-13
Fix import of multi-segment relation files

lubennikovaav:


pageserver/src/import_datadir.rs
```diff
@@ -192,23 +192,25 @@ fn import_rel<R: Repository, Reader: Read>(
...
         dbnode: dboid,
         relnode,
         forknum,
     };
-    modification.put_rel_creation(rel, nblocks as u32)?;
+
+    // If that is the first segment, also put rel creation entry
+    if segno == 0 {
+        modification.put_rel_creation(rel, 0 as u32)?;
+    }
```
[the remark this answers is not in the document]
bojanserafimov: Just realized I make the same assumption in [wal import](https://github.com/neondatabase/neon/pull/1925/files#r895877288) so we'll have to do some work client-side to correctly order things, and maybe we should take care when exporting basebackup to follow the same order.

lubennikovaav:
@bojanserafimov , in this PR i updated the test to upload huge multi-segment relation. 
Maybe it would be better to create a separate test for this case. WDYT?

pageserver/src/import_datadir.rs
```diff
@@ -192,19 +192,26 @@ fn import_rel<R: Repository, Reader: Read>(
...
     };
-    modification.put_rel_creation(rel, nblocks as u32)?;
 
     let mut blknum: u32 = segno * (1024 * 1024 * 1024 / pg_constants::BLCKSZ as u32);
+
+    // Call put_rel_creation for every segment of the relation,
+    // because there is no guarantee about the order in which we are processing segments.
+    // ignore "relation already exists" error
+    if let Err(e) = modification.put_rel_creation(rel, nblocks as u32) {
+        if e.to_string().contains("already exists") {
```
bojanserafimov, line 208: Currently `put_rel_creation` errors if alredy exists, but `put_rel_extend` succeeds if it's already extended. A more consistent api would be nice. Maybe both of them return `bool`, whether any change was made? Or something similar to [insert](https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.insert).

pageserver/src/import_datadir.rs
```diff
@@ -216,7 +223,19 @@ fn import_rel<R: Repository, Reader: Read>(
...
+                        blknum - segno * (1024 * 1024 * 1024 / pg_constants::BLCKSZ as u32);
+
+                    // Update relation size
+                    // This update is important,
+                    // because rel_creation above could have set
+                    // incorrect relsize for multi-segment relations
+                    //
+                    // If we process rel segments out of order,
+                    // put_rel_extend will skip the update.
+                    modification.put_rel_extend(rel, blknum)?;
```
bojanserafimov, line 236: Took me a while to understand this. It might be more readable if moved after the for loop. That way the function will have 3 parts:
1. Create relation if not exists
2. Put all the images
3. Extend size

Where 2 and 3 are unrelated

pageserver/src/import_datadir.rs
lubennikovaav: I am not sure, which API would be better here.
Maybe we can just introduce more informative error messages and handle them properly.
Let's leave it as TODO.

Issue #1623, opened 2022-05-03
restart dead threads after error/panic


LizardWizzard:
Needs to be applied to per tenant threads, such as GC and compactor threads.

Changes should go to `thread_mgr.rs`

Options: 
1. Try to change FnOnce to Fn or FnMut
2. Implement something like a factory trait

zxcjhg:
Hey, is it still relevant? I want to give it a go. Looking at the code, I was thinking to write another function to spawn a thread but with restart_on_error flag instead shutdown_process_on_error. What do you think?

LizardWizzard:
Hi! Thank you for your interest! The problem is likely to be fixed with either #1794 or #1933 TLDR we're planning to switch to a threadpool because when the number of tenants is high so is the number of threads which already affected us because of the limit on a number of threads in some of the `tracing` dependencies

PR #1933, opened 2022-06-15
Gc/compaction thread pool, take 2

bojanserafimov:
Resolves https://github.com/neondatabase/neon/issues/1646

Resolves https://github.com/neondatabase/neon/issues/1623

Previous attempt: https://github.com/neondatabase/neon/pull/1794


pageserver/src/tenant_threads.rs
```diff
@@ -1,77 +1,173 @@
...
-        std::thread::sleep(compaction_period);
-        trace!("compaction thread for tenant {} waking up", tenantid);
+        tokio::time::sleep(compaction_period).await;
+        trace!("compaction loop for tenant {} waking up", tenantid);
 
         // Compact timelines
         let repo = tenant_mgr::get_repository_for_tenant(tenantid)?;
-        repo.compaction_iteration()?;
+        let compaction_result =
+            tokio::task::spawn_blocking(move || repo.compaction_iteration()).await;
```
SomeoneToIgnore, line 40: > tokio::task::spawn_blocking

Since you've placed this one, it would be good to notice that all function calls above this line are also blocking:
* get_tenant_state
* get_repository_for_tenant
* get_compaction_period
* get_repository_for_tenant
* and even `get_tenant_state` below, that's called for trace logging only..

This spawn looks a bit confusing tbh, I'd rather slap a TODO/warning about the blocking calls and move on without extra related changes inside, wrapping it into `spawn_blocking` somewhere above instead.

pageserver/src/tenant_threads.rs
```diff
@@ -1,77 +1,173 @@
...
     Ok(())
 }
 
+static START_GC_LOOP: OnceCell<Sender<ZTenantId>> = OnceCell::new();
+static START_COMPACTION_LOOP: OnceCell<Sender<ZTenantId>> = OnceCell::new();
+
+/// Spawn a task that will periodically schedule garbage collection until
+/// the tenant becomes inactive. This should be called on tenant
+/// activation.
+pub fn start_gc_loop(tenantid: ZTenantId) -> Result<()> {
```
SomeoneToIgnore, line 68: > Result

but there are no result handling inside, only `unwrap`s. 
I think we should leave either, but not both. I'd go with `anyhow::Result` as more safe approach and since its `.context` works on `Option` just as well.


----

As a NOT style note, I find `-> anyhow::Result<()>` more readable since there's already an `std::result::Result` implicitly imported with the std prelude.

pageserver/src/tenant_threads.rs
```diff
@@ -1,77 +1,173 @@
...
+    START_GC_LOOP.set(gc_send).unwrap();
+
+    let (compaction_send, mut compaction_recv) = mpsc::channel::<ZTenantId>(100);
+    START_COMPACTION_LOOP.set(compaction_send).unwrap();
+
+    thread_mgr::spawn(
+        ThreadKind::TenantTaskManager,
+        None,
+        None,
+        "WAL receiver manager main thread",
```
SomeoneToIgnore, line 109: 🙂

pageserver/src/tenant_threads.rs
```diff
@@ -1,77 +1,173 @@
...
+        true,
```
SomeoneToIgnore, line 110: We did not bail on compaction and GC thread errors, so IMO neither should we now.

pageserver/src/tenant_threads.rs
```diff
@@ -1,77 +1,173 @@
...
+pub fn init_tenant_task_pool() -> Result<()> {
+    let runtime = tokio::runtime::Builder::new_multi_thread()
+        .thread_name("tenant-task-worker")
+        .worker_threads(40) // Way more than necessary
+        .max_blocking_threads(100) // Way more than necessary
+        .enable_all()
+        .build()?;
+
+    let (gc_send, mut gc_recv) = mpsc::channel::<ZTenantId>(100);
+    START_GC_LOOP.set(gc_send).unwrap();
```
SomeoneToIgnore, line 100: > -> Result<
> .unwrap(

One more semi-random agitation for better error handling. I also prefer `.expect(` in non-obvious places where we would want to panic for some reason, but even that is very close to `.context(..)?` already.

pageserver/src/tenant_threads.rs
```diff
@@ -1,77 +1,173 @@
...
+        ThreadKind::TenantTaskManager,
+        None,
+        None,
+        "WAL receiver manager main thread",
+        true,
+        move || {
+            runtime.block_on(async move {
+                loop {
+                    tokio::select! {
+                        _ = thread_mgr::shutdown_watcher() => break,
```
SomeoneToIgnore, line 115: > break

That would never stop either of the `select!` ed tasks: we drop their handles, returned from `tokio::spawn` and since don't pass any other cancellation channels to those loops, loose any cancellation/await control on those tasks.

Neither we are now able to determine if any of those tasks had finished abruptly with some panic before (and restart those), due to similar reasons.

I think your neighbour comment is related to this, but it's safer to notice.

pageserver/src/tenant_threads.rs
SomeoneToIgnore: Same for the compaction loop.

pageserver/src/tenant_threads.rs
```diff
@@ -1,77 +1,173 @@
...
-            std::thread::sleep(Duration::from_secs(1));
+            match gc_result {
+                Ok(Ok(_gc_result)) => {
+                    // Gc success, do nothing
+                }
+                Ok(Err(e)) => {
+                    anyhow::bail!(e.context("Gc failed"));
+                }
+                Err(e) => {
+                    anyhow::bail!("Gc join error {}", e);
```
SomeoneToIgnore, line 162: NIT: The join error is now happens due to panic in the underlying task only, so we could get more specific.

---

IMO, we should not `bail!` in both `Err` branches, since previously both threads were specifically set to not to fail the system and they should try to finish their work, otherwise the resources will be depleted?

pageserver/src/tenant_threads.rs
SomeoneToIgnore, line 40: > spawn_blocking

This looks a bit confusing, considering that

* `tenant_mgr::get_tenant_state`
* `tenant_mgr::get_repository_for_tenant`
* `repo.get_compaction_period`
* `tenant_mgr::get_repository_for_tenant`
* and even `tenant_mgr::get_tenant_state`

are blocking.
I'd slap a warning comment on top of the function and swapped it into `spawn_blocking` on a higher level.

SomeoneToIgnore: Same applies to the gc code.

pageserver/src/tenant_threads.rs
SomeoneToIgnore: To elaborate on the cancellations more: we cannot do `handle.abort` on the blocking task, but we can
* periodically `(&mut handle).await` the task to see if it's finished or not
* pass the `tokio::sync::watch::Receiver<()>`/`oneshot`/whatnot channels into the threads and to the cancellation manually there

SomeoneToIgnore (commented):
I like this approach more than the previous one and find it simpler generally, but I'm biased so take me with a grain of salt.

I find somewhat concerning the new gc and compaction task management, but if it fixes the thread issues on staging, never mind my concerns.
Looks good otherwise.

pageserver/src/tenant_threads.rs
LizardWizzard: > we cannot do handle.abort on the blocking task

cant we? It should be possible, but this would be a noop, so await on a handle will wait for task completion anyway

> periodically (&mut handle).await

I would suggest using `handle.is_finished`. 

> into the threads and to the cancellation manually there

Maybe we can integrate this with our thread_mgr and use shutdown checks from there

pageserver/src/tenant_mgr.rs
```diff
@@ -330,44 +328,9 @@ pub fn set_tenant_state(tenant_id: ZTenantId, new_state: TenantState) -> anyhow:
...
-                m.get_mut(&tenant_id)
-                    .with_context(|| format!("Tenant not found for id {tenant_id}"))?
-                    .state = old_state;
-                drop(m);
-                error!("Failed to start GC thread for tenant {tenant_id}, stopping its checkpointer thread: {e:?}");
-                thread_mgr::shutdown_threads(Some(ThreadKind::Compactor), Some(tenant_id), None);
-                return gc_spawn_result;
-            }
+            crate::tenant_threads::start_compaction_loop(tenant_id)?;
+            crate::tenant_threads::start_gc_loop(tenant_id)?;
```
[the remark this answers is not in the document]
LizardWizzard: I think worth a comment on how and when the tasks are terminated. Tasks should stop by themselves when the state becomes any other than Active (because there is a `break` with this condition inside them). It looks like this should work even without this patch, but for some reason, it doesn't

LizardWizzard:
Thanks! I like the approach. To me, it is a better option than take 1, but in that regard I'm as biased as @SomeoneToIgnore :) +1 for @SomeoneToIgnore comments

LizardWizzard (commented):
Thanks! I like the approach. To me, it is a better option than take 1, but in that regard I'm as biased as @SomeoneToIgnore :) +1 for @SomeoneToIgnore comments

pageserver/src/tenant_mgr.rs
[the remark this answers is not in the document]
bojanserafimov: My theory is that currently it doesnt because our load tester creates timelines and doesn't detach them later. Tenant only becomes idle when all its timelines are detached. IMO it should be based on whether there's any recent wal, or we should rename Idle to Detached

pageserver/src/tenant_threads.rs
bojanserafimov: When a task fails we should log and move on (and maybe mark the tenant as stuck), but the manager thread shouldn't fail. If it does we will run out of disk space and fail anyway. Same severity but harder to debug.

SomeoneToIgnore: Indeed, confused with the regular tasks.

pageserver/src/tenant_threads.rs
bojanserafimov: I used `expect`, which is fine in this case. `.context(..)?` is a bit clunky on non-anyhow errors

pageserver/src/tenant_threads.rs
```diff
@@ -1,78 +1,229 @@
...
+                            }
+                            // TODO wait for tasks to die?
+                            break;
+                        },
+                        tenantid = gc_recv.recv() => {
+                            let tenantid = tenantid.expect("Gc task channel closed unexpectedly");
+
+                            // Spawn new task, request cancellation of the old one if exists
+                            let (cancel_send, cancel_recv) = watch::channel(());
+                            let _handle = tokio::spawn(gc_loop(tenantid, cancel_recv))
```
SomeoneToIgnore, line 151: `gc_loop` does a bunch of `anyhow::bail!` and might even panic when calling `tenant_mgr::get_tenant_state` for logging.

The only way to know about this is to do `_handle.await` (or call other methods on the `_handle`) and to process the end result, otherwise we will never know how did the task end: we can add channels inside and know that the task had ended by the fact that the channel counterpart got dropped, but the end result (due to panic possibilities) has to be always checked on the handle.

This brings two implications:
* if anything fails in the GC/compaction thread now, we would never see the end error logged
* we don't restart such task before another set of `set_tenant_state` calls, meaning we may silently loose both compaction and gc tasks if very unlucky


Same applies to compaction (quite a routine comment in this PR related to tenant task management, I wonder if the logic is worth deduplicating already at this point)

SomeoneToIgnore (approved):
Thank you for fixing all this.

I think we're getting somewhat on par with what was before, but lack some error observability: let's add more error logging before `anyhow::bail` in both gc_loop & compaction_loop.

AFAIK, we were not restarting errored gc and compaction threads too before, so all my other comments are nice to have, but nothing required for this PR.

pageserver/src/tenant_threads.rs
```diff
@@ -1,78 +1,229 @@
...
-            break;
-        }
-        let repo = tenant_mgr::get_repository_for_tenant(tenantid)?;
-        let compaction_period = repo.get_compaction_period();
+        trace!("waking up");
 
-        std::thread::sleep(compaction_period);
-        trace!("compaction thread for tenant {} waking up", tenantid);
+        // Run blocking part of the task
+        let period: Result<Result<_, anyhow::Error>, _> = tokio::task::spawn_blocking(move || {
```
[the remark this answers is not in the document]
bojanserafimov: It gets confused without the `anyhow::Error` hint.

pageserver/src/tenant_threads.rs
```diff
@@ -1,78 +1,229 @@
...
+
+/// Spawn a task that will periodically schedule garbage collection until
+/// the tenant becomes inactive. This should be called on tenant
+/// activation.
+pub fn start_gc_loop(tenantid: ZTenantId) -> anyhow::Result<()> {
+    START_GC_LOOP
+        .get()
+        .context("Failed to get START_GC_LOOP")?
+        .blocking_send(tenantid)
+        .map_err(|e| anyhow::anyhow!(e))
```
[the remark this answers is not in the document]
bojanserafimov: apparently not :)

pageserver/src/tenant_threads.rs
bojanserafimov: Added an await with `FuturesUnordered`. I tested with a manual failpoint to see that the panic gets logged

bojanserafimov:
Tenant state metrics would be great too, but that's out of scope

SomeoneToIgnore (approved):

pageserver/src/tenant_tasks.rs
```diff
@@ -0,0 +1,253 @@
...
+                        },
+                        result = futures.next() => {
+                            // Log any errors and panics
+                            match result {
+                                Some(Ok(Ok(()))) => {
+                                    TENANT_TASK_EVENTS.with_label_values(&["stop"]).inc();
+                                },
+                                Some(Ok(Err(e))) => {
+                                    TENANT_TASK_EVENTS.with_label_values(&["fail"]).inc();
+                                    // TODO which tenant failed?
```
bojanserafimov, line 187: would be great to know. not sure if it's possible with this approach of error reporting

SomeoneToIgnore (approved):

pageserver/src/tenant_tasks.rs
```diff
@@ -0,0 +1,257 @@
...
+                    tokio::select! {
+                        _ = thread_mgr::shutdown_watcher() => {
+                            // TODO do this from thread_mgr? Extend it to work with tasks?
+                            for (_, cancel) in gc_loops.drain() {
+                                cancel.send(()).ok();
+                            }
+                            for (_, cancel) in compaction_loops.drain() {
+                                cancel.send(()).ok();
+                            }
+                            // TODO wait for tasks to die?
```
LizardWizzard, line 150: I would say that this is important. For tasks like detach/delete it is desired to join threads to be sure that they no longer access the filesystem. It is needed to have this on per-tenant basis too. So it is possible to cancel task and join it. Cancellation in this case is implicitly performed via setting tenant state to stopping, so the lacking piece is something `join`-like for these tasks. I can imagine two possible solutions, store join handles somewhere globally, or have a special pair of channels so ne can send join request and block on receiving confirmation message. Example of the latter approach can be found here: https://github.com/neondatabase/neon/pull/1936/files#diff-0d4539a7193c24a7f3349df105b8bf31a2767e74bcd4898532cb7d44838bf499R439 

cc @SomeoneToIgnore wdyt?

LizardWizzard (commented):
Looks nice! See one comment regarding joining tasks. To me this feature is already needed

pageserver/src/tenant_tasks.rs
bojanserafimov: Also we should be asserting that "no gc/compaction" is running, rather than "no gc/compaction **threads/tasks** are running". People are free to call `gc_iteration` outside of `thread_mgr` and `tenant_tasks`, and that will mess up detach/delete.

bojanserafimov: The sync/async boundary makes this so complicated. Neither of the two approaches suggested are easy:
1. Can't store join handles because the `FuturesUnordered` eats them
2. Can't use a channel because we could have multiple loops for a tenant and we want all of them to shut down

So I wrote something different in my [latest commit](https://github.com/neondatabase/neon/pull/1933/commits/bf76f43ea4c9d1cb0d5bcd47fafa12747fe98b36). I don't like it that much.. it's not very simple or readable, but if it's workable I'll polish it up a bit and we'll refine later. It has the benefit of also waiting for gc/compaction tasks that circumvent tenant_mgr or thread_mgr, for example [this one](https://github.com/neondatabase/neon/blob/tenant-tasks/pageserver/src/page_service.rs#L1071).

bojanserafimov: One way to simplify, maybe: https://github.com/neondatabase/neon/compare/tenant-tasks...tenant-tasks-cleanup?expand=1

LizardWizzard: >  It seems counter-intuitive to me that the "active -> idle" state transition would block on closing tenant tasks.

It shouldn't be a problem because active -> idle transition is not on the hot path, it happens in the background. It may be a problem if threads we're waiting for do not respond/are stuck. Then we cannot start the tenant again, or we can but it will be in a weird state when there is one semi-dead gc task and one that is new and trying to work.

> Can't store join handles because the FuturesUnordered eats them

Join handles can be stored separately without using FuturesUnordered like it is done in walreceiver.

> Can't use a channel because we could have multiple loops for a tenant and we want all of them to shut down

There should be only one gc/compaction loop for a tenant at a time, or I'm missing something? If we need to notify multiple consumers we can use something like tokio watch or broadcast channel.

pageserver/src/layered_repository.rs
```diff
@@ -158,6 +158,13 @@ pub struct LayeredRepository {
     // Global pageserver config parameters
     pub conf: &'static PageServerConf,
 
+    // Allows us to gracefully cancel operations that edit the directory
+    // that backs this layered repository. Usage:
+    //
+    // Use `let _guard = file_lock.try_read()` while writing any files.
+    // Use `let _guard = file_lock.write().unwrap()` to wait for all writes to finish.
+    pub file_lock: RwLock<()>,
```
LizardWizzard, line 166: With ithis wording we actually need to take this lock for walreceiver too, because it flushes data to disk with its separate thread

pageserver/src/tenant_tasks.rs
bojanserafimov: > Join handles can be stored separately without using FuturesUnordered like it is done in walreceiver.

How do we catch and report errors/panics from tasks if we don't await them? Catch unwind inside the task? Or is there some tokio settings to auto-report?

> There should be only one gc/compaction loop for a tenant at a time, or I'm missing something?

A few commits ago it was possible to have many loops per tenant because when we cancel any existing tasks [we don't wait](https://github.com/neondatabase/neon/blob/2b4c3cb9329046bdbbedf86ec1a2f3959d0c8b66/pageserver/src/tenant_tasks.rs#L183) for them to shutdown. I guess that is fixed now that we wait for shutdown when we go "active -> idle".

But anyway, there are two other problems I already mentioned:
1. We don't wait for gc/compaction that's not part of a thread/task
2. Race conditions between state transitions. Even if we solve these, there's another race condition in detach/delete. We only make sure that there are no gc/compactions at the start of the detach/delete, but gc/compaction can start during detach/delete. It could be possible to end up with a semi-deleted active tenant without noticing. We need to hold a lock (or equivalent) while deleting.

pageserver/src/layered_repository.rs
bojanserafimov: I guess also for branch creation and checkpoint. Seems like an issue for a separate PR.

LizardWizzard:
Personally I'm ok with this rwlock approach, though I'm for unification where possible, so if walreceiver approach with cancellation channels works here too I would use it here ass well. Or if we're changing it to something different lets change it in both places. Currently tenant state management already has some problems we've discussed possible improvements with @SomeoneToIgnore so we can polish it separately. What is important this patch should solve the thread count issue. Do you have ideas on tests we can add to check that all this works correctly?

bojanserafimov:
I'll write a test to assert that tasks started and stopped at least

pageserver/src/tenant_mgr.rs
```diff
@@ -379,8 +345,11 @@ pub fn set_tenant_state(tenant_id: ZTenantId, new_state: TenantState) -> anyhow:
                 Some(tenant_id),
                 None,
             );
-            thread_mgr::shutdown_threads(Some(ThreadKind::GarbageCollector), Some(tenant_id), None);
-            thread_mgr::shutdown_threads(Some(ThreadKind::Compactor), Some(tenant_id), None);
+
+            // Wait until all gc/compaction tasks finish
+            // TODO send cancellation signal too, or make the state a watch
```
bojanserafimov, line 350: This is necessary because otherwise we'll wait an entire gc_period. Will fix before merge

pageserver/src/tenant_tasks.rs
```diff
@@ -0,0 +1,289 @@
...
+                            // Update metrics, remember handle
+                            TENANT_TASK_EVENTS.with_label_values(&["start"]).inc();
+                            futures.push(handle);
+                        },
+                        tenantid = compaction_recv.recv() => {
+                            let tenantid = tenantid.expect("Compaction task channel closed unexpectedly");
+
+                            // Spawn new task, request cancellation of the old one if exists
+                            let (cancel_send, cancel_recv) = watch::channel(());
+                            // TODO this instrument doesn't work
```
bojanserafimov, line 195: Not sure why. `info!` traces from inside the future don't show up with `compaction loop` as context. Will fix before merging

test_runner/batch_others/test_tenant_tasks.py
```diff
@@ -0,0 +1,69 @@
...
+            client.timeline_detach(tenant, t)
+
+    # Create tenant, start compute
+    tenant, _ = env.neon_cli.create_tenant()
+    timeline = env.neon_cli.create_timeline(name, tenant_id=tenant)
+    pg = env.postgres.create_start(name, tenant_id=tenant)
+    assert (get_state(tenant) == "Active")
+
+    # Stop compute, detach timelines
+    # TODO tenant should go idle even if we don't explicitly detach
```
bojanserafimov, line 41: I'll create an issue if this seems like desired behavior

test_runner/batch_others/test_tenant_tasks.py
```diff
@@ -0,0 +1,69 @@
...
+    # Read metrics
+    import time
+    time.sleep(1)
+    tasks_started = get_metric_value('pageserver_tenant_task_events{event="start"}')
+    tasks_ended = get_metric_value('pageserver_tenant_task_events{event="stop"}')
+    tasks_panicked = get_metric_value('pageserver_tenant_task_events{event="panic"}')
+
+    # TODO this fails because the "active -> idle" transition only waits for gc to
+    #      finish without cancelling it, and gc waits for 100 seconds.
+    # assert tasks_started == tasks_ended
```
bojanserafimov, line 68: Will resolve before merging

pageserver/src/tenant_mgr.rs
bojanserafimov: Fixed the test instead. It's not a big problem

pageserver/src/tenant_tasks.rs
bojanserafimov: Replacing `trace_span` with `info_span` fixes the issue. What are the logging level semantics here? Does a span with a certain level apply only to events of an equal or stronger level? The docs don't say anything.

test_runner/batch_others/test_tenant_tasks.py
LizardWizzard: Please do. Tenant should eventually (when all the wal is consumed by pageserver) become idle when compute is down

pageserver/src/layered_repository.rs
LizardWizzard: The flush is a checkpoint basically. Branch creation holds gc lock for that reason :) 

Lets mention here that it does not cover all situations with a todo or something

pageserver/src/tenant_tasks.rs
LizardWizzard: > How do we catch and report errors/panics from tasks if we don't await them?

Hmm, you're right, it is possible to await on join handles, but this way you need to have some communication from the task itself because otherwise you wont know if it is finished or not so without awaiting this join handle will be kept forever, and FuturesUnordered solves this. I think walreceiver uses this message passing mechanism when spawned task communicates updates back to its parent. But here it is not needed, tasks are rather short lived, and this is another difference with walreceiver, where tasks live for whole compute lifetime. 

> Catch unwind inside the task? Or is there some tokio settings to auto-report?

Tokio will catch panic inside task, it uses catch unwind around the poll method of a future and dumps it into logs, but it wont increment metrics and so on. So you're right :)

pageserver/src/tenant_tasks.rs
LizardWizzard: Hmm, thats weird, I tried the example from docs here https://docs.rs/tracing/latest/tracing/index.html#configuring-attributes and it worked for me, the info message was emitted. Though it does not use futures. I'll try reproduce it tomorrow the code from this branch

LizardWizzard (approved):
Looks good to me, thanks! 

One nit: I think the place with the new file lock may benefit from an extended comment mentioning cases which are not covered by the lock. Maybe even point to an issue

It should conflict with my PR #1936, so go ahead and I'll rebase my patch on top of yours :)

bojanserafimov:
stage metrics, fyi https://observer.zenith.tech/d/GOx33ve7z/tenant-tasks?orgId=1

PR #1937, opened 2022-06-15
Rework walreceiver connections

SomeoneToIgnore:
Addresses the follow-ups:
* https://github.com/neondatabase/neon/pull/1892#discussion_r891033987 to handle timeline changes from etcd per one entry
* https://github.com/neondatabase/neon/pull/1619#discussion_r892119620 to not to close the WAL connections eagerly
* https://github.com/neondatabase/neon/pull/1619#discussion_r888610670 (partially) by adjusting the spans so that there are no duplicates now (?)
* the review comment from @arssher about more task-based approach instead of a bunch of channels we have now

I've extracted the connection logic into a separate file, where added most of these changes (etcd ones were done in a different place).

### DIFF
--- a/pageserver/src/walreceiver.rs
+++ b/pageserver/src/walreceiver.rs
@@ -15,66 +15,38 @@
 //!
 //! * handle the actual connection and WAL streaming
 //!
-//! Handle happens dynamically, by portions of WAL being processed and registered in the server.
+//! Handling happens dynamically, by portions of WAL being processed and registered in the server.
 //! Along with the registration, certain metadata is written to show WAL streaming progress and rely on that when considering safekeepers for connection.
 //!
-//! ## Implementation details
-//!
-//! WAL receiver's implementation consists of 3 kinds of nested loops, separately handling the logic from the bullets above:
-//!
-//! * [`init_wal_receiver_main_thread`], a wal receiver main thread, containing the control async loop: timeline addition/removal and interruption of a whole thread handling.
-//! The loop is infallible, always trying to continue with the new tasks, the only place where it can fail is its initialization.
-//! All of the code inside the loop is either async or a spawn_blocking wrapper around the sync code.
-//!
-//! * [`timeline_wal_broker_loop_step`], a broker task, handling the etcd broker subscription and polling, safekeeper selection logic and [re]connects.
-//! On every concequent broker/wal streamer connection attempt, the loop steps are forced to wait for some time before running,
-//! increasing with the number of attempts (capped with some fixed value).
-//! This is done endlessly, to ensure we don't miss the WAL streaming when it gets available on one of the safekeepers.
-//!
-//! Apart from the broker management, it keeps the wal streaming connection open, with the safekeeper having the most advanced timeline state.
-//! The connection could be closed from safekeeper side (with error or not), could be cancelled from pageserver side from time to time.
-//!
-//! * [`connection_handler::handle_walreceiver_connection`], a wal streaming task, opening the libpq connection and reading the data out of it to the end.
-//! Does periodic reporting of the progress, to share some of the data via external HTTP API and to ensure we're able to switch connections when needed.
-//!
-//! Every task is cancellable via its separate cancellation channel,
-//! also every such task's dependency (broker subscription or the data source channel) cancellation/drop triggers the corresponding task cancellation either.
+//! The current module contains high-level primitives used in the submodules; general synchronization, timeline acknowledgement and shutdown logic.
 
-mod connection_handler;
+mod connection_manager;
+mod walreceiver_connection;
 
-use crate::config::PageServerConf;
-use crate::http::models::WalReceiverEntry;
-use crate::repository::Timeline;
-use crate::tenant_mgr::{self, LocalTimelineUpdate, TenantState};
-use crate::thread_mgr::ThreadKind;
-use crate::{thread_mgr, DatadirTimelineImpl};
 use anyhow::{ensure, Context};
-use chrono::{NaiveDateTime, Utc};
-use etcd_broker::{
-    subscription_key::SubscriptionKey, subscription_value::SkTimelineInfo, BrokerSubscription,
-    Client,
-};
+use etcd_broker::Client;
 use itertools::Itertools;
 use once_cell::sync::Lazy;
 use std::cell::Cell;
 use std::collections::{hash_map, HashMap, HashSet};
+use std::future::Future;
 use std::num::NonZeroU64;
-use std::ops::ControlFlow;
 use std::sync::Arc;
 use std::thread_local;
 use std::time::Duration;
-use tokio::select;
 use tokio::{
+    select,
     sync::{mpsc, watch, RwLock},
     task::JoinHandle,
 };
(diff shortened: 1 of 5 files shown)


SomeoneToIgnore:
During test restarts, I've also noted that the new model makes one test to flack: https://app.circleci.com/pipelines/github/neondatabase/neon/7180/workflows/4ea8bf6c-dff2-41f5-98f4-fcc2a2b0734f/jobs/73553/tests

I will push a fix to that shortly.

pageserver/src/walreceiver.rs
```diff
@@ -162,6 +134,97 @@ pub fn init_wal_receiver_main_thread(
...
+            }
+        }
+    })
+    .await
+    {
+        error!("Failed to spawn a task to make all tenants idle: {e:?}");
+    }
+}
+
+/// A handle of an asynchronousl task.
```
petuhovskiy, line 173:
```suggestion
/// A handle of an asynchronous task.
```

LizardWizzard (commented):
Looks good to me. Though I think that the code is complex enough that it might be a good Idea to insert mermaid diagrams in doc comments...

I've left some minor comments. I've also checked coverage report and some reconnect reasons are not covered (and few other places). Maybe for the sake of completeness worth it co cover these places

pageserver/src/walreceiver/connection_manager.rs
```diff
@@ -0,0 +1,988 @@
...
+                    Some(broker_update) => walreceiver_state.register_timeline_update(broker_update),
+                    None => {
+                        info!("Broker sender end was gropped, ending current broker loop step");
+                        return;
+                    }
+                }
+            },
+        }
+
+        // Fetch more etcd timeline updates, but limit ourselves since they may arrive quickly.
```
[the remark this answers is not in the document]
SomeoneToIgnore: No, you are not, same calculations I've had. 
I can decrease it to a 100, but decided that this way is good enough.

pageserver/src/walreceiver/connection_manager.rs
```diff
@@ -0,0 +1,988 @@
...
+                                                wal_producer_connstr: new_wal_producer_connstr,
+                                                reason: ReconnectReason::LaggingWal { current_lsn, new_lsn, threshold: self.max_lsn_wal_lag },
+                                            });
+                                    }
+                                }
+                                None => debug!("Best SK candidate has its commit Lsn behind the current timeline's latest consistent Lsn"),
+                            }
+                    }
+                    None => {
+                        return Some(NewWalConnectionCandidate {
```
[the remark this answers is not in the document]
SomeoneToIgnore: Hm, I've added the tests on both, you can see the `NoEtcdDataForExistingConnection` usage there clearly, but the coverage link from the PR still shows it's uncovered for some reason.

pageserver/src/walreceiver/connection_manager.rs
```diff
@@ -0,0 +1,1134 @@
...
+    wal_connect_timeout: Duration,
+    /// The timeout to use to determine when the current connection is "stale" and reconnect to the other one.
+    lagging_wal_timeout: Duration,
+    /// The Lsn lag to use to determine when the current connection is lagging to much behind and reconnect to the other one.
+    max_lsn_wal_lag: NonZeroU64,
+    /// Current connection to safekeeper for WAL streaming.
+    wal_connection: Option<WalConnection>,
+    wal_connection_attempts: HashMap<NodeId, u32>,
+    /// Data about all timelines, available for connection, fetched from etcd.
+    wal_stream_candidates: HashMap<SubscriptionFullKey, EtcdSkTimeline>,
```
arssher, line 266: Why key here is `SubscriptionFullKey`, won't just safekeeper node id be simpler?

pageserver/src/walreceiver/connection_manager.rs
```diff
@@ -0,0 +1,1134 @@
...
+
+    let mut broker_subscription =
+        subscribe_for_timeline_updates(etcd_client, broker_prefix, id).await;
+    info!("Subscribed for etcd timeline changes, waiting for new etcd data");
+
+    loop {
+        select! {
+            // the order of the polls is especially important here, since the first task to complete gets selected and the others get dropped (cancelled).
+            // place more frequetly updated tasks below to ensure the "slow" tasks are also reacted to.
+            biased;
```
arssher, line 101: I don't get the idea of using biased here. Without it, select picks random branch, and I think it works for us here. With it, we must think about order of arms + still get some probability of lower branches never being picked up.

pageserver/src/walreceiver/connection_manager.rs
```diff
@@ -0,0 +1,1134 @@
...
+
+            // First, check if the broker connection is intact, otherwise we need to restart the whole step
+            broker_connection_result = &mut broker_subscription.watcher_handle => {
```
arssher, line 104: Since closing watch channel and task ending don't happen atomically, there is a chance of getting out through `broker_subscription.value_updates.recv()` arm `returning None`, in which case again error is swallowed.

pageserver/src/walreceiver/connection_manager.rs
```diff
@@ -0,0 +1,1134 @@
...
+                    }
+                }
+            },
+
+            // If nothing else happens, ensure we are not stuck waiting and still react on etcd changes.
+            // Register the value from the broker or abort the loop, if the subscription got cancelled.
+
+            // XXX: etcd changes are frequent, so don't place it above the slower source events such as WAL receiver one, otherwise
+            // the slow ones might get ignored constantly.
+            broker_update = broker_subscription.value_updates.recv() => {
```
arssher, line 169: I realized with this patch watch is never canceled, so task is leaked if connection manager is canceled. Perhaps it would be ok to add cancel to `BrokerSubscription` impl Drop. Hmm, on the second thought this leak is very short living as it will fail sending to closed channel.

arssher (commented):
Looks mostly good!

pageserver/src/walreceiver/connection_manager.rs
SomeoneToIgnore: Had thought of it more and looks like we can rely on basic select's fairness indeed, removed it and a bunch of comments.

pageserver/src/walreceiver/connection_manager.rs
SomeoneToIgnore: I think it's ok to `imp Drop`, thank you for spotting.

Indeed, we're very much guarded buy all these shared channel ends being dropped as signals to stop working.
IMO, it's a bit dangerous since on the top level, where we register new timelines with endless tasks, we spawn channel-less endless tasks to poll etcd and if they panic, we won't even know that now, since we never call `handle.await` before the shutdown.

The only solution I see now is to have a separate loop (that also may panic) to check for such tasks constantly, restarting the abruptly stopped ones, but that requires way more changes that I'm ready to do now.

arssher (approved):

SomeoneToIgnore:
New red tests come from https://github.com/neondatabase/neon/issues/1414 , CircleCI run looks ok, so merging this.

PR #1951, opened 2022-06-17
Backported 36ee182d. Implement page servise 'fullbackup' endpoint

lubennikovaav:
that works like basebackup, but also sends relational files

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -246,7 +246,7 @@ dependencies = [
 [[package]]
 name = "bookfile"
 version = "0.3.0"
-source = "git+https://github.com/zenithdb/bookfile.git?branch=generic-readext#d51a99c7a0be48c3d9cc7cb85c9b7fb05ce1100c"
+source = "git+https://github.com/neondatabase/bookfile.git?branch=main#bf6e43825dfb6e749ae9b80e8372c8fea76cec2f"
 dependencies = [
  "aversion",
  "byteorder",
@@ -1467,6 +1467,7 @@ dependencies = [
  "hex-literal",
  "humantime",
  "hyper",
+ "itertools",
  "lazy_static",
  "log",
  "nix",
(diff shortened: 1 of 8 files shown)


bojanserafimov (approved):

lubennikovaav:
I've attempted to implement a hotfix endpoint `import empty rel` that would call `put_rel_creation` for provided relpath,
 just like `import_rel()` from `import_datadir.rs`.
 
But it didn't work out, because we do not allow to insert data at earlier lsns.
I got this error: `cannot modify relation after advancing last_record_lsn (incoming_lsn={}, last_record_lsn={})`
Which makes a lot of sense.

So another idea is to correct the data (load lost relations) while uploading basebackup.
That is awful, but I can't come up with any better solution(

bojanserafimov:
> So another idea is to correct the data (load lost relations) while uploading basebackup.

Is there a way to get a list of those rels, other than running a compute node and looking for errors? I'm thinking maybe load the exported basebackup into a local vanilla pg, and run some query, hoping we get an answer before postgres notices that rel files are missing?

I tried running `test_fullbackup` with a broken exporter that skips empty files, and postgres didn't complain. I checked that there was at least one skipped empty rel

lubennikovaav:
>I tried running test_fullbackup with a broken exporter that skips empty files, and postgres didn't complain. I checked that there was at least one skipped empty rel

That's understandable - fullbackup successfully returns what was loaded. And we haven't loaded missing file.

>Is there a way to get a list of those rels, other than running a compute node and looking for errors? I'm thinking maybe load the exported basebackup into a local vanilla pg, and run some query, hoping we get an answer before postgres notices that rel files are missing?

I did it very straightforward and just hardcoded lists of such files from `initdb`.
But now I understand that it won't help with user tables, that were created and truncated to 0

and we still need to find a way to fix it after data is uploaded (

PR #1953, opened 2022-06-17
Fix WAL timeout in test_s3_wal_replay

petuhovskiy:
I did that by reimplementing wait_for_lsn in python with a higher timeout. walreceiver seems to be triggered by itself, because safekeepers are pushing to etcd.

@SomeoneToIgnore please force push


SomeoneToIgnore (approved):
Here's another way to get info about pageserver's Lsns, now from walreceiver directly: https://github.com/neondatabase/neon/blob/e2a5a3159556341ef53cfe4a87180c0ffa5dadbd/test_runner/batch_others/test_pageserver_api.py#L73-L97

PR #1955, opened 2022-06-17
basebackup import fixes

lubennikovaav:
It doesn't merges clearly, because I've rebased `basebackup-import` on top of latest main.
Feel free to just cherry-pick last commit.


lubennikovaav:
Fixed pg_control checks. They were too tight and didn't apply to import_from_vanilla case


>does it mean zenith.signal should be at the end of the tar? Let's just document ordering expectations, since it's getting complicated

Not necessarily. It happens before `modification.commit()?;` that updates lsn to the final one, so everything should be ok.

>Will it work with bail!?

Probably yes, but why would we need it? Let's just skip the file

bojanserafimov (approved):
Looks good, thanks! Left a few questions.

You can merge the parent branch into this one, then "squash and merge"

PR #1975, opened 2022-06-22
Don't require project name for link auth

bojanserafimov:
There's not much point to link auth if the user already has a project.


kelvich (approved):

bojanserafimov:
We should also make sure e2e test doesn't provide project name. Seems like currently it does, only in the default config, which might not be used for stage e2e (i'm not 100% sure)

PR #1981, opened 2022-06-23
Fix zero timeline_start_lsn

### LINKED ISSUE #1980: Fix safekeeper connection for migrated timeline.
After moving from old to new pageserver (ps-1 --> ps-3), the compute fails to start with `--sync-safekeepers` failure.

experimental project is  https://console.stage.neon.tech/admin/projects/wispy-voice-850996

petuhovskiy:
Currently there is a problem with old timelines, which are created before we introduced `timeline_start_lsn`. This problem is zero `timeline_start_lsn` and incorrect `local_start_lsn`. More specifically, we have the following code:

https://github.com/neondatabase/neon/blob/84b9fcbbd59d7cfaa4181f2f1b6869d248070be0/safekeeper/src/safekeeper.rs#L707-L717

This code updates `local_start_lsn` to proposer's start streaming at point. This is ok for new timelines, because `timeline_start_lsn` is not zero and we initialize everything only once. But this is working incorrectly for old timelines, because we don't know `timeline_start_lsn` and this `timeline_start_lsn` is always zero. That's why we change `local_start_lsn` every time, which is wrong.

Fix sets LSNs to Lsn(1), which is very hacky, but works for now and will allow us to fix this later.

Should fix #1980

### DIFF
--- a/safekeeper/src/control_file_upgrade.rs
+++ b/safekeeper/src/control_file_upgrade.rs
@@ -247,6 +247,7 @@ pub fn upgrade_control_file(buf: &[u8], version: u32) -> Result<SafeKeeperState>
         }
 
         // set special timeline_start_lsn because we don't know the real one
+        info!("setting timeline_start_lsn and local_start_lsn to Lsn(1)");
         oldstate.timeline_start_lsn = Lsn(1);
         oldstate.local_start_lsn = Lsn(1);
 
--- a/safekeeper/src/control_file_upgrade.rs
+++ b/safekeeper/src/control_file_upgrade.rs
@@ -239,6 +239,18 @@ pub fn upgrade_control_file(buf: &[u8], version: u32) -> Result<SafeKeeperState>
             remote_consistent_lsn: Lsn(0),
             peers: Peers(vec![]),
         });
+    } else if version == 5 {
+        info!("reading safekeeper control file version {}", version);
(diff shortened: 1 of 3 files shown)


lubennikovaav:
looks good to me, but I'd wait for @arssher review

arssher (approved):
I'd insert this hack at 
```
 if state.timeline_start_lsn == Lsn(0) { 
     // Remember point where WAL begins globally. 
     state.timeline_start_lsn = msg.timeline_start_lsn; 
```
place, as msg.timeline_start_lsn can't be zero for new timelines. 

Current version works too, but slightly risky, as valid empty control file is possible. We can remove this after all old timelines are fixed.

petuhovskiy:
Sounds good, I'll move that hack to that place. If so, we need to run `--sync-safekeepers` for this to be applied, right?

petuhovskiy:
> I'd insert this hack at handle_elected

Looked again, I think this will not work. Currently `--sync-safekeepers` fails on recovery, because `START REPLICATION` is not working because of `local_start_lsn`. So we'll never reach `handle_elected` and this fix will not be applied. I think it's safe to merge and deploy on staging. If anything, we'll fix it before this code will reach production.

petuhovskiy:
Ready to merge. If you want to deploy it to staging, please press Squash and merge :)
I cannot do that because of circleci ban

cc @lubennikovaav

PR #1986, opened 2022-06-24
Migrate from CircleCI to Github Actions: docker build and deploy

SergeyMelnikov:


SomeoneToIgnore:
Have been doing partially intersecting work, following up the comments from https://github.com/neondatabase/neon/pull/1647

https://github.com/neondatabase/neon/pull/1987

I've not added Docker builds, but had fixed the coverage reports, presumably and it's done differently than here (since we have to collect coverage from both Python and Rust test runs).

.circleci/ansible/neon-stress.hosts
```diff
@@ -12,6 +12,7 @@ pageservers
 safekeepers
```
SomeoneToIgnore, line 12: Since `ansible` files are now only used in GH Actions, we could move it and remove the whole `.circleci` directory (if we move the `release` deploys too, as suggested)

.circleci/config.yml
```diff
@@ -448,106 +389,6 @@ jobs:
...
-            curl -s https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash
-            helm repo add neondatabase https://neondatabase.github.io/helm-charts
-      - run:
-          name: Re-deploy proxy
-          command: |
-            DOCKER_TAG=$(git log --oneline|wc -l)
-            helm upgrade neon-stress-proxy       neondatabase/neon-proxy --install -f .circleci/helm-values/neon-stress.proxy.yaml --set image.tag=${DOCKER_TAG} --wait
-            helm upgrade neon-stress-proxy-scram neondatabase/neon-proxy --install -f .circleci/helm-values/neon-stress.proxy-scram.yaml --set image.tag=${DOCKER_TAG} --wait
-
   deploy-release:
```
SomeoneToIgnore, line 392: Should we also move the release deploys out of here, using something like https://github.com/neondatabase/neon/pull/1454/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R104-R114
?

This way, we can remove the CircleCI files altogether, given that `staging` release works.

SomeoneToIgnore (approved):
If I'm not mistaken, to fully finish the transition, we only need to do similar deployment steps for the `release` branch, we can do that too and finally close the issue.

.circleci/config.yml
SergeyMelnikov: Yes, I also moved release deploy here

SomeoneToIgnore (approved):
lgtm, but I'm bad with deploy scripts and might have missed something.

I think we can remove the entire `.circleci` dir with all the configs now.

PR #1988, opened 2022-06-24
remove unused imports, make more use of pathlib.Path

LizardWizzard:


test_runner/batch_others/test_wal_restore.py
```diff
@@ -22,13 +23,13 @@ def test_wal_restore(neon_env_builder: NeonEnvBuilder,
...
     port = port_distributor.get_port()
-    data_dir = os.path.join(test_output_dir, 'pgsql.restored')
+    data_dir = test_output_dir / 'pgsql.restored'
     with VanillaPostgres(data_dir, PgBin(test_output_dir), port) as restored:
         pg_bin.run_capture([
             os.path.join(base_dir, 'libs/utils/scripts/restore_from_wal.sh'),
             os.path.join(pg_distrib_dir, 'bin'),
-            os.path.join(test_output_dir, 'repo/safekeepers/sk1/{}/*'.format(tenant_id)),
-            data_dir,
+            str(test_output_dir / 'repo/safekeepers/sk1/{}/*'.format(tenant_id)),
```
SomeoneToIgnore, line 31: NIT: right part could be an f-string for more readability?

SomeoneToIgnore (approved):
Always wondered, if there's any automated way to lint such missing imports, we could use it at some point if we stick with Python.

test_runner/batch_pg_regress/test_neon_regress.py
```diff
@@ -17,9 +16,10 @@ def test_neon_regress(neon_simple_env: NeonEnv, test_output_dir, pg_bin, capsys)
     pg.safe_psql('CREATE DATABASE regression')
 
     # Create some local directories for pg_regress to run in.
-    runpath = os.path.join(test_output_dir, 'regress')
-    mkdir_if_needed(runpath)
-    mkdir_if_needed(os.path.join(runpath, 'testtablespace'))
+    runpath = test_output_dir / 'regress'
+    runpath.mkdir()
+
+    (runpath / 'testtablespace').mkdir()
```
funbringer, line 22:
```python
(runpath / 'testtablespace').mkdir(parents=True)
```

test_runner/fixtures/neon_fixtures.py
```diff
@@ -1998,14 +2001,14 @@ def get_test_output_dir(request: Any) -> str:
 # this fixture ensures that the directory exists.  That works because
 # 'autouse' fixtures are run before other fixtures.
 @pytest.fixture(scope='function', autouse=True)
-def test_output_dir(request: Any) -> str:
+def test_output_dir(request: Any) -> pathlib.Path:
```
funbringer, line 2004: nit: It seems that both `Path` and `pathlib.Path` are used arbitrarily. Could we stick with just one, perhaps `Path`?

funbringer (approved):

Issue #1191, opened 2022-01-27
UDF in postgres to current build (for postgres and zenith extension)


kelvich:

ololobus:
Closing, there is an equivalent Jira ticket for that https://databricks.atlassian.net/browse/LKB-392

PR #1992, opened 2022-06-24
WalStreamDecoder refactoring

yeputons:
Based on #1990, it should be merged afterward.

This PR aims to straighten the logic in `WalStreamDecoder` and add more checks that are currently present in `find_end_of_wal` only. These changes have already exposed a bug, #1191 originally came from there.

It should be much easier to review with `git show HEAD -w --color-moved`

### DIFF
--- a/libs/postgres_ffi/src/waldecoder.rs
+++ b/libs/postgres_ffi/src/waldecoder.rs
@@ -13,6 +13,7 @@ use super::xlog_utils::*;
 use super::XLogLongPageHeaderData;
 use super::XLogPageHeaderData;
 use super::XLogRecord;
+use super::XLOG_PAGE_MAGIC;
 use bytes::{Buf, BufMut, Bytes, BytesMut};
 use crc32c::*;
 use log::*;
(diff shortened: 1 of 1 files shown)


aome510:
> @aome510, since you were just performance testing WAL ingestion speed in https://github.com/neondatabase/neon/pull/2010, can you also check what effect this has on it?

Take a quick look. I'm not familiar with the WalStreamDecoder codes, so couldn't review the changes. 

#2010 doesn't optimize the performance for WalStreamDecoder, it optimizes the code after receiving the WAL bytes from the stream decoder. Given this PR is just a refactoring, I guess there shouldn't be much of an effect on the performance.

yeputons:
I expect to fix the test failure via #2051

libs/postgres_ffi/src/waldecoder.rs
```diff
@@ -76,128 +129,121 @@ impl WalStreamDecoder {
...
+                    }
+                    // Fast path for the common case that the whole record fits on the page.
+                    let pageleft = self.lsn.remaining_in_block() as u32;
+                    if self.inputbuf.remaining() >= xl_tot_len as usize && xl_tot_len <= pageleft {
+                        self.lsn += xl_tot_len as u64;
+                        let recordbuf = self.inputbuf.copy_to_bytes(xl_tot_len as usize);
+                        return Ok(Some(self.complete_record(recordbuf)?));
+                    } else {
+                        // Need to assemble the record from pieces. Remember the size of the
+                        // record, and loop back. On next iteration, we will reach the 'else'
```
arssher, line 200: this is not 'else' branch anymore

arssher (approved):

PR #1996, opened 2022-06-27
Fix walreceiver reconnection attempts maintenance

arssher:


arssher:
@SomeoneToIgnore I see a branch fixing the same, feel free to choose version you like better.

PR #1997, opened 2022-06-27
Console split RFC

petuhovskiy:
[Rendered](https://github.com/neondatabase/neon/blob/rfc-console-split/docs/rfcs/017-console-split.md)


docs/rfcs/017-console-split.md
```diff
@@ -0,0 +1,420 @@
...
+
+- using k8s API
+- doing requests to any of the storage services (proxy, compute, safekeeper, pageserver, etc..)
+- tracking current status of tenants/timelines, managing lifetime of computes
+
+Based on that idea, we can say that new control-plane service should have the following components:
+
+- single HTTP API for everything
+    - Create and manage tenants and timelines
+    - Manage global settings and storage configuration (regions, platforms, safekeepers, pageservers)
```
petuhovskiy, line 169: > regions, platforms

Stas Kelvich:
I think we can assume that control plane is in one region, but console knows list of deployed control planes in different regions.

docs/rfcs/017-console-split.md
```diff
@@ -0,0 +1,420 @@
...
+```
+GET /events/<cursor>
+
+{
+  "events": [...],
+  "next_cursor": 123
+}
+```
+
+It should be possible to replay event logs from some point of time, to get a state of almost anything from the storage services. That means that if we maintain some state in the control-plane database and we have a reason to have the same state in the console database, it is possible by polling events from the control-plane API and changing the state in the console database according to the events.
```
petuhovskiy, line 322: Stas Kelvich:
Sounds a bit like kafka :) Should we just embrace kafka? Can it help with the current operation executer so that ops do live in kafka instead of postgres?

I would like to avoid kafka, but will be cool to discuss pros/cons.

hlinnaka (approved):
+1, I love it!

docs/rfcs/017-console-split.md
```diff
@@ -0,0 +1,420 @@
...
+GET /tenants/{tenant_id}/operations
+GET /tenants/{tenant_id}/operations/{operation_id}
+POST /tenants/{tenant_id}/query
+GET /tenants/{tenant_id}/roles
+POST /tenants/{tenant_id}/roles
+GET /tenants/{tenant_id}/roles/{role_name}
+DELETE /tenants/{tenant_id}/roles/{role_name}
+POST /tenants/{tenant_id}/roles/{role_name}/reset_password
+POST /tenants/{tenant_id}/start
+POST /tenants/{tenant_id}/stop
```
ololobus, line 286: I think that we need similar to project concept/abstraction in the control-plane too.

Tenant (or project in the current console) is a collection of many things: timelines, settings, compute conninfo, etc. Just renaming project -> tenant doesn't feel as a good abstraction to me.

I'd propose that we introduce the `compute` entity here. I'll try to elaborate:
- you can have many timelines, but some could be kept as backups / PITR marks
- we need some place where to store postgres settings and connection info
- yes, for the first time we can keep all of this in the timeline, but later we will anyway need read-replicas and it will be a separate compute on the same timeline
- we need a place to keep and track auto-scaling limits
- we need a place to keep connectivity info (e.g. allow connections, allow proxy link auth, etc)

docs/rfcs/017-console-split.md
```diff
@@ -0,0 +1,420 @@
...
+POST /projects/{project_id}/stop
+POST /psql_session/{psql_session_id}
+```
+
+It looks fine and we probably already have clients relying on it. So we should not change it, at least for now. But most of these endpoints (if not all) are related to storage, and it can suggest us what control-plane API should look like:
+
+```
+GET /tenants/{tenant_id}
+PATCH /tenants/{tenant_id}
+POST /tenants/{tenant_id}/branches
```
ololobus, line 269: Do we need branches in the control-plane? Or we can explicitly call them timelines? And then branch is just a timeline out of timeline (non-root-timeline)

And console will be the place that knows mapping of the timelines, tenants and computes to projects and branches

docs/rfcs/017-console-split.md
```diff
@@ -0,0 +1,420 @@
...
+GET /tenants/{tenant_id}/databases
+POST /tenants/{tenant_id}/databases
```
ololobus, line 271: This should likely be in the timeline/branch scope, as different timelines may have different sets of roles and databases

docs/rfcs/017-console-split.md
```diff
@@ -0,0 +1,420 @@
...
+GET /tenants/{tenant_id}/databases/{database_id}
+PUT /tenants/{tenant_id}/databases/{database_id}
+DELETE /tenants/{tenant_id}/databases/{database_id}
+POST /tenants/{tenant_id}/delete
+GET /tenants/{tenant_id}/issue_token
```
ololobus, line 276: Remove, outdated

docs/rfcs/017-console-split.md
```diff
@@ -0,0 +1,420 @@
...
+All of the above cases can happen without using the console, just by accessing compute through the proxy.
+
+To solve this, we can have a log of events occurring in the storage (event logs). That is very similar to operations table we have right now, the only difference is that events are immutable and we cannot change them after saving to the database. For example, we might want to have events for the following activities:
+
+- We finished processing some HTTP API query, such as resetting the password
+- We changed some state, such as started or stopped a compute
+- Operation is created
+- Operation is started for the first time
+- Operation is failed for the first time
+- Operation is finished
```
ololobus, line 309: This events section is the most questionable to me. Here you mention that 'operation failed' is event. But bellow you propose that we may have some sort of PITR for events, i.e. replay events log to get into some state of the control-plane. Yet, 'operation failed' event isn't needed to be replayed, the actual operation is what we replay, isn't it?

I'd propose that we continue iterating on the concept of operations, but use events as an information stream, similar to combination of the websockets and segment we use right now. Because, for example, we will need a life stream of the updates from the control-planes to the console and finally UI, and that's where events concept fits well, IMO

ololobus (commented):
Generally, strong +1 from me! I put a bunch of comments bellow, though

docs/rfcs/017-console-split.md
```diff
@@ -0,0 +1,420 @@
...
+## Motivation
+
+These are the two most important problems we want to solve:
+
+- Publish open-source implementation of all our cloud/storage features
+- Make a unified control-plane that is used in all cloud (serverless) and local (tests) setups
+
+Right now we have some closed-source code in the cloud repo. That code contains implementation for running Neon computes in k8s and without that code it’s impossible to automatically scale PostgreSQL computes. That means that we don’t have an open-source serverless PostgreSQL at the moment.
+
+After splitting and open-sourcing control-plane service we will have source code and Docker images for all storage services. That control-plane service should have HTTP API for creating and managing tenants (including all our storage features), while proxy will listen for incoming connections and create computes on-demand.
```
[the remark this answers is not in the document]
ololobus: Ah, I got you. Yes, we do `start_compute` operation if it was idle, but all the compute info is already inside db, so we just spawn a k8s deployment for it

docs/rfcs/017-console-split.md
```diff
@@ -0,0 +1,420 @@
...
+POST /tenants/{tenant_id}/roles
+GET /tenants/{tenant_id}/roles/{role_name}
+DELETE /tenants/{tenant_id}/roles/{role_name}
+POST /tenants/{tenant_id}/roles/{role_name}/reset_password
+POST /tenants/{tenant_id}/start
+POST /tenants/{tenant_id}/stop
+POST /psql_session/{psql_session_id}
+```
+
+One of the options here is to use gRPC instead of the HTTP, which has some useful features, but there are some strong points towards using plain HTTP:
```
SomeoneToIgnore, line 290: We also use gRPC implicitly already due to etcd depencency

agalitsyn: Let's keep using HTTP :) gRPC is heavy stack with many caveats

docs/rfcs/017-console-split.md
```diff
@@ -0,0 +1,420 @@
...
+We should support starting several instances of the new control-plane service at the same time.
+
+At the same time, it should be possible to use only single instance of control-plane, which can be useful for local tests.
+
+## Security implications
+
+New control-plane service is an internal service, so no external requests can reach it. But at the same time, it contains API to do absolutely anything with any of the tenants. That means that bad internal actor can potentially read and write all of the tenants. To make this safer, we can have one of these:
+
+- Simple option is to protect all requests with a single private key, so that no one can make requests without having that one key.
+- Another option is to have a separate token for every tenant and store these tokens in another secure place. This way it’s harder to access all tenants at once, because they have the different tokens.
```
agalitsyn, line 347: We can do both, if we will protect these handlers with JWT, and sign them using using a secret (with the HMAC algorithm). All context, user_id, tenant_id for example, can be passed in jwt token.

docs/rfcs/017-console-split.md
```diff
@@ -0,0 +1,420 @@
...
+    - API Keys
+    - App Controller, serving UI HTML
+    - Auth Controller
+    - Projects
+    - User Metrics
+    - Invites
+    - Users
+    - Passwordless login
+- Other things
+    - segment analytics integration
```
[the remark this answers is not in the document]
petuhovskiy: In my understanding, segment is also a state, which is changed by calling segment API. If there is a misunderstanding in the "state change" terminology, then we can denote an event as everything that happens in control-plane and is needed to be pushed to the console. So "proxy connection auth" can also be an event.

I don't know much about segment, but I suppose it accepts all possible analytics events with a `user_id`, which is unknown in control-plane. Also from my experience, analytics is changed more frequently than other code, and it can be inconvenient to patch control-plane every time we need to change something in analytics.

An ideal schema (from my perspective) would be to have a well-defined list of all events in control-plane, that console can subscribe to and listen for new events in real-time. It gives a clear separation of concerns, where control-plane don't need to do anything beyond operating storage (so no need to know segment credentials, no need to know about users, etc.), and it allows to do anything with the storage just by emitting events and implementing HTTP API. So, once an event has arrived at the console, it can be enriched with user_id and other info, and then sent to the segment.

docs/rfcs/017-console-split.md
```diff
@@ -0,0 +1,420 @@
...
+    - sentry integration
```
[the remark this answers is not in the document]
petuhovskiy: Haven't used sentry, but if it will help us in debugging and reporting issues, then we can add it. Right now I'm not sure if sentry has helped us to catch any bug.

stepashka (approved):

docs/rfcs/017-console-split.md
```diff
@@ -0,0 +1,420 @@
...
+## Security implications
+
+New control-plane service is an internal service, so no external requests can reach it. But at the same time, it contains API to do absolutely anything with any of the tenants. That means that bad internal actor can potentially read and write all of the tenants. To make this safer, we can have one of these:
+
+- Simple option is to protect all requests with a single private key, so that no one can make requests without having that one key.
+- Another option is to have a separate token for every tenant and store these tokens in another secure place. This way it’s harder to access all tenants at once, because they have the different tokens.
+
+## Alternative implementation
+
+There was an idea to create a k8s operator for managing storage services and computes, but author of this RFC is not really familiar with it.
```
lassizci, line 351: Operator pattern could potentially be a good fit here. In the most simplest terms it means having 1) [custom resource definition (CRD)](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/), which can extend kubernetes api with own object definitions, i.e we could have Compute instead of Pod and 2) application that runs in kubernetes and acts on changes to those specified objects.

So in this case the database could be replaced with CRD (as kubernetes already has the same database for any kind of kubernetes objects such as pods, deployments and so on)

There are frameworks to reduce the boilerplate (i.e code that  watches for changes to the objects), such as https://operatorframework.io/what/.

Additionally, it would be possible to leverage kubernetes RBAC and avoid the confused deputy problem discussed in the security implications: We could have the tenants in their separate namespace (as opposed to default), create credentials that would only allow operations in that specific namespace and then use those credentials. We could have separate credentials for creating those namespaces and limit the operations only to necessary ones.

DavidSpek: If any help is needed in terms of operator development or Kubernetes deployments I’d be willing to help wherever needed. I’ve had an eye on NeonDB for a while now, and would love to onboard as a potential alternative to the Zalando Postgres operator on our Marketplace https://www.plural.sh/. A lot of applications on our marketplace rely on Postgres, and when trying to use YugabyteDB as an alternative the first few apps we tested were not compatible. So we’d be very interested to see if NeonDB can fit this use case.

LizardWizzard:
As discussed moving to draft.

kelvich (approved):

PR #1998, opened 2022-06-27
replace extra urlencode dependency with already present url library

LizardWizzard:


compute_tools/src/compute.rs
```diff
@@ -268,21 +268,25 @@ impl ComputeNode {
...
         // but we can create a new one and grant it all privileges.
-        let mut client = match Client::connect(&self.connstr, NoTls) {
+        let mut client = match Client::connect(self.connstr.as_str(), NoTls) {
             Err(e) => {
                 info!(
                     "cannot connect to postgres: {}, retrying with `zenith_admin` username",
                     e
                 );
-                let zenith_admin_connstr = self.connstr.replacen("cloud_admin", "zenith_admin", 1);
+                let mut new_connstr = self.connstr.clone();
```
ololobus, line 277: This is actually not new connstr, but old one :) I'd really keep the long old name `zenith_admin_connstr` as it makes it easier to grep and delete later

compute_tools/src/spec.rs
```diff
@@ -231,9 +230,13 @@ pub fn handle_role_deletions(node: &ComputeNode, client: &mut Client) -> Result<
...
         if db.owner != *role_name {
-            let db_name_encoded = format!("/{}", encode(&db.name));
-            let db_connstr = node.connstr.replacen("/postgres", &db_name_encoded, 1);
-            let mut client = Client::connect(&db_connstr, NoTls)?;
+            let mut connstr = node.connstr.clone();
+            connstr
+                .path_segments_mut()
+                .map_err(|_| anyhow::anyhow!("invalid connstr: url is not a base"))?
+                .pop()
+                .push(&db.name);
```
ololobus, line 238: Maybe just use `set_path` https://docs.rs/url/latest/url/struct.Url.html#method.set_path ?

You anyway assumes that dbname is at the end, but `set_path` at least will preserve `?` parameters if we ever use them

compute_tools/src/compute.rs
```diff
@@ -35,7 +35,7 @@ use crate::spec::*;
 /// Compute node info shared across several `compute_ctl` threads.
 pub struct ComputeNode {
     pub start_time: DateTime<Utc>,
-    pub connstr: String,
+    pub connstr: url::Url,
```
ololobus, line 38: I'd make a comment that it automatically escapes all needed url parts (I hope) with https://github.com/servo/rust-url/blob/master/percent_encoding/src/lib.rs

And when you call `.as_str()` it's already escaped

compute_tools/src/spec.rs
LizardWizzard: The current code will preserve parameters as well. See the example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8d252b4a9a1508e863e4e6fbc8cab2f5 

But I agree, set_path better describes the intention

ololobus: Yeah, really, `[path_segments_mut](https://docs.rs/url/latest/url/struct.Url.html#method.path_segments_mut)` only manipulates with path segments. For some reason I thought that it will split the whole url into segments, sorry, then doesn't matter much

ololobus (approved):
Hm, e2e ci trigger moved to the actions already, but it fails to build. I'd prefer to see e2e passed before merging

LizardWizzard:
For some reason it cannot find `protoc` binary during build of etcd client library

LizardWizzard:
And disabling cachepot seems to fix the problem: https://github.com/neondatabase/neon/runs/7093592215?check_suite_focus=true 

I see that we have two caches now, github actions cache, and cachepot. And something is clearly wrong with caching or our self hosted runners setup. In general I think it would be good to have only one cahing mechanism. Cachepot is used to be able to feed the build cache to docker builds that happen in console e2e tests. If we can build binaries outside of docker and produce docker images by simply copying binaries inside we can avoid cachepot entirely. 
 
Any objections to disable cachepot for now and figure out the path forward separately? @SomeoneToIgnore

SomeoneToIgnore:
Yes, let's use it in Docker images only then, since there's no other way around currently.

SomeoneToIgnore:
Just to finish this thread, we've left cachepot in the builds, purged its cache that contained older rustc build results among other and set its cache retention to 30 days instead of 183.

There's also an environment variable to allow [overriding](https://github.com/paritytech/cachepot#overwriting-the-cache) the cache during build, so that's worth trying next time the situation appears.

SomeoneToIgnore (approved):

PR #1999, opened 2022-06-27
Fix wal reconnects

SomeoneToIgnore:
Closes https://github.com/neondatabase/neon/issues/1985

* make `TaskHandle` to detect when its inner tokio task handle finishes
* fix the connection attempts increment logic
* lower priority for SK nodes with numerous unsuccessful connection attempts 

Does not address the high-level etcd subscription tasks' premature finish: now, there's nothing to check if such task had panicked and restart it. Either some channels (that detect that the other pard had been dropped) or `FuturesUnordered` that contain all handles should be introduced, but both of them require some extra code for case of cancelling all tasks and waiting on them to finish. 
In order to fix the CI issue first, a small PR with the fix is created instead.

### DIFF
--- a/pageserver/src/walreceiver.rs
+++ b/pageserver/src/walreceiver.rs
@@ -178,7 +178,7 @@ async fn shutdown_all_wal_connections(
 /// That may lead to certain events not being observed by the listener.
 #[derive(Debug)]
 struct TaskHandle<E> {
-    handle: JoinHandle<()>,
+    handle: JoinHandle<Result<(), String>>,
     events_receiver: watch::Receiver<TaskEvent<E>>,
     cancellation: watch::Sender<()>,
 }
@@ -205,8 +205,8 @@ impl<E: Clone> TaskHandle<E> {
 
         let sender = Arc::clone(&events_sender);
         let handle = tokio::task::spawn(async move {
-            let task_result = task(sender, cancellation_receiver).await;
-            events_sender.send(TaskEvent::End(task_result)).ok();
+            events_sender.send(TaskEvent::Started).ok();
+            task(sender, cancellation_receiver).await
         });
 
         TaskHandle {
@@ -216,11 +216,34 @@ impl<E: Clone> TaskHandle<E> {
         }
     }
 
+    async fn next_task_event(&mut self) -> TaskEvent<E> {
+        select! {
+            next_task_event = self.events_receiver.changed() => match next_task_event {
+                Ok(()) => self.events_receiver.borrow().clone(),
+                Err(_task_channel_part_dropped) => join_on_handle(&mut self.handle).await,
+            },
+            task_completion_result = join_on_handle(&mut self.handle) => task_completion_result,
+        }
+    }
+
     /// Aborts current task, waiting for it to finish.
     async fn shutdown(self) {
         self.cancellation.send(()).ok();
         if let Err(e) = self.handle.await {
             error!("Task failed to shut down: {e}")
+        }
+    }
+}
+
+async fn join_on_handle<E>(handle: &mut JoinHandle<Result<(), String>>) -> TaskEvent<E> {
+    match handle.await {
+        Ok(task_result) => TaskEvent::End(task_result),
+        Err(e) => {
+            if e.is_cancelled() {
+                TaskEvent::End(Ok(()))
+            } else {
+                TaskEvent::End(Err(format!("WAL receiver task panicked: {e}")))
+            }
         }
     }
 }
--- a/pageserver/src/walreceiver/connection_manager.rs
+++ b/pageserver/src/walreceiver/connection_manager.rs
@@ -104,49 +104,29 @@ async fn connection_manager_loop_step(
 
             Some(wal_connection_update) = async {
                 match walreceiver_state.wal_connection.as_mut() {
-                    Some(wal_connection) => {
-                        let receiver = &mut wal_connection.connection_task.events_receiver;
-                        Some(match receiver.changed().await {
-                            Ok(()) => receiver.borrow().clone(),
-                            Err(_cancellation_error) => TaskEvent::End(Ok(())),
-                        })
-                    }
+                    Some(wal_connection) => Some(wal_connection.connection_task.next_task_event().await),
                     None => None,
                 }
             } => {
-                let (connection_update, reset_connection_attempts) = match &wal_connection_update {
-                    TaskEvent::Started => (Some(Utc::now().naive_utc()), true),
-                    TaskEvent::NewEvent(replication_feedback) => (Some(DateTime::<Local>::from(replication_feedback.ps_replytime).naive_utc()), true),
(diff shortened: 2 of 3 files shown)


SomeoneToIgnore:
I have rerun the GH Actions pipeline a few times and have not noticed the "test_restarts_under_load" failures so far, the current logs seem to be ok too in both CI systems.

During restarts, `test_normal_work[3-1]` had flacked:
https://github.com/neondatabase/neon/runs/7077634173?check_suite_focus=true#step:3:475
that seems unrelated, since caused by inability to remove the timeline in time.
Seem to fail on GH Actions only too, but might be already fixed by https://github.com/neondatabase/neon/pull/1936

pageserver/src/walreceiver/connection_manager.rs
```diff
@@ -476,6 +478,38 @@ impl WalreceiverState {
         None
     }
 
+    fn select_connection_candidate(
```
arssher, line 481: - This function deserves a comment.
- I initially doubted that considering only smallest_attempts_count is good. It bears a risk of choosing lagged node. The problem we are trying to solve is avoiding reconnecting to node which just gone dead. Another possible way would be immediately after getting connection failure to delete `wal_stream_candidates` entry; dead node won't send updates anymore. It has its own faults though; an obvious race -- we might still get update which was just slow to arrive -- and it doesn't consider split situations where etcd is visible but direct connection fails. The middle between these options can be augment this PR with forgetting `wal_connection_attempts` after period of time with no failures. But I'm ok merging as is, it is good enough currently.
- Nitpicking, but perhaps in more functional style (fold wal_connection_attempts to find min, then filter candidates) it would be simpler.
- s/assempts/attempts

arssher (changes requested):

pageserver/src/walreceiver.rs
```diff
@@ -225,6 +236,19 @@ impl<E: Clone> TaskHandle<E> {
...
 }
 
+async fn join_on_task_handle<E>(handle: &mut JoinHandle<()>) -> TaskEvent<E> {
+    match handle.await {
+        Ok(()) => TaskEvent::End(Ok(())),
+        Err(e) => {
+            if e.is_panic() {
+                TaskEvent::End(Err(format!("WAL receiver task panicked: {e}")))
+            } else {
+                TaskEvent::Cancelled
```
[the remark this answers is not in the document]
SomeoneToIgnore: It was a bit of a confusing code indeed, where the actual task processed here could not error:
```
tokio::task::spawn(async move {
    let task_result = task(sender, cancellation_receiver).await;
    events_sender.send(TaskEvent::End(task_result)).ok();
});
```

I've untangled this and now the result is passed on in the `TaskEvent::End`, but either way it was not ignored.
Removed the new enum variant.

pageserver/src/walreceiver/connection_manager.rs
```diff
@@ -104,49 +104,54 @@ async fn connection_manager_loop_step(
...
+                        reset_connection = true;
+                        None
                     },
                 };
 
-                if let Some(connection_update) = connection_update {
-                    match &mut walreceiver_state.wal_connection {
-                        Some(wal_connection) => {
-                            wal_connection.latest_connection_update = connection_update;
+                match &mut walreceiver_state.wal_connection {
```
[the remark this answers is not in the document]
SomeoneToIgnore: Had tried to reuse the best from [the PR](https://github.com/neondatabase/neon/pull/1996/files), hopefully now it's better.

pageserver/src/walreceiver/connection_manager.rs
SomeoneToIgnore: > I initially doubted that considering only smallest_attempts_count is good.
I've decided to leave the candidates selection logic as is, without removals or expirations, just since it's simpler.
I plan to revisit the global task restarts here at some point, can continue with this one too unless any problems pop up before.

Otherwise reworked.

pageserver/src/walreceiver.rs
```diff
@@ -205,8 +205,8 @@ impl<E: Clone> TaskHandle<E> {
 
         let sender = Arc::clone(&events_sender);
         let handle = tokio::task::spawn(async move {
-            let task_result = task(sender, cancellation_receiver).await;
-            events_sender.send(TaskEvent::End(task_result)).ok();
+            events_sender.send(TaskEvent::Started).ok();
```
SomeoneToIgnore, line 208: Noticed that there's no `TaskEvent::Started` posted otherwise, the initial channel value is not propagated.
Also, `async move` is needed here, otherwise the closure won't capture the vars for long enough.

So we have to have this logic in one way or another. I can return the `::End` here too and make this task infallible after all these discoveries, if you think it's more appropriate.

pageserver/src/walreceiver/connection_manager.rs
```diff
@@ -476,6 +453,49 @@ impl WalreceiverState {
...
     }
 
+    /// Selects the best possible cancidate out of all data, stored from etcd updates about the safeekeepers.
+    /// Optionally, omits the node given, to support the connections witch logic.
+    ///
+    /// The candidate that is chosen:
+    /// * has smallest possible connection attempts from pageserver to safekeeper node (reset every time the WAL replication feedback is sent)
+    /// * has biggest data Lsn among the ones that are left
+    ///
+    /// We do not evict timeline data received from etcd, so there's a chance to select the node with good information that did not send etcd updates for a long time.
```
arssher, line 463: Well, we do evict in `cleanup_old_candidates`. Another problem is possible: choosing lagging node which has 0 attempts counter while there might be more advanced one with non-zero counter (thought it might be already be healthy).

arssher (approved):

hlinnaka (commented):
I must admit I don't understand the underlying problem or how this fixes it; I don't have enough context on this code. But I nevertheless left some comments on the implementation and code comments.

This is kind of urgent to fix the failing CI, so if this fixes the CI, +1 on committing this sooner rather than later..

pageserver/src/walreceiver/connection_manager.rs
```diff
@@ -476,6 +453,52 @@ impl WalreceiverState {
...
+    /// Optionally, omits the node given, to support the connections witch logic.
+    ///
+    /// The candidate that is chosen:
+    /// * has smallest possible connection attempts from pageserver to safekeeper node (reset every time the WAL replication feedback is sent)
+    /// * has biggest data Lsn among the ones that are left
+    ///
+    /// We evict timeline data received from etcd based on time passed since it was registered,
+    /// minifying a chance to select the node with advanced enough information that did not send etcd updates for a long time is.
+    /// Yet, we might select more lagging nodes due to this heuristics, since not altering the errors before a successful reconnection,
+    /// that might not happen before all other nodes accumulate big enough connection attempts counts.
```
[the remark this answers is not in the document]
SomeoneToIgnore: Reworded.

PR #2007, opened 2022-06-29
Disable cachepot for GH Actions builds

SomeoneToIgnore:
~5 failures happend recently in different places, in Docker builds and regular GH Actions build after we've started transitioning from the CI.

Recent one:
https://github.com/neondatabase/neon/runs/7112481569?check_suite_focus=true#step:6:200

We have to cache `target/` directory already for coverage report, so I've removed `hash(Cargo.lock)` for its key to use this as the build cache.


LizardWizzard (approved):
LGTM

PR #2008, opened 2022-06-29
Add tests for different Postgres client libraries

bayandin:
We want to make sure Neon is compatible with the vast majority of popular Postgres client libraries.

Currently, we want to find out which libraries do or do not support [SNI](https://www.cloudflare.com/en-gb/learning/ssl/what-is-sni/) (since libpq supports `SNI` we are interested in non-libpq implementations. See  https://wiki.postgresql.org/wiki/List_of_drivers)

This PR adds several examples for different languages/libraries to `test_runner/pg-clients/<language>/<library>` and `test_runner/pg_clients/test_pg_clients.py` — to run them as a test.

All examples use `NEON_HOST`/`NEON_DATABASE`/`NEON_USER`/`NEON_PASSWORD` environment variables and execute simple `SELECT 1`.

Current status:

| Language | Library | Status | Comment |
|----------|--------|-------|-----------|
| C# | [npgsql](https://github.com/npgsql/npgsql) | ✅ | |
| Java | [JDBC](https://github.com/pgjdbc/pgjdbc) | ✅  | |
| Python | [asyncpg](https://github.com/MagicStack/asyncpg) | ✅  | |
| Python | [pg8000](https://github.com/tlocke/pg8000) | ❌ | `scramp.core.ScramException: The channel_binding parameter must be None if channel binding is not required.`|
| Swift    | [PostgresClientKit](https://github.com/codewinsdotcom/PostgresClientKit) | ❌ | `Project name is not specified. EITHER please upgrade the postgres client library (libpq) for SNI support OR pass the project name as a parameter: '&options=project%3D<project-name>'` |
| TypeScript | [postgresql-client](https://github.com/panates/postgresql-client) | ✅ | |


test_runner/pg_clients/test_pg_clients.py
```diff
@@ -0,0 +1,51 @@
...
+@pytest.mark.remote_cluster
+@pytest.mark.parametrize(
+    "client",
+    [
+        "csharp/npgsql",
+        "java/jdbc",
+        "python/asyncpg",
+        pytest.param("python/pg8000", marks=pytest.mark.xfail(reason="Doesn't support SNI")),
+        pytest.param("swift/PostgresClientKit",
+                     marks=pytest.mark.xfail(reason="Doesn't support SNI")),
```
bojanserafimov, line 21: Does it work with connection options `'&options=project%3D<project-name>'`, as suggested by our error message?

test_runner/pg_clients/test_pg_clients.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+@pytest.mark.remote_cluster
+@pytest.mark.parametrize(
+    "client",
+    [
+        "csharp/npgsql",
+        "java/jdbc",
+        "python/asyncpg",
+        pytest.param("python/pg8000", marks=pytest.mark.xfail(reason="Doesn't support SNI")),
```
bojanserafimov, line 19: I don't understand the error we get `scramp.core.ScramException: The channel_binding parameter must be None if channel binding is not required.` but it seems like it's pointing at a bug in our scram server-side protocol maybe? cc @funbringer

bojanserafimov (commented):
Great start! Thanks for the patch. Do the tests currently pass?

What do you think about eventually syncing with the [snippets](https://github.com/neondatabase/cloud/tree/3027f6983911321df37f256cf804f5e63eb461f5/web/modules/console/components/ProjectWidgets/ProjectConnectionWidget/snippets) we display in the console? We'll need to simplify some of them and add a few comments.

.github/workflows/pg_clients.yml
```diff
@@ -0,0 +1,64 @@
+name: Test Postgres client libraries
+
+on:
```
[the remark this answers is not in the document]
bayandin: We'll probably do it after [the Console split](https://github.com/neondatabase/neon/pull/1997).

test_runner/pg_clients/test_pg_clients.py
bayandin: As far as I can see the library doesn't support connection string or any kind of additional options [1]; but there's no validation for these parameters, so we could try to inject it into the StartupRequest

- [1] https://github.com/codewinsdotcom/PostgresClientKit/blob/v1.4.3/Sources/PostgresClientKit/ConnectionConfiguration.swift
- [2] https://github.com/codewinsdotcom/PostgresClientKit/blob/v1.4.3/Sources/PostgresClientKit/Request/StartupRequest.swift#L43-L66

test_runner/pg_clients/test_pg_clients.py
bayandin: Here actually I'm not sure if it's a bug in our code or in `pg8000`.
If I comment out this block:
https://github.com/tlocke/pg8000/blob/1.29.1/pg8000/core.py#L264-L267 it starts to work for me (creating custom `ssl_context` with `request_ssl=False` doesn't really work)

bayandin:
> Do the tests currently pass?

Yep, here's an example: https://github.com/neondatabase/neon/actions/runs/2592269753
And if it fails (with removed `xfail`), then it looks like this: https://github.com/neondatabase/neon/runs/7138147182

.github/workflows/pg_clients.yml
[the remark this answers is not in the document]
bojanserafimov: A local (even if imperfect) test makes it easier to iterate, and requires less bootstrapping. I'd try to investigate the pg8000 issue if I didn't need to get a secret connection string to run this code.

Doesn't need to happen in this PR, bu I just want to point out that there's a lot of value in making this work locally, and the Console split might take a while.

test_runner/pg_clients/test_pg_clients.py
bojanserafimov: Yeah looks like the library assumes that if ssl is requested then channel binding is required, which is not the case for our proxy. Let's leave a comment to clarify.

test_runner/pg_clients/test_pg_clients.py
bojanserafimov: injection is sometimes the only solution, but in this case it seems especially brittle, snice it wouldn't be surprising if they added validation. Let's leave a comment with this context, to explain that they don't have SNI **and** they don't support parameters

bojanserafimov (approved):
Looks good to merge (see my comments).

One more thing: would we get notified of failures if we don't manually check github actions?

test_runner/pg_clients/test_pg_clients.py
funbringer: I'm busy with some other stuff, but with my superficial understanding this should be fixed on the proxy's side. I'll take a closer look later.

bayandin:
> One more thing: would we get notified of failures if we don't manually check github actions?

I've added slack notifications for failed job

PR #2010, opened 2022-06-29
Reduce memory allocations for page server

aome510:
## Overview

This patch reduces the number of memory allocations when running the page server under a heavy write workload. This mostly helps improve the speed of WAL record ingestion. 

## Changes
- modified `DatadirModification` to allow reuse the struct's allocated memory after each modification
- modified `decode_wal_record` to allow passing a `DecodedWALRecord` reference. This helps reuse the struct in each `decode_wal_record` call
- added a reusable buffer for serializing object inside the `InMemoryLayer::put_value` function
- added a performance test simulating a heavy write workload for testing the changes in this patch

### Semi-related changes
- remove redundant serializations when calling `DeltaLayer::put_value` during `InMemoryLayer::write_to_disk` function call [1]
- removed the info span `info_span!("processing record", lsn = %lsn)` during each WAL ingestion [2]

## Notes
- [1]: in `InMemoryLayer::write_to_disk`, a deserialization is called
  ```
  let val = Value::des(&buf)?;
  delta_layer_writer.put_value(key, *lsn, val)?;
  ``` 
  `DeltaLayer::put_value` then creates a serialization based on the previous deserialization
  ```
  let off = self.blob_writer.write_blob(&Value::ser(&val)?)?;
  ```
- [2]: related: https://github.com/neondatabase/neon/issues/733


aome510:
# Performance investigation notes:

- Removing the span improves the write throughput by 1.5 times
- Other changes in this patch improves the write throughput by about 10%

## Flamegraphs

- The flamegraph of this patch when running the new heavy-write workload (search for `malloc` call):

Screen Shot 2022-06-29 at 4 26 25 PM


- Flamegraph of `main` with `info_span!("processing record", lsn = %lsn)` removed (search for `malloc` call):
Screen Shot 2022-06-29 at 4 26 55 PM

- Flamegraph of `main`. Noticed that lots of CPU time are spent on `tracing::span::Span::new`.

Screen Shot 2022-06-29 at 4 27 39 PM

pageserver/src/walreceiver/walreceiver_connection.rs
```diff
@@ -151,14 +155,14 @@ pub async fn handle_walreceiver_connection(
                 waldecoder.feed_bytes(data);
 
                 while let Some((lsn, recdata)) = waldecoder.poll_decode()? {
-                    let _enter = info_span!("processing record", lsn = %lsn).entered();
+                    // let _enter = info_span!("processing record", lsn = %lsn).entered();
```
aome510, line 158: I see this as free performance improvement. Can revert the change in this PR if needed

pageserver/src/walreceiver/walreceiver_connection.rs
```diff
@@ -150,19 +151,23 @@ pub async fn handle_walreceiver_connection(
 
                 waldecoder.feed_bytes(data);
 
-                while let Some((lsn, recdata)) = waldecoder.poll_decode()? {
-                    let _enter = info_span!("processing record", lsn = %lsn).entered();
+                {
+                    let mut decoded = DecodedWALRecord::default();
+                    let mut modification = timeline.begin_modification();
+                    while let Some((lsn, recdata)) = waldecoder.poll_decode()? {
+                        // let _enter = info_span!("processing record", lsn = %lsn).entered();
```
aome510, line 158: I see this as a free performance improvement. Can revert the change in this PR if needed

hlinnaka: The span provides very useful context to any errors or warnings that happen while processing the record. It would be a pity to just outright remove it. However, maybe we can compensate by adding the information another way. I think we could do something like this in the ingest_record call below:
```
    walingest.ingest_record(recdata, lsn, &mut modification, &mut decoded)
  .with_context("could not ingest record at {lsn}")
```

aome510: Yeah, that would be better.

pageserver/src/layered_repository/inmemory_layer.rs
```diff
@@ -41,6 +41,10 @@ pub struct InMemoryLayer {
     ///
     start_lsn: Lsn,
 
+    /// A buffer for serializing object during [`InMemoryLayer::put_value`].
+    /// This buffer is reused for each serialization to avoid additional malloc calls.
+    ser_buffer: Mutex<Vec<u8>>,
```
hlinnaka, line 46: This seems weird to me. Avoiding a malloc() is nice, but locking a Mutex isn't free either. Let's move this to InMemoryLayerInner, so that it doesn't need a separate Mutex to protect it. Or maybe even better, use `thread_local`. If it's thread-local, then we could potentially reuse the same scratch buffer in many different places.

pageserver/src/layered_repository/inmemory_layer.rs
```diff
@@ -342,8 +352,8 @@ impl InMemoryLayer {
             // Write all page versions
             for (lsn, pos) in vec_map.as_slice() {
                 cursor.read_blob_into_buf(*pos, &mut buf)?;
-                let val = Value::des(&buf)?;
-                delta_layer_writer.put_value(key, *lsn, val)?;
+                let will_init = Value::des(&buf)?.will_init();
```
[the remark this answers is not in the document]
aome510: Yeah same though =) and the `will_init` flag is only available in one variant `ZenithWalRecord::Postgres`. I was thinking using some tricks like encoding the `will_init` value inside the `pos` variable so `pos*2` denotes `will_init=false` and `pos*2+1` denotes `will_init=true`. Still, I don't think that's worth the effort and changes.

pageserver/src/layered_repository/inmemory_layer.rs
aome510: I used to store the buffer inside InMemoryLayerInner, but when calling `inner.file.write_blob(&buf)?`, it'll return error like: cannot borrow `buf` as immutable as `inner.file` is also borrowed as mutable.

`thread_local` looks good. Will try that out.

aome510:
So I ran [this patch](https://github.com/neondatabase/neon/tree/thang/exp-06-29) vs [`main`](https://github.com/neondatabase/neon/tree/thang/exp-06-28) (both with the span `info_span!("processing record", lsn = %lsn)` removed).

Results when running the new test on CircleCI machines:

```
patch:
- test_compare_pg_stats_wo_with_heavy_write[neon-10-10].pg_stat_database.tup_inserted:
    + runs: [2303000, 2164000, 2174000, 2266000, 2059000, 1812000]
    + avg: 2129666.6666666665 
- test_compare_pg_stats_wo_with_heavy_write[neon-10-1].pg_stat_database.tup_inserted:
    + runs: [2809000, 2430000, 2630000, 2748000, 2142000, 2272000]
    + avg: 2505166.6666666665

main:
- test_compare_pg_stats_wo_with_heavy_write[neon-10-10].pg_stat_database.tup_inserted:
    + runs: [1874000, 2187000, 2065000, 1968000, 2145000, 1824000]
    + avg: 2010500.0 
- test_compare_pg_stats_wo_with_heavy_write[neon-10-1].pg_stat_database.tup_inserted:
    + runs: [2288000, 2627000, 2379000, 2454000, 2635000, 2498000]
    + avg: 2480166.6666666665 
```

Look like the performance improvement is not that significant. Not sure if the 10% percent improvement I got when running locally is because `malloc` on MacOS is expensive compared to other OSes. Will need others to confirm this by running the test in their local environment.

Also note that the performance when running CIs is quite random as the compute resources are not fixed across different runs/tests, which makes measuring the net improvement more difficult.

aome510:
The failed test `test_pageserver_recovery` in this PR should be fixed by https://github.com/neondatabase/neon/pull/2060.  It's better to wait for that PR to get in.

aome510:
@hlinnaka can you take a look into this PR again?

pageserver/src/layered_repository/inmemory_layer.rs
```diff
@@ -270,10 +277,17 @@ impl InMemoryLayer {
...
-
         inner.assert_writeable();
 
-        let off = inner.file.write_blob(&Value::ser(val)?)?;
+        let off = {
+            SER_BUFFER.with(|x| -> Result<_> {
+                let mut buf = x.borrow_mut();
+                val.ser_into(&mut (*buf))?;
+                let off = inner.file.write_blob(&buf)?;
+                buf.clear();
```
hlinnaka, line 287: Does `ser_into` overwrite existing data in the buffer, or append? If the `write_blob` call fails, `buf.clear()` is not called, so on the next call there will already be some data in the buffer. Best to call `buf.clear()` before `ser_into`.

hlinnaka (approved):
The [`buf.clear()` thing](https://github.com/neondatabase/neon/pull/2010/files#r926347112) needs to be checked/fixed. Other than that, LGTM

PR #2017, opened 2022-07-01
Prefix WAL files in s3 with environment name.

arssher:
It wasn't merged to prod yet, so safe to enable.


SergeyMelnikov (approved):

cicdteam:
Why we need it if we are using different S3 buckets for different environments ?

arssher:
Well, from bucket names I wasn't sure this is true, especially in the future. But otherwise it is indeed not needed; but neither harmful, I think.

PR #2024, opened 2022-07-04
Fix 1.62 Clippy errors

SomeoneToIgnore:


pageserver/src/profiling.rs
```diff
@@ -81,6 +81,12 @@ mod profiling_impl {
 
     pub struct DummyProfilerGuard;
 
+    impl Drop for DummyProfilerGuard {
+        fn drop(&mut self) {
+            // do nothing, this exists to calm Clippy down
```
LizardWizzard, line 86: I wonder what lint triggered this?

LizardWizzard (approved):

pageserver/src/profiling.rs
SomeoneToIgnore:
```
error: call to `std::mem::drop` with a value that does not implement `Drop`. Dropping such a type only extends its contained lifetimes
   --> pageserver/src/page_service.rs:537:13
    |
537 |             drop(profiling_guard);
    |             ^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `-D clippy::drop-non-drop` implied by `-D warnings`
note: argument has type `std::option::Option<profiling::profiling_impl::DummyProfilerGuard>`
   --> pageserver/src/page_service.rs:537:18
    |
537 |             drop(profiling_guard);
    |                  ^^^^^^^^^^^^^^^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#drop_non_drop
```

PR #2025, opened 2022-07-04
Return download stream directly from the remote storage API

SomeoneToIgnore:
Follow-up of https://github.com/neondatabase/neon/pull/1932#discussion_r897833419


safekeeper/src/wal_storage.rs
```diff
@@ -604,8 +604,7 @@ impl WalReader {
 
         // Try to open remote file, if remote reads are enabled
         if self.enable_remote_read {
-            let (reader, _) = read_object(wal_file_path, xlogoff as u64).await;
-            return Ok(Box::pin(reader));
+            return read_object(wal_file_path, xlogoff as u64).await;
```
LizardWizzard, line 607: Will it read straight from s3 without saving it locally? So if we have a replica that constantly connects and craches after reading a few bytes it will every time connect to s3? Also if we have more than one replica, will they share the file or every replica will get its own download stream from s3?

LizardWizzard (commented):
Looks good to me. Minor comments:

* NIT: IMO fully qualified path to anyhow shortcuts looks more consistent. There were multiple discussions to have `anyhow::Result` instead of importing it and using without the path to anyhow. So I think it might be better for consistency to use things like bail/ensure with full path as well.
* Initially I imagined that NotFound will be an error, with the current approach every consumer needs to match on the result and throw an error on NotFound variant. So the option might be to introduce associated type Error to the RemoteStorage trait so the error itself can be better typed (using thiserror for example, or it may be possible to use just RusotoError for s3 impl) and include NotFound variant. There may be several options, just suggesting ideas. 
* Another idea. dyn AsyncRead in return position for download can be avoided using associated type. We anyway match for the remote storage impl. It may be inconvenient for callers that abstract away the storage impl, so may not worth it :)

safekeeper/src/wal_storage.rs
LizardWizzard: Not introduced with this PR though, so probably question to @petuhovskiy or @arssher

SomeoneToIgnore:
Thank you for the proposals, I've tried to avoid the associated types since they bloat the code more than needed now maybe, and implemented all but the 3rd point.

I've also tried using `thiserror` but the remote_storage lib has way more classes of errors I could cover with the `download` case, so I decided to use basic enum with impls for simplicity.

pageserver/src/storage_sync/download.rs
```diff
@@ -197,7 +211,7 @@ where
                 drop(destination_file);
 
                 fail::fail_point!("remote-storage-download-pre-rename", |_| {
-                    anyhow::bail!("remote-storage-download-pre-rename failpoint triggered")
+                    bail!("remote-storage-download-pre-rename failpoint triggered")
```
LizardWizzard, line 214: nit: if we decided to fully qualify anyhow imports should be `anyhow::bail`

LizardWizzard (approved):
Looks good to me, thanks!

safekeeper/src/wal_storage.rs
arssher: Yes, it reads directly from s3 and doesn't save to disk which allows to not think about garbage collecting files. This proxying to s3 is sort of emergency code path; replica shouldn't regularly crash after receiving some bytes. So while it can be improved, currently it's ok in IMV.

PR #2032, opened 2022-07-05
Fix panic while waiting for Postgres readiness in the compute_ctl (#2021)

### LINKED ISSUE #2021: Panic in the compute_ctl while reading the pid file
Got this in Docker once:
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', compute_tools/src/pg_helpers.rs:252:18
stack backtrace:
   0: rust_begin_unwind
             at ./rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/std/src/panicking.rs:498:5
   1: core::panicking::panic_fmt
             at ./rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/panicking.rs:107:14
   2: core::panicking::panic
             at ./rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/panicking.rs:48:5
   3: core::option::Option<T>::unwrap
             at ./rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/option.rs:746:21
   4: compute_tools::pg_helpers::wait_for_postgres
             at ./home/runner/compute_tools/src/pg_helpers.rs:249:26
   5: compute_tools::compute::ComputeNode::run
             at ./home/runner/compute_tools/src/compute.rs:264:9
   6: compute_tools::compute::ComputeNode::prepare_and_run
             at ./home/runner/compute_tools/src/compute.rs:348:9
   7: compute_ctl::main
             at ./home/runner/compute_tools/src/bin/compute_ctl.rs:153:11
   8: core::ops::function::FnOnce::call_once
             at ./rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/ops/function.rs:227:5
```

It happened around the first unwrap() here:
https://github.com/neondatabase/neon/blob/65704708fa922b524d3ab75995c39afc5c4f562e/compute_tools/src/pg_helpers.rs#L247-L260

The log looks like this:
```
2022-07-01 17:45:20.013 UTC [main] INFO: updated pg_hba.conf to allow external connections
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', compute_tools/src/pg_helpers.rs:252:18
stack backtrace:
2022-07-01 17:45:20.028 GMT [19] LOG:  starting PostgreSQL 14.2 on x86_64-pc-linux-gnu, compiled by gcc (Debian 8.3.0-6) 8.3.0, 64-bit
```

So we panicked right in the moment when Postgres started.

Maybe we tried to read pid file while it was modified by Postgres? Probably we should handle this unwrap() gracefully.

ololobus:
We were reading Postgres pid file and looking for the 'ready' status,
but it could be empty or we could not read it. So add all the checks.

Resolve #2021

### DIFF
--- a/compute_tools/src/pg_helpers.rs
+++ b/compute_tools/src/pg_helpers.rs
@@ -248,18 +248,20 @@ pub fn wait_for_postgres(pg: &mut Child, port: &str, pgdata: &Path) -> Result<()
             bail!("Postgres exited unexpectedly with code {}", code);
         }
 
-        if pid_path.exists() {
-            let file = BufReader::new(File::open(&pid_path)?);
-            let status = file
-                .lines()
-                .last()
-                .unwrap()
-                .unwrap_or_else(|_| "unknown".to_string());
-            let can_connect = TcpStream::connect_timeout(&addr, timeout).is_ok();
+        // Check that we can open pid file first.
+        if let Ok(file) = File::open(&pid_path) {
+            let file = BufReader::new(file);
+            let last_line = file.lines().last();
 
-            // Now Postgres is ready to accept connections
-            if status.trim() == "ready" && can_connect {
-                break;
+            // Pid file could be there and we could read it, but it could be empty, for example.
+            if let Some(Ok(line)) = last_line {
+                let status = line.trim();
+                let can_connect = TcpStream::connect_timeout(&addr, timeout).is_ok();
+
+                // Now Postgres is ready to accept connections
+                if status == "ready" && can_connect {
+                    break;
+                }
             }
         }
 


compute_tools/src/pg_helpers.rs
```diff
@@ -248,18 +248,20 @@ pub fn wait_for_postgres(pg: &mut Child, port: &str, pgdata: &Path) -> Result<()
...
-            if status.trim() == "ready" && can_connect {
-                break;
+        // Check that we can open pid file first.
+        if let Ok(file) = File::open(&pid_path) {
+            let file = BufReader::new(file);
+            let last_line = file.lines().last();
+
+            // Pid file could be there and we could read it, but it could be empty, for example.
+            if let Some(line) = last_line {
+                let status = line.unwrap_or_else(|_| "unknown".to_string());
```
SomeoneToIgnore, line 258: NIT (not your PR caused this anyway): `to_string` will cause an allocation, and `trim` is actually available on `&str`, so we could do below smth. like
`if line.map(|line| line.trim()) == Some("ready")`.

SomeoneToIgnore (approved):

compute_tools/src/pg_helpers.rs
SomeoneToIgnore: My bad, I've thought it's a pure String in the `Some(line)` code hence all the suggestions.
We can use `as_deref()` on Option/Result to override this:

```rust
// Pid file could be there and we could read it, but it could be empty, for example.
if let Some(line) = last_line {
    let status = line.as_deref().unwrap_or("unknown").trim();
    let can_connect = TcpStream::connect_timeout(&addr, timeout).is_ok();

    // Now Postgres is ready to accept connections
    if status == "ready" && can_connect {
        break;
    }
}
```

Either way, not important, so feel free to ignore if it takes too much time.

SomeoneToIgnore: None, I was slightly suspicious that the semantics changes a bit: now, we throw away file reading errors, while before we substituted it with `"unknown"`, but IMO there's not much of a difference after I've given it a thought.

PR #2036, opened 2022-07-05
Add export/import test

bojanserafimov:
Add test for export/import script.

This is different from the existing separate export and import tests because:
1. It works with two simultaneously running pageservers, so we can test the whole script, inclucing glue code
2. We get to test with and without a workload

### DIFF
--- a/test_runner/batch_others/test_tenant_relocation.py
+++ b/test_runner/batch_others/test_tenant_relocation.py
@@ -10,8 +10,8 @@ from typing import Optional
 import signal
 import pytest
 
-from fixtures.neon_fixtures import PgProtocol, PortDistributor, Postgres, NeonEnvBuilder, Etcd, NeonPageserverHttpClient, assert_local, wait_until, wait_for_last_record_lsn, wait_for_upload, neon_binpath, pg_distrib_dir
-from fixtures.utils import lsn_from_hex
+from fixtures.neon_fixtures import PgProtocol, PortDistributor, Postgres, NeonEnvBuilder, Etcd, NeonPageserverHttpClient, assert_local, wait_until, wait_for_last_record_lsn, wait_for_upload, neon_binpath, pg_distrib_dir, base_dir
+from fixtures.utils import lsn_from_hex, subprocess_capture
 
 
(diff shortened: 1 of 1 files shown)


test_runner/batch_others/test_tenant_relocation.py
```diff
@@ -184,19 +200,45 @@ def test_tenant_relocation(neon_env_builder: NeonEnvBuilder,
...
-        # when load is active these checks can break because lsns are not static
-        # so lets check with some margin
-        assert_abs_margin_ratio(lsn_from_hex(new_timeline_detail['local']['disk_consistent_lsn']),
-                                lsn_from_hex(timeline_detail['local']['disk_consistent_lsn']),
-                                0.03)
+        # Migrate either by attacking from s3 or import/export basebackup
+        if method == "major":
+            cmd = [
+                "python",
+                os.path.join(base_dir, "scripts/export_import_betwen_pageservers.py"),
```
lubennikovaav, line 207:
```suggestion
                os.path.join(base_dir, "scripts/export_import_between_pageservers.py"),
```

test_runner/batch_others/test_tenant_relocation.py
```diff
@@ -184,19 +200,45 @@ def test_tenant_relocation(neon_env_builder: NeonEnvBuilder,
...
+                "--from-pg-port",
+                str(env.pageserver.service_port.pg),
+                "--to-host",
+                "localhost",
+                "--to-http-port",
+                str(new_pageserver_http_port),
+                "--to-pg-port",
+                str(new_pageserver_pg_port),
+                "--psql-path",
+                os.path.join(pg_distrib_dir, "bin", "psql"),
```
lubennikovaav, line 223:
```suggestion
                "--psql-path",
                os.path.join(pg_distrib_dir, "bin", "psql"),
                 "--work-dir",
                os.path.join(test_output_dir),
```

test_runner/batch_others/test_tenant_relocation.py
bojanserafimov: The script itself needs to be renamed too :)

lubennikovaav (approved):

PR #2042, opened 2022-07-06
Fix signal file parsing

bojanserafimov:
This allows us to import empty neon timelines.

### DIFF
--- a/pageserver/src/import_datadir.rs
+++ b/pageserver/src/import_datadir.rs
@@ -516,9 +516,17 @@ pub fn import_file<R: Repository, Reader: Read>(
         // Parse zenith signal file to set correct previous LSN
         let bytes = read_all_bytes(reader)?;
         // zenith.signal format is "PREV LSN: prev_lsn"
-        let zenith_signal = std::str::from_utf8(&bytes)?;
-        let zenith_signal = zenith_signal.split(':').collect::<Vec<_>>();
-        let prev_lsn = zenith_signal[1].trim().parse::<Lsn>()?;
+        // TODO write serialization and deserialization in the same place,
+        //      write tests.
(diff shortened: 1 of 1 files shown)


pageserver/src/import_datadir.rs
```diff
@@ -516,9 +516,19 @@ pub fn import_file<R: Repository, Reader: Read>(
...
+                let split = other.split(':').collect::<Vec<_>>();
+                split[1]
+                    .trim()
+                    .parse::<Lsn>()
+                    .context("can't parse zenith.signal")?
+            }
+        };
 
         let writer = modification.tline.tline.writer();
         writer.finish_write(prev_lsn);
```
lubennikovaav, line 534: Why do we need this code in the first place?
comment to `finish_write` says "Call this after you have finished writing all the WAL up to 'lsn'."

Is it always true, that `zenith.signal` is the last file? I don't think so.

pageserver/src/import_datadir.rs
```diff
@@ -516,9 +516,19 @@ pub fn import_file<R: Repository, Reader: Read>(
...
+            other => {
+                let split = other.split(':').collect::<Vec<_>>();
+                split[1]
+                    .trim()
+                    .parse::<Lsn>()
+                    .context("can't parse zenith.signal")?
+            }
+        };
 
         let writer = modification.tline.tline.writer();
```
lubennikovaav, line 533:
```suggestion
        // zenith.signal is not necessarily the last file, that we handle
        // but it is ok to call `finish_write()`, because final `modification.commit()`
        // will update lsn once more to the final one.
        let writer = modification.tline.tline.writer();
```

lubennikovaav (approved):

PR #2047, opened 2022-07-07
update timeout for proxy deploy

cicdteam:
it should solve https://github.com/neondatabase/cloud/issues/1765


cicdteam:
> Let's try, but why it takes that long?

may be 10 min enough but I've decided to use 15. Its because AWS ALB target (de)registration flow is slow and now pod readiness depends on it (it used for console, but proxy resides in the same namespace in the same EKS cluster for staging). By other words - we need wait more time to ensure pod is in Ready state (helm waiting that as well)

PR #2050, opened 2022-07-07
postgres_ffi/wal_generate: refactor and cleanup

yeputons:
Based on #1990

Previously there was a bunch of functions with the same interface, but it was an informal convention. Now it is a formal trait.

Note that this trait is not "object-safe", i.e. one cannot use it for dynamic dispatching like `&dyn Generator`. So instead of, say, creating an array of all generators, I still have to write manual `match`es in the `wal_generate` binary. I think it's ok as long as no other similar boilerplate places emerge.

Easier to review with `git diff -w`


yeputons:
Maybe rename `wal_generate` to `wal_craft`/`craft_wal` as well, WDYT?

arssher:
Yes, wal_craft sounds better.

yeputons:
@arssher Rebased and ready for review.

PR #2056, opened 2022-07-08
do not ignore errors during downloading of tenant index parts

LizardWizzard:
Instead of swallowing errors it catches them but this state is not recoverable other than through detach/attach sequence

Misses tests, I wonder should we have a Mock impl of remote storage trait that serves like a proxy to another implementation but randomly (or semi randomly) throws errors. WDYT @SomeoneToIgnore?

### DIFF
--- a/pageserver/src/storage_sync.rs
+++ b/pageserver/src/storage_sync.rs
@@ -1557,6 +1557,7 @@ fn schedule_first_sync_tasks(
     local_timeline_init_statuses
 }
 
+/// bool in return value stands for awaits_download
 fn compare_local_and_remote_timeline(
     new_sync_tasks: &mut VecDeque<(ZTenantTimelineId, SyncTask)>,
     sync_id: ZTenantTimelineId,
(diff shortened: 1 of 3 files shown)


SomeoneToIgnore (approved):
Indeed feels like more negative tests with failures are great to have, so we can try and work on that too.

So far though, the `Poisoned` state looks a bit more global than I've wanted to? 
I like last TODO in this PR and I wonder, if the first step would be to actually collect as many index parts as we can, since those are mapped as one per timeline: throwing them all away due to one error seems wasteful.

Errors could be rescheduled for later downloads, actually, but that's definitely out of scope of this PR.

LizardWizzard:
Yeah, I agree with your point, I can change poisoned variant so it still contains the data without throwing it away. It can collect both, failed and successful downloads, so the later part would be to actually reschedule failed tasks

SomeoneToIgnore (approved):
Thank you for updating this.

PR #2059, opened 2022-07-08
Add OpenAPI spec for safekeeper HTTP API

### LINKED ISSUE #2061: Make safekeeper REST API more consistent
As proposed in https://github.com/neondatabase/neon/pull/2059.

ololobus:
I was thinking about doing this issue https://github.com/neondatabase/cloud/issues/1264. But we now try to use generated http clients for all components, so I added one for safekeeper too.

Yet, along the way I noticed that its API is a bit messy and not that RESTy. So I propose to do following changes in this PR:
- GET "/v1/timeline/:tenant_id/:timeline_id" -> "/v1/tenants/:tenant_id/timeline/:timeline_id"
- POST "/v1/timeline" -> "/v1/tenant/:tenant_id/timeline" + moving `tenant_id` out of JSON payload to path params

As far as I see, nobody is using it beside tests, so it's fine to break. @arssher @yeputons what do you think?

Also, if you look on the spec itself, it'd be great. You can copy-paste it here https://editor.swagger.io/ to have a GUI

BTW, I prefer plural scopes like `tenants` and `timelines`, but it seems that pageserver uses singular form, so let's keep it as is.

Resolve #2061

### DIFF
--- a/safekeeper/src/http/openapi_spec.yaml
+++ b/safekeeper/src/http/openapi_spec.yaml
@@ -1,0 +1,365 @@
+openapi: "3.0.2"
+info:
+  title: Safekeeper control API
+  version: "1.0"
+
+
+servers:
+  - url: "http://localhost:7676"
+
+
+paths:
+  /v1/status:
+    get:
+      tags:
+      - "Info"
+      summary: Get safekeeper status
+      description: ""
+      operationId: v1GetSafekeeperStatus
+      responses:
+        "200":
+          description: Safekeeper status
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/SafekeeperStatus"
+        "403":
+          $ref: "#/components/responses/ForbiddenError"
+        default:
+          $ref: "#/components/responses/GenericError"
+
+
+  /v1/tenant/{tenant_id}:
+    parameters:
+      - name: tenant_id
+        in: path
+        required: true
+        schema:
+          type: string
+          format: hex
+
+    delete:
+      tags:
+      - "Tenant"
+      summary: Delete tenant and all its timelines
+      description: "Deletes tenant and returns a map of timelines that were deleted along with their statuses"
+      operationId: v1DeleteTenant
+      responses:
+        "200":
+          description: Tenant deleted
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/TenantDeleteResult"
+        "403":
+          $ref: "#/components/responses/ForbiddenError"
+        default:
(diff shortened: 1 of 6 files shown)


arssher:
+1 on restructuring (and spec).
https://github.com/neondatabase/neon/issues/2061

ololobus:
> +1 on restructuring (and spec). https://github.com/neondatabase/neon/issues/2061

I think that I can do it right in this PR, so assigning the issue on me. Will generate a client and test with console as well

safekeeper/src/http/openapi_spec.yaml
```diff
@@ -0,0 +1,389 @@
...
+  title: Safekeeper control API
+  version: "1.0"
+
+
+servers:
+  - url: "http://localhost:7676"
+
+
+paths:
+  /status:
```
petuhovskiy, line 12:
```suggestion
  /v1/status:
```

safekeeper/src/http/routes.rs
```diff
@@ -126,7 +126,7 @@ async fn timeline_create_handler(mut request: Request<Body>) -> Result<Response<
     let request_data: TimelineCreateRequest = json_request(&mut request).await?;
 
     let zttid = ZTenantTimelineId {
-        tenant_id: request_data.tenant_id,
+        tenant_id: parse_request_param(&request, "tenant_id")?,
```
petuhovskiy, line 129: Having `tenant_id` in the url and `timeline_id` in the body looks strange. I'm interested what's the reasoning behind this? It seems that in all other places they are together in the url like `/tenant/:tenant_id/timeline/:timeline_id`.

ololobus: Maybe I'm wrong, but I think that it's just usual REST convention:
- when you use `/resources/:id` you address some existing resource, so it's usually GET, PATCH, DELETE
- when you create new resource, you use `POST or PUT /resources/`, because i) you may put an arbitrary payload as body and even if `id` is known at creation time it doesn't have much sense to split a single entity payload into path and body parts; and ii) `id` is usually not known at creation time (not our case, though)

ololobus: BTW, as far as I see, this endpoint is never used. And anyway we don't have tenant creation endpoint, so does tenant represent some separate entity on pageserver side? Or it always goes along with timeline? So it doesn't matter whether there are some timelines for this tenant already or not

petuhovskiy (approved):

petuhovskiy:
Some tests have failed in CI, maybe rebase on latest main will help

PR #2060, opened 2022-07-08
Add `close_fds` for `initdb` command and add close fd test

aome510:
This PR adds a test for https://github.com/neondatabase/neon/pull/1834 and fixes the error in https://app.circleci.com/pipelines/github/neondatabase/neon/7753/workflows/94d1b796-10a3-4989-b23c-4c1eb4a49cf5/jobs/79586, which happens because `pageserver.pid` is held by `initdb` command on restart.

Because the test requires `lsof` to be installed in the docker image, this PR also updates the caches and docker image specified in CircleCI config file.


test_runner/batch_others/test_close_fds.py
```diff
@@ -0,0 +1,51 @@
...
+import subprocess
+import os.path
+
+from cached_property import threading
+from fixtures.neon_fixtures import NeonEnv
+from fixtures.log_helper import log
+
+
+def lsof_path() -> str:
+    path_output = shutil.which("lsof")
```
bojanserafimov, line 13: If this is not installed by default on osx, let's add it to the "getting started" readme

bojanserafimov (approved):
Thanks!

Let's pick up https://github.com/neondatabase/docker-images/pull/11 to see if tests pass

test_runner/batch_others/test_close_fds.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def start_workload():
+        env.neon_cli.create_branch("test_lsof_pageserver_pid")
+        pg = env.postgres.create_start("test_lsof_pageserver_pid")
+        with closing(pg.connect()) as conn:
+            with conn.cursor() as cur:
+                cur.execute("CREATE TABLE foo as SELECT x FROM generate_series(1,100000) x")
+                cur.execute("update foo set x=x+1")
+
+    workload_thread = threading.Thread(target=start_workload, args=())
```
LizardWizzard, line 33: lets have `daemon=True` so the thread exits if main thread finishes

SomeoneToIgnore:
lgtm, thanks for adjusting those.

Issue #1667, opened 2022-05-10
test_tenant_relocation is flaky


arssher:
https://app.circleci.com/pipelines/github/neondatabase/neon/6205/workflows/7d0eb517-8275-47fa-84b5-f927544ac02a/jobs/61751/steps

```
test_tenant_relocation[with_load]

[FLAKY](https://app.circleci.com/insights/github/neondatabase/neon/workflows/build_and_test/tests?branch=cleanup-serde)
test_runner.batch_others.test_tenant_relocation
psycopg2.errors.InternalError_: [ZENITH_SMGR] could not get response from pageserver: server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
test_runner/batch_others/test_tenant_relocation.py:147: in test_tenant_relocation
    cur.execute("CREATE TABLE load(value text)")
E   psycopg2.errors.InternalError_: [ZENITH_SMGR] could not get response from pageserver: server closed the connection unexpectedly
E   	This probably means the server terminated abnormally
E   	before or while processing the request.
```

```
test_tenant_relocation[without_load]

[FLAKY](https://app.circleci.com/insights/github/neondatabase/neon/workflows/build_and_test/tests?branch=cleanup-serde)
test_runner.batch_others.test_tenant_relocation
test_runner/batch_others/test_tenant_relocation.py:54: in new_pageserver_helper
    yield
test_runner/batch_others/test_tenant_relocation.py:233: in test_tenant_relocation
    with pg_cur(tenant_pg) as cur:
../.pyenv/versions/3.9.2/lib/python3.9/contextlib.py:117: in __enter__
    return next(self.gen)
test_runner/batch_others/test_tenant_relocation.py:63: in pg_cur
    with closing(pg.connect()) as conn:
test_runner/fixtures/zenith_fixtures.py:284: in connect
    conn = psycopg2.connect(**self.conn_options(**kwargs))
../.cache/pypoetry/virtualenvs/zenith-3aSsmiER-py3.9/lib/python3.9/site-packages/psycopg2/__init__.py:122: in connect
    conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
E   psycopg2.OperationalError: connection to server at "localhost" (127.0.0.1), port 15112 failed: FATAL:  [ZENITH_SMGR] could not get response from pageserver: server closed the connection unexpectedly
E   	This probably means the server terminated abnormally
E   	before or while processing the request.
```

CC @LizardWizzard

[...]

PR #2062, opened 2022-07-08
test_tenant_relocation: add more logging

LizardWizzard:
Trying to reproduce https://github.com/neondatabase/neon/issues/1667#issuecomment-1179001544 but without success. I've rerun job 5 times in CI and it worked fine. So lets merge it and continue investigation once it fails again.

Just to note. The failure occurs s before anything related to relocation actually starts. Its spins a branch, inserts a bunch of rows in it, then forks it, adds another portion of rows and checks if these numbers add up and this is the point where it breaks. So it may be a problem or some sort of a race condition during branch creation

### DIFF
--- a/test_runner/batch_others/test_tenant_relocation.py
+++ b/test_runner/batch_others/test_tenant_relocation.py
@@ -14,6 +14,7 @@ from fixtures.neon_fixtures import (
     NeonEnv,
     NeonEnvBuilder,
     NeonPageserverHttpClient,
+    PageserverPort,
     PortDistributor,
     Postgres,
     assert_no_in_progress_downloads_for_tenant,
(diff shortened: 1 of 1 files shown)


test_runner/batch_others/test_tenant_relocation.py
```diff
@@ -115,14 +116,24 @@ def load(pg: Postgres, stop_event: threading.Event, load_ok_event: threading.Eve
...
     # insert some data
     with pg_cur(pg) as cur:
         cur.execute("SHOW neon.timeline_id")
         timeline_id = UUID(cur.fetchone()[0])
         log.info("timeline to relocate %s", timeline_id.hex)
 
+        cur.execute("SELECT pg_current_wal_flush_lsn()")
+        log.info("pg_current_wal_flush_lsn() %s", lsn_from_hex(cur.fetchone()[0]))
+        log.info("timeline detail %s",
+                 ps_http.timeline_detail(tenant_id=tenant_id, timeline_id=timeline_id))
```
yeputons, line 135: Does it really logs as expected? I'd expect f-strings or `"timeline detail {}".format(.....)` instead.

yeputons (approved):

test_runner/batch_others/test_tenant_relocation.py
LizardWizzard: Yeah, it works :) This method works better for integration with various logging middleware. For example if you use Sentry it will be able do dedup events by taking into account only format string, not the parameters. If parameters are pre filled like you proposed then it wont be able to do so and will create separate tickets for every combination of user_id/request_id/whatever that came with the event. It is highly unlikely that we'll need to use something like Sentry in tests, but this is the way that I used to cook python logging just to adhere to least surprise principle

PR #2064, opened 2022-07-08
Refactor and test `find_end_of_wal` with non-zero `start_lsn`

yeputons:
Preparation for #2066 so ensure the current behavior does not break.

I've checked that replacing `start_lsn` with `Lsn(0)` in `find_end_of_wal` breaks the test because it cannot parse the erased beginning of the WAL.


libs/postgres_ffi/src/xlog_utils.rs
```diff
@@ -622,26 +621,50 @@ mod tests {
...
-        info!(
-            "find_end_of_wal returned (wal_end={}, tli={})",
-            wal_end, tli
-        );
-        assert_eq!(wal_end, expected_end_of_wal_non_partial);
+        // Check find_end_of_wal on the initial WAL
+        let last_segment = get_last_segment();
+        check_pg_waldump_end_of_wal(&cfg, &last_segment, expected_end_of_wal_partial);
+        for start_lsn in std::iter::once(Lsn(0)).chain(intermediate_lsns) {
+            // TODO: zero WAL files before start_lsn
```
yeputons, line 649: It is to make sure we're really reading from `start_lsn`, and not just from the start every time.

libs/postgres_ffi/src/xlog_utils.rs
```diff
@@ -622,26 +621,50 @@ mod tests {
...
-        let wal_end = Lsn(wal_end);
-        info!(
-            "find_end_of_wal returned (wal_end={}, tli={})",
-            wal_end, tli
-        );
-        assert_eq!(wal_end, expected_end_of_wal_non_partial);
+        // Check find_end_of_wal on the initial WAL
+        let last_segment = get_last_segment();
+        check_pg_waldump_end_of_wal(&cfg, &last_segment, expected_end_of_wal_partial);
+        for start_lsn in std::iter::once(Lsn(0)).chain(intermediate_lsns) {
```
yeputons, line 648: Chain `expected_end_of_wal_partial` as well.

arssher (approved):

PR #2067, opened 2022-07-09
Make Postgres installation dir configurable

bayandin:
This PR allows configuring the installation directory for Postgres (by setting `POSTGRES_DISTRIB_DIR`, default is `tmp_install`).

This is my first Rust code, so it's highly inspired by rust-cookbook and numerous Google results.

ihave

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,8 @@
+ROOT_PROJECT_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
+
+# Where to install Postgres, default is ./tmp_install, maybe useful for package managers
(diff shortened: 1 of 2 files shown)


hlinnaka (approved):
The Rust changes look OK to me. I'm curious, why do you need to change the installation dir?

bayandin:
> I'm curious, why do you need to change the installation dir?

It's for package managers mostly (and in other cases when download location and install location are different). I've played with a [Homebrew tap](https://github.com/bayandin/homebrew-tap/blob/main/Formula/neon.rb), but I guess others require something similar.

Issue #2058, opened 2022-07-08
Give access to the public schema to db owner


ololobus:
Some users report that db owner cannot drop the public schema. Indeed:
```sql
main=> drop schema public cascade;
ERROR:  must be owner of schema public
```

```sql
main=> SELECT nspname, nspowner::regrole::text FROM pg_catalog.pg_namespace;
      nspname       |  nspowner   
--------------------+-------------
 pg_toast           | cloud_admin
 pg_catalog         | cloud_admin
 public             | cloud_admin
 information_schema | cloud_admin
 mypublic           | web_access
```

The workaround could be to grant permissions to `public` to db owner
```sql
ALTER SCHEMA public OWNER TO <db_owner>;
```

Let's try it.

zomars:
Thanks for the follow up @ololobus I suspect that I cannot run this myself right? Trying to run `ALTER SCHEMA public OWNER TO <db_owner>;` returns the same error: 

```
SQL Error [42501]: ERROR: must be owner of schema public
```

ololobus:
Yeah, we need to adjust our compute node code. Just tried with the local Postgres instance, worked well. Not sure how it'll play with our roles management, though, will try

ololobus:
@zomars just bumped the compute image version on prod. Database owner now should have a full control of the scheme `public`. So you can drop and create it again. Can you try to restore your dump again?

stepashka:
feel free to reopen if this doesn't work for you, @zomars

zomars:
Are these changes live? I'm afraid I'm seeing the same issue:

<img width="820" alt="image" src="https://user-images.githubusercontent.com/3504472/180882752-a28863ba-9269-44b5-951e-b6a7eb3a34c7.png">

ololobus:
> Are these changes live?

Yes, it is public. Just tried with a new project and database owner is owner of schema public, so I can drop and create it again:
```
ololobus@main=> SELECT nspname, nspowner::regrole::text FROM pg_catalog.pg_namespace;
      nspname       |  nspowner   
--------------------+-------------
 pg_toast           | cloud_admin
 pg_catalog         | cloud_admin
 information_schema | cloud_admin
 public             | ololobus
(4 rows)

ololobus@main=> drop schema public cascade;
DROP SCHEMA
```

> I'm afraid I'm seeing the same issue

[...]

PR #2072, opened 2022-07-11
[compute_tools] Set public schema owner to db owner (#2058)

ololobus:
Otherwise, it does not have a control on it, which is reasonable thing
to have and some users already hit it.

It'd be nice to also ask Anastasia for review, but she is on vacation now.


compute_tools/src/spec.rs
```diff
@@ -380,5 +382,20 @@ pub fn handle_grants(spec: &ComputeSpec, client: &mut Client) -> Result<()> {
         client.execute(query.as_str(), &[])?;
     }
 
+    // Do some per-database access adjustments
+    for db in &node.spec.cluster.databases {
+        let mut db_connstr = node.connstr.clone();
```
SomeoneToIgnore, line 387: I wonder if that could be created before the loop once, and then `set_path` could be called on the same data.

SomeoneToIgnore (commented):
lgtm, but I don't have the whole context here

compute_tools/src/spec.rs
ololobus: Changed, thanks

SomeoneToIgnore (approved):

Issue #2016, opened 2022-07-01
Epic: set up infra for periodic perf tests with bigger databases


kelvich:
Right now we have some infrastructure to run periodic tests, but it needs more love. For some amount of time safekeepers were not capable of automatically deleting WAL, so we avoided big databases as they were causing disk overflow on staging. Now safekeepers can clean their WAL so let's start doing more realistic tests.

As result we want to have the following:
* Grafana overview dashboard with one panel for each test -- panel shows each test result as a single number (seconds or transactions) against test competition time.
* Links to per-timeline dashboard for each test run.

Approximate steps:
- [x] bring existing periodic perf tests to life: https://github.com/neondatabase/neon/pull/2037
- [x] update runner (or use k8s runners): https://github.com/neondatabase/neon/pull/2037
- [x] put results to s3 instead of git (or it is stored in postgres? -- find out)
- [x] increase scale up to 50GB https://github.com/neondatabase/neon/pull/2612
- [ ] increase scale up to 300GB
- [x] increase run time to a few hours: https://github.com/neondatabase/neon/pull/2077
- [x] set up / fix grafana dashboards

bayandin:
https://github.com/neondatabase/neon/issues/2221 is required to continue

stepashka:
during the meeting today @hlinnaka  mentioned that  we're ready  to increase the scale up more towards the 300GB, e.g. 50GB first?

stepashka:
is this still blocked by neondatabase/neon#2221 ?

bayandin:
> during the meeting today @hlinnaka mentioned that we're ready to increase the scale up more towards the 300GB, e.g. 50GB first?

I've experemented with `pgbench -i`, setting up a 50 GB (scale 3424) database took ~ 1h (with enabled prefetch):
```
------------------------------ Benchmark results -------------------------------
test_pgbench_remote_init[3600-3424].scale: 3424 
test_pgbench_remote_init[3600-3424].init.start_timestamp: 1665488570 
test_pgbench_remote_init[3600-3424].init.end_timestamp: 1665492349 
test_pgbench_remote_init[3600-3424].init.duration: 3,778.830 s
test_pgbench_remote_init[3600-3424].init.drop_tables: 0.030 s
test_pgbench_remote_init[3600-3424].init.create_tables: 0.130 s
test_pgbench_remote_init[3600-3424].init.client_side_generate: 1,419.680 s
[...]

PR #2077, opened 2022-07-12
Set up a workflow to run pgbench against captest

bayandin:
I've added a `pgbench-compare` job to benchmark workflow that uses a project on captest and rds.
Along with it:
- Split init of pgbench into stages
- Added parameterization for `test_pgbench_remote` test make it possible to run only init/select-only/simple-update or all of them
- Added parameterization of test by human-readable size (rather than scale) [ref](https://www.cybertec-postgresql.com/en/a-formula-to-calculate-pgbench-scaling-factor-for-target-db-size/)

See https://observer.zenith.tech/d/B1g2eWm4k/wip-perf-test-results?orgId=1&from=now-7d&to=now


test_runner/fixtures/utils.py
```diff
@@ -34,8 +34,12 @@ def subprocess_capture(capture_dir: str, cmd: List[str], **kwargs: Any) -> str:
...
     with open(stdout_filename, 'w') as stdout_f:
         with open(stderr_filename, 'w') as stderr_f:
-            log.info('(capturing output to "{}.stdout")'.format(base))
-            subprocess.run(cmd, **kwargs, stdout=stdout_f, stderr=stderr_f)
+            log.info(f'(capturing output to "{base}.stdout" and "{base}.stderr")')
+            try:
+                subprocess.run(cmd, **kwargs, stdout=stdout_f, stderr=stderr_f)
+            finally:
+                stdout_f.flush()
+                stderr_f.flush()
```
bojanserafimov, line 42: the `open` context doesn't do this?

test_runner/performance/test_perf_pgbench.py
```diff
@@ -129,6 +176,7 @@ def test_pgbench_flamegraph(zenbenchmark, pg_bin, neon_env_builder, scale: int,
 # Run the pgbench tests against an existing Postgres cluster
 @pytest.mark.parametrize("scale", get_scales_matrix())
 @pytest.mark.parametrize("duration", get_durations_matrix())
+@pytest.mark.parametrize("filter", ["init", "simple-update", "select-only"])
```
bojanserafimov, line 179: can I run `simple-update` without running `init` first?

.github/workflows/benchmarking.yml
```diff
@@ -113,3 +116,91 @@ jobs:
...
+      TEST_PG_BENCH_SCALES_MATRIX: "10gb"
+      REMOTE_ENV: "1"
+      POSTGRES_DISTRIB_DIR: /usr
+
+    strategy:
+      fail-fast: false
+      matrix:
+        connstr: [ BENCHMARK_CAPTEST_CONNSTR, BENCHMARK_RDS_CONNSTR ]
+
+    runs-on: dev
```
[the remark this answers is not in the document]
bayandin: `eu-central-1`  
Ref https://github.com/neondatabase/aws/tree/main/terraform/environment/dev

.github/workflows/benchmarking.yml
```diff
@@ -113,3 +116,91 @@ jobs:
...
+        git --version
+
+    - name: Install poetry & python deps
+      run: |
+        python3 -m pip install --upgrade poetry
+        ./scripts/pysync
+
+    - name: Benchmark init
+      env:
+        PLATFORM: ${{ (matrix.connstr == 'BENCHMARK_RDS_CONNSTR' && 'rds-aurora') || 'neon-captest' }}
```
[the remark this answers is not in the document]
bayandin: Nice! Like how symmetric it is.

.github/workflows/benchmarking.yml
```diff
@@ -113,3 +116,91 @@ jobs:
...
+        BENCHMARK_CONNSTR: ${{ secrets[matrix.connstr] }}
+      run: |
+        mkdir -p perf-report-captest
+
+        psql $BENCHMARK_CONNSTR -c "SELECT 1;"
+        ./scripts/pytest test_runner/performance/test_perf_pgbench.py -v -m "remote_cluster" -k "init" --skip-interfering-proc-check --out-dir perf-report-captest --timeout 21600
```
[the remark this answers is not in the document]
bayandin: Because the name also depends on other parameters, it will require to be changed if we change the duration or DB size. And I will definitely forget about it. So I think I'll keep it as it is.

test_runner/fixtures/utils.py
bayandin: It should, those are leftovers from debugging, will delete it

test_runner/performance/test_perf_pgbench.py
bayandin: Currently, we use pre-created DB, so yes

.github/workflows/benchmarking.yml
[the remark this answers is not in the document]
bojanserafimov: I'd add a comment `defined here: <link>`. Maybe it's obvious if you're fluent in yaml configs, but it would take me a while to figure it out :)

test_runner/performance/test_perf_pgbench.py
bojanserafimov: Where is it created? Does the scale parameter affect `simple-update`?

.github/workflows/benchmarking.yml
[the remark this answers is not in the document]
bojanserafimov: Ah yes that makes sense. It can still backfire, but it's better.

Let's come back to this after the readme is updated. I'd ask "why not run **all** remote tests, in one step, just like we do against staging", but I'll get my answer from the readme :)

test_runner/performance/test_perf_pgbench.py
bayandin: > Where is it created?

Wherever we pass in connection string, this PR uses `BENCHMARK_CAPTEST_CONNSTR` and `BENCHMARK_RDS_CONNSTR`, so it's on captest env and on AWS RDS.

> Does the scale parameter affect simple-update?

Not directly, and if we run only `simple-update` test it doesn't affect it at all

bojanserafimov: We use the same connstr for init and simple-update. So the database size is effectively "whatever was created by the last init run"?

bayandin: Yes (and if we run only `simple-update`, then the database will come from the previous run)

bayandin: I probably misunderstood your first question, and now I got lost 😄 

So

> can I run simple-update without running init first?

Yes, using `-k simple-update`, as we use in our workflow. And in this case, we won't know anything about real size

.github/workflows/benchmarking.yml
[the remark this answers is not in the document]
bayandin: > why not run all remote tests, in one step, just like we do against staging

Long story short: Because we want to run them separately 😄 

This PR is a part of https://github.com/neondatabase/neon/issues/2016. 
Initially, the PR was about adding like 300GB database, but during the process of setting it up, I've undercovered a couple of issues[1], [2] that make it hardly possible to do with pgbench initialize. 
I've split the test into steps to make it possible to set up the database once and then run only `simple-update`/`select-only` flows. Also, such separation makes it more convenient to see what's going on when you open a GitHub workflow run (is it still initializing or doing some real tests).

And this PR has evolved into something like, "let's start collecting initialize results and compare them with RDS, then we will be able to set up a big database as it was intended originally"

- [1] https://github.com/neondatabase/cloud/issues/1872
- [2] https://github.com/neondatabase/neon/issues/2221

test_runner/performance/test_perf_pgbench.py
bojanserafimov: Ok, I understand. Currently we run 3 steps:
1. `-k init`
2. `-k simple-update`
3. `-k select-only`

The duration+scale matrix for the `bench-captest` job is 1x1, so this is fine.

But the `bench` job runs all `remote_cluster` tests, in whatever order pytest decides on, and it's matrix is 2x1. Is it possible that It runs two inits with different sizes that override each other, and then two simple-updates that are intended to run with different sizes but actually run on whatever was left from the last init?

There's also some non-determinisim if simple-update and select-only are run in a different order with each other. Maybe select-only is slower if it runs on a database that has been simple-update'd?

FYI [this](https://pytest-dependency.readthedocs.io/en/stable/) might be helpful

bayandin: > But the bench job runs all remote_cluster tests, in whatever order pytest decides on, and it's matrix is 2x1. Is it possible that It runs two inits with different sizes that override each other, and then two simple-updates that are intended to run with different sizes but actually run on whatever was left from the last init?

You're right. `bench` job is a bit outdated for now (uses runner manually configured runner, for example). I will update it in the following PRs.

> FYI [this](https://pytest-dependency.readthedocs.io/en/stable/) might be helpful

Yep, I think I'll use either this or [that](https://pypi.org/project/pytest-order/)

.github/workflows/benchmarking.yml
```diff
@@ -113,3 +116,91 @@ jobs:
         slack-message: "Periodic perf testing: ${{ job.status }}\n${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
       env:
         SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
+
+  bench-captest:
```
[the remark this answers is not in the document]
bayandin: We actually have it in `test_runner/performance/README.md`, I've added a not about using captest env.

.github/workflows/benchmarking.yml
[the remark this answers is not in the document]
bayandin: We're working to make this our default environment; I don't think we need to mention it in this particular file, but probably should be added to docs/wiki after https://github.com/neondatabase/neon/pull/2128

bayandin:
@bojanserafimov I believe I've addressed all your comments; here are the new changes from the last review https://github.com/neondatabase/neon/compare/d4dab395bef42b98b3f31819d9c6143766b4c854..b87f94b91fa55f8c56dc1420b2d2a9b7f5ce2873 (sorry, had to rebase due to merge conflicts)

.github/workflows/benchmarking.yml
```diff
@@ -96,7 +104,7 @@ jobs:
         # since it might generate duplicates when calling ingest_perf_test_result.py
         rm -rf perf-report-staging
         mkdir -p perf-report-staging
-        ./scripts/pytest test_runner/performance/ -v -m "remote_cluster" --skip-interfering-proc-check --out-dir perf-report-staging --timeout 3600
+        ./scripts/pytest test_runner/performance/ -v -m "remote_cluster" --sparse-ordering --skip-interfering-proc-check --out-dir perf-report-staging --timeout 3600
```
bojanserafimov, line 107: Does `--sparse-ordering` also imply "run tests in order of definition"? I don't see any `@pytest.mark.order` calls.

If so, let's add a comment by the test definitions that the order is important.

.github/workflows/benchmarking.yml
```diff
@@ -113,3 +121,106 @@ jobs:
         slack-message: "Periodic perf testing: ${{ job.status }}\n${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
       env:
         SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
+
+  pgbench-compare:
```
bojanserafimov, line 125: I like this new name. Implicitly answers many questions.

test_runner/performance/test_perf_pgbench.py
```diff
@@ -70,31 +69,75 @@ def run_pgbench(env: PgCompare, prefix: str, cmdline):
 # the test database.
 #
 # Currently, the # of connections is hardcoded at 4
-def run_test_pgbench(env: PgCompare, scale: int, duration: int):
-
-    # Record the scale and initialize
+def run_test_pgbench(env: PgCompare, scale: int, duration: int, filter: str = "all"):
```
bojanserafimov, line 72: `"all"` doesn't run all tests, which is maybe why `test_pgbench` and `test_pgbench_flamegraph` are failing in CI

test_runner/performance/test_perf_pgbench.py
```diff
@@ -70,38 +77,84 @@ def run_pgbench(env: PgCompare, prefix: str, cmdline):
...
 
 
 # Run the pgbench tests against vanilla Postgres and neon
 @pytest.mark.parametrize("scale", get_scales_matrix())
 @pytest.mark.parametrize("duration", get_durations_matrix())
 def test_pgbench(neon_with_baseline: PgCompare, scale: int, duration: int):
-    run_test_pgbench(neon_with_baseline, scale, duration)
+    run_test_pgbench(neon_with_baseline, scale, duration, PgBenchLoadType.INIT)
+    run_test_pgbench(neon_with_baseline, scale, duration, PgBenchLoadType.SIMPLE_UPDATE)
+    run_test_pgbench(neon_with_baseline, scale, duration, PgBenchLoadType.SELECT_ONLY)
```
bojanserafimov, line 157: Nice, I like this change. Nit: now the `filter` argument could be renamed to a more self-explanatory name, `workload_type`.

test_runner/performance/test_perf_pgbench.py
```diff
@@ -123,12 +176,29 @@ def test_pgbench_flamegraph(zenbenchmark, pg_bin, neon_env_builder, scale: int,
...
+@pytest.mark.remote_cluster
+def test_pgbench_remote_simple_update(remote_compare: PgCompare, scale: int, duration: int):
+    run_test_pgbench(remote_compare, scale, duration, filter=PgBenchLoadType.SIMPLE_UPDATE)
+
+
+@pytest.mark.parametrize("scale", get_scales_matrix())
+@pytest.mark.parametrize("duration", get_durations_matrix())
+@pytest.mark.remote_cluster
+def test_pgbench_remote_select_only(remote_compare: PgCompare, scale: int, duration: int):
+    run_test_pgbench(remote_compare, scale, duration, filter=PgBenchLoadType.SELECT_ONLY)
```
bojanserafimov, line 204: Can we comment that the relative order of these 3 tests shouldn't change, and no new tests should be added in between? Maybe at the top of all 3 tests, comment: "This test runs on an existing database as it was set up by previous tests, and leaves the database in a state that would be used in the next tests. Modifying the definition order of these functions or adding other remote tests in between will alter results. See usage of `--sparse-ordering` flag in the pytest invocation in the CI workflow"

bojanserafimov (approved):
Looks good. Hopefully I noticed all changes since last review, but can't be sure because the force push makes it hard. Lmk if I should notice some specific change

bojanserafimov (approved):

bayandin:
Thanks a lot for the code review, @bojanserafimov!

PR #2080, opened 2022-07-13
Migrate from CircleCI to Github Actions: Send deploy notifications from github actions workflow

SergeyMelnikov:


.github/workflows/build_and_test.yml
```diff
@@ -599,6 +600,16 @@ jobs:
...
 
+      - name: Post to a Slack channel
+        if: always()
+        id: slack
+        uses: slackapi/slack-github-action@v1
+        with:
+          channel-id: ${{ matrix.slack_channel_id }}
+          slack-message: "Neon deploy to ${{ matrix.env_name }}: ${{ job.status }}\nJob details: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+        env:
+          SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
```
ololobus, line 611: It will only post if deploy job started, but if it failed before that we didn't know that the whole workflow failed. It could be a separate job that runs always and reports success only if deploy succeeded, otherwise failed including scipped

bayandin: So, does the deploy job finish successfully even if the deploy isn't successful?
Maybe we should add polling for its result here as well?

.github/workflows/build_and_test.yml
```diff
@@ -577,6 +577,7 @@ jobs:
           pip install --progress-bar off --user ansible boto3
 
       - name: Redeploy
```
bayandin, line 579:
```suggestion
      - name: Re-deploy
```

.github/workflows/build_and_test.yml
ololobus: > So, does the deploy job finish successfully even if the deploy isn't successful?

No, I suggest that there will be a separate job `report-deploy-status`, that should be always executed. Similar to what we have in the cloud repo for remote CI status https://github.com/neondatabase/cloud/blob/0f34c89cc71c903106d4950fa85c8e6726136d66/.github/workflows/testing.yml#L980

PR #2084, opened 2022-07-13
Add zenith-1-ps-3 to prod inventory

petuhovskiy:


SergeyMelnikov (approved):

cicdteam (approved):

petuhovskiy:
Yes, there is a special patched version deployed to ps-3. I see no problems with merging it now, because we deploy to production only in release branches. So this patched version will live there until we create a new commit in `release` branch. Am I missing something?

PR #2087, opened 2022-07-13
Update README.md

kelvich:
I think installing postgres client is optional, isn't it? And StackOverflow link looks a bit strange there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,13 +62,6 @@ brew install protobuf etcd openssl
 curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
 ```
 
-3. Install PostgreSQL Client
-```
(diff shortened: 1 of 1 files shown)


aome510:
> I think installing postgres client is optional, isn't it?

If you mean optional in terms of running `neon`, then yes. But, the example instructions in https://github.com/neondatabase/neon/tree/fix-flaky-branch-tests#running-neon-database requires to have a `psql` command in `$PATH`.  We can change that to `tmp_install/bin/psql` if needed.

Note there is also a dependency note in the readme:
> To run the `psql` client, install the `postgresql-client` package or modify `PATH` and `LD_LIBRARY_PATH` to include `tmp_install/bin` and `tmp_install/lib`, respectively.

Issue #2086, opened 2022-07-13
test_ancestor_branch and test_branch_and_gc are flaky


aome510:

yeputons:
Examples of `test_ancestor_branch`:

https://app.circleci.com/pipelines/github/neondatabase/neon/8032/workflows/f017e5f0-940d-41ce-bb71-7197930ceb17/jobs/82345
https://app.circleci.com/pipelines/github/neondatabase/neon/8032/workflows/c6c23a56-dc0b-41e7-96c8-112317be5fc4/jobs/82328/tests#failed-test-0

Logs are the almost same, here is the first one:

```
Exception:             Run ['/tmp/zenith/bin/neon_local', 'timeline', 'branch', '--branch-name', 'branch2', '--tenant-id', 'f87be4beaf274b768268ce637e189cf5', '--ancestor-branch-name', 'branch1', '--ancestor-start-lsn', '0/2C4C898'] failed:
              stdout: 
              stderr: command failed: error sending request for url (http://localhost:15004/v1/tenant/f87be4beaf274b768268ce637e189cf5/timeline): operation timed out

Caused by:
    operation timed out
subprocess.CalledProcessError: Command '['/tmp/zenith/bin/neon_local', 'timeline', 'branch', '--branch-name', 'branch2', '--tenant-id', 'f87be4beaf274b768268ce637e189cf5', '--ancestor-branch-name', 'branch1', '--ancestor-start-lsn', '0/2C4C898']' returned non-zero exit status 1.

The above exception was the direct cause of the following exception:
test_runner/batch_others/test_ancestor_branch.py:72: in test_ancestor_branch
    env.neon_cli.create_branch('branch2', 'branch1', tenant_id=tenant, ancestor_start_lsn=lsn_200)
test_runner/fixtures/neon_fixtures.py:1100: in create_branch
    res = self.raw_cli(cmd)
test_runner/fixtures/neon_fixtures.py:981: in raw_cli
    raise Exception(msg) from subprocess.CalledProcessError(res.returncode,
E   Exception:             Run ['/tmp/zenith/bin/neon_local', 'timeline', 'branch', '--branch-name', 'branch2', '--tenant-id', 'f87be4beaf274b768268ce637e189cf5', '--ancestor-branch-name', 'branch1', '--ancestor-start-lsn', '0/2C4C898'] failed:
E                 stdout: 
[...]

PR #2091, opened 2022-07-14
Do not cancel in-progress checks on the `main` branch.

yeputons:
See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency

* Previously there was a single concurrency group per each branch.
  As the `main` branch got pushed into frequently, very few commits got
  tested to the end. It resulted in "broken" `main` branch as there were
  no fully successful workflow runs.
  Now the `main` branch gets a separate concurrency group for each commit.
* As GitHub Actions syntax does not have the conditional operator, it is
  emulated via logical and/or operations. Although undocumented, they
  return one of their operands instead of plain true/false.
* Replace 3-space indentation with 2-space indentation while we are here
  to be consistent with the rest of the file.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -11,8 +11,9 @@ defaults:
     shell: bash -ex {0}
 
 concurrency:
-   group: ${{ github.workflow }}-${{ github.ref }}
-   cancel-in-progress: true
+  # Allow only one workflow per any non-`main` branch.
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.ref == 'refs/heads/main' && github.sha || 'anysha' }}
+  cancel-in-progress: true
 
 env:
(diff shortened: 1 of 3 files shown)


yeputons:
As one can see, GitHub Actions on this PR were successfully cancelled after I force-pushed, see https://github.com/neondatabase/neon/actions/runs/2669472431 . The resulting concurrency group is `Check code style and build-refs/pull/2091/merge-anysha`.

yeputons:
I've also tested that it works the other way by temporarily replacing `main` with `concurrency-in-branches-1`.

SomeoneToIgnore (approved):
Thank you for fixing that, stumbled on it yesterday myself.

yeputons:
`test_ancestor_branch` keeps failing, I consider it unrelated to this PR. See #2086. Merging.

PR #2095, opened 2022-07-14
[proxy] Add the `password hack` authentication flow

funbringer:
See https://github.com/neondatabase/cloud/issues/1620#issuecomment-1165332290


kelvich:
Looks good, thanks!

I've set up a proxy from this branch alongside the local console and confirm that it works as desired: the usual password works with SNI, and an altered password works when SNI is disabled.

With this double base64 encoding it is hard to encode the password by hands. What do you think about using non-base64 encoded password? I think URL encoding is only needed when one uses `postres://...` type connection string but not needed in case of `host=... pass=...` type of string. So we can get away with a more transparent option.

Let's also impose a stricter schema on json just `project` and `password` (I would also suggest field password instead of token).

proxy/src/auth/password_hack.rs
```diff
@@ -0,0 +1,51 @@
...
+pub struct PasswordHackPayload {
+    pub project: String,
+
+    /// Token is base64-encoded because it may contain arbitrary byte sequences.
+    #[serde(deserialize_with = "deserialize_base64")]
+    pub token: Vec<u8>,
+}
+
+fn deserialize_base64<'a, D: Deserializer<'a>>(des: D) -> Result<Vec<u8>, D::Error> {
+    struct Visitor;
```
LizardWizzard, line 14: Should be correct to replace the body of the function with the following code:

```rust
let base64: &str = Deserialize::deserialize(des)?;
base64::decode(base64.as_bytes()).map_err(|e| serde::de::Error::custom(e))
```

LizardWizzard (commented):
I've looks through almost everything, looks good to me so far, will continue on monday

proxy/src/auth/password_hack.rs
funbringer: Indeed!

kelvich (approved):

LizardWizzard (approved):
Looks good to me, thanks!

PR #2105, opened 2022-07-15
Post periodic benchmarking failures to slack

bayandin:
We'd like to know when periodic perf tests fail: https://github.com/neondatabase/neon/actions/workflows/benchmarking.yml


LizardWizzard (approved):
Looks good to me, thanks!

PR #2113, opened 2022-07-18
Release 2022-07-18

petuhovskiy:
ref release guide: https://github.com/neondatabase/cloud/wiki/Release:-general-flow

Optional actions:
- [ ] One thing that I can think of as a preparation is take some tenant's files from ps-2 and safekeepers and try to run it locally with new binaries. Does that makes sense to do before the release?
- [ ] Copy old storage binaries to have them in case of rollback
- [ ] Check if S3 offloading is enabled for safekeepers?

To-do and checks:
- [ ] No serious issues in the `main`
- [X] Pageservers can access the etcd
- [X] Safekeepers are uploading data to etcd
- [X] Pageservers have correct etcd config
- [x] We have all prod pageservers and safekeepers in the ansible inventory
- [ ] Our CI is properly configured to deploy `release` branch to production
- [ ] Resolve merge conflicts #2114
- [ ] Check that `vendor/postgres` points to the main branch
- [ ] Prepare the changelog
- [ ] Something else?
- [ ] Press **Create a merge commit** once all checkboxes above are checked
- [ ] Wait for CI to finish and trigger deployment to prod
- [ ] Check notification in Slack
- [ ] Check that everything is working

PR #2115, opened 2022-07-18
Release 2022-07-18

petuhovskiy:
overrides https://github.com/neondatabase/neon/pull/2113, using temporary branch to avoid conflicts

ref release guide: https://github.com/neondatabase/cloud/wiki/Release:-general-flow

Optional actions:
- [ ] One thing that I can think of as a preparation is take some tenant's files from ps-2 and safekeepers and try to run it locally with new binaries. Does that makes sense to do before the release?
- [ ] Copy old storage binaries to have them in case of rollback
- [ ] Check if S3 offloading is enabled for safekeepers?

To-do and checks:
- [ ] No serious issues in the `main`
- [X] Pageservers can access the etcd
- [X] Safekeepers are uploading data to etcd
- [X] Pageservers have correct etcd config
- [x] We have all prod pageservers and safekeepers in the ansible inventory
- [ ] Our CI is properly configured to deploy `release` branch to production
- [X] Resolve merge conflicts
- [X] ~~Check that `vendor/postgres` points to the main branch~~ It's not, but it works
- [ ] Something else?
- [ ] Press **Create a merge commit** once all checkboxes above are checked
- [ ] Wait for CI to finish and trigger deployment to prod
- [ ] Check notification in Slack
- [ ] Check that everything is working
- [ ] Prepare the changelog


.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,650 @@
...
+          pip install --progress-bar off --user ansible boto3
+
+      - name: Redeploy
+        run: |
+          cd "$(pwd)/.github/ansible"
+
+          if [[ "$GITHUB_REF_NAME" == "main" ]]; then
+            ./get_binaries.sh
+          elif [[ "$GITHUB_REF_NAME" == "release" ]]; then
+            RELEASE=true ./get_binaries.sh
```
petuhovskiy, line 600: Should we pass ${VERSION} here, or it's already passed at some place of this yml? 
ref https://github.com/neondatabase/neon/blob/main/.github/ansible/get_binaries.sh#L15

cc @SergeyMelnikov @cicdteam

petuhovskiy: I looked at the staging CI logs, it seems that version is ok

PR #2117, opened 2022-07-18
Copy logical size from ancestor timeline when branching on the last position

knizhnik:
See https://neondb.slack.com/archives/C036U0GRMRB/p1657747809541039

### DIFF
--- a/pageserver/src/pgdatadir_mapping.rs
+++ b/pageserver/src/pgdatadir_mapping.rs
@@ -76,6 +76,12 @@ impl<R: Repository> DatadirTimeline<R> {
         Ok(())
     }
 
+    /// Copy the logical size of the database from ancestor branch.
+    pub fn copy_logical_size(&self, size: usize) {
+        self.current_logical_size
+            .store(size as isize, Ordering::SeqCst);
+    }
+
     /// Start ingesting a WAL record, or other atomic modification of
     /// the timeline.
     ///
--- a/pageserver/src/tenant_mgr.rs
+++ b/pageserver/src/tenant_mgr.rs
@@ -5,6 +5,7 @@ use crate::config::PageServerConf;
 use crate::layered_repository::{load_metadata, LayeredRepository};
 use crate::pgdatadir_mapping::DatadirTimeline;
 use crate::repository::Repository;
+use crate::repository::Timeline;
 use crate::storage_sync::index::{RemoteIndex, RemoteTimelineIndex};
 use crate::storage_sync::{self, LocalTimelineInitStatus, SyncStartupData};
 use crate::tenant_config::TenantConfOpt;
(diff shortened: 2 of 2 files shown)


aome510:
Looks like https://github.com/neondatabase/neon/pull/2094 is related. @knizhnik what is the status of the other PR? Should we merge that or wait for @hlinnaka for another review?

aome510:
> Yes. Optimization in https://github.com/neondatabase/neon/pull/2094 may help to reduce logical database size calculation after restart, bot not after branching (because branch is separate timeline and cache of ancestor timeline relations will not help)

Thanks for the clarification. Changes in this patch generally LGTM. However, I don't understand why you need to implement separate `Tenant::get_local_timeline_with_load` and `Tenant::load_local_timeline` functions. Why don't just modify the `load_local_timeline` function.

knizhnik:
> > Yes. Optimization in #2094 may help to reduce logical database size calculation after restart, bot not after branching (because branch is separate timeline and cache of ancestor timeline relations will not help)
> 
> Thanks for the clarification. Changes in this patch generally LGTM. However, I don't understand why you need to implement separate `Tenant::get_local_timeline_with_load` and `Tenant::load_local_timeline` functions. Why don't just modify the `load_local_timeline` function?
 
I have to make `load_local_timeline` because otherwise  `get_local_timeline_with_load` is called recursively (because I need to locate ancestor timeline from `init_logical_size` and it cause error on attempt of recursive exclusive lock of  `TENANTS`. 

y original idea was to completely eliminate `load_local_timeline` function and call method instead of it. But it requires mute reference to tenant cause Rust error in usage of this function in `attach_downloaded_tenant`.
I will be pleased if you can suggest way how to solve this issue.

pageserver/src/pgdatadir_mapping.rs
```diff
@@ -76,6 +76,12 @@ impl<R: Repository> DatadirTimeline<R> {
         Ok(())
     }
 
+    /// Copy the logical size of the database from ancestor branch.
```
aome510, line 79: nit: this function is quite general purpose. Maybe we should update it to `update_logical_size` and its description in case we may use it for other cases

pageserver/src/tenant_mgr.rs
```diff
@@ -104,6 +105,64 @@ struct Tenant {
...
+        let repartition_distance = self.repo.get_checkpoint_distance() / 10;
+        let mut ancestor_logical_size: Option<usize> = None;
+        if let Some(ancestor_id) = inmem_timeline.get_ancestor_timeline_id() {
+            if let Ok(ancestor_pgdir) = self.get_local_timeline_with_load(ancestor_id) {
+                if ancestor_pgdir.tline.get_last_record_lsn() == inmem_timeline.get_ancestor_lsn() {
+                    ancestor_logical_size = Some(ancestor_pgdir.get_current_logical_size());
+                    debug!("Branching: copy ancestor logical size");
+                } else {
+                    debug!(
+                        "Brancing: ancestor LSN={}, branch LSN={}",
```
aome510, line 141:
```suggestion
                        "Branching: ancestor LSN={}, branch LSN={}",
```

pageserver/src/tenant_mgr.rs
```diff
@@ -104,6 +105,64 @@ struct Tenant {
...
+                        ancestor_pgdir.tline.get_last_record_lsn(),
+                        inmem_timeline.get_ancestor_lsn()
+                    );
+                }
+            }
+        };
+        let pgdir = Arc::new(DatadirTimelineImpl::new(
```
aome510, line 148: nit: for consistency with the `load_local_timeline`, this should be named `page_tline` or `tline`. Not sure if `pgdir` fits the variable

pageserver/src/tenant_mgr.rs
```diff
@@ -104,6 +105,64 @@ struct Tenant {
...
+        &mut self,
+        timeline_id: ZTimelineId,
+    ) -> anyhow::Result<Arc<DatadirTimeline<LayeredRepository>>> {
+        let inmem_timeline = self.repo.get_timeline_load(timeline_id).with_context(|| {
+            format!("Inmem timeline {timeline_id} not found in tenant's repository")
+        })?;
+        let repartition_distance = self.repo.get_checkpoint_distance() / 10;
+        let mut ancestor_logical_size: Option<usize> = None;
+        if let Some(ancestor_id) = inmem_timeline.get_ancestor_timeline_id() {
+            if let Ok(ancestor_pgdir) = self.get_local_timeline_with_load(ancestor_id) {
```
aome510, line 135: nit: same with `pgdir`, maybe this should be changed to `ancestor_tline`

pageserver/src/tenant_mgr.rs
knizhnik: Ehhh... This is really mess with all this different timelines which are not actually timelines (doesn't implement `Timeline` trait. So `page_tline`seems to be very confusing (at least for me): it is not strictly a timeline and why "page"?

pageserver/src/tenant_mgr.rs
knizhnik: Once again - there are two different things, ancestor timeline 'LayeredTime' and ancestor `DatadirTimelineImpl`. Just what to somehow distinguish them by variable name. May be bad idea, especially in case of using IDe which shows variable types.

knizhnik:
> Just tried to run this patch with new `test_branch_creation_heavy_write` and `test_branch_creation_many`. The result is slower than/equal to what we have in `main`. This seems to be because we copy the logical size **on load** [1] not **on branch creation** [2]:
> 
> [1]:
> 
> ```rust
> if ancestor_pgdir.tline.get_last_record_lsn() == inmem_timeline.get_ancestor_lsn() {
>      ancestor_logical_size = Some(ancestor_pgdir.get_current_logical_size());
>      debug!("Branching: copy ancestor logical size");
>  }
> ```
> 
> [2]: (in `branch_timeline` function)
> 
> ```rust
> // If no start LSN is specified, we branch the new timeline from the source timeline's last record LSN
> let start_lsn = start_lsn.unwrap_or_else(|| {
>     let lsn = src_timeline.get_last_record_lsn();
>     info!("branching timeline {dst} from timeline {src} at last record LSN: {lsn}");
>     lsn
> });
> ```
> 
> The if statement in [1] is mostly false because in the test, upon loading the child, the ancestor branch is being updated. IMHO, it's better to move the logical size copy step to **on branch creation**. With that, we can make sure that the logical size is in sync and use that to construct the child branch.
> 
> That said, the loading speedup for child branch should be an improvement. The question is how much?

I do not understand how this patch may slowdown something. It just allows to eliminate some non-incremental calculations of logical database size. Even if check [1] is always false, then... we do exactly the same as main branch. The only added overhead is check itself, for which we need to locate and load ancestor timeline. Can this extra timeline lookup significantly slowdown tests? If so, then there is something wrong with timeline lookup and loading...

aome510:
> I do not understand how this patch may slowdown something. It just allows to eliminate some non-incremental calculations of logical database size. Even if check [1] is always false, then... we do exactly the same as main branch. The only added overhead is check itself, for which we need to locate and load ancestor timeline. Can this extra timeline lookup significantly slowdown tests? If so, then there is something wrong with timeline lookup and loading..

You're right. Parameters of the test are initialized randomly, so maybe that's why there is a difference. I'll need to measure the performance more carefully.

aome510:
@knizhnik as https://github.com/neondatabase/neon/pull/2139 is merged, should we close this one?

knizhnik:
Replaced by #2139

PR #2126, opened 2022-07-19
Add timeline physical size tracking

aome510:
Ref #1902.

- Track the layered timeline's `physical_size` using `pageserver_current_physical_size` metric when updating the layer map.
- Report the local timeline's `physical_size` in timeline GET APIs.
- Add `include-non-incremental-physical-size` URL flag to also report the local timeline's `physical_size_non_incremental` (similar to `logical_size_non_incremental`)
- Add a `UIntGaugeVec` and `UIntGauge` to represent `u64` prometheus metrics


pageserver/src/layered_repository.rs
```diff
@@ -2121,6 +2139,8 @@ impl LayeredTimeline {
 
         let mut layers = self.layers.write().unwrap();
         for l in image_layers {
+            self.physical_size
```
aome510, line 2142: I wonder if we should update the `PERSISTENT_BYTES_WRITTEN` metric here when creating an image layer similar to when we create a L0 delta layer

pageserver/src/layered_repository.rs
```diff
@@ -2379,19 +2399,28 @@ impl LayeredTimeline {
...
             layers.insert_historic(Arc::new(l));
         }
 
         // Now that we have reshuffled the data to set of new delta layers, we can
         // delete the old ones
         let mut layer_paths_do_delete = HashSet::with_capacity(deltas_to_compact.len());
-        for l in &deltas_to_compact {
-            l.delete()?;
+        drop(all_keys_iter);
+        for l in deltas_to_compact {
```
aome510, line 2416: Previously, `deltas_to_compact` needs to be borrowed here because `all_keys_iter` still exists. I'm not sure why Rust doesn't detect that `all_keys_iter` is not referred till the end the function to allow to move `deltas_to_compact`, so I add an explicit drop here.

@knizhnik can you review this change?

knizhnik: > I'm not sure why Rust doesn't detect that all_keys_iter is not referred till the end the function to allow to move deltas_to_compact,
Most likely it is because `all_keys_iter` iterator has destructor which according to the specification should be invoced at the end of the scope. And running the destructor is considered a use ‒ obviously the last one.
For example the following fragment of code is also not compiled:

```
let mut data = vec![1, 2, 3];
let x = &data[0];
println!("{}", x);
// This is OK, x is no longer needed
data.push(4);

```

So your change is ok, 
But I just wonder why new version is better than old one?

aome510: > But I just wonder why new version is better than old one?

`deltas_to_compact` is moved, so we don't need to clone in `layers.remove_historic(l.clone());`. I know `Arc` clone is not that expensive, but it's still nice not to have that.

LizardWizzard (commented):
Looks good to me. May worth to turn physical size to a metric as well so we can set up alerts on it

Another point, for logical size we have option to calculate it non incrementally to check whether the incremental diverges from the true one. What do you think, do we need it here as well?

Thought a bit more about that, we have a bit confusing situation, because physical size does not take into account ancestor branches, but physical size does. It is more flexible though, so you can construct the needed value by adding to it, but worth documenting I think. Probably a good idea to add a test for that

pageserver/src/layered_repository.rs
```diff
@@ -1099,6 +1099,9 @@ pub struct LayeredTimeline {
     // It can be unified with latest_gc_cutoff_lsn under some "first_valid_lsn",
     // though lets keep them both for better error visibility.
     initdb_lsn: Lsn,
+
+    /// Current physical size of the timeline at the lastest LSN
+    physical_size: AtomicU64,
```
LizardWizzard, line 1104: Worth noting which kind of files we actually take into account. E g we do not count temp download files, metadata, etc

test_runner/batch_others/test_timeline_size.py
```diff
@@ -176,3 +178,116 @@ def test_timeline_size_quota(neon_env_builder: NeonEnvBuilder):
...
+    env.pageserver.safe_psql(f"do_gc {env.initial_tenant.hex} {new_timeline_id.hex} 0")
+    assert_physical_size(env, env.initial_tenant, new_timeline_id)
+
+
+def assert_physical_size(env: NeonEnv, tenant_id: UUID, timeline_id: UUID):
+    """Check the current physical size returned from timeline API
+    matches the total physical size of the timeline on disk"""
+    client = env.pageserver.http_client()
+    res = assert_timeline_local(client, tenant_id, timeline_id)
+    timeline_path = f"{env.repo_dir}/tenants/{tenant_id.hex}/timelines/{timeline_id.hex}/"
```
LizardWizzard, line 281: Can be turned into pathlib.Path so in get_timeline_dir_size you can call methods on path objects directly instead of using functions from os module

test_runner/batch_others/test_timeline_size.py
```diff
@@ -176,3 +178,116 @@ def test_timeline_size_quota(neon_env_builder: NeonEnvBuilder):
...
+    assert res["local"]["current_physical_size"] == get_timeline_dir_size(timeline_path)
+
+
+def get_timeline_dir_size(path: str) -> int:
+    """Get the timeline directory's total size, which only counts the layer files' size."""
+    sz = 0
+    for f in os.listdir(path):
+        f_path = os.path.join(path, f)
+        # "simple" check if a file is a layer file
+        if os.path.isfile(f_path) and (not f.endswith('temp')) and ("__" in f):
```
LizardWizzard, line 291: Should we explicitly specify which kind of files we ignore? Or other way around specify that we count only layer files :) E g s3 download code creates temporary files with extension `temp_download` and it will go through this filter. Maybe entirely avoid files with extensions?

pageserver/src/layered_repository.rs
aome510: Yeah, this variable is constructed based on the files in `layer_map`. So it won't take into account temp file, metadata.

test_runner/batch_others/test_timeline_size.py
aome510: Yeah, we should. The current version is good enough for the new tests but I guess it's still better to be more correct on this, so future tests can use this function.

aome510:
> Another point, for logical size we have option to calculate it non incrementally to check whether the incremental diverges from the true one. What do you think, do we need it here as well?

You mean the option when making an API request right? If yes, I guess we may need one for the same purpose as logical size.

LizardWizzard:
> guess we may need one for the same purpose as that of logical size.

Yeah, thats exactly what I meant :)

test_runner/batch_others/test_timeline_size.py
aome510: Updated in https://github.com/neondatabase/neon/pull/2126/commits/2cdc14709411f1eb077426c2be596ac7ce4d914c. Now it filters by parsing the delta files, similar to what we do when loading the layer map.

pageserver/src/layered_repository.rs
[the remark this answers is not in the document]
aome510: Updated in https://github.com/neondatabase/neon/pull/2126/commits/52dabf812e5fc0bf9342687b85b03ee9c1eb0e28

test_runner/fixtures/utils.py
```diff
@@ -89,3 +90,40 @@ def get_dir_size(path: str) -> int:
                 pass  # file could be concurrently removed
 
     return totalbytes
+
+
+def get_timeline_dir_size(path: pathlib.Path) -> int:
+    """Get the timeline directory's total size, which only counts the layer files' size."""
+    sz = 0
+    for f_name in os.listdir(path):
```
LizardWizzard, line 98: nit: Path has handy `iterdir` method that yields directly Path instances so there is no need to construct full path with directory manually.

So this is the full snippet:
```python
    for dir_entry in path.iterdir():
        with contextlib.suppress(Exception):
            # file is an image layer
            _ = parse_image_layer(dir_entry.name)
            sz += dir_entry.stat().st_size
            continue
        
        with contextlib.suppress(Exception):
            # file is a delta layer
            _ = parse_delta_layer(dir_entry.name)
            sz += dir_entry.stat().st_size
            continue
```
To me it looks a bit cleaner with suppress but feel free to keep try except

LizardWizzard (commented):
Thanks! I left few minor comments. What do you think about the option to use custom UIntGauge wrapper? This way we can keep size in u64 instead of i64

aome510:
> What do you think about the option to use custom UIntGauge wrapper? This way we can keep size in u64 instead of i64

LGTM. I cherry picked your commit and also address your comments. Lmk if I miss anything

LizardWizzard (approved):
Thanks!

PR #2128, opened 2022-07-20
Gen2 GH runner

zoete:
https://github.com/neondatabase/neon/issues/2158


bayandin:
It looks like `Unsupported instrumentation profile format version` is caused by a mismatch in llvm versions — rust uses llvm-13 ([since 1.56](https://github.com/rust-lang/rust/pull/87570)), but we have llvm-10 in our image (should be fixed by https://github.com/neondatabase/build/pull/7)

LizardWizzard:
Is it part of the https://github.com/neondatabase/neon/issues/2158? So I can pick it up, from there, right?

bayandin:
> Is it part of the #2158? So I can pick it up, from there, right?

I work on this PR in the background (I need it for https://github.com/neondatabase/neon/pull/2071, so I do changes here from time to time, hopefully, it will be done today/tomorrow).

Sorry, it should have been assigned to me.


But I haven't checked/migrated any deploy-related jobs, so you can pick them @LizardWizzard

Dockerfile.compute-node
```diff
@@ -0,0 +1,60 @@
+ARG COMPUTE_TOOLS_TAG=latest
```
zoete, line 1: Will be used by follow-up merge: https://github.com/neondatabase/cloud/issues/1963 - currently inert.

.github/workflows/build_and_test.yml
```diff
@@ -7,10 +7,6 @@ on:
       - release
   pull_request:
 
-defaults:
-  run:
-    shell: bash -euxo pipefail {0}
-
```
bayandin, line 13: Is it removed for any particular reason?

.github/workflows/build_and_test.yml
```diff
@@ -23,7 +19,7 @@ env:
 jobs:
   build-neon:
     runs-on: dev
-    container: 369495373322.dkr.ecr.eu-central-1.amazonaws.com/rustlegacy:2746987948
+    container: 369495373322.dkr.ecr.eu-central-1.amazonaws.com/rust:latest
```
bayandin, line 22: Let's use a specific tag, not just the latest one.
Also, we can move the image tag to the workflow env and use it `${{ env.RUST_TAG }}` to not to repeat it for each job

.github/workflows/build_and_test.yml
```diff
@@ -369,150 +375,166 @@ jobs:
...
     outputs:
-      build-tag: ${{steps.build-tag.outputs.tag}}
+      value: ${{ steps.dockerfile-check.outputs.any_changed }}
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+
+      - name: Get specific changed files
+        id: dockerfile-check
+        uses: tj-actions/changed-files@v24.1
```
bayandin, line 391: Let's use a commit hash instead of a tag

.github/workflows/build_and_test.yml
```diff
@@ -369,150 +375,166 @@ jobs:
...
+    # force building for now
+    if: needs.dockerfile-check.outputs.value != 'true'
+    runs-on: dev
+    needs: [ dockerfile-check ]
+    container: gcr.io/kaniko-project/executor:debug
+    environment: dev
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
```
bayandin, line 409: Do we use `@v1` for a particular reason? I think on other jobs we use `@v3`?

.github/workflows/build_and_test.yml
```diff
@@ -369,150 +375,166 @@ jobs:
...
+    runs-on: dev
+    needs: [ promote-images ]
+    container: golang:bullseye
+    environment: dev
+
+    steps:
+      - name: Install Crane & ECR helper
+        run: |
+          go install github.com/google/go-containerregistry/cmd/crane@main
+          go install github.com/awslabs/amazon-ecr-credential-helper/ecr-login/cli/docker-credential-ecr-login@latest
```
bayandin, line 487: Can we pin some versions/commits here as well?

.github/workflows/build_and_test.yml
```diff
@@ -556,8 +581,14 @@ jobs:
           submodules: true
           fetch-depth: 0
 
+      - name: Setup python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
```
bayandin, line 587: Can we use `rust` image here which comes with python?

Dockerfile.compute-node
```diff
@@ -0,0 +1,60 @@
...
+
+# Build PostGIS from upstream the upstream PostGIS mirror. PostGIS compiles against neon postgres sources without changes.
+# Perhaps we could even use the upstream binaries, compiled against vanilla Postgres, but it would require some
+# investigation to check that it works, and also keeps working in the future. So for now, we compile our own binaries.
+FROM build-deps AS postgis-build
+COPY --from=pg-build /usr/local/pgsql/ /usr/local/pgsql/
+RUN apt update && \
+    apt install -y gdal-bin libgdal-dev libprotobuf-c-dev protobuf-c-compiler
+
+RUN git clone -b stable-3.2 https://github.com/postgis/postgis.git postgis && \
```
bayandin, line 27: I'd say `stable-3.2` should be a build arg, and probably it should be a tag. 
And if we don't need to get it exactly from git, we can download https://download.osgeo.org/postgis/source/postgis-3.2.2.tar.gz, then `autogen.sh` won't be needed

.github/workflows/build_and_test.yml
```diff
@@ -369,150 +375,166 @@ jobs:
...
+        id: dockerfile-check
+        uses: tj-actions/changed-files@v24.1
+        with:
+          files: |
+            Dockerfile
+            Dockerfile.compute-tools
+            ./vendor/postgres/Dockerfile
+
+  build-neon-image:
+    name: 'kaniko-neon'
```
bayandin, line 399: Do we really need to set `name` for each step?

.github/workflows/build_and_test.yml
zoete: Yes as kaniko & aws-cli don't have bash, I the run-shell for the jobs that need it

.github/workflows/build_and_test.yml
zoete: For v3 you need node, which didn't work for this image

.github/workflows/build_and_test.yml
zoete: Would conflict with lifecycle rule on ECR: Only keep 32 non latest images - latest will be behind an approval step on the build repository.

bayandin: > Would conflict with lifecycle rule on ECR: Only keep 32 non latest images

Wow, this is weird from them. Should we consider using other images storages? [GitHub container registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry), for example?

I'd expect running the same code to produce the same output (or about the same), which is nearly impossible to do with using `latest` tags. It also makes it harder to debug CI failures (when images switch occasionally)

.github/workflows/build_and_test.yml
bayandin: Ahhh, okay, worth adding a comment then.
Also we should use a tag for this image as well

.github/workflows/build_and_test.yml
zoete: No this is my added ECR rule, I guess I can put it to 128 non latest images but that would invalidate older pipelines, would that be preferred?

.github/workflows/build_and_test.yml
zoete: Both projects are strict on forward & backwards compatibility, similar to aws-cli I won't pin these.

.github/workflows/build_and_test.yml
zoete: To be done after this is merged as that would be a larger refactoring of the deployment script

.github/workflows/build_and_test.yml
bayandin: Is it 128 across in total or per image?

.github/workflows/build_and_test.yml
zoete: Kaniko is strict on forward & backwards compatibility, similar to aws-cli I won't pin this.

.github/workflows/build_and_test.yml
zoete: Per ECR repo, so per rust \ rustlegacy \ etc. Can go up much higher, but cost will increase.

.github/workflows/build_and_test.yml
bayandin: > Kaniko is strict on forward & backwards compatibility, similar to aws-cli I won't pin this.

Ok, can we pin a major version at least? Using just `debug` makes it unclear what is running (and why `debug`?)

zoete: Debug is needed else there is no shell to configure the ECR login, pinned on most recent release

.github/workflows/build_and_test.yml
bayandin: > Both projects are strict on forward & backwards compatibility, similar to aws-cli I won't pin these.

Does using the latest version of these tools benefit us somehow? I don't see advantages in it but can see potential problems.

Btw, the latest `golang` from the image can be incompatible with these tools.

.github/workflows/build_and_test.yml
bayandin: It depends on cost, but a stable build infrastructure is probably worth it

.github/workflows/build_and_test.yml
zoete: Updated with pins

.github/workflows/build_and_test.yml
zoete: What about this approach: https://github.com/neondatabase/build/pull/12 - then we have predictability but not inefficient levels of storage used

bayandin: Could you please explain what it does? 😄

zoete: Check if any images have tag "pinned" - if so, keep at least 3. - continue with rule #3. As posted in that pull, not sure what makes sense for workflow to "tag" an image as such, as it is infrequent.

zoete: Added the pinned tag, which for now is set manual. Will add this workflow to build readme

bayandin: Hmm, I think with `pinned` we'll have the same issues as with `latest` — workflow will pick up new images implicitly without any changes in the workflow itself.
I'm for having a new tag for each image version and letting the old one expire

zoete: Yes, however comparing the balancing act of keeping enough images (inefficient) that are almost all identical so we can have one specific numbered image available, say `rust:2746987948` - versus a workflow that will highlight what needs to be done before someone promotes an image to become the newest `rust:pinned` I would say the later one offers more control.

bayandin: Ok, let's switch it back to the `latest` and discuss it separately (to not block this PR further)

zoete: I got it on pinned now, which is outside of any automated workflow.

bayandin (approved):
If you have tested that the deploy part works as well, I'm 👍

.github/workflows/build_and_test.yml
```diff
@@ -369,150 +375,160 @@ jobs:
...
+    needs: [ dockerfile-check ]
+    container: gcr.io/kaniko-project/executor:v1.9.0-debug
+    environment: dev
 
-      - name: Build neon Docker image
-        uses: docker/build-push-action@v2
-        with:
-          context: .
-          build-args: |
-            GIT_VERSION="${{github.sha}}"
```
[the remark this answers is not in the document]
bayandin: > Curious, how these access keys are now passed? I see that current builds have working cachepot, but don't see any env vars or build args like these anymore.

Access keys are "bundled" with runners we use (not sure I use the right terminology).

> We seem to miss GIT_VERSION since this PR, I'm trying to add it back.

Let me know if you need any help with it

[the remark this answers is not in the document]
SomeoneToIgnore: Thank you, I have been explained that IAM is set-up there, so that answers the first part.

And I've fixed the versions in https://github.com/neondatabase/neon/pull/2578 so we're good for now.

I was thinking to add some extra step after kaniko build to ensure that `pageserver --version` does not contain a default Docker version, but we need `docker run` for that.

PR #2135, opened 2022-07-21
performance/test_dup_key: cleanup tables before test

bayandin:
For remote env, we use a preexisting database.

Ref https://github.com/neondatabase/neon/runs/7442734950?check_suite_focus=true
```
>                   cur.execute("create table t (i integer, filler text);")
E                   psycopg2.errors.DuplicateTable: relation "t" already exists
```

### DIFF
--- a/test_runner/performance/test_dup_key.py
+++ b/test_runner/performance/test_dup_key.py
@@ -17,6 +17,8 @@ def test_dup_key(env: PgCompare):
 
     with closing(env.pg.connect()) as conn:
         with conn.cursor() as cur:
+            cur.execute('drop table if exists t, f;')
+
             cur.execute("SET synchronous_commit=off")
             cur.execute("SET statement_timeout=0")
 


bayandin:
> > For remote env, we use a preexisting database.
> 
> That is surprising, I would not have expected that. Is that intentional?

Just to avoid confusion "remove env" == marked with `remote_cluster`. 
I think this is intentional, at least for now:
- https://github.com/neondatabase/neon/tree/main/test_runner/performance#remote-tests
- https://github.com/neondatabase/neon/blob/main/.github/workflows/benchmarking.yml#L56-L59

LizardWizzard (approved):
Yeah, remote tests were added before the scram thing, before the proper api for managing databases was added. AFAIK these problems no longer exist so should be possible to recreate database for every test run.

Another point, if we reuse database it can help to detect bugs, because this way the cluster is lives longer so can trigger more edge cases, data bloat etc etc

PR #2164, opened 2022-07-26
Switch staging storage to dedicated etcd

SergeyMelnikov:


.github/ansible/staging.hosts
```diff
@@ -17,4 +17,4 @@ env_name = us-stage
 console_mgmt_base_url = http://console-staging.local
 bucket_name           = zenith-staging-storage-us-east-1
 bucket_region         = us-east-1
-etcd_endpoints        = etcd-staging.local:2379
+etcd_endpoints        = zenith-us-stage-etcd.local
```
cicdteam, line 20: is it ok do not specify port there ? our storage will use 2379 as default ?

cicdteam (approved):

PR #2172, opened 2022-07-27
Fix bug when import large (>1GB) relations

### LINKED ISSUE #2097: import basebackup bug with large (>1GB) relations
A large relation will have multiple segment files ("12345", "12345.1", "12345.2", etc). After importing one of those files we flush the datadir modification, but don't call `writer.flush_write`, so we don't advance `last_record_lsn`. When we import the next file, we try to read the current relation size to make sure not to decrease it, in case files were imported out of order. Because we flushed, the relation size read will read from the actual timeline, [at](https://github.com/neondatabase/neon/blob/f88fe021-import-patch/pageserver/src/pgdatadir_mapping.rs#L986) `last_record_lsn`. This is lower than the lsn the relation size was written at, so we won't find the relation size, and we will panic.

One solution is to read at `self.lsn` instead of `last_record_lsn`. Another is to use `writer.flush_write`. Another is to skip relation size keys when flushing. Not sure which one's the best.

aome510:
Resolves #2097 

- use timeline modification's `lsn` and timeline's `last_record_lsn` to determine the corresponding LSN to query data in `DatadirModification::get`
- update `test_import_from_pageserver`. Split the test into 2 variants: `small` and `multisegment`. 
  + `small` is the old test
  + `multisegment` is to simulate #2097 by using a larger number of inserted rows to create multiple segment files of a relation. `multisegment` is configured to only run with a `release` build


aome510:
> update `test_import_from_pageserver` to use a large number of rows (**when in release build**) to simulate the scenario described in the linked issue

Can use `BUILD_TYPE=release NEON_BIN=(pwd)/target/release ./scripts/pytest -k test_import_from_pageserver` to run the test to confirm that it fails without this patch

pageserver/src/pgdatadir_mapping.rs
```diff
@@ -919,8 +931,8 @@ impl<'a, T: DatadirTimeline> DatadirModification<'a, T> {
                 bail!("unexpected pending WAL record");
             }
         } else {
-            let last_lsn = self.tline.get_last_record_lsn();
-            self.tline.get(key, last_lsn)
+            let lsn = Lsn::max(self.tline.get_last_record_lsn(), self.last_flush_lsn);
+            self.tline.get(key, lsn)
```
bojanserafimov, line 935: Does it work if we read at `self.lsn`? Seems simpler unless I'm missing something

test_runner/batch_others/test_import.py
```diff
@@ -105,19 +105,25 @@ def import_tar(base, wal):
...
+def test_import_from_pageserver(test_output_dir,
+                                pg_bin,
+                                vanilla_pg,
+                                neon_env_builder: NeonEnvBuilder):
+    built_type = os.environ.get("BUILD_TYPE")
+    if built_type is not None and built_type == "release":
+        log.info("Detected release build type, use a larger number of rows...")
+        num_rows = 30000000
+    else:
+        num_rows = 3000
```
bojanserafimov, line 117: nit: These are now different tests, testing different edge cases. I'd rather be explicit about that. Maybe parameterize the test into two variants: `small` and `multisegment`, and skip `multisegment` in debug with `pytest.mark.skipif`?

bojanserafimov (commented):
Thanks for the patch!

pageserver/src/pgdatadir_mapping.rs
aome510: Yes, it should work. Prior to this PR, `self.lsn` was removed from the struct but added back in https://github.com/neondatabase/neon/pull/2094.

aome510: updated in https://github.com/neondatabase/neon/pull/2172/commits/b90832dd981be98492f6779d9a1be6a0edc31694

test_runner/batch_others/test_import.py
aome510: updated in https://github.com/neondatabase/neon/pull/2172/commits/fa5806a5f44f1a3076eb5a7ebf7121f5b529f576

test_runner/batch_others/test_import.py
```diff
@@ -130,12 +136,16 @@ def test_import_from_pageserver(test_output_dir, pg_bin, vanilla_pg, neon_env_bu
             lsn = cur.fetchone()[0]
             log.info(f"start_backup_lsn = {lsn}")
 
+    logical_size = env.pageserver.http_client().timeline_detail(
+        env.initial_tenant, timeline)['local']['current_logical_size']
+    log.info(f"timeline logical size = {logical_size / (1024 ** 2)}MB")
```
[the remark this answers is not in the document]
aome510: updated in https://github.com/neondatabase/neon/pull/2172/commits/76ecdb4999d439c90b95dcb1f346a4a0cf66d720

bojanserafimov (approved):
Looks good

bojanserafimov (approved):

PR #2173, opened 2022-07-27
Add current physical size to tenant status endpoint

aome510:
Ref #1902

### DIFF
--- a/pageserver/src/http/routes.rs
+++ b/pageserver/src/http/routes.rs
@@ -444,23 +444,31 @@ async fn tenant_status(request: Request<Body>) -> Result<Response<Body>, ApiErro
             false
         });
 
-    // Retrieves local timelines of the tenant
-    let local_timeline_infos = tokio::task::spawn_blocking(move || {
+    let current_physical_size = match tokio::task::spawn_blocking(move || {
         crate::timelines::get_local_timelines(tenant_id, false, false)
     })
     .await
-    .map_err(ApiError::from_err)??;
-
-    let current_physical_size = local_timeline_infos
-        .into_iter()
-        .fold(0, |acc, x| acc + x.1.current_physical_size.unwrap());
+    .map_err(ApiError::from_err)?
+    {
+        Err(err) => {
+            // Getting local timelines can fail when no local repo is on disk (e.g, during tenant relocation task).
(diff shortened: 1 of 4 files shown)


pageserver/src/http/routes.rs
```diff
@@ -438,14 +438,29 @@ async fn tenant_status(request: Request<Body>) -> Result<Response<Body>, ApiErro
     let index_accessor = remote_index.read().await;
     let has_in_progress_downloads = index_accessor
         .tenant_entry(&tenant_id)
-        .ok_or_else(|| ApiError::NotFound("Tenant not found in remote index".to_string()))?
```
aome510, line 441: I don't think this should return an error here. A newly created tenant may not be attached to a remote index. Got the error whenever I tried to call `tenant_status` in test.

Can you confirm this behaviour @LizardWizzard ?

pageserver/src/tenant_mgr.rs
```diff
@@ -508,6 +508,7 @@ pub struct TenantInfo {
     #[serde_as(as = "DisplayFromStr")]
     pub id: ZTenantId,
     pub state: Option<TenantState>,
+    pub current_physical_size: Option<u64>, // physical size is only included in `tenant_status` endpoint
```
LizardWizzard, line 511: Please add new field to the swagger spec too (unfortunately this is manual work for now)

pageserver/src/http/routes.rs
```diff
@@ -444,23 +444,31 @@ async fn tenant_status(request: Request<Body>) -> Result<Response<Body>, ApiErro
...
     .await
-    .map_err(ApiError::from_err)??;
-
-    let current_physical_size = local_timeline_infos
-        .into_iter()
-        .fold(0, |acc, x| acc + x.1.current_physical_size.unwrap());
+    .map_err(ApiError::from_err)?
+    {
+        Err(err) => {
+            // Getting local timelines can fail when no local repo is on disk (e.g, during tenant relocation task).
```
LizardWizzard, line 454: nit:

> (e.g, during tenant relocation task) 

When tenant data is being downloaded to be precise

LizardWizzard (approved):
LGTM! Thanks!

PR #2194, opened 2022-07-31
Deactivate sk timelines when no compute activity

petuhovskiy:
In-memory timelines will be deactivated after there will be no compute connections for more than 10 minutes. Deactivation here will set `active = false` in `SharedState` struct and that will stop pushing info about this timeline to etcd.

Timeline deactivation happens in `update_status`, which is called on every etcd message from the peer.

### DIFF
--- a/safekeeper/src/timeline.rs
+++ b/safekeeper/src/timeline.rs
@@ -15,6 +15,7 @@ use std::collections::{HashMap, HashSet};
 use std::fs::{self};
 
 use std::sync::{Arc, Mutex, MutexGuard};
+use std::time::Instant;
 
 use tokio::sync::mpsc::Sender;
 use tracing::*;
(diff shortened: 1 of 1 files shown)


petuhovskiy:
Tested locally with a test that counts active timelines:
```
def test_plenty_timelines(neon_env_builder: NeonEnvBuilder):
    neon_env_builder.num_safekeepers = 3

    # start postgres on each timeline
    pgs = []
    branch_idx = 0

    def create_branch():
        nonlocal branch_idx
        branch_name = f"abc_{branch_idx}"
        branch_idx += 1
        new_timeline_id = env.neon_cli.create_branch(branch_name)
        pg = env.postgres.create_start(branch_name)
        pg.safe_psql("SELECT 1")
        pg.stop()
        pgs.append(pg)

    def report_metrics():
        # collect metrics from other servers
        for sk in env.safekeepers:
            all_metrics = parse_metrics(sk.http_client().get_metrics_str())
            active_tlis = sum(map(lambda f: int(f.value), all_metrics.query_all("safekeeper_timeline_active", {})))

            log.info(f"Safekeeper {sk.id} have {active_tlis} active timelines")

    for i in range(100):
        create_branch()

    report_metrics()

    # sleep until timelines are deactivated
    time.sleep(50)

    while True:
        for i in range(10):
            create_branch()

        report_metrics()
        time.sleep(10)

```

arssher:
Superseded by [26746](https://github.com/neondatabase/neon/pull/2228)

PR #2195, opened 2022-07-31
refactor: replace lazy-static with once-cell

### LINKED ISSUE #1147: Replace `lazy_static` with `once_cell`
According to @funbringer  and @LizardWizzard, It's a more modern and elegant way to handle globals which doesn't involve macros.

ansrivas:
- Replacing all the occurrences of lazy-static with `once-cell::sync::Lazy`
- fixes #1147

**NOTE**: `pageserver/src/storage_sync.rs:202:1`  `FATAL_TASK_FAILURES` isn't being used anywhere (as clippy reported)


ansrivas:
Hey folks, clippy reported failure, as noted in the comment above, should I remove it?


image

bojanserafimov:
> Hey folks, clippy reported failure, as noted in the comment above, should I remove it?

cc @LizardWizzard Are we missing an increment somewhere? The metric was added with [this PR](https://github.com/neondatabase/neon/commit/7738254f83c86e46795b34db834d18af97197d8d) and was never incremented. Clippy never complained because probably the macro made it look used.

I think it's fine to remove so we can move on with this PR. No metric is better than a misleading metric. If we need it we can add it back with proper increments.

bojanserafimov (approved):
Thanks for the patch! Looks good. Let's fix the clippy failure and merge.

LizardWizzard:
> Are we missing an increment somewhere?

I think the increment was removed during some of the refactorings, should be ok to remove here and bring it back separately if needed

bojanserafimov (approved):
Looks good. Tests fail because they rely on infra you have no access to. Sorry about that, we'll work around it somehow to merge this PR and solve the issue for the next external contributor. Thanks for contributing.

Issue #1863, opened 2022-06-02
How do we accept contributions and run tests from forked repositories?


yeputons:
We have quite a bunch of checks which use some secret tokens:

* `coverage-report` [uploads coverage reports to our repository](https://github.com/neondatabase/neon/blob/de7eda2dc6a6dbad3c3ec96e71673c5a8a48bb79/.circleci/config.yml#L406)
* `remote-ci-trigger` [triggers end-to-end tests with our cloud tests](https://github.com/neondatabase/neon/blob/de7eda2dc6a6dbad3c3ec96e71673c5a8a48bb79/.circleci/config.yml#L724)

If an external contributor sends us a pull request from their forked repository, these checks cannot succeed. CI typically does not share secrets with such pull requests and [for a good reason](https://circleci.com/docs/2.0/oss/#pass-secrets-to-builds-from-forked-pull-requests): otherwise, a malicious actor could easily extract them by crafting a special script, altering `config.yml`, and sending us a PR, obtaining secret credentials minutes after.

However, we still need to accept contributions from external users, and probably run them against our internal tests as well before merging.

Any ideas on how to do it? I've seen other projects having a special bot that looks for maintainer's comments like `@helpfulbot please test` and triggers CI jobs with secrets.

yeputons:
Note: this may get partially resolved with #1414

LizardWizzard:
Still has some problems with triggering e2e tests: https://github.com/neondatabase/neon/runs/7610981072?check_suite_focus=true

PR #2196, opened 2022-07-31
docs: added poetry related step in tests

ansrivas:
Added the link to the dependencies which should be installed
before running the tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,8 @@ postgres=# select * from t;
 
 ## Running tests
 
+Ensure your dependencies are installed as described [here](https://github.com/neondatabase/neon#dependency-installation-notes).
+
 ```sh
 git clone --recursive https://github.com/neondatabase/neon.git
 make # builds also postgres and installs it to ./tmp_install


README.md
```diff
@@ -205,8 +205,12 @@ postgres=# select * from t;
 ## Running tests
 
 ```sh
+# install poetry if not available on the system.
+# pip install poetry
+
 git clone --recursive https://github.com/neondatabase/neon.git
 make # builds also postgres and installs it to ./tmp_install
+poetry install # install the test dependencies
```
LizardWizzard, line 213: It shouldnt work without `--no-root` unless newer poetry versions deal with that automatically. There is shortcut script for that: [scripts/pysync](https://github.com/neondatabase/neon/blob/main/scripts/pysync)

LizardWizzard: Maybe worth linking this paragraph from earlier part of the readme https://github.com/neondatabase/neon#dependency-installation-notes

ansrivas: aha, thanks for pointing it out. 
Somehow I missed the dependency installation notes. I will link it in the PR.

LizardWizzard (approved):
Thanks!

LizardWizzard:
Some checks have failed because they need to trigger some actions in private repos, and unfortunately this doesnt work for contributors outside neon. The progress of solving this is tracked here https://github.com/neondatabase/neon/issues/1863

Thanks!

PR #2201, opened 2022-08-01
run benchmarks conditionally, on main or if run-benchmarks label is set

LizardWizzard:
Benchmarking job takes around 35 minutes and for significant part of the PR's is not needed so this PR makes it configurable via run-benchmarks label

- [x] check it does not trigger without the label
- [x] check triggers with label
- [ ] check triggers on main


zoete (approved):
LGTM

Issue #2192, opened 2022-07-29
Refactor pageserver config handling


LizardWizzard:
Currently it has a few problems. First the distinction between init phase and normal operation. Init is needed to be run separately and that requires separate handling in docker-entrypoint.sh we have here https://github.com/neondatabase/neon/blob/main/docker-entrypoint.sh

This script requires updates to support new parameters. 

Options:
* Match safekeeper behavior, avoid config altogether, patth everything via command line or add env var support via clap feature https://docs.rs/clap/latest/clap/struct.Arg.html#method.env
* use pre specified config file passed via --config

cc @SomeoneToIgnore

PR #2206, opened 2022-08-01
support node id and remote storage params in docker_entrypoint.sh

LizardWizzard:


docker-entrypoint.sh
```diff
@@ -1,17 +1,21 @@
 #!/bin/sh
 set -eux
 
+pageserver_id_param="${NODE_ID:-10}"
```
petuhovskiy, line 4: That 10 looks weird, but let's keep it in case someone depends on it :)

It looks like no one used this docker image except console, because it didn't support some arguments.

petuhovskiy (approved):

docker-entrypoint.sh
ololobus: I think this is because in the console's docker-compose.yaml NODE_ID isn't set. We can add it there and remove here. Also we always run etcd now, so maybe we can remove condition on broker endpoints?

So the only condition left is presence of tenants dir, which is probably ok to put into `command` section in the docker-compose.yaml

LizardWizzard: 10 was chosen because there was a bug in regex that was handling only single digit node ids. 

This script is used in console docker-compose setup to run proxy as well as storage components so unless we can completely separate pageserver config from we need to keep these things optional. 

I think we need to refactor this part inside pageserver. See https://github.com/neondatabase/neon/issues/2192

LizardWizzard: So for now I would rather keep the script mostly as it is. Having tenant dir condition in line is a bit messy to me

Issue #2200, opened 2022-08-01
Safekeeper panicked while holding TIMELINES_STATE mutex


petuhovskiy:
After the panic TIMELINES_STATE mutex became poisoned and all subsequent timeline gets resulted in PoisonError. I wasn't able to find the panic itself, we probably should tweak our logger to print stacktrace for all panics.

The issue is probably connected to calling delete tenant endpoint, because we didn't see this issue before enabling that. There is a possible race condition in deleting a timeline, because it could be implicitly created concurrently with deletion, but it's not yet clear how it could cause panic and if it's indeed the issue.

The first PoisonError appeared on staging at `2022-07-31T13:19:35.490718Z`, the previous log before that was at `2022-07-31T13:19:13.663980Z`. I couldn't find logs in between, it's very strange, could they have been dropped?

arssher:
> could they have been dropped?

Yes, see lines like `2022-08-03 07:09:59 | Suppressed 18238 messages from safekeeper.service`. That was due to ill cluster trying to fetch in busyloop WAL segment which wasn't found, I fixed that. Let's see if this repeats now.

I also increased RateLimit to reduce chance of losing logs again:
https://github.com/neondatabase/cloud/pull/1970

arssher:
2022-08-04 02:25:02 | 2022-08-04T02:25:02.619537Z  INFO broker:{timeline=4a8dd72104349afe5e4e135c63ebf4d6}: timeline 53466ed253a1c866860c476e9bef1bbc restored thread 'broker thread' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', libs/postgres_ffi/src/xlog_utils.rs:399:41 stack backtrace:

This is the problem.

petuhovskiy:
More context:
```
2022-08-04T02:25:02.619317Z  INFO deleting all timelines for tenant 4a8dd72104349afe5e4e135c63ebf4d6
2022-08-04T02:25:02.619408Z  INFO broker:{timeline=4a8dd72104349afe5e4e135c63ebf4d6}: loading control file ./4a8dd72104349afe5e4e135c63ebf4d6/53466ed253a1c866860c476e9bef1bbc/safekeeper.control
2022-08-04T02:25:02.619537Z  INFO broker:{timeline=4a8dd72104349afe5e4e135c63ebf4d6}: timeline 53466ed253a1c866860c476e9bef1bbc restored
[...]

PR #2214, opened 2022-08-04
Check find_end_of_wal errors instead of unwrap

petuhovskiy:
Fixes https://github.com/neondatabase/neon/issues/2200

The issue was etcd message loading timeline from disk after deletion. On timeline restore (load) we called `find_end_of_wal` which panicked if there's no directory on disk. Panic is fixed by returning `Err(...)`.

### DIFF
--- a/libs/postgres_ffi/src/xlog_utils.rs
+++ b/libs/postgres_ffi/src/xlog_utils.rs
@@ -16,7 +16,7 @@ use crate::XLogRecord;
 use crate::XLOG_PAGE_MAGIC;
 
 use crate::pg_constants::WAL_SEGMENT_SIZE;
-use anyhow::{bail, ensure};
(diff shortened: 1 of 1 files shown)


arssher (approved):

petuhovskiy:
Let's merge this PR once it passes the tests, we better roll out deletion to production soon.

Note that there is also https://github.com/neondatabase/neon/pull/2066, which can solve the same issue by rewriting `find_end_of_wal`. Let's add a test for loading deleted timeline there, to ensure the issue will be fixed there as well.

arssher:
Failed REAL_S3 tests clearly don't have anything to do with this; and on the first glance failure reason is just slowness of s3. In safekeeper test segment failed to upload in 10s and I don't see any errors.

PR #2225, opened 2022-08-05
Add more buckets to pageserver latency metrics

bojanserafimov:
Our top bucket is currently non-empty, so we should add more buckets.

### DIFF
--- a/pageserver/src/layered_repository/timeline.rs
+++ b/pageserver/src/layered_repository/timeline.rs
@@ -67,7 +67,7 @@ fn get_buckets_for_critical_operations() -> Vec<f64> {
     // adding this bucket we'll never know.
     let mut buckets = vec![];
     for exp in -12..=4 {
-        buckets.push((10 as f64).powf(exp as f64 / 2.0))
+        buckets.push(10_f64.powf(exp as f64 / 2.0))
     }
     buckets
 }
--- a/pageserver/src/layered_repository/timeline.rs
+++ b/pageserver/src/layered_repository/timeline.rs
@@ -58,12 +58,30 @@ use crate::walredo::WalRedoManager;
 use crate::CheckpointConfig;
 use crate::{page_cache, storage_sync};
 
+/// Prometheus histogram buckets (in seconds) that capture the majority of
+/// latencies in the microsecond range but also extend far enough up to capture
+/// all latencies.
+fn get_buckets_for_critical_operations() -> Vec<f64> {
+    let mut buckets = vec![];
+    for exp in -12..=4 {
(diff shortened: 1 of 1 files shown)


pageserver/src/layered_repository/timeline.rs
```diff
@@ -58,12 +58,27 @@ use crate::walredo::WalRedoManager;
...
+fn get_buckets_for_critical_operations() -> Vec<f64> {
+    // Generate exponential buckets from 1 microsecond to ~5 minutes, sqrt(10) apart.
+    // Hopefully no critical operation will last 5 minutes. But without explicitly
+    // adding this bucket we'll never know.
+    let mut buckets = vec![];
+    for exp in -12..=4 {
+        buckets.push(10_f64.powf(exp as f64 / 2.0))
+    }
+    buckets
+}
```
ololobus, line 73: Maybe I miss something, but two problems I see:
- the buckets are significantly different now, they start from 1, 3, 10, 30, 100, 300 micro secs, then quickly grows to 1, 3, 10, 30, 100, 300 ms, then 1+ sec already, so it misses the most heat range for get page right now (100-150 micro secs), maybe it's ok
- iiuc, due to the math bucket boundaries will be too uneven, maybe do rounding till 1 significant figure? So it was e.g. exactly 3 micro sec instead of the current 3.162277660168379e-06

bojanserafimov: Just found a [builtin function](https://docs.rs/prometheus/latest/prometheus/fn.exponential_buckets.html) so I'll rewrite this

On your points:
- Good point, let's pick a smaller base. With 10^(1/4), it will look like `1, 1.7, 3.1, 5.6, 10, ...` when rounded. Maybe we should go even smaller. The reason I'm picking roots of 10 is that we also get the human readable `0.01, 0.1, 1, 10, ...` on the axis, but it's not a big deal if not
- The *current* buckets we have are uneven, and you can see hotspots that are just artifacts of the distribution. 1-2.5 ms is suspiciously thick. Here's how I'm thinking about it: Exponential buckets on the raw data points is equivalent to linear buckets on log(data points). So I'm subdividing the log of the result in prefectly equal buckets.

bojanserafimov: btw grafana can round the decimals on the y axis. IMO it's better to do it there, but lmk if I'm missing something

bojanserafimov: Decided against using the builtin. It's not any easier to use, and it generates numbers like 9.99999, which I'm only 50% sure will round well in grafana

ololobus (approved):

PR #2229, opened 2022-08-08
Report test results to Allure

bayandin:
This PR adds reporting tests to Allure (for now, it is enabled for batch_others/pg_regression tests only). This should help us debug failed tests on CI and track flaky tests.

I group tests by build type (ie release/debug). Also, we keep raw data, so it's possible to repopulate previous reports.

Links:
- Debug build test results: https://neon-github-public-dev.s3.amazonaws.com/reports/pr-2229/debug/latest/index.html
- Release build test results: https://neon-github-public-dev.s3.amazonaws.com/reports/pr-2229/release/latest/index.html

Some particular examples:
- [An example of a flaky test](https://neon-github-public-dev.s3.amazonaws.com/reports/pr-2229/debug/2860339707/index.html#suites/8f76013924fd6509310c270f125fad47/c96846a0e50843ae/history)
- [Failed test](https://neon-github-public-dev.s3.amazonaws.com/reports/pr-2229/debug/2856063987/index.html#testresult/73e9d8855e8ccc5d)
- [Getting logs from a test](https://neon-github-public-dev.s3.amazonaws.com/reports/pr-2229/debug/2860339707/index.html#suites/88419a57728fab056d570baf85366bef/d20f7a086170204e/): expand the "Tear down" section, expand `test_output_dir` — it will show a list of different logs produced by the test 
- [A run with reruns](https://neon-github-public-dev.s3.amazonaws.com/reports/pr-2229/release/2860339707/index.html)

Next ToDo:
- Add a link to the report to PR comment or as a commit status (like we attach cloud e2e test results)
- Find out if it is possible to get a link to a particular failed test (this should make it easier to verify)
- Report benchmark and pg_clients tests to Allure


.github/actions/allure-report/action.yml
```diff
@@ -0,0 +1,195 @@
...
+      if: ${{ inputs.action == 'generate' && always() }}
+      shell: bash -euxo pipefail {0}
+      env:
+        LOCK_FILE: reports/${{ steps.calculate-key.outputs.KEY }}/lock.txt
+        BUCKET: neon-github-public-dev
+      run: |
+        aws s3 cp --only-show-errors "s3://${BUCKET}/${LOCK_FILE}" ./lock.txt || exit 0
+
+        if [ "$(cat lock.txt)" = "${GITHUB_RUN_ID}-${GITHUB_RUN_ATTEMPT}-${{ inputs.test_selection }}" ]; then
+          aws s3 rm "s3://${BUCKET}/${LOCK_FILE}"
```
SomeoneToIgnore, line 194: Most probably you're aware, but S3 provides [object locks](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lock.html) for certainly configured buckets.

Seems a bit more native here, but definitely nothing serious.

.github/actions/allure-report/action.yml
```diff
@@ -0,0 +1,195 @@
...
+        fi
+        echo "::set-output name=KEY::${key}"
+
+    - uses: actions/setup-java@v3
+      if: ${{ inputs.action == 'generate' }}
+      with:
+        distribution: 'temurin'
+        java-version: '17'
+
+    - name: Install Allure
```
SomeoneToIgnore, line 54: I would imagine, this can be installed on an agent later, can't it? 
Sure now it's a prototype PR and it's generally not a big issue anyway.

SomeoneToIgnore (commented):
Great to have a UI for tests!
I've noticed no extra statuses or comments, how a person who creates a PR gets to know the link to the test results?

LGTM in general, not approving right away since not proficient with yaml development 🙂

cicdteam:
just minor question should we have more related (and short) domain name (as example `tests.neon.text`) instead of `neon-github-public-dev.s3.amazonaws.com` ?

bayandin:
> I've noticed no extra statuses or comments, how a person who creates a PR gets to know the link to the test results?

Added!

Screenshot 2022-08-19 at 15 05 21

.github/actions/allure-report/action.yml
bayandin: It won't because the installation is wrapped in `if ! which allure; then`.
Yes, I'm going to create a separate image with java and allure only, but for now, it'll do.

.github/actions/allure-report/action.yml
bayandin: I've never used it, but it does a bit different things from the description. It blocks object changes, but I only need to block object changes from other jobs. Also, it requires versioned buckets, which we don't really need.

bayandin:
> just minor question should we have more related (and short) domain name (as example tests.neon.text) instead of neon-github-public-dev.s3.amazonaws.com ?

I think while we're experimenting with it, it's ok to use `github-public-dev.s3.amazonaws.com`, we may want to make it prettier later

test_runner/fixtures/neon_fixtures.py
```diff
@@ -2237,6 +2239,13 @@ def get_test_output_dir(request: Any) -> pathlib.Path:
     return test_dir
 
 
+ATTACHMENT_SUFFIXES = frozenset((
+    '.log',
+    '.stderr',
+    '.stdout',
+))
```
hlinnaka, line 2246:
```suggestion
ATTACHMENT_SUFFIXES = frozenset((
    '.log',
    '.stderr',
    '.stdout',
    '.diffs',
))
```
The Postgres regression tests output `regression.diffs` on failure. It contains the diff between expected and actual results.

hlinnaka (approved):
The Allure reports look really nice, thanks!

kelvich (approved):

SergeyMelnikov (approved):

PR #2236, opened 2022-08-09
[proxy] Rework wire format of the password hack and some errors

funbringer:
```
The new format has a few benefits: it's shorter, simpler and
human-readable as well. We don't use base64 anymore, since
url encoding got us covered.
    
We also show a better error in case we couldn't parse the
payload; the users should know it's all about passing the
correct project name.
```

### DIFF
--- a/proxy/src/auth.rs
+++ b/proxy/src/auth.rs
@@ -45,6 +45,14 @@ pub enum AuthErrorImpl {
     #[error("Malformed password message: {0}")]
     MalformedPassword(&'static str),
 
+    #[error(
+        "Project name is not specified. \
+        Either please upgrade the postgres client library (libpq) for SNI support \
+        or pass the project name as a parameter: '&options=project%3D<project-name>'. \
+        See more at https://neon.tech/sni"
+    )]
+    MissingProjectName,
+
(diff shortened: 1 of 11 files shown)


kelvich (approved):

proxy/src/auth.rs
```diff
@@ -22,51 +22,53 @@ pub type Result<T> = std::result::Result<T, AuthError>;
...
+    BadAuthMethod(Box<str>),
+
     #[error("Malformed password message: {0}")]
     MalformedPassword(&'static str),
 
-    /// Errors produced by [`crate::stream::PqStream`].
+    #[error(
+        "Project name is not specified. \
+        EITHER please upgrade the postgres client library (libpq) for SNI support \
+        OR pass the project name as a parameter: '&options=project%3D<project-name>'."
```
kelvich, line 51: Let's also add `See more at https://neon.tech/sni`

kelvich (approved):

proxy/src/auth.rs
ololobus: NIT: do we really need CAPS in the message? :)

kelvich: Huh, + 1 for downcasing. I think i am just used to the old looks of this mesage.

PR #2248, opened 2022-08-11
increase timeout in wait_for_upload to avoid spurious failures when testing with real s3

LizardWizzard:
ref #2245

### DIFF
--- a/test_runner/fixtures/neon_fixtures.py
+++ b/test_runner/fixtures/neon_fixtures.py
@@ -2440,7 +2440,7 @@ def wait_for_upload(pageserver_http_client: NeonPageserverHttpClient,
                     timeline: uuid.UUID,
                     lsn: int):
     """waits for local timeline upload up to specified lsn"""
-    for i in range(10):
+    for i in range(20):
         current_lsn = remote_consistent_lsn(pageserver_http_client, tenant, timeline)
         if current_lsn >= lsn:
             return


test_runner/fixtures/neon_fixtures.py
```diff
@@ -460,6 +460,8 @@ class RemoteStorageKind(enum.Enum):
 
 
 def available_remote_storages() -> List[RemoteStorageKind]:
+    return [RemoteStorageKind.LOCAL_FS]
```
aome510, line 463: Is this for debugging?

test_runner/fixtures/neon_fixtures.py
```diff
@@ -2440,7 +2442,7 @@ def wait_for_upload(pageserver_http_client: NeonPageserverHttpClient,
                     timeline: uuid.UUID,
                     lsn: int):
     """waits for local timeline upload up to specified lsn"""
-    for i in range(10):
+    for i in range(20):
```
aome510, line 2445: How do we make sure that `20` is enough but not other bigger numbers? Did you test this change in another branch?

test_runner/fixtures/neon_fixtures.py
LizardWizzard: Yeah, thanks. Cleaned it up

test_runner/fixtures/neon_fixtures.py
LizardWizzard: I checked it locally and it didnt fail several times, it didnt've failed with old value either... I'll rerun the job here several times to make sure that this is enough. 

I think here we hit some temporary slowdown with s3 maybe because of the limits. So Incrementing it to twice the value should be enough

aome510 (approved):

PR #2254, opened 2022-08-11
Bump Postgres version

knizhnik:


knizhnik:
Fix is merged in main of vendor/postgres repository

PR #2258, opened 2022-08-12
Disable `test_import_from_pageserver_multisegment`

aome510:
This test failed consistently on `main` now. It's better to temporarily disable it to avoid blocking others' PRs while investigating the root cause for the test failure.

See: #2255, #2256


zoete (approved):
LGTM

PR #2260, opened 2022-08-12
get_binaries uses DOCKER_TAG taken from docker image build step

cicdteam:


SomeoneToIgnore (approved):
Great.
I would disable the autodiscovery logic by default, so no one gets odd images unintended for sure, but not sure if worth it to do now.

PR #2262, opened 2022-08-12
Find end of WAL on safekeepers using WalStreamDecoder.

arssher:
We could make it inside wal_storage.rs, but taking into account that
 - wal_storage.rs reading is async
 - we don't need s3 here
 - error handling is different; error during decoding is normal
I decided to put it separately.

Test
cargo test test_find_end_of_wal_last_crossing_segment
prepared earlier by @yeputons passes now.

Fixes https://github.com/neondatabase/neon/issues/544 and https://github.com/neondatabase/cloud/issues/2004
Supersedes https://github.com/neondatabase/neon/pull/2066


knizhnik (approved):

arssher:
Failed test_large_schema if caused by separate issue, being fixed at
https://github.com/neondatabase/postgres/pull/193/
so I'm merging this.

PR #2267, opened 2022-08-12
log: adding context to wal timeout errors

### LINKED ISSUE #2106: Add context for `Timed out while waiting for WAL` error
Distinguish between 4 cases:
a) We have a walreceiver and we reached end of stream
b) We're actively processing wal
c) There's no walreceiver
d) Requested LSN is ahead of commit LSN

Or possibly measure and report breakdown of what all the time was spent on. Ideally we'd use tracing infrastructure but we can improvise until then.

ansrivas:
Extend the `wait_lsn()` to accept a `str` context.

This is simply helpful to quickly figure out, which operation
was exactly in progress when the timeout really happened.

fixes #2106

### DIFF
--- a/pageserver/src/basebackup.rs
+++ b/pageserver/src/basebackup.rs
@@ -80,7 +80,12 @@ where
         let (backup_prev, backup_lsn) = if let Some(req_lsn) = req_lsn {
             // Backup was requested at a particular LSN. Wait for it to arrive.
             info!("waiting for {}", req_lsn);
-            timeline.wait_lsn(req_lsn)?;
+
+            let ctx = format!(
+                "Backup was already requested at lsn {}. Waiting for it to arrive.",
+                req_lsn
+            );
+            timeline.wait_lsn(req_lsn, &ctx)?;
 
             // If the requested point is the end of the timeline, we can
             // provide prev_lsn. (get_last_record_rlsn() might return it as
--- a/pageserver/src/layered_repository/timeline.rs
+++ b/pageserver/src/layered_repository/timeline.rs
@@ -418,7 +418,7 @@ impl Timeline for LayeredTimeline {
     }
 
     /// Wait until WAL has been received up to the given LSN.
-    fn wait_lsn(&self, lsn: Lsn) -> anyhow::Result<()> {
+    fn wait_lsn(&self, lsn: Lsn, ctx: &str) -> anyhow::Result<()> {
         // This should never be called from the WAL receiver thread, because that could lead
         // to a deadlock.
         ensure!(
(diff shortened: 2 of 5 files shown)


pageserver/src/basebackup.rs
```diff
@@ -80,7 +80,12 @@ where
         let (backup_prev, backup_lsn) = if let Some(req_lsn) = req_lsn {
             // Backup was requested at a particular LSN. Wait for it to arrive.
             info!("waiting for {}", req_lsn);
-            timeline.wait_lsn(req_lsn)?;
+
+            let ctx = format!(
+                "Backup was already requested at lsn {}. Waiting for it to arrive.",
+                req_lsn
+            );
+            timeline.wait_lsn(req_lsn, &ctx)?;
```
petuhovskiy, line 88: We already have basebackup context, it's captured by our logger. Take a look at this log for example:

```
2022-08-15T09:32:03.915134Z ERROR query handler for 'basebackup 8c39fbeaa128d63bc0b2ce5918e55f78 ae25ca1d50b9d17d112d39f1187d991b 0/1845070' failed: Timed out while waiting for WAL record at LSN 0/1845070 to arrive, last_record_lsn 0/17FA000 disk consistent LSN=0/16B5A38
Caused by:
    SeqWaitError
Stack backtrace:
   0: <pageserver::layered_repository::timeline::LayeredTimeline as pageserver::repository::Timeline>::wait_lsn::{{closure}}
             at /home/runner/pageserver/src/layered_repository/timeline.rs:430:16
             at /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/prometheus-0.13.1/src/histogram.rs:716:19
      <pageserver::layered_repository::timeline::LayeredTimeline as pageserver::repository::Timeline>::wait_lsn
             at /home/runner/pageserver/src/layered_repository/timeline.rs:429:9
   1: pageserver::basebackup::Basebackup<W,T>::new
             at /home/runner/pageserver/src/basebackup.rs:83:13
...
```

The same also applies to other `wait_lsn` callers, because we have a full stacktrace in the logs.

Also note that passing context as a string can be replaced with calling [with_context](https://docs.rs/anyhow/1.0.36/anyhow/trait.Context.html#tymethod.with_context), which can look like:

```rust
timeline.wait_lsn(req_lsn).with_context(|| "waiting for basebackup")?;
```

petuhovskiy (commented):
Thanks for the contribution!

Issue https://github.com/neondatabase/neon/issues/2106 was about adding a context about walreceiver, which is mostly stored here:
- https://github.com/neondatabase/neon/blob/main/pageserver/src/walreceiver/connection_manager.rs
- https://github.com/neondatabase/neon/blob/main/pageserver/src/walreceiver/walreceiver_connection.rs

This PR adds context about the caller, which we already have in the logs, so I would suggest closing this PR. If you still want to try fixing the original issue, you can always open a new PR.

pageserver/src/basebackup.rs
ansrivas: Hi, @petuhovskiy thanks for your comments.
While implementation itself I was unsure, but it seemed way too approachable so I gave it a try.

I will close this one and check the linked code. Thank you.

PR #2278, opened 2022-08-15
split out timeline metrics, track layer map loading and size calculation

LizardWizzard:


hlinnaka (approved):
Makes sense

LizardWizzard:
Failure caused by clippy triggering ICE :)

PR #2281, opened 2022-08-16
Check for entire range during sasl validation

SomeoneToIgnore:
1.63 clippy pointed out an interesting pattern in the code, I've decided to separate it from the rest of the clippy fixes to ensure it's the right thing to do.


funbringer (approved):
Good catch!

PR #2290, opened 2022-08-16
Update cachepot endpoint

zoete:


SomeoneToIgnore (approved):

Dockerfile
```diff
@@ -1,8 +1,6 @@
 # Build Postgres
-FROM neondatabase/rust:1.58 AS pg-build
-WORKDIR /pg
-
-USER root
+FROM 369495373322.dkr.ecr.eu-central-1.amazonaws.com/rust:pinned AS pg-build
```
SomeoneToIgnore, line 2: How is that connected to our old `neondatabase/rust:1.58` images?
Would I still be able to push changes to https://github.com/neondatabase/docker-images and get the updated image?

SomeoneToIgnore:
The PR became a bit more than the title, which is usually ok, but it would be really really great to have a better commit history, especially in such cases.

From this it's not really clear what's up with the new Docker images at least:

image

Dockerfile
zoete: They are closely related to the new images of rust & rustlegacy https://github.com/neondatabase/build/blob/main/rust/Dockerfile & https://github.com/neondatabase/build/blob/main/rustlegacy/Dockerfile

SomeoneToIgnore: Looks like the repo I've linked is obsolete now, if the new ones are used.

Would be nice to hear later a digest on our recent CI and infra changes, that seems to be big at least.

PR #2291, opened 2022-08-16
Make local mypy behave like CI mypy

bojanserafimov:
Running `make fmt` sometimes returns error locally, but runs fine on CI.

To narrow down the problem, running mypy on an individual file returns errors that are not found when running mypy on the project. The `make fmt` script runs mypy on individual files, but ci runs `poetry run mypy .`

Example, running on main branch:
```
$ poetry run mypy test_runner/batch_others/test_pageserver_restart.py
test_runner/batch_others/test_pageserver_restart.py:1: error: Cannot find implementation or library stub for module named "fixtures.neon_fixtures"
test_runner/batch_others/test_pageserver_restart.py:1: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
test_runner/batch_others/test_pageserver_restart.py:2: error: Cannot find implementation or library stub for module named "fixtures.log_helper"
```

My solution is to explicitly tell mypy where to look for imports.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -18,6 +18,10 @@ exclude = ^vendor/
 # some tests don't typecheck when this flag is set
 check_untyped_defs = false
 
+# Help mypy find imports when running against list of individual files.
+# Without this line it would behave differently when executed on the entire project.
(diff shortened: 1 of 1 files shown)


LizardWizzard (approved):
Thanks!

PR #2293, opened 2022-08-17
Enable import test

bojanserafimov:
Resolves https://github.com/neondatabase/neon/issues/2255

Let's see if it works with the new runners


bojanserafimov:

PR #2295, opened 2022-08-17
Fix walreceiver and safekeeper bugs

petuhovskiy:
- There was an issue with zero commit_lsn `reason: LaggingWal { current_commit_lsn: 0/0, new_commit_lsn: 1/6FD90D38, threshold: 10485760 } }`. The problem was in `send_wal.rs`, where we initialized `end_pos = Lsn(0)` and in some cases sent it to the pageserver.
- IDENTIFY_SYSTEM previously returned `flush_lsn` as a physical end of WAL. I'm not sure if walproposer actually relies on it, but now it returns `flush_lsn` (as it was) to walproposer and `commit_lsn` to everyone else including pageserver.
- There was an issue with backoff where connection was cancelled right after initialization: `connected!` -> `safekeeper_handle_db: Connection cancelled` -> `Backoff: waiting 3 seconds`. The problem was in sleeping before establishing the connection. This is fixed by reworking retry logic.
- There was an issue with getting `NoKeepAlives` reason in a loop. The issue is probably the same as the previous.
- There was an issue with filtering safekeepers based on retry attempts, which could filter some safekeepers indefinetely. This is fixed by using retry cooldown duration instead of retry attempts.
- Some `send_wal.rs` connections failed with errors without context. This is fixed by adding a timeline to safekeepers errors.

New retry logic works like this:
- Every candidate has a `next_retry_at` timestamp and is not considered for connection until that moment
- When walreceiver connection is closed, we update `next_retry_at` using exponential backoff, increasing the cooldown on every disconnect.
- When `last_record_lsn` was advanced using the WAL from the safekeeper, we reset the retry cooldown and exponential backoff, allowing walreceiver to reconnect to the same safekeeper instantly.

### DIFF
--- a/pageserver/src/walreceiver/connection_manager.rs
+++ b/pageserver/src/walreceiver/connection_manager.rs
@@ -128,7 +128,7 @@ async fn connection_manager_loop_step(
                             Ok(()) => debug!("WAL receiving task finished"),
                             Err(e) => warn!("WAL receiving task failed: {e}"),
                         };
-                        walreceiver_state.drop_old_connection().await;
+                        walreceiver_state.drop_old_connection(false).await;
                     },
                 }
             },
@@ -319,7 +319,7 @@ impl WalreceiverState {
 
     /// Shuts down the current connection (if any) and immediately starts another one with the given connection string.
     async fn change_connection(&mut self, new_sk_id: NodeId, new_wal_source_connstr: String) {
-        self.drop_old_connection().await;
+        self.drop_old_connection(true).await;
 
         let id = self.id;
         let connect_timeout = self.wal_connect_timeout;
@@ -357,13 +357,15 @@ impl WalreceiverState {
 
     /// Drops the current connection (if any) and updates retry timeout for the next
     /// connection attempt to the same safekeeper.
-    async fn drop_old_connection(&mut self) {
+    async fn drop_old_connection(&mut self, needs_shutdown: bool) {
         let wal_connection = match self.wal_connection.take() {
             Some(wal_connection) => wal_connection,
             None => return,
         };
 
-        wal_connection.connection_task.shutdown().await;
+        if needs_shutdown {
+            wal_connection.connection_task.shutdown().await;
+        }
 
         let retry = self
             .wal_connection_retries
--- a/pageserver/src/walreceiver/connection_manager.rs
+++ b/pageserver/src/walreceiver/connection_manager.rs
@@ -11,7 +11,6 @@
 
 use std::{
     collections::{hash_map, HashMap},
-    future::pending,
     num::NonZeroU64,
     sync::Arc,
     time::Duration,
@@ -129,7 +128,7 @@ async fn connection_manager_loop_step(
                             Ok(()) => debug!("WAL receiving task finished"),
                             Err(e) => warn!("WAL receiving task failed: {e}"),
                         };
-                        walreceiver_state.drop_old_connection(false).await;
+                        walreceiver_state.drop_old_connection().await;
                     },
                 }
             },
@@ -154,13 +153,7 @@ async fn connection_manager_loop_step(
                 }
             },
 
-            _ = async {
-                if let Some(time_until_next_retry) = time_until_next_retry {
-                    tokio::time::sleep(time_until_next_retry).await
-                } else {
-                    pending().await
-                }
-            } => {}
+            _ = tokio::time::sleep(time_until_next_retry.unwrap()), if time_until_next_retry.is_some() => {}
         }
 
         // Fetch more etcd timeline updates, but limit ourselves since they may arrive quickly.
@@ -326,7 +319,7 @@ impl WalreceiverState {
 
     /// Shuts down the current connection (if any) and immediately starts another one with the given connection string.
     async fn change_connection(&mut self, new_sk_id: NodeId, new_wal_source_connstr: String) {
-        self.drop_old_connection(true).await;
+        self.drop_old_connection().await;
 
         let id = self.id;
         let connect_timeout = self.wal_connect_timeout;
@@ -364,15 +357,13 @@ impl WalreceiverState {
 
     /// Drops the current connection (if any) and updates retry timeout for the next
     /// connection attempt to the same safekeeper.
-    async fn drop_old_connection(&mut self, needs_shutdown: bool) {
+    async fn drop_old_connection(&mut self) {
         let wal_connection = match self.wal_connection.take() {
             Some(wal_connection) => wal_connection,
             None => return,
         };
 
-        if needs_shutdown {
-            wal_connection.connection_task.shutdown().await;
-        }
+        wal_connection.connection_task.shutdown().await;
 
         let retry = self
             .wal_connection_retries
@@ -621,12 +612,7 @@ impl WalreceiverState {
     ///
     /// The candidate that is chosen:
     /// * has no pending retry cooldown
-    /// * has greatest data Lsn among the ones that are left
-    ///
-    /// NOTE:
-    /// We evict timeline data received from etcd based on time passed since it was registered, along with its connection attempts values, but
-    /// otherwise to reset the connection attempts, a successful connection to that node is needed.
-    /// That won't happen now, before all nodes with less connection attempts are connected to first, which might leave the sk node with more advanced state to be ignored.
+    /// * has greatest commit_lsn among the ones that are left
     fn select_connection_candidate(
         &self,
         node_to_omit: Option<NodeId>,
--- a/pageserver/src/walreceiver/connection_manager.rs
+++ b/pageserver/src/walreceiver/connection_manager.rs
@@ -153,7 +153,7 @@ async fn connection_manager_loop_step(
                 }
             },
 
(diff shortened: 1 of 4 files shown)


arssher:
> I'm not sure if walproposer actually relies on it
do, it doesn't

pageserver/src/walreceiver/connection_manager.rs
```diff
@@ -354,6 +362,71 @@ impl WalreceiverState {
...
 
+    /// Drops the current connection (if any) and updates retry timeout for the next
+    /// connection attempt to the same safekeeper.
+    async fn drop_old_connection(&mut self, needs_shutdown: bool) {
+        let wal_connection = match self.wal_connection.take() {
+            Some(wal_connection) => wal_connection,
+            None => return,
+        };
+
+        if needs_shutdown {
```
SomeoneToIgnore, line 373: NIT: this looks a bit dangerous, since unattended walreceiver connection is not what we want generally.
`drop_old_connection(false)` is only called on the task end and boils down to `TaskHandle`:
```rust
async fn shutdown(self) {
    self.cancellation.send(()).ok();
    if let Err(e) = self.handle.await {
        error!("Task failed to shut down: {e}")
    }
}
```

to me, this code is safer to run twice than never.
I wonder, should we put it the code above into `Drop` impl.

pageserver/src/walreceiver/connection_manager.rs
```diff
@@ -354,6 +362,71 @@ impl WalreceiverState {
...
+        let now = Utc::now().naive_utc();
+
+        // Schedule the next retry attempt. We want to have exponential backoff for connection attempts,
+        // and we add backoff to the time when we started the connection attempt. If the connection
+        // was active for a long time, then next_retry_at will be in the past.
+        retry.next_retry_at =
+            wal_connection
+                .started_at
+                .checked_add_signed(chrono::Duration::milliseconds(
+                    (retry.retry_duration_seconds * 1000.0) as i64,
```
SomeoneToIgnore, line 394: Quite interesting, there's no `f64 <-> i64` conversion via [TryFrom](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) is implemented.

f64's integer part range seems narrow enough ok to cast into i64, but I wonder if I'm missing anything.

pageserver/src/walreceiver/connection_manager.rs
```diff
@@ -558,41 +631,31 @@ impl WalreceiverState {
...
             .filter(|(_, info)| info.timeline.commit_lsn.is_some())
+            .filter(move |(sk_id, _)| {
+                let next_retry_at = self
+                    .wal_connection_retries
+                    .get(sk_id)
+                    .and_then(|retry_info| {
+                        retry_info.next_retry_at
+                    });
+
+                next_retry_at.is_none() || next_retry_at.unwrap() <= now
```
SomeoneToIgnore, line 657: NIT: `None < Some(T)` for any `T`, so this code could be rewritten as `next_retry_at <= Some(now)`

pageserver/src/walreceiver/connection_manager.rs
```diff
@@ -558,41 +631,31 @@ impl WalreceiverState {
         &self,
         node_to_omit: Option<NodeId>,
     ) -> Option<(NodeId, &SkTimelineInfo, String)> {
-        let all_candidates = self
-            .applicable_connection_candidates()
+        self.applicable_connection_candidates()
```
SomeoneToIgnore, line 634: NIT: looks line we can inline the method now, it's name does not add more sense and the method is used once only.

pageserver/src/walreceiver/connection_manager.rs
```diff
@@ -547,7 +620,7 @@ impl WalreceiverState {
     /// Optionally, omits the given node, to support gracefully switching from a healthy safekeeper to another.
     ///
     /// The candidate that is chosen:
-    /// * has fewest connection attempts from pageserver to safekeeper node (reset every time we receive a WAL message from the node)
+    /// * has no pending retry cooldown
```
SomeoneToIgnore, line 623: `next_connection_candidate`'s rustdoc now needs some adjustments too, at least

> pick a candidate ... and lowest failed connection attemps

needs removal/rephrasing.

SomeoneToIgnore (commented):
Great idea about the timeouts.

I cannot comment Lsn and safekeeper code changes much, look ok to me.
Pageserver part has some nits and one odd place in the `select!` loop that IMO does nothing and could be removed.

pageserver/src/walreceiver/connection_manager.rs
```diff
@@ -154,6 +153,14 @@ async fn connection_manager_loop_step(
                     }
                 }
             },
+
+            _ = async {
+                if let Some(time_until_next_retry) = time_until_next_retry {
+                    tokio::time::sleep(time_until_next_retry).await
+                } else {
+                    pending().await
```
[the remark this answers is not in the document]
petuhovskiy: Yes, etcd and walreceiver events are processed in the same select, but there will be only ~1 event per second in the worst case when all events are triggered simultaneously.

Adding more selectors in `select!` will just call `walreceiver_state.next_connection_candidate()` more often, and that's what I want to happen.

Let's say we have some wal_connection and other candidates with `next_retry_at` in the future. For example, the time until nearest `next_retry_at` will be `3.1s`. Without this extra `async {}` we will call `walreceiver_state.next_connection_candidate()` in `0s`, `1s`, `2s`, `3s`, `4s` and trigger reconnection after `4s`. With this extra `async {}` we can be more precise and wait extra `100ms` after `3s`, to trigger reconnection faster.

arssher (approved):
I like moving retries at this level.

There is a small risk of not waking up when we need this if retry timeout expires immediately before we calculate time to sleep, but this is negligible.

pageserver/src/walreceiver/connection_manager.rs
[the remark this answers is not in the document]
arssher: arguably `async { sleep(time_until_next_retry.as_mut().unwrap()) }, if time_until_next_retry.is_some()` would be more clear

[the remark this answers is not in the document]
SomeoneToIgnore: Ah, so another soft cap to ensure that all other events' backoffs won't become too big.
I remember having something similar in the initial version of the reconnection, nice.

SomeoneToIgnore (approved):

pageserver/src/walreceiver/connection_manager.rs
petuhovskiy: Fixed, looks safe to call twice, but I wonder what will happen if `self.handle.await` is called twice?

pageserver/src/walreceiver/connection_manager.rs
petuhovskiy: So, do you know any better way to convert f64 seconds to chrono::Duration? I don't like multiplying to 1000.0 like it's done now, looks very hacky.

pageserver/src/walreceiver/connection_manager.rs
petuhovskiy: It seems that it'd be harder to understand because it requires knowing Option comparison rules.

pageserver/src/walreceiver/connection_manager.rs
petuhovskiy: Fixed, but I found writing comments about new retry logic somehow hard and may have missed some places. If you know other places with inappropriate or missing comments, let me know.

pageserver/src/walreceiver/connection_manager.rs
arssher: just tried,
`thread 'main' panicked at 'JoinHandle polled after completion', /home/ars/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.12.0/src/runtime/task/core.rs:208:22
`

SomeoneToIgnore: It should return the same result as it returned on the first invocation, at least that's what I've noted in my experiments and there could be some exceptional situations.

SomeoneToIgnore: Looks like I was wrong, we can try https://docs.rs/tokio/latest/tokio/task/struct.JoinHandle.html#method.is_finished then.

pageserver/src/walreceiver/connection_manager.rs
SomeoneToIgnore: I'm afraid, there's no simple float <-> integer conversion anywhere I've looked at, same for overflow checks, so not sure how to make it better conceptually.

pageserver/src/walreceiver/connection_manager.rs
petuhovskiy: `is_finished` returns true if the task was finished, not if the result was consumed. I see no better way than reverting to `needs_shutdown`.

arssher: Well, I'd just maintain 'if we awaited, then `self.wal_connection` is None'; so we always await if there is a handle, even if we know through `TaskEvent::End` that task is going to finish.

PR #2296, opened 2022-08-17
Fix: Always build images

zoete:
Disable the output check for now pending further discussion on image building

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -435,8 +435,7 @@ jobs:
             ./vendor/postgres/Dockerfile
 
   neon-image:
-    # force building for all 3 images
-    if: needs.dockerfile-check.outputs.value == 'true'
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -435,8 +435,7 @@ jobs:
             ./vendor/postgres/Dockerfile
 
   neon-image:
-    # force building for all 3 images
-    if: needs.dockerfile-check.outputs.value == 'true'
+    #if: needs.dockerfile-check.outputs.value == 'true'
```
bayandin, line 438: Let's delete the whole `dockerfile-check` job if we don't use it

zoete: Done

bayandin (approved):

lubennikovaav:
What's the reason of this change? In my understanding this check was there to speed up the build.
Is there some other cache that does that job now?

PR #2301, opened 2022-08-18
Remove main pageserver traits with single implementation 

SomeoneToIgnore:
As discussed in https://github.com/neondatabase/neon/pull/2152 and personally, it feels that our API borders will shift drastically, it makes no sense to carry traits around.

To ease the future refactorings, this PR removes all "interface" traits and proposes basic, module-based, API where we expose a certain `pub struct` or `pub enum` with a certain set of `pub fn` methods. 
When we notice an API pattern we would want to make more generic, then we introduce a trait for that, if needed.

Otherwise, we could consider adding traits for our unit tests, but we do not use this approach much in the current code and might go with https://github.com/nrxus/faux instead for traitless mocking approach.

### DIFF
--- a/pageserver/src/pgdatadir_mapping.rs
+++ b/pageserver/src/pgdatadir_mapping.rs
@@ -7,8 +7,8 @@
 //! Clarify that)
 //!
 use crate::keyspace::{KeySpace, KeySpaceAccum};
+use crate::layered_repository::Timeline;
 use crate::reltag::{RelTag, SlruKind};
-use crate::repository::Timeline;
 use crate::repository::*;
 use crate::walrecord::ZenithWalRecord;
 use anyhow::{bail, ensure, Result};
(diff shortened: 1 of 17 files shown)


SomeoneToIgnore:
This PR is deliberately mechanical and does generic changes to the codebase to avoid excessive conflicts with https://github.com/neondatabase/neon/pull/2231

Ideally, we would want to consider merging or renaming `layered_repository.rs` and `repository.rs` module files and might consider removing `Repository` into `Tenant` as it's more or less the same "something that contains all timelines" entity in pageserver, but that would bloat the diff a lot.

pageserver/src/pgdatadir_mapping.rs
```diff
@@ -51,7 +51,7 @@ pub enum LsnForTimestamp {
 /// versions, if there are big changes in how the data is organized in the data
 /// directory, or if new special files are introduced.
 ///
-pub trait DatadirTimeline: Timeline {
```
LizardWizzard, line 54: Comment above needs to be adjusted, it says "This trait ..." which is no longer true

LizardWizzard (approved):
Looks good to me. Though having multiple `impl Timeline` in different modules is a bit confusing at first. As we discussed in DM possible next step would be to split them into different entities

PR #2304, opened 2022-08-18
Fix dependencies issue between compute-tools and compute node docker images

MMeent:
Compute node docker image requires compute-tools to build, but this
dependency (and the argument for which image to pick) weren't described in the
workflow file. This lead to out-of-date binaries in latest builds, which
subsequently broke these images.


SomeoneToIgnore (approved):
The promoting script looks worth splitting into multiple lines.

ololobus (approved):
LGTM as a current fix, thanks

I still think that there could be a risky race here:
- `push-docker-hub` pulls `latest` images from our registry and pushes them with version tag to docker hub
- it depends on the concurrency settings, but if we allow parallel builds, some parallel build can promote another version to `latest`, so we will pull and publish an older or newer version

Previously, we were passing only the local images around (same as it's currently done in the cloud) and it was way less bug prone, IMO. And this incident just a confirmation to that. cc @zoete

PR #2314, opened 2022-08-19
actions/run-python-test-set: run tests in more parallel processes

bayandin:
WIP

### DIFF
--- a/.github/actions/run-python-test-set/action.yml
+++ b/.github/actions/run-python-test-set/action.yml
@@ -104,7 +104,7 @@ runs:
           exit 1
         fi
         if [[ "${{ inputs.run_in_parallel }}" == "true" ]]; then
-          EXTRA_PARAMS="-n4 $EXTRA_PARAMS"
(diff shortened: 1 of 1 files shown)


.github/actions/run-python-test-set/action.yml
```diff
@@ -104,7 +104,7 @@ runs:
           exit 1
         fi
         if [[ "${{ inputs.run_in_parallel }}" == "true" ]]; then
-          EXTRA_PARAMS="-n4 $EXTRA_PARAMS"
+          EXTRA_PARAMS="-n$(nproc) $EXTRA_PARAMS"
```
LizardWizzard, line 107: FYI nproc is typically too much because one test process spawns many processes under the hood, one pageserver (sometimes two), three safekeepers, etcd, one or more computes. One way to check would be to ssh to the runner and see what the cpu usage is. For 4 physical cores and 8 with hyper threads I used 4 test processes. Worth experimenting with different values

bayandin:
We have increased it to 16 for new runners in https://github.com/neondatabase/neon/pull/3220.

PR #2319, opened 2022-08-22
Remove interferring proc check

LizardWizzard:
We do not need it anymore because ports_distributor checks
whether the port can be used before giving it to service


hlinnaka (approved):

bayandin (approved):
Nice!

PR #2323, opened 2022-08-23
Update WAL redo histograms

MMeent:
Previously, it could only distinguish REDO task durations down to 5ms, which
equates to approx. 200pages/sec or 1.6MB/sec getpage@LSN traffic. 
This patch improves to 200'000 pages/sec or 1.6GB/sec, allowing for
much more precise performance measurement of the redo process.


petuhovskiy:
I wonder should `WAL_REDO_TIME` be a histogram at all? I looked at the code which updates it and the logic is the following:

```
fn apply_batch_postgres(records: &[(Lsn, ZenithWalRecord)]) {
  let start_time = Instant::now();
  
  take_a_lock();
  let lock_time = Instant::now();
  WAL_REDO_WAIT_TIME.observe(lock_time.duration_since(start_time));

  // apply N records
  process.apply_wal_records(records);
  WAL_REDO_TIME.observe(now.duration_since(lock_time));
}

```

The problem is that if N is arbitrary, then we're measuring random duration. Generally, we have different timelines generating a different number of WAL records, and these WAL records can have different sizes in bytes. But here we have a single histogram for a whole pageserver, and the results will tell a story about average values.

So the main question is what information do we want to get by looking at these metrics? I can think of following:
- average walredo speed, in bytes/s
- average walredo speed, in n_records/s
- total number of processed bytes and records
- total time spent in walredo

I don't know how useful that information is, perhaps it can be used to compare speed of walredo on different hardware.

My concern here is that to get information from above we don't need histograms at all, all we need is this:
- `pageserver_wal_redo_seconds_total: Counter` – total time spent in walredo
- `pageserver_wal_redo_records_total: Counter` – total count of processed wal records
- `pageserver_wal_redo_bytes_total: Counter` – total size of processed wal records

I think that using histograms adds no value here and we can replace them with only three Counter metrics. But maybe I'm missing something, what do you think? Maybe there are other useful metrics that we are missing in walredo?

It's not an issue with your patch, because these metrics were already there, just asking here to think about how to get better metrics in the future.

petuhovskiy:
> I think it should, because it allows us to detect performance improvements in a more precise manner than just 'average speed went up' and 'average speed went down'.

Are you sure we can measure performance based on histogram buckets of `WAL_REDO_TIME`? It shows a distribution of time it takes to run `apply_wal_records(records)`, which is exactly a distribution of `records.len()` and I don't think it has a big correlation to performance. That is, more records we have – longer it takes to process them in walredo.

For example, if users will run a lot of projects that insert a single row into a table once a second, we will have less WAL to process in each call to `apply_wal_records(records)` and distribution will change – it will be shifted towards smaller walredo durations. But the performance will stay the same.

The opposite example is someone running a `pgbench init` which generates a lot of WAL. At that moment we will have a lot of WAL to process and distribution will shift to larger durations. But the performance will stay the same.

hlinnaka (approved):

PR #2324, opened 2022-08-23
Do timeline initial size calculation in background

SomeoneToIgnore:
Reworks the way timeline initial size is calculated: 
* never overwrite the data coming the "incremental" part of that size, adjusted after every walreceiver flush/commit
* prepare the context, needed to calculate the "base" part and calculate it once per timeline only. 
* calculate the timeline logical size no more than once, starting from the first access, in the background

I've commented on how some edge cases appeared due to the logic changed, yet I think it's "fine" to merge those as is and iterate on top of the PR to make things better: in particular, entire `LayeredTimelineEntry::Unloaded` has to go and that will change the logic around the same files, so I want to make this change first and only then fix the potential corner cases found here.

### DIFF
--- a/pageserver/src/http/routes.rs
+++ b/pageserver/src/http/routes.rs
@@ -75,7 +75,7 @@ fn get_config(request: &Request<Body>) -> &'static PageServerConf {
 // Helper functions to construct a LocalTimelineInfo struct for a timeline
 
 fn local_timeline_info_from_loaded_timeline(
-    timeline: &Timeline,
+    timeline: &Arc<Timeline>,
     include_non_incremental_logical_size: bool,
     include_non_incremental_physical_size: bool,
 ) -> anyhow::Result<LocalTimelineInfo> {
@@ -106,7 +106,11 @@ fn local_timeline_info_from_loaded_timeline(
         prev_record_lsn: Some(timeline.get_prev_record_lsn()),
         latest_gc_cutoff_lsn: *timeline.get_latest_gc_cutoff_lsn(),
         timeline_state: LocalTimelineState::Loaded,
-        current_logical_size: Some(timeline.get_current_logical_size()),
+        current_logical_size: Some(
+            timeline
+                .get_current_logical_size()
+                .context("Timeline info creation failed to get current logical size")?,
+        ),
         current_physical_size: Some(timeline.get_physical_size()),
         current_logical_size_non_incremental: if include_non_incremental_logical_size {
             Some(timeline.get_current_logical_size_non_incremental(last_record_lsn)?)
@@ -212,7 +216,7 @@ async fn timeline_create_handler(mut request: Request<Body>) -> Result<Response<
         ) {
             Ok(Some((new_timeline_id, new_timeline))) => {
                 // Created. Construct a TimelineInfo for it.
-                let local_info = local_timeline_info_from_loaded_timeline(new_timeline.as_ref(), false, false)?;
+                let local_info = local_timeline_info_from_loaded_timeline(&new_timeline, false, false)?;
                 Ok(Some(TimelineInfo {
                     tenant_id,
                     timeline_id: new_timeline_id,
--- a/pageserver/src/http/routes.rs
+++ b/pageserver/src/http/routes.rs
@@ -11,8 +11,7 @@ use super::models::{
     StatusResponse, TenantConfigRequest, TenantCreateRequest, TenantCreateResponse, TenantInfo,
     TimelineCreateRequest,
 };
-use crate::layered_repository::{metadata::TimelineMetadata, Timeline};
-use crate::repository::{LocalTimelineState, RepositoryTimeline};
+use crate::layered_repository::Timeline;
 use crate::storage_sync;
 use crate::storage_sync::index::{RemoteIndex, RemoteTimeline};
 use crate::tenant_config::TenantConfOpt;
@@ -74,7 +73,7 @@ fn get_config(request: &Request<Body>) -> &'static PageServerConf {
 
 // Helper functions to construct a LocalTimelineInfo struct for a timeline
 
-fn local_timeline_info_from_loaded_timeline(
+fn local_timeline_info_from_timeline(
     timeline: &Arc<Timeline>,
     include_non_incremental_logical_size: bool,
     include_non_incremental_physical_size: bool,
@@ -105,7 +104,6 @@ fn local_timeline_info_from_loaded_timeline(
         last_record_lsn,
         prev_record_lsn: Some(timeline.get_prev_record_lsn()),
         latest_gc_cutoff_lsn: *timeline.get_latest_gc_cutoff_lsn(),
-        timeline_state: LocalTimelineState::Loaded,
         current_logical_size: Some(
             timeline
                 .get_current_logical_size()
@@ -129,61 +127,20 @@ fn local_timeline_info_from_loaded_timeline(
     Ok(info)
 }
 
-fn local_timeline_info_from_unloaded_timeline(metadata: &TimelineMetadata) -> LocalTimelineInfo {
-    LocalTimelineInfo {
-        ancestor_timeline_id: metadata.ancestor_timeline(),
-        ancestor_lsn: {
-            match metadata.ancestor_lsn() {
-                Lsn(0) => None,
-                lsn @ Lsn(_) => Some(lsn),
-            }
-        },
-        disk_consistent_lsn: metadata.disk_consistent_lsn(),
-        last_record_lsn: metadata.disk_consistent_lsn(),
-        prev_record_lsn: metadata.prev_record_lsn(),
-        latest_gc_cutoff_lsn: metadata.latest_gc_cutoff_lsn(),
-        timeline_state: LocalTimelineState::Unloaded,
-        current_logical_size: None,
-        current_physical_size: None,
-        current_logical_size_non_incremental: None,
-        current_physical_size_non_incremental: None,
-        wal_source_connstr: None,
-        last_received_msg_lsn: None,
-        last_received_msg_ts: None,
-    }
-}
-
-fn local_timeline_info_from_repo_timeline(
-    repo_timeline: &RepositoryTimeline<Timeline>,
-    include_non_incremental_logical_size: bool,
-    include_non_incremental_physical_size: bool,
-) -> anyhow::Result<LocalTimelineInfo> {
-    match repo_timeline {
-        RepositoryTimeline::Loaded(timeline) => local_timeline_info_from_loaded_timeline(
-            timeline,
-            include_non_incremental_logical_size,
-            include_non_incremental_physical_size,
-        ),
-        RepositoryTimeline::Unloaded { metadata } => {
-            Ok(local_timeline_info_from_unloaded_timeline(metadata))
-        }
-    }
-}
-
 fn list_local_timelines(
     tenant_id: ZTenantId,
     include_non_incremental_logical_size: bool,
     include_non_incremental_physical_size: bool,
 ) -> Result<Vec<(ZTimelineId, LocalTimelineInfo)>> {
     let repo = tenant_mgr::get_repository_for_tenant(tenant_id)
-        .with_context(|| format!("Failed to get repo for tenant {}", tenant_id))?;
+        .with_context(|| format!("Failed to get repo for tenant {tenant_id}"))?;
     let repo_timelines = repo.list_timelines();
 
     let mut local_timeline_info = Vec::with_capacity(repo_timelines.len());
     for (timeline_id, repository_timeline) in repo_timelines {
         local_timeline_info.push((
             timeline_id,
-            local_timeline_info_from_repo_timeline(
+            local_timeline_info_from_timeline(
                 &repository_timeline,
                 include_non_incremental_logical_size,
                 include_non_incremental_physical_size,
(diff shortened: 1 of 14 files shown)


pageserver/src/http/routes.rs
```diff
@@ -114,7 +114,10 @@ fn local_timeline_info_from_loaded_timeline(
         prev_record_lsn: Some(timeline.get_prev_record_lsn()),
         latest_gc_cutoff_lsn: *timeline.get_latest_gc_cutoff_lsn(),
         timeline_state: LocalTimelineState::Loaded,
-        current_logical_size: Some(timeline.get_current_logical_size()),
+        current_logical_size: timeline
+            .get_logical_size_if_initialized()
```
SomeoneToIgnore, line 118: Before, we used to return some odd logical size there: if no `init_logical_size` was called, we would return the intermediate increments added after walreceivers' flush/commit calls.
After it was called, we've started sending something close to reality.


Now, we send `None` if no init was done before: we do send back this `TimelineInfo` on every timeline creation/branching, so always calling init here slows things drastically.
After the init is called, we start to send the "proper" size, meanwhile `current_logical_size_non_incremental` is available for the whole size if needed.

pageserver/src/layered_repository/timeline.rs
```diff
@@ -1448,7 +1498,15 @@ impl Timeline {
         Ok(new_delta_path)
     }
 
-    pub fn compact(&self) -> Result<()> {
+    pub fn compact(&self) -> anyhow::Result<()> {
+        let last_record_lsn = self.get_last_record_lsn();
+
+        // Last record Lsn could be zero in case the timelie was just created
+        if !last_record_lsn.is_valid() {
```
SomeoneToIgnore, line 1505: I'm not sure how this thing was supposed to work before, presumably, if failed on timeline load for newely created branches and that failure was gracefully handled.

Since now we calculate the timeline size lazily and don't need to call timeline load every time, we stop failing eagerly and get down to compaction to fail.

pageserver/src/tenant_mgr.rs
```diff
@@ -374,15 +373,7 @@ pub fn get_local_timeline_with_load(
     tenant_id: ZTenantId,
     timeline_id: ZTimelineId,
 ) -> anyhow::Result<Arc<Timeline>> {
-    let repository = get_repository_for_tenant(tenant_id)?;
```
SomeoneToIgnore, line 377: This was the main function that kept our state intact despite some logic caps, either initializing the timeline's initial logical size or eagerly failing if some layer was corrupt.

Now this is not true anymore and we're calculating the size lazily, which means we **do not check if the layers are corrupt during timeline load**(!).
I think it's bad, but loading every layer to check if those are corrupt also seems odd to me. IMO, we should add checksum verification around here.

pageserver/src/tenant_mgr.rs
```diff
@@ -485,9 +465,11 @@ pub fn list_tenants(remote_index: &RemoteTimelineIndex) -> Vec<TenantInfo> {
         .iter()
         .map(|(id, tenant)| {
             let has_in_progress_downloads = remote_index
-                .tenant_entry(id)
-                .map(|entry| entry.has_in_progress_downloads());
+            .tenant_entry(id)
+            .map(|entry| entry.has_in_progress_downloads());
 
+            // TODO this is not correct when we might have remote storage sync disabled:
```
SomeoneToIgnore, line 471: Not really related, but found in the logs that we error when remote storage is disabled, so I've left a comment.

pageserver/src/walreceiver/walreceiver_connection.rs
```diff
@@ -315,18 +315,20 @@ pub async fn handle_walreceiver_connection(
 
             // Send zenith feedback message.
             // Regular standby_status_update fields are put into this message.
-            let zenith_status_update = ReplicationFeedback {
-                current_timeline_size: timeline.get_current_logical_size() as u64,
+            let status_update = ReplicationFeedback {
+                current_timeline_size: timeline
+                    .get_current_logical_size()
```
SomeoneToIgnore, line 320: This is now the only place where we actually init the timeline logical size.
This could have changed the performance due to the way we calculated the logical size before and due to the fact that we used to load timeline everywhere, so the size init could have happened before the walreceiver process had to send the feedback.

Now, we do that here and it seems that tests are fine, yet I think we could improve it (maybe, calculate the size in the background on walreceiver start?)

SomeoneToIgnore:
As a general note, I'm not happy how simply and unnoticeable the timeline state could be "wrong" due to `Lsn` being invalid for `disk_consistent_lsn` or similar.

I think, we would need some enum or better logic to avoid such pitfalls, but I'd rather try to address those during `LayeredTimelineEntry::Unloaded` removal or after it's done.

pageserver/src/layered_repository/timeline.rs
```diff
@@ -391,6 +391,29 @@ pub struct Timeline {
...
+    /// yet.
+    pub last_received_wal: Mutex<Option<WalReceiverInfo>>,
+
+    /// Relation size cache
+    pub rel_size_cache: RwLock<HashMap<RelTag, (Lsn, BlockNumber)>>,
+}
+
+/// Internal structure to hold all data needed for logical size calculation.
+/// Calculation consists of two parts: initial size calculation, that potentially might take a lot of time,
+/// since requires all layers up to the `initial_part_end` to be read.
```
hlinnaka, line 407: Not all layers, only the ones containing relation sizes. The point is still valid that it could take a long time, though

pageserver/src/layered_repository/timeline.rs
```diff
@@ -391,6 +391,29 @@ pub struct Timeline {
...
+struct LogicalSize {
+    /// Size, potentially slow to compute, derived from all layers located locally on this node's FS.
+    /// Might require multiple layers read or even ancestor's ones to be read to collect the size.
```
hlinnaka, line 410: grammar
```suggestion
    /// Might require reading multiple layers, and even ancestor's layers, to collect the size.
```

pageserver/src/walreceiver/walreceiver_connection.rs
hlinnaka: Yeah, would be nice to calculate it in the background. As it is, this blocks digesting new WAL.

Hmm, come to think of it, if we call get_current_logical_size() in the walreceiver, doesn't that render the whole lazy calculation pointless? The only thing that ingests new WAL to the timeline, and updates the logical size, is the WAL receiver. If the WAL receiver is also the only thing that performs the calculation, we can never be performing the initial calculation and digesting new WAL at the same time.

P.S. It actually feels weird to me that we send the logical size to the safekeepers, and the safekeeper forwards it to the compute node. It would feel more natural to send it directly from the pageserver to the compute node. But we don't have a feedback channel to send updates from the pageserver to the compute node, currently. Would need some redesigning of the protocol. (I don't suggest doing that in this PR, but maybe later)

SomeoneToIgnore: I've adjusted the code to run a single background thread per timeline to calculate its logical size.

I'm not super happy with the current impl due to obvious reasons:
* it's not a timeline's responsibility to run and orchestrate tasks, yet it's doing that
* a task is run in a separate thread without any cancellation and the managing code around is somewhat dirty compared to async or rust 1.61

I think those all should be fixed when we agree on the sync-async more and define better our task manager, the latter should be on my list next week.

pageserver/src/layered_repository/timeline.rs
```diff
@@ -837,92 +930,153 @@ impl Timeline {
...
+            if ancestor.get_last_record_lsn() == ancestor_branch_off_lsn {
+                let ancestor_logical_size = ancestor
+                    // Note, that while we calculate some logical size based on the ancestor layers only,
+                    // we don't save this result to ancestor's data: ancestor could have its `initial_part_end`
+                    // not equal to `up_to_lsn`/`ancestor_branch_off_lsn`.
+                    //
+                    // We also don't have a function that accepts `from_lsn: Lsn` to calculate a logical size,
+                    // limited with the lower value.
+                    //
+                    // Due to these limitations, we never reuse other logical size calculation results and recompute
```
SomeoneToIgnore, line 1032: I've rewritten the code so that the size init is done only once, as all the comments before and the naming implied.
Due to such approach, we traverse ancestors and record their sizes for init once only.

Before, we used the following code to do that:

https://github.com/neondatabase/neon/blob/1a666a01d672298a1da12771c4b68a28c1d5ebed/pageserver/src/layered_repository/timeline.rs#L854

and this way, we received incomplete ancestors' sizes, hence potentially reporting less timeline size as it is in reality and never updated that value with ancestors' changes.

Now, we wait for all changes to arrive, to be precise, yet that fails our `debug` tests: we're unable to calculate some sizes for half a minute at least, see the corresponding failures.

I'm not quite sure what to do now, except changing the test and adding some TODO, since this looks like a big separate task.

pageserver/src/walreceiver/walreceiver_connection.rs
SomeoneToIgnore: I was wrong, the code is adjusted and works now.

pageserver/src/layered_repository/timeline.rs
```diff
@@ -835,92 +928,157 @@ impl Timeline {
...
 
-    /// (Re-)calculate the logical size of the database at the latest LSN.
+    /// Retrieve current logical size of the timeline.
     ///
-    /// This can be a slow operation.
-    pub fn init_logical_size(&self) -> Result<()> {
-        if self.logical_size_initialized.load(AtomicOrdering::Acquire) {
-            return Ok(());
+    /// The size could be lagging behind the actual number, in case
+    /// the initial size calculation has not been run (gets triggered on the first size access).
```
hlinnaka, line 934: So if the initial size hasn't been calculated yet, this returns a bogus value?

Also, if the current "partial size" is negative, current_size() will return error, and this function will also return an Error.

I'd suggest always returning 0 or None if the initial calculation hasn't finished yet.

pageserver/src/layered_repository/timeline.rs
```diff
@@ -407,22 +434,76 @@ pub struct Timeline {
...
+    size_added_after_initial: AtomicI64,
+}
 
-    /// Information about the last processed message by the WAL receiver,
-    /// or None if WAL receiver has not received anything for this timeline
-    /// yet.
-    pub last_received_wal: Mutex<Option<WalReceiverInfo>>,
+#[derive(Debug, Clone, Copy)]
+enum CurrentSize {
+    Partial(u64),
```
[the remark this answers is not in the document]
SomeoneToIgnore: It is a public interface, so I expected to give away a normalized result, just noting if that is a final one or not with the enum values.

Will add more comments and think of naming more.

pageserver/src/layered_repository/timeline.rs
```diff
@@ -407,22 +434,76 @@ pub struct Timeline {
...
+            }
+            None => {
+                let non_negative_size_increment = size_increment.max(0);
+                u64::try_from(non_negative_size_increment)
+                    .with_context(|| {
+                        format!(
+                            "Failed to convert size increment {non_negative_size_increment} to u64"
+                        )
+                    })
+                    .map(CurrentSize::Partial)
```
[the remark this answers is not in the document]
SomeoneToIgnore: Still, better than panicking 🙂

pageserver/src/layered_repository/timeline.rs
```diff
@@ -389,6 +389,33 @@ pub struct Timeline {
...
+    /// yet.
+    pub last_received_wal: Mutex<Option<WalReceiverInfo>>,
+
+    /// Relation size cache
+    pub rel_size_cache: RwLock<HashMap<RelTag, (Lsn, BlockNumber)>>,
+}
+
+/// Internal structure to hold all data needed for logical size calculation.
+/// Calculation consists of two parts: initial size calculation, that potentially might take a lot of time,
+/// since requires all layers, containing relation sizes, up to the `initial_part_end` to be read.
```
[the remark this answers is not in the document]
SomeoneToIgnore: Taken + added the second part.

pageserver/src/layered_repository/timeline.rs
LizardWizzard: > Not sure how realistic this scenario is, but same doubts I have about the whole init in the background approach (with the new layer sizes, seems that init happens in reasonable time for it to be moved away from the background task to the main code).

FYI https://github.com/neondatabase/neon/pull/2265#discussion_r946129370

I tried to check this for bigger project size on staging, but custom limit setting is broken https://github.com/neondatabase/cloud/issues/2198

For smaller databases IMO this is definitely not a problem anymore

LizardWizzard (approved):
Looks good to me

pageserver/src/layered_repository/timeline.rs
```diff
@@ -407,22 +372,82 @@ pub struct Timeline {
...
+                    initial_size.checked_sub(absolute_size_increment)
+                } else {
+                    initial_size.checked_add(absolute_size_increment)
+                }.with_context(|| format!("Overflow during logical size calculation, initial_size: {initial_size}, size_increment: {size_increment}"))
+                .map(CurrentLogicalSize::Exact)
+            }
+            None => {
+                let non_negative_size_increment = size_increment.max(0);
+                u64::try_from(non_negative_size_increment)
+                    .with_context(|| {
```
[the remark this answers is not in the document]
SomeoneToIgnore: I'm just inconvenient with panics, that's why writing it this way, no good reason behind it, no strong opposition to other approach either, but would leave it as is either 🙂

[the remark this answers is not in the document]
LizardWizzard: See https://rust.godbolt.org/z/cfY9xjWWP

Compiler managed to collapse the code and cut the possible panic. 

If you check with `cargo asm --asm-style=intel --rust pageserver::layered_repository::timeline::LogicalSize::current_size` there is no error message from this with_context anyway. So I think here we unnecessarily increase amount of code and if compiler stops seeing this behind layers of anyhow errors it will generate worse code. It is not on the hot code path, but anyway. Worth a comment at least. We can separate this to a separate function to decrease possibility of someone coming and changing something so panic becomes possible again. WDYT?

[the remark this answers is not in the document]
SomeoneToIgnore: Interesting, yes, we could do something better there indeed.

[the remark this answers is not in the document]
hlinnaka: This might be more clear as:
```
let non_negative_size_increment if size_increment < 0 {
    0
} else {
    size_increment as u64    
};
```
Or:
```
let non_negative_size_increment if size_increment < 0 {
    0
} else {
    u64::try_from(size_increment).unwrap("not negative, cannot fail")
};
```

Issue #2204, opened 2022-08-01
Refactor safekeeper timeline handling


petuhovskiy:
Safekeepers now have several per-timeline tasks (like S3 upload, trimming, broker) and the code behind it got complicated a bit. So, we have some tech debt and possibly some bugs.

Ideas:
- [x] late initialization of `wal_seg_size` is not convenient, better to postpone the whole timeline initialization until we know `wal_seg_size`
- [ ] we often need to get the latest commit_lsn/flush_lsn which are managed inside `SafeKeeper`, we can try to propagate all updates up to timeline, to know precise values without holding mutex over `SafeKeeper`
- [ ] several tasks require coordination and data communication, we may try to use something like `TaskHandle` (https://github.com/neondatabase/neon/blob/main/pageserver/src/walreceiver.rs), to have a unified way for communication and spawning/killing tasks
-  [x] we can try to introduce basic timeline state enum (Uninitialized, Active, Deleted) to solve some problems with delete race conditions and #2166
- [ ] after we have better control over timeline tasks, we can start/stop them dynamically and offload a timeline from memory when all tasks have stopped

Steps:
- [x] require `wal_seg_size` in constructor of `safekeeper.rs`
- [x] require `wal_seg_size` in constructor of `wal_storage.rs`
- [x] postpone timeline initialization until `wal_seg_size` is known
- [x] refactor timeline tasks
- [ ] ... more?

petuhovskiy:
#2329 changes timeline handling (restore/create/delete lifecycle) and fixes wal_seg_size late initialization, but leaves most of the other things untouched.

The next step after merging that PR can be migration to async tasks in safekeeper, and better management for timeline-related tasks.

petuhovskiy:
This was done in https://github.com/neondatabase/neon/pull/7768

Issue #2166, opened 2022-07-26
Safekeeper tries to load timeline control file several times in a second


ololobus:
Noticed locally that safekeeper with one running, yet inactive, compute constantly writes in the log that it reads the control file:
```
safekeeper4_1  | 2022-07-26T16:44:04.359805Z  INFO broker:{timeline=1ecb38997a16069b58ab742109cc5377}: loading control file ./1ecb38997a16069b58ab742109cc5377/29f4c56c3e77eb693143e92a4f98774f/safekeeper.control
safekeeper4_1  | 2022-07-26T16:44:04.420440Z  INFO broker:{timeline=1ecb38997a16069b58ab742109cc5377}: loading control file ./1ecb38997a16069b58ab742109cc5377/29f4c56c3e77eb693143e92a4f98774f/safekeeper.control
safekeeper4_1  | 2022-07-26T16:44:04.746911Z  INFO broker:{timeline=1ecb38997a16069b58ab742109cc5377}: loading control file ./1ecb38997a16069b58ab742109cc5377/29f4c56c3e77eb693143e92a4f98774f/safekeeper.control
safekeeper4_1  | 2022-07-26T16:44:05.362961Z  INFO broker:{timeline=1ecb38997a16069b58ab742109cc5377}: loading control file ./1ecb38997a16069b58ab742109cc5377/29f4c56c3e77eb693143e92a4f98774f/safekeeper.control
safekeeper4_1  | 2022-07-26T16:44:05.423777Z  INFO broker:{timeline=1ecb38997a16069b58ab742109cc5377}: loading control file ./1ecb38997a16069b58ab742109cc5377/29f4c56c3e77eb693143e92a4f98774f/safekeeper.control
safekeeper4_1  | 2022-07-26T16:44:05.750068Z  INFO broker:{timeline=1ecb38997a16069b58ab742109cc5377}: loading control file ./1ecb38997a16069b58ab742109cc5377/29f4c56c3e77eb693143e92a4f98774f/safekeeper.control
```

Comment from @petuhovskiy:
> It’s a bug, I’ve already seen this in staging. This safekeeper doesn’t have this timeline, but some other safekeeper pushes this timeline to etcd every second. When this safekeeper receives message from etcd it tries to load timeline from disk, but doesn’t have it on disk. We don’t log error about loading from disk, and that’s why we have a lot of loading control file INFO logs.

petuhovskiy:
Fixed in #2329

PR #2329, opened 2022-08-24
Refactor safekeeper timelines handling

petuhovskiy:
All timelines are kept in the global in-memory map and loaded from the disk on startup. This way we know all valid timelines and there are no race conditions with timeline creation/deletion.

This fixes race condition where timeline tried to restore SharedState from disk after deletion. It also fixes the issue where we try to restore one timeline from the disk multiple times, but it's not possible because the timeline is not found or corrupted.

Based on https://github.com/neondatabase/neon/issues/2204 but not limited to, what I wanted to fix in this PR:
- Postpone timeline initialization until `wal_seg_size` is known. Now we don't create a Timeline until we know `wal_seg_size`.
- https://github.com/neondatabase/neon/issues/2166. We had a lot of places where we try to restore timeline from disk, the most often case was another peer pushing tli info through broker. Note that this can be especially annoying when each safekeeper have different set of timelines, but they communicate through the same broker. This is fixed by restoring timelines from disk only on startup.
- We have some code running while holding `TIMELINES_STATE` mutex. This could block other threads from getting `Arc<Timeline>` fast, but was especially unpleasant when panics poisoned that mutex. This is fixed by using map only to get `Arc<Timeline>` and don't hold this mutex for other blocking and potentially dangerous operations.
- `truncate_wal` deleted WAL segments from less to higher, this could lead to data corruption if `truncate_wal` is aborted in the middle. Fixed by getting a list of all files in a directory.
- `delete_tenant` now returns an error if anything went wrong. It should tell console to retry the delete operation and fix the issue where we could not fully delete some timelines in corner cases.

<details>
<summary>Previous version of this PR</summary>

This code is not compiling yet, PR is a PoC.

This PR adds a TimelineState:
```
pub enum TimelineState {
    /// Timeline on-disk state is unknown. We either haven't tried to restore
    /// timeline state from disk or got an error.
    Uninitialized(UninitializedState),
    /// Timeline exists on disk and loaded in memory.
    Loaded(SharedState),
    /// Timeline was deleted and cannot be used anymore.
    Deleted,
}
```

All timelines are stored in a single map and they're not deleted from it until safekeeper restart. This allows to fix race condition where timeline tried to restore SharedState from disk after deletion. It also fixes the issue where we try to restore one timeline from the disk multiple times, but it's not possible because the timeline is not found or corrupted.

Based on https://github.com/neondatabase/neon/issues/2204 but not limited to, what I wanted to fix in this PR:
- Postpone timeline initialization until `wal_seg_size` is known. The timeline will have `Uninitialized` state and most methods will return an error until compute tell us `wal_seg_size`.
- https://github.com/neondatabase/neon/issues/2166. We had a lot of places where we try to restore timeline from disk, the most often case was another peer pushing tli info through broker. Note that this can be especially annoying when each safekeeper have different set of timelines, but they communicate through the same broker. This is fixed by not trying to restore timeline from disk more than once.
- We have some code running while holding `TIMELINES_STATE` mutex. This could block other threads from getting `Arc<Timeline>` fast, but was especially unpleasant when panics poisoned that mutex. This is fixed by using map only to get `Arc<Timeline>` and nothing else.
- `truncate_wal` deleted WAL segments from less to higher, this could lead to data corruption if `truncate_wal` is aborted in the middle. Fixed by getting a list of all files in a directory.
- `delete_tenant` now returns an error if anything went wrong. It should tell console to retry the delete operation and fix the issue where we could not fully delete some timelines in corner cases.

I tried to keep changes minimal, but that's not easy and PR turned out bigger than I wanted. I currently don't really like how it turned out in terms of complexity, maybe @arssher knows how to make it better.

</details>


arssher:
I propose the following changes in this area:

1. Just don't create timeline at all while we don't know wal seg size. In current coding that would mean for `receive_wal.rs` to start working without having timeline yet, manually get first message and then create it, which I think is ok. For dynamic membership change safekeepers must know their peers anyway (and probably for good peer recovery), so more explicit init seems unavoidable anyway. Motivation: half-baked state where timeline is already there but doesn't have wal seg size is really messy.
2. Load to all timelines on boot to the hashmap. Motivation: lookups (under global lock) will never need to touch fs, and "using map only to get Arc<Timeline> and nothing else." is achieved. Creation is fast, and even here we can avoid special not-initialized-yet-state and not to hold global lock during creation, if desired: we can create timeline, insert it into global map, take timeline mutex, release global mutex and proceed with creation (reversing the procedure on failure).
3. With deletions, it seems simpler to me to 1) remove timeline from the map to prevent new users 2) optionally wait for all users (threads, tasks) termination through cancel channel (and some feedback channels/counter); 3) drop the directory. But currently I think step 2 is nice-to-have but not super important, as doing something while directory deleted won't have criminal consequences. What is important is to delete both directory and map entry, as we create and delete a lot of timelines.

All these allow to never have timeline in 'invalid state'. All 'normal' code works without caring about it, which avoids cluttering many places with 'is timeline valid' checks.

+1 on moving map to separate module.

+1 on handling delete error properly.

+1 on truncating WAL in reversed order (though being paranoid this is not 100% trouble free, see https://github.com/neondatabase/neon/pull/847#discussion_r750131730 -- but better than currently).

petuhovskiy:
Generally agree and support your proposed changes, however, there are some details and corner cases that should be considered.


> Motivation: half-baked state where timeline is already there but doesn't have wal seg size is really messy.

+1, now in the PR timeline is written to disk only after wal_seg_size is known, but Uninitialized timeline still can stay in memory.

>  we can avoid special not-initialized-yet-state and not to hold global lock during creation, if desired: we can create timeline, insert it into global map, take timeline mutex, release global mutex and proceed with creation (reversing the procedure on failure).

Sounds good, let's look at pseudocode and try to break it virtually. New logic would be:

```go
// sorry for rust-go :)

// Let's assume that process can be killed at any moment, leaving any tenant/timeline in invalid state.
// If we can handle this, then we can handle most of our possible bugs also.

fn create_timeline(zttid, args) {
    // create timeline only in memory
    arc := new_empty_timeline(zttid, args)?

    // take a lock for SharedState, so all actions would be blocked until we exit this function
    arc.lock()
    defer arc.unlock()

    // insert tli into global map, if not exists yet
    global_map.try_insert(zttid, arc)?

    // we created new tli, now we should persist it and start background processes
    err := arc.write_to_disk_and_spawn_bg_tasks()
    if err != nil {
        // XXX: What could be a reason for failure here? Perhaps disk issues:
        // - no space left on disk / inodes
        // - directory already exists
        //
        // Already exists == corrupted timeline on disk which is not loaded to memory,
        // what should we do about this? Trying to load(create) it every 1s compute reconnects
        // to safekeeper is not very good.

        // remove tli from global map; note that delete could happen concurrently,
        // so tli may already be absent from the map.
        global_map.remove(zttid)

        // mark timeline as deleted, if someone already got it from the global map
        arc.mark_deleted()

        // XXX: here we rollback creation, i.e. deleting directory. It seems that it shouldn't
        // fail, but maybe it could? What should we do in this case?
        arc.delete_from_disk()?

        // XXX: If we create/delete the same tli concurrently a lot of times, something can go
        // wrong, but we don't do that, right?

        return err
    }


    return arc
}

fn delete_timeline(zttid) {
    arc := global_map.try_remove(zttid)?

    // take a lock for SharedState, so all actions would be blocked until we exit this function
    arc.lock()
    defer arc.unlock()

    // XXX(optional): arc.wait_for_bg_tasks(). What should we do if wait fails, block indefinetely?

    err := arc.delete_dir_from_disk()
    if err != nil {
        // XXX: What should we do here? I guess this can happen if someone holding a lock for dir/file
    }
}

fn on_startup() {
    tenant_entries = listdir(workdir)
    for tenant_id in tenant_entries.map_uuid() {
        timeline_entries := listdir(workdir.join(tenant_id))
        for timeline_id in timeline_entries.map_uuid() {
            // load timeline from disk
            arc, err := load_timeline(tenant_id, timeline_id)
            if err != nil {
                // XXX: We found corrupted timeline on disk, it cannot be loaded to memory at the moment.
                // What should we do about it? I wouldn't allow to recreate/delete it, and kept it in 
                // the global map for that.
                //
                // Also would be nice to restore it automatically if we fix it manually on disk, but can
                // also be done just by restarting the process.

                // XXX: previously we could have wal_seg_size == 0, but now we require it to be set.
                // We can fail here because of that, what's the best way to handle it?

                return err
            }

            // insert into global map
            global_map.try_insert(tenant_id, arc)?
        }
    }
}
```

What do you think about comments starting from XXX? I'd like to remove Uninitialized state too, but I don't know how to handle corrupted timelines without it.


> Load to all timelines on boot to the hashmap.

It has benefits, but can become a problem when we will have a large number of timelines.

First possible problem is memory consumption. It's unlikely that we will have OOM anytime soon, let's do basic calculations.
`size_of::<Timeline>()` is 1184 bytes. Currently we store `size_of::<SafeKeeperConf>() == 312` twice in shared state, and it's easy to use single Arc for all timelines. New sizeof without redundant fields will be **560 bytes**. This is enough to fit millions of timelines in a single gigabyte. One WAL segment is 16MB on disk and my calculations show that 1 gigabyte of RAM timelines corresponds to ~30TB of disk space. 
The conclusion is that we can store all timelines in a single map without worries about memory consumption.

Second possible problem is startup time. Long startup time can cause downtime and there are also [other reasons against it](https://12factor.net/disposability).
I don't know how long it takes to load a single safekeeper timeline from disk, but we can have estimate from the other side. If we don't want more than 30s of downtime and we have 1'000'000 timelines, then single timeline load should take no longer than `0.03ms`.

arssher:
I thought a bit more and concluded it is indeed better to have 'timeline is under deletion' state, but restrict its existence inside the map only -- If timeline is returned to normal user (to e.g. walsender), it is good. In this case, answers would be

>         // XXX: What could be a reason for failure here? Perhaps disk issues:
>         // - no space left on disk / inodes
>         // - directory already exists
>         //
>         // Already exists == corrupted timeline on disk which is not loaded to memory,
>         // what should we do about this? Trying to load(create) it every 1s compute reconnects
>         // to safekeeper is not very good.

Yes, this should clearly error out. Well, what other options are here except retries? Retries are ok, but note that
- Such error is not normal at all, so it should be report to sentry and attract someone's attention soon.
- At some point compute startup will give up and retries will be stopped (after timeout later operation is retried and so on).
- If desired we can add backoff to increase retry period.

> // remove tli from global map; note that delete could happen concurrently,

Now I think it is better to make deletion from map the last action during deletion (and first during insertion), in this case concurrent delete is not possible. Deletion from map can't fail; also, in this way creation and deletion are totally isolated, avoiding concurrency issues. Lastly, if deletion fails, it can be retried. 

Though it is very-very unlikely, deletion can fail, making manipulations potentially unsafe (however semantics of 'unsafe' operations once delete button was pressed is blurry). Well, tli will be kept in 'deletion in progress' state until it succeeds (or restart; we can protection even here, but I doubt we want).

> // XXX: If we create/delete the same tli concurrently a lot of times, something can go
>         // wrong,

no, if we isolate them, see above

 > // XXX(optional): arc.wait_for_bg_tasks(). What should we do if wait fails, block indefinetely?

I think so. It is better to wait and fail request by timeout rather than leave orphaned tasks. But as said above, I think waiting is not the first priority.

>  // XXX: We found corrupted timeline on disk, it cannot be loaded to memory at the moment.
>                 // What should we do about it? I wouldn't allow to recreate/delete it, and kept it in 
>                 // the global map for that.

As you wrote, new timeline creation will fail due to directory existence. Seems like stable enough situation to wait for manual intervention.

>  // XXX: previously we could have wal_seg_size == 0, but now we require it to be set.
>                 // We can fail here because of that, what's the best way to handle it?

it is for so short living period (wal_seg_size could be zero only during tiny safekeeper internal work, i.e. we don't wait for new messages of something to set it), so I don't think we have such timelines. Just regard as corrupted, then.

> The conclusion is that we can store all timelines in a single map without worries about memory consumption.
+1

> Second possible problem is startup time. Long startup time can cause downtime and there are also [other reasons against it](https://12factor.net/disposability).

Yes. In theory, loading a lot of timelines (big term history, long WAL end searching) might take time, though I think this is not the utmost concertn atm. In this case, we can add lazy loading -- on startup only scan ttids, and add one more intermediate state 'loading' with access blocked while it is in progress.


I'm ok with the idea of adding such states, but only within timeline creation/deletion -- end user always gets Arc to healthy tli, so we don't need 'is timeline ok' checks everywhere.

arssher (approved):

petuhovskiy:
Logs after deploying this to staging:
```
	
2022-09-20T11:19:07.649721Z  INFO found 10455 tenants directories, successfully loaded 10483 timelines
2022-09-20T11:19:07.649678Z  INFO initialized storage for timeline 2c91f30157a3f5864cd077f9812ee4c5, flush_lsn=0/1731BA8, commit_lsn=0/17283E8, peer_horizon_lsn=0/16C4CC8
2022-09-20T11:19:07.649664Z  INFO find_end_of_wal reached end at 0/1731BA8, decode error: WalDecodeError { msg: "invalid xl_tot_len 0", lsn: 0/1731BA8 }
2022-09-20T11:19:07.649521Z  INFO initialized storage for timeline 40fd5830185f9990ee98986aeb304069, flush_lsn=0/1731AF8, commit_lsn=0/1728368, peer_horizon_lsn=0/16C7F08
2022-09-20T11:19:07.649508Z  INFO find_end_of_wal reached end at 0/1731AF8, decode error: WalDecodeError { msg: "invalid xl_tot_len 0", lsn: 0/1731AF8 }
2022-09-20T11:19:07.649363Z  INFO initialized storage for timeline 051d5a1f467a0315fdd061292a95307e, flush_lsn=0/1731BB0, commit_lsn=0/17283E8, peer_horizon_lsn=0/16EE068
2022-09-20T11:19:07.649349Z  INFO find_end_of_wal reached end at 0/1731BB0, decode error: WalDecodeError { msg: "invalid xl_tot_len 0", lsn: 0/1731BB0 }
2022-09-20T11:19:07.649212Z  INFO initialized storage for timeline a77ce6121ebf4a8ac61d4908ff45188d, flush_lsn=0/1731BA8, commit_lsn=0/17283E8, peer_horizon_lsn=0/16C4CC8
2022-09-20T11:19:07.649199Z  INFO find_end_of_wal reached end at 0/1731BA8, decode error: WalDecodeError { msg: "invalid xl_tot_len 0", lsn: 0/1731BA8 }
2022-09-20T11:19:07.649061Z  INFO initialized storage for timeline 9587b46b26ec3bfaa24ce0ed9edc1ed1, flush_lsn=0/16C5130, commit_lsn=0/16C5130, peer_horizon_lsn=0/16C5130
2022-09-20T11:19:07.649049Z  INFO find_end_of_wal reached end at 0/16C5130, decode error: WalDecodeError { msg: "invalid xl_tot_len 0", lsn: 0/16C5130 }
2022-09-20T11:19:07.648963Z  INFO initialized storage for timeline 3889052eeaf886ad19bd94353c38e145, flush_lsn=0/1731A10, commit_lsn=0/17280F0, peer_horizon_lsn=0/16C7F08
2022-09-20T11:19:07.648950Z  INFO find_end_of_wal reached end at 0/1731A10, decode error: WalDecodeError { msg: "invalid xl_tot_len 0", lsn: 0/1731A10 }
2022-09-20T11:19:07.648795Z  INFO initialized storage for timeline 43243d856287e27c070514f0aa918103, flush_lsn=0/16C9568, commit_lsn=0/16C7F08, peer_horizon_lsn=0/16C4CC8
2022-09-20T11:19:07.648782Z  INFO find_end_of_wal reached end at 0/16C9568, decode error: WalDecodeError { msg: "invalid xl_tot_len 0", lsn: 0/16C9568 }
2022-09-20T11:19:07.648685Z  INFO initialized storage for timeline 6281462bdf99f8ffc8e12b5c2ed5e465, flush_lsn=0/16C4CC8, commit_lsn=0/16C4CC8, peer_horizon_lsn=0/16C4CC8
2022-09-20T11:19:07.648672Z  INFO find_end_of_wal reached end at 0/16C4CC8, decode error: WalDecodeError { msg: "invalid xl_tot_len 0", lsn: 0/16C4CC8 }
2022-09-20T11:19:07.648584Z  INFO initialized storage for timeline f9b2416d13864306f8de07938632a9bd, flush_lsn=0/16C5130, commit_lsn=0/16C5130, peer_horizon_lsn=0/16C5130
2022-09-20T11:19:07.648571Z  INFO find_end_of_wal reached end at 0/16C5130, decode error: WalDecodeError { msg: "invalid xl_tot_len 0", lsn: 0/16C5130 }
2022-09-20T11:19:07.648482Z  INFO initialized storage for timeline 5a5f3b2b00382a9454f86589247f33b5, flush_lsn=0/16C5130, commit_lsn=0/16C5130, peer_horizon_lsn=0/16C5130
2022-09-20T11:19:07.648470Z  INFO find_end_of_wal reached end at 0/16C5130, decode error: WalDecodeError { msg: "invalid xl_tot_len 0", lsn: 0/16C5130 }
2022-09-20T11:19:07.648417Z  INFO initialized storage for timeline 5a69ea12c5164954a555afef459ce402, flush_lsn=0/16C4CC8, commit_lsn=0/16C4CC8, peer_horizon_lsn=0/16C4CC8
2022-09-20T11:19:07.648410Z  INFO find_end_of_wal reached end at 0/16C4CC8, decode error: WalDecodeError { msg: "invalid xl_tot_len 0", lsn: 0/16C4CC8 }
2022-09-20T11:19:07.648362Z  INFO initialized storage for timeline 267b2dcaa84087dd1ea7e2b03a41ea57, flush_lsn=0/16C50F8, commit_lsn=0/16C50F8, peer_horizon_lsn=0/16C50F8
2022-09-20T11:19:07.648354Z  INFO find_end_of_wal reached end at 0/16C50F8, decode error: WalDecodeError { msg: "invalid xl_tot_len 0", lsn: 0/16C50F8 }
2022-09-20T11:19:07.648281Z  INFO initialized storage for timeline 51c6c95879f599e5eac2534b37279f9f, flush_lsn=0/1734818, commit_lsn=0/172AE60, peer_horizon_lsn=0/16C5130
2022-09-20T11:19:07.648268Z  INFO find_end_of_wal reached end at 0/1734818, decode error: WalDecodeError { msg: "invalid xl_tot_len 0", lsn: 0/1734818 }
2022-09-20T11:19:07.648109Z  INFO initialized storage for timeline 5187895e5c77218cc2666e93f78accfd, flush_lsn=0/16C4CC8, commit_lsn=0/16C4CC8, peer_horizon_lsn=0/16C4CC8
2022-09-20T11:19:07.648092Z  INFO find_end_of_wal reached end at 0/16C4CC8, decode error: WalDecodeError { msg: "invalid xl_tot_len 0", lsn: 0/16C4CC8 }
2022-09-20T11:19:07.648031Z  INFO initialized storage for timeline 4b027736de17c83aa12e34f709f190ea, flush_lsn=0/16C4CC8, commit_lsn=0/16C4CC8, peer_horizon_lsn=0/16C4CC8
2022-09-20T11:19:07.648023Z  INFO find_end_of_wal reached end at 0/16C4CC8, decode error: WalDecodeError { msg: "invalid xl_tot_len 0", lsn: 0/16C4CC8 }
2022-09-20T11:19:07.647967Z  INFO initialized storage for timeline bb16dfe4091e5a517abe6c28187c2826, flush_lsn=0/16C50F8, commit_lsn=0/16C50F8, peer_horizon_lsn=0/16C50F8
2022-09-20T11:19:07.647954Z  INFO find_end_of_wal reached end at 0/16C50F8, decode error: WalDecodeError { msg: "invalid xl_tot_len 0", lsn: 0/16C50F8 }
2022-09-20T11:19:07.647869Z  INFO initialized storage for timeline c8de064c1504682c9775a0b477495c92, flush_lsn=0/16C5130, commit_lsn=0/16C5130, peer_horizon_lsn=0/16C5130
2022-09-20T11:19:07.647857Z  INFO find_end_of_wal reached end at 0/16C5130, decode error: WalDecodeError { msg: "invalid xl_tot_len 0", lsn: 0/16C5130 }
2022-09-20T11:19:07.647778Z  INFO initialized storage for timeline bd44fb0f4ee91dc0914b7906d1e42dc9, flush_lsn=0/1734388, commit_lsn=0/172AA08, peer_horizon_lsn=0/16F01C8
2022-09-20T11:19:07.647767Z  INFO find_end_of_wal reached end at 0/1734388, decode error: WalDecodeError { msg: "invalid xl_tot_len 0", lsn: 0/1734388 }
2022-09-20T11:19:07.647664Z  INFO initialized storage for timeline 9cea585c9a552a5d0a6db82c3cdfa7cd, flush_lsn=0/1729C80, commit_lsn=0/1729C80, peer_horizon_lsn=0/16960E8
2022-09-20T11:19:07.647654Z  INFO find_end_of_wal reached end at 0/1729C80, decode error: WalDecodeError { msg: "invalid xl_tot_len 0", lsn: 0/1729C80 }
2022-09-20T11:19:07.647539Z  INFO initialized storage for timeline d8fe15e31a2ca09025c7d5fee961322d, flush_lsn=0/16C5130, commit_lsn=0/16C5130, peer_horizon_lsn=0/16C5130
2022-09-20T11:19:07.647532Z  INFO find_end_of_wal reached end at 0/16C5130, decode error: WalDecodeError { msg: "invalid xl_tot_len 0", lsn: 0/16C5130 }
```

PR #2330, opened 2022-08-25
Remove deprecated notification channel

zoete:
As we got Slack I don't see a reason to keep this separate notification channel


kelvich (approved):

cicdteam:
never used that channel (tg) but as i remember  some guys like it (not sure who)

PR #2332, opened 2022-08-25
docs/sourcetree: add info about IDE config

yeputons:


hlinnaka (approved):
I don't use CLion so cannot comment on the details of those instructions, but looks reasonable at a quick glance.

PR #2334, opened 2022-08-25
Integration tests: make sure there are no JWT tokens in logs

yeputons:


yeputons:
@SomeoneToIgnore WDYT about this?

SomeoneToIgnore (commented):
Generally, I think it makes sense, yet the place is indeed could be different.
We have a similar functionality in our Python tests: https://github.com/neondatabase/neon/blob/42c6ddef8edcd1e0dae8cba5a4c8c1f2b3d70589/test_runner/fixtures/neon_fixtures.py#L1994

and I think we'd better for such things there, since later we might want to add something else (passwords, aws creds, etc.) and CI yaml bash scripting is not fun for this.

Another reason to use it in our tests would be to test for JTW tokens better than looking for `eyJ` if possible: for instance, we can remember the generated key in the beginning of the test and grep for it directly.

yeputons:
Together with @arssher decided to close this PR and not create any follow-up issues, assuming the issue is a minor one.

PR #2335, opened 2022-08-25
Copy binaries to /usr/local

zoete:


Dockerfile.compute-node
```diff
@@ -88,8 +88,9 @@ RUN echo "deb http://ftp.debian.org/debian testing main" >> /etc/apt/sources.lis
     apt install -y --no-install-recommends -t testing binutils && \
     rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
-# "temporary" symlink for compatibility with old control-plane
-RUN ln -s /usr/local/bin/compute_ctl /usr/local/bin/zenith_ctl
+# "temporary" symlink for old control-plane & postgres
+RUN ln -s /usr/local/bin/compute_ctl /usr/local/bin/zenith_ctl && \
+    ln -s /usr/local/pgsql/bin/postgres /usr/local/bin/postgres
```
ololobus, line 93: I think it can break in many other ways. Postgres relies by default that some things like extension files and locales are in the /usr/local/share, IIRC

I could be wrong, but I had various problems with Postgres installation, until I just copied the whole installation into /usr/local/* and it worked perfectly well for quite a while

zoete: If it still breaks I'll revert to that approach, all postgres files are on the path

ololobus: > And I don't see that you do any configuration to take this non-default installation path (/usr/local/pgsql), like setting LD_LIBRARY_PATH, etc

Ah, I see that you do `echo '/usr/local/lib' >> /etc/ld.so.conf && /sbin/ldconfig`, but actually in your setup postgres shared libs should be in the `/usr/local/pgsql/lib` or somewhere around, IIUC

zoete: Implemented suggested approach

antonyc (approved):

PR #2343, opened 2022-08-26
Fix proxy tests

funbringer:
There might be different psql & locale configurations, therefore we should explicitly reset them to defaults.


KlimentSerafimov (approved):
Thanks for the compression!

PR #2367, opened 2022-08-30
Store Allure test stats in a database

bayandin:
Let's store data from each test run in DB. It will allow us to create a Grafana chart with test duration (and some other statistics) in the future.

```
CREATE TABLE IF NOT EXISTS regress_test_results (
    id SERIAL PRIMARY KEY,
    reference CHAR(255),
    revision CHAR(40),
    build_type CHAR(16),
    data JSONB
)
```

An example of data https://neon-github-public-dev.s3.amazonaws.com/reports/main/release/2964997750/data/suites.json

### DIFF
--- a/scripts/ingest_regress_test_result.py
+++ b/scripts/ingest_regress_test_result.py
@@ -1,0 +1,97 @@
+#!/usr/bin/env python3
+import argparse
+import os
+import sys
+from contextlib import contextmanager
+from pathlib import Path
+
+import psycopg2
+
+CREATE_TABLE = """
+CREATE TABLE IF NOT EXISTS regress_test_results (
+    id SERIAL PRIMARY KEY,
+    reference CHAR(255),
+    revision CHAR(40),
+    build_type CHAR(16),
+    data JSONB
+)
+"""
+
+
+def err(msg):
+    print(f"error: {msg}")
+    sys.exit(1)
+
+
+@contextmanager
+def get_connection_cursor():
+    connstr = os.getenv("DATABASE_URL")
+    if not connstr:
+        err("DATABASE_URL environment variable is not set")
+    with psycopg2.connect(connstr, connect_timeout=30) as conn:
+        with conn.cursor() as cur:
+            yield cur
+
+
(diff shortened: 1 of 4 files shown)


scripts/ingest_regress_test_result.py
```diff
@@ -0,0 +1,97 @@
...
+        build_type,
+        data_file.read_text(),
+    )
+    cursor.execute(
+        """
+        INSERT INTO regress_test_results (
+            reference,
+            revision,
+            build_type,
+            data
```
LizardWizzard, line 55: What are the plans on how to work with that? Here we store plain json data without parsing into different columns. For perf test it is parsed and grafana can pick this up easily. We can write queries with manual extraction from jsonb, but it has its pros and cons

scripts/ingest_regress_test_result.py
```diff
@@ -0,0 +1,97 @@
...
+"""
+
+
+def err(msg):
+    print(f"error: {msg}")
+    sys.exit(1)
+
+
+@contextmanager
+def get_connection_cursor():
```
LizardWizzard, line 27: Probably a good idea to move utility functions to separate module. It can be injected via adjusting PYTHONPATH. Or we can move this to separate package and runsscripts via poetry run. Not a big problem right now, but at some point we’ll need something like that

scripts/ingest_regress_test_result.py
bayandin: Yes, the original idea was to collect data "as is" and visualise useful things in Grafana using Postgres JSON functions. The current implementation allows us to start to collect data and do experiments with it without worrying too much about a proper data format. If we find out that JSON won't work for our case, it'll be possible to change the DB schema and migrate the data.

scripts/ingest_regress_test_result.py
bayandin: Agree. Will revise once we have another script for uploading data or if we need to make major changes to these two.

LizardWizzard (approved):

PR #2371, opened 2022-08-31
Recreate timeline on pageserver in s3_wal_replay test.

arssher:
That's closer to real usage than switching to brand new pageserver.


test_runner/regress/test_wal_acceptor.py
```diff
@@ -526,25 +518,21 @@ def test_s3_wal_replay(neon_env_builder: NeonEnvBuilder, remote_storage_kind: Re
...
     pg.stop_and_destroy()
-    env.pageserver.stop()
 
-    log.info(f"Removing current pageserver state at {pageserver_tenants_dir}")
-    shutil.rmtree(pageserver_tenants_dir)
-    log.info(f"Copying fresh pageserver state from {pageserver_fresh_copy}")
-    shutil.move(pageserver_fresh_copy, pageserver_tenants_dir)
+    # recreate timeline on pageserver from scratch
+    ps_cli.timeline_delete(UUID(tenant_id), UUID(timeline_id))
+    ps_cli.timeline_create(UUID(tenant_id), UUID(timeline_id))
```
petuhovskiy, line 534: It also may be useful to "simulate" a real scenario when pageserver has crashed and fully replaced with a fresh one. To do that we can restart/delete pageserver dir between `timeline_delete` and `timeline_create`. But the current test is ok already, so this is just an idea for additional test.

petuhovskiy (approved):

petuhovskiy:
> That's closer to real usage than switching to brand new pageserver.

I don't know how often we will need to repair data for a single timeline, probably both cases (single timeline and full pageserver) are possible.

PR #2375, opened 2022-09-01
Fix PERF_TEST_RESULT_CONNSTR for benchmark init

SergeyMelnikov:
…param

### DIFF
--- a/.github/workflows/benchmarking.yml
+++ b/.github/workflows/benchmarking.yml
@@ -179,6 +179,7 @@ jobs:
       env:
         PLATFORM: ${{ steps.calculate-platform.outputs.PLATFORM }}
         BENCHMARK_CONNSTR: ${{ secrets[matrix.connstr] }}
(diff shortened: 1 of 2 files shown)


.github/workflows/benchmarking.yml
```diff
@@ -179,6 +179,7 @@ jobs:
       env:
         PLATFORM: ${{ steps.calculate-platform.outputs.PLATFORM }}
         BENCHMARK_CONNSTR: ${{ secrets[matrix.connstr] }}
+        PERF_TEST_RESULT_CONNSTR: "${{ secrets.PERF_TEST_RESULT_CONNSTR }}"
```
bayandin, line 182: It looks like `VIP_VAP_ACCESS_TOKEN` is required as well (there should be no difference between this step and the following ones)

```suggestion
        VIP_VAP_ACCESS_TOKEN: "${{ secrets.VIP_VAP_ACCESS_TOKEN }}"
        PERF_TEST_RESULT_CONNSTR: "${{ secrets.PERF_TEST_RESULT_CONNSTR }}"
```

bayandin (approved):
Thanks for fixing it!

PR #2381, opened 2022-09-02
Return safekeeper remote storage logging during downloads

SomeoneToIgnore:


safekeeper/src/wal_backup.rs
```diff
@@ -437,16 +437,23 @@ pub async fn read_object(
...
+        file_path.display(),
+        offset
+    );
+    let download = storage
         .download_storage_object(Some((offset, None)), &file_path)
         .await
         .with_context(|| {
             format!(
-                "Failed to open WAL segment download stream for local storage path {}",
+                "Failed to open WAL segment download stream for local path {}",
```
SomeoneToIgnore, line 456: Just realised, that it's a bit odd that we log "local path" here, but querying the storage path.
Not sure how I can do it better right now, so decided to log as is for a while.

petuhovskiy (approved):

PR #2383, opened 2022-09-02
Small fixes

SomeoneToIgnore:
Fixes 

* Benchmark Python test failure
https://github.com/neondatabase/neon/actions/runs/2977846982
https://neon-github-public-dev.s3.amazonaws.com/reports/main/release/2977846982/index.html#suites/792a596a01c566b9215997335bf18812/8185c7bf6360013c/

```
AttributeError: 'ZTenantId' object has no attribute 'hex'
test_runner/performance/test_bulk_insert.py:27: in test_bulk_insert
    env.flush()
test_runner/fixtures/compare_fixtures.py:115: in flush
    self.pscur.execute(f"do_gc {self.env.initial_tenant.hex} {self.timeline} 0")
E   AttributeError: 'ZTenantId' object has no attribute 'hex'
```

* Takes https://github.com/neondatabase/neon/pull/2368/files#r959375753 and again embeds it into another PR 🙂

### DIFF
--- a/pageserver/src/storage_sync/download.rs
+++ b/pageserver/src/storage_sync/download.rs
@@ -141,8 +141,7 @@ async fn download_index_part(
     sync_id: ZTenantTimelineId,
 ) -> Result<IndexPart, DownloadError> {
     let index_part_path = metadata_path(conf, sync_id.timeline_id, sync_id.tenant_id)
-        .with_file_name(IndexPart::FILE_NAME)
-        .with_extension(IndexPart::FILE_EXTENSION);
+        .with_file_name(IndexPart::FILE_NAME);
     let mut index_part_download = storage
         .download_storage_object(None, &index_part_path)
         .await?;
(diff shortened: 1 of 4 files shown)


LizardWizzard (approved):
Heh, you've opened a PR faster :) Just did the same for benchmarks and was about to open a PR but saw this one

LizardWizzard:
I've added run-benchmarks label, but workflow was already marked as skipped, worth running it to check that it helped. Maybe it will work with a retry (cannot check yet, because workflow is still running)

hlinnaka (approved):
Thanks!

SomeoneToIgnore:
I've tried rerunning it a few times, but looks like benchmarks won't start for such jobs:

image

LizardWizzard:
> but looks like benchmarks won't start for such jobs:

Eh, it means it evaluates skips once per push, so may be we should check for benchmark label on label creation event, if thats even possible

PR #2385, opened 2022-09-02
Proposal: Introduce downcasting for GenericRemoteStorage instead of `as_local` implementation

vadim2404:
Hey Neon!

I want to propose removing `as_local` implementation in favor of having the ability to convert between types. WDYT?

### DIFF
--- a/libs/remote_storage/src/lib.rs
+++ b/libs/remote_storage/src/lib.rs
@@ -158,13 +158,6 @@ impl GenericRemoteStorage {
         }
     }
 
-    pub fn as_local(&self) -> Option<&LocalFs> {
-        match self {
-            Self::Local(local_fs) => Some(local_fs),
-            _ => None,
-        }
-    }
-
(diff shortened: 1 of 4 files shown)


funbringer:
Hi! The benefits of your proposal are not obvious: it requires 4x more code and hurts clarity (`try_into` is much more vague, and you have to know that the instance exists while writing new code), code completion and type inference (you had to specify a type). Could you elaborate?

SomeoneToIgnore:
Thank you for your idea.

That is indeed plenty of extra code for the compiler to swallow for the test purposes only, so not sure it this approach is feasible.
We've discussed the alternatives more, https://github.com/neondatabase/neon/pull/2368 has a good main thread (see last three messages, at least).

I'm closing this now, since personally thinking that we'd better try https://github.com/neondatabase/neon/pull/2368#pullrequestreview-1099644840 instead.
If you feel like trying that approach out instead, I'd be more than happy to review the PR.

PR #2386, opened 2022-09-03
Fix negative usize parsing

SomeoneToIgnore:
Follow-up of https://github.com/neondatabase/neon/pull/2324#discussion_r961484551

### DIFF
--- a/pageserver/src/layered_repository/timeline.rs
+++ b/pageserver/src/layered_repository/timeline.rs
@@ -432,14 +432,12 @@ impl LogicalSize {
                 .map(CurrentLogicalSize::Exact)
             }
             None => {
-                let non_negative_size_increment = size_increment.max(0);
-                u64::try_from(non_negative_size_increment)
-                    .with_context(|| {
-                        format!(
-                            "Failed to convert size increment {non_negative_size_increment} to u64"
-                        )
-                    })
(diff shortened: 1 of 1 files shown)


pageserver/src/layered_repository/timeline.rs
```diff
@@ -432,14 +432,12 @@ impl LogicalSize {
...
-                    .with_context(|| {
-                        format!(
-                            "Failed to convert size increment {non_negative_size_increment} to u64"
-                        )
-                    })
-                    .map(CurrentLogicalSize::Approximate)
+                let non_negative_size_increment = if size_increment < 0 {
+                    0
+                } else {
+                    u64::try_from(size_increment).expect("not negative, cannot fail")
```
LizardWizzard, line 438: Maybe worth mentioning that compiler should be able to remove panic code, but up to you, it may be redundant

LizardWizzard (approved):
Thanks!

pageserver/src/layered_repository/timeline.rs
SomeoneToIgnore: Given how outdated the comments may become after big refactorings, I'd rater keep it as is, it's pretty self-explanatory and atomic enough.

PR #2388, opened 2022-09-04
incremental image builds

antonyc:
This using Docker BuildKit might help with local incremental builds. It uses incremental build for release profile, also it might conflict with cachepot (see it's output below). Could anyone who knows cachepot and rust incremental builds have a look? Local rebuild after changing a file takes 2:16.11 total on my VM.

```
#24 39.93   Downloaded ansi_term v0.12.1
#24 40.08    Compiling prost-build v0.10.4
#24 42.12    Compiling tonic-build v0.7.2
#24 43.52    Compiling etcd-client v0.9.2
#24 50.34    Compiling etcd_broker v0.1.0 (/home/nonroot/libs/etcd_broker)
#24 50.77    Compiling pageserver v0.1.0 (/home/nonroot/pageserver)
#24 50.77    Compiling safekeeper v0.1.0 (/home/nonroot/safekeeper)
#24 58.95    Compiling control_plane v0.1.0 (/home/nonroot/control_plane)
#24 96.57     Finished release [optimized + debuginfo] target(s) in 1m 36s
#24 96.66 Compile requests                     15
#24 96.66 Compile requests executed             3
#24 96.66 Cache hits                            0
#24 96.66 Cache misses                          3
#24 96.66 Cache misses (Rust)                   3
#24 96.66 Cache timeouts                        0
#24 96.66 Cache read errors                     0
#24 96.66 Forced recaches                       0
#24 96.66 Cache write errors                    0
#24 96.66 Compilation failures                  0
#24 96.66 Cache errors                          3
#24 96.66 Cache errors (Rust)                   3
#24 96.66 Non-cacheable compilations            0
#24 96.66 Non-cacheable calls                  12
#24 96.66 Non-compilation calls                 0
#24 96.66 Unsupported compiler calls            0
#24 96.66 Average cache write               2.484 s
#24 96.66 Average cache read miss           2.484 s
#24 96.66 Average cache read hit            0.000 s
#24 96.66 Failed distributed compilations       0
#24 96.66
#24 96.66 Non-cacheable reasons:
#24 96.66 crate-type                            6
#24 96.66 incremental                           4
#24 96.66 -                                     2
#24 96.66
#24 96.66 Cache location                  S3, bucket: neon-github-dev
#24 DONE 117.3s
```

Also could use `COPY --link . .` to use [MergeOp and DiffOp](https://www.docker.com/blog/mergediff-building-dags-more-efficiently-and-elegantly/), but seems it needs some tuning on Kaniko side, which I am not well aware of.

Also during `mold cargo build` we build
```
#23 5.828    Compiling prost-build v0.10.4
#23 7.919    Compiling tonic-build v0.7.2
#23 9.340    Compiling etcd-client v0.9.2
#23 16.11    Compiling etcd_broker v0.1.0 (/home/nonroot/libs/etcd_broker)
#23 16.53    Compiling pageserver v0.1.0 (/home/nonroot/pageserver)
#23 16.53    Compiling safekeeper v0.1.0 (/home/nonroot/safekeeper)
#23 24.85    Compiling control_plane v0.1.0 (/home/nonroot/control_plane)
```
but in ./Dockefile we only need 
* pageserver
* safekeeper
* proxy
So why build the others?
The second build with changed pageserver/src/config.rs takes 2 minutes compared to 8 minutes of the first build.


Cargo.toml
```diff
@@ -13,6 +13,7 @@ members = [
 # This is useful for profiling and, to some extent, debug.
 # Besides, debug info should not affect the performance.
 debug = true
+incremental = true
```
SomeoneToIgnore, line 16: That's [the default](https://doc.rust-lang.org/cargo/reference/config.html#configuration-format) and we might want to disable it from time to time with an env var setting, not sure how this would interact with the env var and not sure I want to know now 🙂

Dockerfile
```diff
@@ -40,9 +41,15 @@ COPY . .
 
 # Show build caching stats to check if it was used in the end.
 # Has to be the part of the same RUN since cachepot daemon is killed in the end of this RUN, losing the compilation stats.
-RUN set -e \
-&& mold -run cargo build --locked --release \
-    && cachepot -s
+RUN --mount=type=cache,uid=1000,target=/usr/local/cargo/registry \
```
SomeoneToIgnore, line 44: I think, we should cache both directories, similar way we do it now with "native" rust builds:

https://github.com/neondatabase/neon/blob/a463749f59c3fe020065c4cacc91df8fa11ffb99/.github/workflows/build_and_test.yml#L113-L125

The way it's defined now, we're mixing text source files (` ~/.cargo/registry/ `) and the `target/` binaries, the latter are platform-specific, so if we copy-paste the whole thing, the cache key should be different.

Dockerfile
```diff
@@ -60,9 +67,9 @@ RUN set -e \
     && useradd -d /data zenith \
     && chown -R zenith:zenith /data
 
-COPY --from=build --chown=zenith:zenith /home/nonroot/target/release/pageserver /usr/local/bin
-COPY --from=build --chown=zenith:zenith /home/nonroot/target/release/safekeeper /usr/local/bin
-COPY --from=build --chown=zenith:zenith /home/nonroot/target/release/proxy      /usr/local/bin
+COPY --from=build --chown=zenith:zenith /home/nonroot/pageserver /usr/local/bin
```
SomeoneToIgnore, line 70: To me that looks an extra `cp` and swapping one conventional directory to another one, would that make sense to keep things as is?

Dockerfile
```diff
@@ -81,4 +88,4 @@ VOLUME ["/data"]
 USER zenith
 EXPOSE 6400
 EXPOSE 9898
-CMD ["/bin/bash"]
+CMD ["/bin/bash"]
```
SomeoneToIgnore, line 91: > no newline in the end

It's not like I'm advocating to that, but the majority of our files end with a newline, for example, `Cargo.toml` above and that was not changed in the PR, so let's keep things aligned and newlined here.

SomeoneToIgnore (commented):
> Local rebuild after changing a file takes 2:16.11 total on my VM.

> #24 96.66 Average cache write               2.484 s

At this rate, we're ready to dump cachepot IMO and start caching the `target/` between agents, otherwise LGTM

-----

Of course, caching `target/` for cargo builds has its disadvantages:
* bigger rebuilds if crate changes
* fast cache bloat (hence, longer downloads)
* occasional rustc incremental bugs would require to fiddle with caches

Yet we get one less tool to break in our pipeline and do builds more or less the same way we're doing builds with tests, so it might be ok to do and consider our performance now.

Dockerfile
antonyc: Cannot leave it as is, because `--mount` directive is valid during the single `RUN` only, and contents of it are not observed by the build machinery after this `RUN` has successfully finished.

Dockerfile
antonyc: I added `/usr/local/cargo/git/target` to cache, if I understood you right.

Cargo.toml
antonyc: Is it by default? Tried to remove that line and see a slight increase in build time (+30sec) and cachepot shows different picture:
```
#25 5.453    Compiling prost-build v0.10.4
#25 7.517    Compiling tonic-build v0.7.2
#25 8.976    Compiling etcd-client v0.9.2
#25 16.03    Compiling etcd_broker v0.1.0 (/home/nonroot/libs/etcd_broker)
#25 17.48    Compiling pageserver v0.1.0 (/home/nonroot/pageserver)
#25 17.48    Compiling safekeeper v0.1.0 (/home/nonroot/safekeeper)
#25 100.4    Compiling control_plane v0.1.0 (/home/nonroot/control_plane)
#25 127.1     Finished release [optimized + debuginfo] target(s) in 2m 06s
#25 127.2 Compile requests                      15
#25 127.2 Compile requests executed              7
#25 127.2 Cache hits                             0
#25 127.2 Cache misses                           7
#25 127.2 Cache misses (Rust)                    7
#25 127.2 Cache timeouts                         0
#25 127.2 Cache read errors                      0
#25 127.2 Forced recaches                        0
#25 127.2 Cache write errors                     0
#25 127.2 Compilation failures                   0
#25 127.2 Cache errors                           7
#25 127.2 Cache errors (Rust)                    7
#25 127.2 Non-cacheable compilations             0
#25 127.2 Non-cacheable calls                    8
#25 127.2 Non-compilation calls                  0
#25 127.2 Unsupported compiler calls             0
#25 127.2 Average cache write               22.705 s
#25 127.2 Average cache read miss           22.705 s
#25 127.2 Average cache read hit             0.000 s
#25 127.2 Failed distributed compilations        0
#25 127.2
#25 127.2 Non-cacheable reasons:
#25 127.2 crate-type                             6
#25 127.2 -                                      2
#25 127.2
#25 127.2 Cache location                  S3, bucket: neon-github-dev
```
compared to this line ON:
```
#23 5.770    Compiling prost-build v0.10.4
#23 7.811    Compiling tonic-build v0.7.2
#23 9.250    Compiling etcd-client v0.9.2
#23 15.98    Compiling etcd_broker v0.1.0 (/home/nonroot/libs/etcd_broker)
#23 16.41    Compiling pageserver v0.1.0 (/home/nonroot/pageserver)
#23 16.41    Compiling safekeeper v0.1.0 (/home/nonroot/safekeeper)
#23 25.70    Compiling control_plane v0.1.0 (/home/nonroot/control_plane)
#23 64.03     Finished release [optimized + debuginfo] target(s) in 1m 03s
#23 64.12 Compile requests                     15
#23 64.12 Compile requests executed             3
#23 64.12 Cache hits                            0
#23 64.12 Cache misses                          3
#23 64.12 Cache misses (Rust)                   3
#23 64.12 Cache timeouts                        0
#23 64.12 Cache read errors                     0
#23 64.12 Forced recaches                       0
#23 64.12 Cache write errors                    0
#23 64.12 Compilation failures                  0
#23 64.12 Cache errors                          3
#23 64.12 Cache errors (Rust)                   3
#23 64.12 Non-cacheable compilations            0
#23 64.12 Non-cacheable calls                  12
#23 64.12 Non-compilation calls                 0
#23 64.12 Unsupported compiler calls            0
#23 64.12 Average cache write               2.468 s
#23 64.12 Average cache read miss           2.468 s
#23 64.12 Average cache read hit            0.000 s
#23 64.12 Failed distributed compilations       0
#23 64.12
#23 64.12 Non-cacheable reasons:
#23 64.12 crate-type                            6
#23 64.12 incremental                           4
#23 64.12 -                                     2
#23 64.12
#23 64.12 Cache location                  S3, bucket: neon-github-dev
```

So, is the release/incremental folder even used when `incremental = true` is unset?

antonyc:
At console we also cache [all intermediate layers](https://github.com/neondatabase/cloud/blob/main/.github/workflows/testing.yml#L71-L72), does Kaniko already do that?

Cargo.toml
SomeoneToIgnore: It should be enabled by default but I've just realized that we're using rustc 1.58.1 to build things, that might have some other default settings due to some bugs with incremental compilation, hence the difference.

Another explanation could be cachepot itself, that could cache interim results differently based on cargo options.

I would propose using `CARGO_INCREMENTAL` instead of editing the toml file instead, since the situation with this feature is relative volatile.

Dockerfile
SomeoneToIgnore: Sorry for cryptic comment, my main concern are the `/home/nonroot/target/release/` files: those are supposed to be cached between agents somehow? So that some other Docker-in-Docker builds reuses the `target/` to build incrementally on top only things that were changed?

In the link I've shared above, we archive and upload that `target/` (among others) and then download it before building things and I would expect for similar things to happen here.

SomeoneToIgnore:
> So why build the others?

No reason, I would expect, so we can change it to use `--bin name_1 --bin name_2` to build only what we need.

> does Kaniko already do that?

Alas, no idea about Kaniko.

Dockerfile
antonyc: The original comment in slack was about local builds, so was aiming to speed up only rebuilds on the local machines. These cache mounts use local machine cache, so they speed up only when:

1. docker build
2. edit local rust file
3. docker build

And on GH agents my changes are unlikely to introduce any significant changes. What kind of agents do we use? Standard GHA agents? Or self-hosted runners?

I also added [sharing=locked](https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/reference.md#run---mounttypecache) to reduce changes to concurrent access to cache.

Cargo.toml
antonyc: Ok, `--build-arg CARGO_INCREMENTAL=0` can turn off the incremental build.

zoete:
> At console we also cache [all intermediate layers](https://github.com/neondatabase/cloud/blob/main/.github/workflows/testing.yml#L71-L72), does Kaniko already do that?

We cache across all image building jobs with `--cache=true` https://github.com/GoogleContainerTools/kaniko#flag---cache

Dockerfile
SomeoneToIgnore: > self-hosted runners

These.

> about local builds

Now I start to understand, never mind me then.
In that case, before we make our mind on this for CI (yet if this approach works on our build agents, I think it's worth trying it), never mind my words about `cachepot`, we should not remove it.
It could be disabled for local builds with `RUSTC_WRAPPER=""` in case it helps more.

SomeoneToIgnore:
> No reason, I would expect, so we can change it to use --bin name_1 --bin name_2 to build only what we need.

Fixed in https://github.com/neondatabase/neon/pull/2462/

SomeoneToIgnore:
It was always marked as "work in progress", so up to you, I guess 🙂 
I think, we might want to use this option for regular Docker builds, so thank you for showing the way. The PR to do that still makes sense.

PR #2389, opened 2022-09-05
Control rustc version from storage repo

SomeoneToIgnore:
Adds `rust-toolchain.toml` file and makes all our CI to install that rustc version instead to build the project.

Also bumps rustc version to 1.60 and adds cargo build times report with `--timings` that became available on 1.60 to ensure that we've updated the version everywhere.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -54,7 +54,11 @@ jobs:
       fail-fast: false
       matrix:
         build_type: [ debug, release ]
-        rust_toolchain: [ 1.58 ]
+        # TODO this version is currently needed to make build statuses more informative
+        # and to clear cargo caches in a more transparent way.
+        # We should rather read this value from the file in the root of the repo, `rust-toolchain.toml` since it's
+        # truly setting what version of compiler the sources are built with
+        rust_toolchain: [ '1.60' ]
 
     env:
       BUILD_TYPE: ${{ matrix.build_type }}
@@ -100,11 +104,11 @@ jobs:
           if [[ $BUILD_TYPE == "debug" ]]; then
             cov_prefix="scripts/coverage --profraw-prefix=$GITHUB_JOB --dir=/tmp/coverage run"
             CARGO_FEATURES=""
-            CARGO_FLAGS="--locked"
+            CARGO_FLAGS="--locked --timings"
           elif [[ $BUILD_TYPE == "release" ]]; then
             cov_prefix=""
             CARGO_FEATURES="--features profiling"
-            CARGO_FLAGS="--locked --release $CARGO_FEATURES"
+            CARGO_FLAGS="--locked --timings --release $CARGO_FEATURES"
           fi
           echo "cov_prefix=${cov_prefix}" >> $GITHUB_ENV
           echo "CARGO_FEATURES=${CARGO_FEATURES}" >> $GITHUB_ENV
@@ -218,6 +222,17 @@ jobs:
           name: neon-${{ runner.os }}-${{ matrix.build_type }}-${{ matrix.rust_toolchain }}-artifact
           path: /tmp/neon
 
+      - name: Prepare cargo build timing stats for storing
+        run: |
+          mkdir -p "/tmp/neon/cargo-timings/$BUILD_TYPE/"
+          cp -r ./target/cargo-timings/* "/tmp/neon/cargo-timings/$BUILD_TYPE/"
+        shell: bash -euxo pipefail {0}
+      - name: Upload cargo build stats
+        uses: ./.github/actions/upload
+        with:
+          name: neon-${{ runner.os }}-${{ matrix.build_type }}-${{ matrix.rust_toolchain }}-build-stats
+          path: /tmp/neon/cargo-timings/
+
       # XXX: keep this after the binaries.list is formed, so the coverage can properly work later
       - name: Merge and upload coverage data
         if: matrix.build_type == 'debug'
@@ -233,7 +248,7 @@ jobs:
       fail-fast: false
       matrix:
         build_type: [ debug, release ]
-        rust_toolchain: [ 1.58 ]
+        rust_toolchain: [ '1.60' ]
     steps:
       - name: Checkout
         uses: actions/checkout@v3
@@ -269,7 +284,7 @@ jobs:
       fail-fast: false
       matrix:
         build_type: [ release ]
-        rust_toolchain: [ 1.58 ]
+        rust_toolchain: [ '1.60' ]
     steps:
       - name: Checkout
         uses: actions/checkout@v3
@@ -341,7 +356,7 @@ jobs:
       fail-fast: false
       matrix:
         build_type: [ debug ]
-        rust_toolchain: [ 1.58 ]
+        rust_toolchain: [ '1.60' ]
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -77,9 +88,9 @@ brew link --force libpq
 git clone --recursive https://github.com/neondatabase/neon.git
 cd neon
 
-# The preferred and default is to make a debug build. This will create a 
+# The preferred and default is to make a debug build. This will create a
 # demonstrably slower build than a release build. If you want to use a release
-# build, utilize "BUILD_TYPE=release make -j`nproc`" 
+# build, utilize "BUILD_TYPE=release make -j`nproc`"
```
hlinnaka, line 93: While we're at it, I'd suggest:
```
# The preferred and default is to make a debug build. This will create a
# demonstrably slower build than a release build. For a release build,
# use "make -j`nproc` BUILD_TYPE=release"
```
and same in the OS X instructions below

hlinnaka (approved):

test_runner/fixtures/utils.py
```diff
@@ -155,7 +155,7 @@ def get_scale_for_db(size_mb: int) -> int:
 
 
 ATTACHMENT_NAME_REGEX = re.compile(
-    r".+\.log|.+\.stderr|.+\.stdout|.+\.filediff|.+\.metrics|flamegraph\.svg|regression\.diffs"
+    r".+\.log|.+\.stderr|.+\.stdout|.+\.filediff|.+\.metrics|flamegraph\.svg|regression\.diffs|.+\.html"
```
bayandin, line 158: To support this properly in Allure we also need to add to `allure_attach_from_dir` below:

```python
elif source.endswith(".html"):
    attachment_type = "text/html"
    extension = "html"
```

bayandin (commented):
Should we also support these changes in https://github.com/neondatabase/build/blob/main/rust/Dockerfile#L118-L132?

rust-toolchain.toml
```diff
@@ -0,0 +1,7 @@
+[toolchain]
+channel = "1.60"
```
bayandin, line 2: If we're bumping it anyway, why not update it to the latest `rustc` 1.63?
And probably we need to add this to matrix workflow builds (where 1.58 is hardcoded)

hlinnaka:
`.github/workflows/build_and_test.yml` and `.github/workflows/codestyle.yml` stil talk about version 1.58. I think that may have been a lie even before this PR, I don't see anything in those files to download or install any particular toolchain version. I guess the image we use had 1.58 installed, but I'm not sure. It would be nice to do `rustc --version` at some point during the build, to print the rustc version that was used to the log.

rust-toolchain.toml
hlinnaka: > If we're bumping it anyway, why not update it to the latest rustc 1.63?

I think I'm the reason for that. I don't use rustup on my laptop, I use the rust toolchain from the Debian repository, and the latest version there, in experimental as of today, is 1.60.

If there's a particular reason, like a feature that is super helpful for us, I'm OK with updating our minimum required toolchain. But otherwise I'd like to stick to something that's more widely available. Admittedly even 1.60 isn't really very widely available yet, debian experimental hardly counts, but I'm ok with that compromise.

It would be good to explain this in a comment here, BTW.

SomeoneToIgnore: Commented.

SomeoneToIgnore:
Thanks for noticing the workflow toolchain version, that's more of a bummer than I've thought due to implicit dependency on the `target/` directory caching, so I've bumped it too and left a TODO for now.
Looks like there's [a way](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-contexts-to-create-matrices) to include github context variables into matrix, so we can think of some script way to parse rustc version out of the toml file later.

SomeoneToIgnore:
I've realised that current Allure set-up deals with tests, which is far beyond the build task step, that generates the needed html files.
So I've added a regular GH upload job instead, yet left the html files' changes in our Allure scripts.

Let me know, if there are better ways to deal with that.

.github/workflows/build_and_test.yml
```diff
@@ -54,7 +54,11 @@ jobs:
       fail-fast: false
       matrix:
         build_type: [ debug, release ]
-        rust_toolchain: [ 1.58 ]
+        # TODO this version is currently needed to make build statuses more informative
+        # and to clear cargo caches in a more transparent way.
+        # We should rather read this value from the file in the root of the repo, `rust-toolchain.toml` since it's
+        # truly setting what version of compiler the sources are built with
+        rust_toolchain: [ '1.60' ]
```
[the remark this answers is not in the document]
LizardWizzard: Version is included in cache key, so when we upgrade we shouldnt try downloading build cache for different version because it wont be used anyway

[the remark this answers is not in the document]
SomeoneToIgnore: I've realised, that we don't need to parse any files:

* GH Actions' rust caches in storage repo would require more vigilance, I've put a comment into the toolchain file about this
* rest of the storage logic just passes the rustc version to use it in the artifact names, removed
* style check does not need to install any toolchain separately: currently, we use standard GH runners there and those come up with rustup preinstalled (commented that in the style check workflow file).
When/If we migrate to our GH Agents, we will have rustup and our toolchain preinstalled there.

SomeoneToIgnore:
e2e tests were failing with compute not starting and erroring with 

>  /usr/local/bin/compute_ctl: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.29' not found (required by /usr/local/bin/compute_ctl)

That was caused by `debian:buster-slim` image used in the legacy compute tools image build:
* it was used in `build-deps` image
https://github.com/neondatabase/neon/blob/05e263d0d36bb2e49aa83b4488361b78ec3016e9/Dockerfile.compute-node.legacy#L25
* which later was used to build some Postgres binaries: 
https://github.com/neondatabase/neon/blob/05e263d0d36bb2e49aa83b4488361b78ec3016e9/Dockerfile.compute-node.legacy#L33-L54

Debian Buster has libc 2.28 installed:
https://github.com/neondatabase/cloud/runs/8270217269?check_suite_focus=true#step:7:1579
that causes e2e to fail.

When switched to Debian bullseye with libc 2.31 (the one we have on our real envs)
https://github.com/neondatabase/cloud/runs/8270621653?check_suite_focus=true#step:7:1677
e2e pass:
https://github.com/neondatabase/cloud/actions/runs/3022604100

This also means we can safely merge https://github.com/neondatabase/build/pull/18 as the error comes from a different source.
This PR is ready to be reviewed now.

rust-toolchain.toml
```diff
@@ -0,0 +1,13 @@
+[toolchain]
+# We try to stick to a toolchain version that is widely available on popular distributions, so that most people
+# can use the toolchain that comes with their operating system. But if there's a feature we miss badly from a later
+# version, we can consider updating. As of this writing, 1.60 is available on Debian 'experimental' but not yet on
+# 'testing' or even 'unstable', which is a bit more cutting-edge than we'd like. Hopefully the 1.60 packages reach
+# 'testing' soon (and similarly for the other distributions).
+# See https://tracker.debian.org/pkg/rustc for more details on Debian rustc package.
+channel = "1.60" # do update GitHub CI cache values for rust builds, when changing this value
```
bayandin, line 8: I guess we don't need this comment anymore?

```suggestion
channel = "1.60"
```

bayandin (approved):
I like it!
Workflows look much cleaner without scattered `rust_toolchain:` everywhere

rust-toolchain.toml
bayandin: Ah, sorry, I've misread the comment. Nevermind

PR #2405, opened 2022-09-07
Pass COPT and PG_CFLAGS  to Extension's CFLAGS

bayandin:
I've noticed, that we started to get a compile warning again:

```
/Users/runner/work/neon/neon//pgxn/neon/libpagestore.c:430:15: warning: incompatible function pointer types assigning to 'dbsize_hook_type' (aka 'long (*)(unsigned int)') from 'const int64 (Oid)' (aka 'const long (unsigned int)') [-Wincompatible-function-pointer-types]
                dbsize_hook = zenith_dbsize;
                            ^ ~~~~~~~~~~~~~
1 warning generated.
```
https://github.com/neondatabase/neon/runs/8225115360?check_suite_focus=true#step:11:4621

This was fixed in https://github.com/neondatabase/postgres/pull/189, but accidentally brought back by the next PR: https://github.com/neondatabase/postgres/pull/188. I don't really know if these changes could affect us in a bad way.

I think it would be nice to catch such kinds of warnings on CI, so I'm adding `-Werror` to Makefiles for the extension. 

Initially, I wanted to pass `-Werror` as a flag on CI, but apparently my make-fu is not strong enough for this, so suggestions on how to do it (and not break the build) are welcome. On the other hand, currently, we don't have any warnings, so having this in makefiles by default should help us to catch them earlier during the local development.

### DIFF
--- a/.github/workflows/codestyle.yml
+++ b/.github/workflows/codestyle.yml
@@ -17,6 +17,7 @@ concurrency:
 
 env:
   RUST_BACKTRACE: 1
+  COPT: '-Werror'
 
 jobs:
   check-codestyle-rust:
--- a/Makefile
+++ b/Makefile
@@ -10,12 +10,12 @@ POSTGRES_INSTALL_DIR ?= $(ROOT_PROJECT_DIR)/pg_install/
 BUILD_TYPE ?= debug
 ifeq ($(BUILD_TYPE),release)
 	PG_CONFIGURE_OPTS = --enable-debug --with-openssl
-	PG_CFLAGS = -O2 -g3 -Werror $(CFLAGS)
+	PG_CFLAGS = -O2 -g3 $(CFLAGS)
 	# Unfortunately, `--profile=...` is a nightly feature
 	CARGO_BUILD_FLAGS += --release
 else ifeq ($(BUILD_TYPE),debug)
 	PG_CONFIGURE_OPTS = --enable-debug --with-openssl --enable-cassert --enable-depend
-	PG_CFLAGS = -O0 -g3 -Werror $(CFLAGS)
+	PG_CFLAGS = -O0 -g3 $(CFLAGS)
 else
 	$(error Bad build type '$(BUILD_TYPE)', see Makefile for options)
 endif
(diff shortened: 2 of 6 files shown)


hlinnaka (approved):
Putting this in `PG_CFLAGS` in the top-level Makefile might be even better. That would also catch warnnigs/errors coming from the changes we've made to vendor/postgres.

bayandin:
> Putting this in `PG_CFLAGS` in the top-level Makefile might be even better. That would also catch warnnigs/errors coming from the changes we've made to vendor/postgres.

Added to `PG_CFLAGS`, and passed to extension explicitly via `CFLAGS`.

bayandin:
Interesting, adding `-Werror` results in configure to fail with `Cannot find a working 64-bit integer type.`. And googling this particular error message gets a lot of Postgres-related questions (mostly from psql-hackers and StackOverflow): https://www.google.com/search?q=Cannot+find+a+working+64-bit+integer+type

funbringer:
> Interesting, adding -Werror results in configure to fail with Cannot find a working 64-bit integer type.

Usually this has to do with the order of such flags, i.e.:

```c
int main(void) {
    int a; // -Werror=unused-variable
    return 0;
}
```

```shell
gcc test.c -Wall                    # OK
gcc test.c -Wall -Werror            # FAIL
gcc test.c -Wall -Werror -Wno-error # OK, the latter takes precedence
gcc test.c -Wall -Wno-error -Werror # FAIL, the latter takes precedence
```

Configure's tests essentially just try to compile the code in order to check whether a feature is available, so this flag might hinder them. You might want to take a look at `configure.log`, it should contain the exact reason.

bayandin:
> > Interesting, adding -Werror results in configure to fail with Cannot find a working 64-bit integer type.
> 
> Usually this has to do with the order of such flags, i.e.:

Thanks for the explanations!

[Postgres doc](https://www.postgresql.org/docs/current/install-procedure.html) recommends setting `-Werror` flag via `COPT`:

> An important example is that gcc's -Werror option cannot be included in the CFLAGS passed to configure, because it will break many of configure's built-in tests. To add such flags, include them in the COPT environment variable while running make. 

So, I've changed the approach: I'm setting `COPT=-Werror` in `codestyle.yml` and passing `COPT` to the extension`s `CFLAGS`

PR #2417, opened 2022-09-08
Use regular agent for triggering e2e tests

SomeoneToIgnore:
e2e triggering is a bunch of `curl` calls on GH API + a few secrets, so could be done from any agent, not a special rust one.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -412,7 +412,7 @@ jobs:
   trigger-e2e-tests:
     runs-on: dev
     container:
-      image: 369495373322.dkr.ecr.eu-central-1.amazonaws.com/rust:pinned
+      image: 369495373322.dkr.ecr.eu-central-1.amazonaws.com/base:pinned
       options: --init
     needs: [ build-neon ]
     steps:


.github/workflows/build_and_test.yml
```diff
@@ -410,10 +410,7 @@ jobs:
         shell: bash -euxo pipefail {0}
 
   trigger-e2e-tests:
-    runs-on: dev
-    container:
-      image: 369495373322.dkr.ecr.eu-central-1.amazonaws.com/rust:pinned
-      options: --init
+    runs-on: 'ubuntu-latest'
```
LizardWizzard, line 413: There is `369495373322.dkr.ecr.eu-central-1.amazonaws.com/base:latest` image which might be preferable if the current policy is "use our runners everywhere except for mac os compile job". @zoete wdyt?

zoete: The rust one might be a bit overkill for this purpose, the most recent `base:pinned` image has curl

SomeoneToIgnore: Is there such a thing as `base:pinned`? Have not found that, so replaced with `base:latest`, as suggested.

zoete: Yes there is as you can see here: https://eu-central-1.console.aws.amazon.com/ecr/repositories/private/369495373322/base?region=eu-central-1

bayandin (approved):
Nice!

PR #2434, opened 2022-09-13
test_runner: redact passwords from log messages

bayandin:


test_runner/fixtures/log_helper.py
```diff
@@ -29,6 +30,17 @@
 }
 
 
+class PasswordFilter(logging.Filter):
+    """Filter out password from logs."""
+
+    # Good enought to filter our passwords produced by PgProtocol.connstr
```
LizardWizzard, line 36:
```suggestion
    # Good enough to filter our passwords produced by PgProtocol.connstr
```

LizardWizzard (approved):

PR #2448, opened 2022-09-14
[proxy] Drop support for legacy cloud API

funbringer:
Apparently, it no longer exists in the cloud.


proxy/src/main.rs
```diff
@@ -52,8 +69,8 @@ async fn main() -> anyhow::Result<()> {
             Arg::new("auth-backend")
                 .long("auth-backend")
                 .takes_value(true)
-                .help("Possible values: legacy | console | postgres | link")
-                .default_value("legacy"),
+                .value_parser(clap::builder::EnumValueParser::<auth::BackendType<()>>::new())
+                .default_value("link"),
```
LizardWizzard, line 73: Since default value has changed, I'm not sure if we specify the parameter in all instances currently deployed. In other words should we update something in deploy config to match this change?

LizardWizzard (commented):
Not sure if we need to change deploy configs to match these changes. Otherwise LGTM

proxy/src/main.rs
funbringer: No, I don't think this should be necessary exactly because we're going to select the next best thing. The legacy auth provider boils down to the link auth flow, and afaik we only use this part (or how else could it work given that it's been dropped from the cloud?).

LizardWizzard (approved):

proxy/src/main.rs
funbringer: A few links to support my hypothesis:

https://github.com/neondatabase/neon/blob/f86ea09323ac0d6f2904dcf603652044cea50664/.github/workflows/build_and_test.yml#L736-L737

https://github.com/neondatabase/neon/blob/f86ea09323ac0d6f2904dcf603652044cea50664/.github/helm-values/staging.proxy.yaml#L7-L9

https://github.com/neondatabase/neon/blob/f86ea09323ac0d6f2904dcf603652044cea50664/.github/helm-values/staging.proxy-scram.yaml#L7-L10

ololobus (approved):

PR #2450, opened 2022-09-14
Validate tenant create options

bojanserafimov:
to avoid debugging config typos

### DIFF
--- a/control_plane/src/storage.rs
+++ b/control_plane/src/storage.rs
@@ -411,7 +411,7 @@ impl PageServerNode {
                 .context("Failed to parse 'max_lsn_wal_lag' as non zero integer")?,
         };
         if !settings.is_empty() {
-            bail!("Unrecognized tenant settings: {:?}", settings)
+            bail!("Unrecognized tenant settings: {settings:?}")
         }
         self.http_request(Method::POST, format!("{}/tenant", self.http_base_url))
             .json(&request)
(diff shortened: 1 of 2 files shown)


test_runner/regress/test_tenant_conf.py
```diff
@@ -1,19 +1,29 @@
...
     neon_env_builder.pageserver_config_override = """
 page_cache_size=444;
 wait_lsn_timeout='111 s';
 tenant_config={checkpoint_distance = 10000, compaction_target_size = 1048576}"""
 
     env = neon_env_builder.init_start()
-    """Test per tenant configuration"""
+
+    # Check that we raise on misspelled configs
+    with pytest.raises(Exception):
```
SomeoneToIgnore, line 20: NIT: maybe worth checking, that we're failing due to the intended reason and `some_invalid_setting_name_blah_blah` is contained in the exception message

control_plane/src/storage.rs
```diff
@@ -371,43 +371,50 @@ impl PageServerNode {
...
+                .remove("lagging_wal_timeout")
+                .map(|x| x.to_string()),
+            max_lsn_wal_lag: settings
+                .remove("max_lsn_wal_lag")
+                .map(|x| x.parse::<NonZeroU64>())
+                .transpose()
+                .context("Failed to parse 'max_lsn_wal_lag' as non zero integer")?,
+        };
+        if !settings.is_empty() {
+            bail!("Unrecognized tenant settings: {:?}", settings)
```
SomeoneToIgnore, line 414:
```suggestion
            bail!("Unrecognized tenant settings: {settings:?}")
```

SomeoneToIgnore (approved):

PR #2452, opened 2022-09-14
Follow-up for neondatabase/neon#2448

ololobus:
* remove `legacy` mode from the proxy readme
* explicitly specify `authBackend` in the link auth proxy helm-values for all envs


funbringer (approved):
Thank you for cleaning up the mess!

ololobus:
> Thank you for cleaning up the mess!

I should've noticed that during review, especially taking into account that Dmitry R made a comment about it :) But I didn't know how these helm parts are scattered across the repos

Issue #2378, opened 2022-09-01
`make` always recompiles the Neon extension


yeputons:
## Steps to reproduce
1. Build neon
2. Run `make neon-pg-ext`

## Expected result
There are no `gcc` commands in the input as no files has been changed.

## Actual result
```
Compiling neon
make PG_CONFIG=/home/yeputons/neon/neon//tmp_install/bin/pg_config \
	-C /home/yeputons/neon/neon//pgxn/neon install
make[1]: Entering directory '/home/yeputons/neon/neon/pgxn/neon'
gcc -Wall -Wmissing-prototypes -Wpointer-arith -Wdeclaration-after-statement -Werror=vla -Wendif-labels -Wmissing-format-attribute -Wimplicit-fallthrough=3 -Wcast-function-type -Wformat-security -fno-strict-aliasing -fwrapv -fexcess-precision=standard -Wno-format-truncation -Wno-stringop-truncation -g -O0 -g3  -fPIC -I/home/yeputons/neon/neon/tmp_install/include -I. -I./ -I/home/yeputons/neon/neon/tmp_install/include/postgresql/server -I/home/yeputons/neon/neon/tmp_install/include/postgresql/internal  -D_GNU_SOURCE   -c -o inmem_smgr.o inmem_smgr.c -MMD -MP -MF .deps/inmem_smgr.Po
gcc -Wall -Wmissing-prototypes -Wpointer-arith -Wdeclaration-after-statement -Werror=vla -Wendif-labels -Wmissing-format-attribute -Wimplicit-fallthrough=3 -Wcast-function-type -Wformat-security -fno-strict-aliasing -fwrapv -fexcess-precision=standard -Wno-format-truncation -Wno-stringop-truncation -g -O0 -g3  -fPIC -I/home/yeputons/neon/neon/tmp_install/include -I. -I./ -I/home/yeputons/neon/neon/tmp_install/include/postgresql/server -I/home/yeputons/neon/neon/tmp_install/include/postgresql/internal  -D_GNU_SOURCE   -c -o libpagestore.o libpagestore.c -MMD -MP -MF .deps/libpagestore.Po
[...]

PR #2325, opened 2022-08-24
Extract neon and neon_test_utils from postgres repo

MMeent:
See #2307

PR #2454, opened 2022-09-15
docs/sourcetree: update CLion set up instructions

yeputons:
After #2325 the old method no longer works as our Makefile does not print compilation commands when run with `--dry-run`, see https://github.com/neondatabase/neon/issues/2378#issuecomment-1241421325

This method is much slower but is hopefully robust.

Add some more notes while we're here.


yeputons:
Actually, may be already fixed by #2430, hold on for a moment

yeputons:
Nope, it does not for me. Sad.

This PR is relevant.

SomeoneToIgnore (approved):

PR #2455, opened 2022-09-15
Add authentication for Safekeeper

yeputons:
Based on #2730

### DIFF
--- a/control_plane/src/compute.rs
+++ b/control_plane/src/compute.rs
@@ -322,6 +322,9 @@ impl PostgresNode {
         conf.append("shared_preload_libraries", "neon");
         conf.append_line("");
         conf.append("neon.pageserver_connstring", &pageserver_connstr);
+        if let AuthType::NeonJWT = auth_type {
+            conf.append("neon.safekeeper_token_env", "$ZENITH_AUTH_TOKEN");
+        }
         conf.append("neon.tenant_id", &self.tenant_id.to_string());
         conf.append("neon.timeline_id", &self.timeline_id.to_string());
         if let Some(lsn) = self.lsn {
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -1,30 +1,154 @@
 ## Authentication
 
 ### Overview
+We use JWT tokens in communication between almost all components (compute, pageserver, safekeeper, CLI) regardless of the protocol used (HTTP/PostgreSQL).
+Etcd currently has no authentication.
+Authentication is optional and is disabled by default for easier debugging.
+It is used in some tests, though.
+Note that we do not cover authentication with `pg.neon.tech` here.
 
-Current state of authentication includes usage of JWT tokens in communication between compute and pageserver and between CLI and pageserver. JWT token is signed using RSA keys. CLI generates a key pair during call to `neon_local init`. Using following openssl commands:
+For HTTP connections we use the Bearer authentication scheme.
+For PostgreSQL connections we expect the token to be passed as a password.
+There is a caveat for `psql`: it silently truncates passwords to 100 symbols, so to correctly pass JWT via `psql` you have to either use `PGPASSWORD` environment variable, or store password in `psql`'s config file.
+
+Current token scopes are described in `utils::auth::Scope`.
+There are no expiration or rotation schemes.
+
+_TODO_: some scopes allow both access to server management API and to the data.
+These probably should be split into multiple scopes.
+
+Tokens should not occur in logs.
+They may sometimes occur in configuration files, although this is discouraged
+because configs may be parsed and dumped into logs.
+
+#### Tokens generation and validation
+JWT tokens are signed using a private key.
+Compute/pageserver/safekeeper use the private key's public counterpart to validate JWT tokens.
+These components should not have access to the private key and may only get tokens from their configuration or external clients. 
+
+The key pair is generated once for an installation of compute/pageserver/safekeeper, e.g. by `neon_local init`.
+There is currently no way to rotate the key without bringing down all components.
+
+### CLI
+CLI generates a key pair during call to `neon_local init` with the following commands:
 
 ```bash
-openssl genrsa -out private_key.pem 2048
-openssl rsa -in private_key.pem -pubout -outform PEM -out public_key.pem
+openssl genrsa -out auth_private_key.pem 2048
+openssl rsa -in auth_private_key.pem -pubout -outform PEM -out auth_public_key.pem
 ```
 
-CLI also generates signed token and saves it in the config for later access to pageserver. Now authentication is optional. Pageserver has two variables in config: `auth_validation_public_key_path` and `auth_type`, so when auth type present and set to `NeonJWT` pageserver will require authentication for connections. Actual JWT is passed in password field of connection string. There is a caveat for psql, it silently truncates passwords to 100 symbols, so to correctly pass JWT via psql you have to either use PGPASSWORD environment variable, or store password in psql config file.
+Configuration files for all components point to `public_key.pem` for JWT validation.
+However, authentication is disabled by default.
+There is no way to automatically enable it everywhere, you have to configure each component individually.
 
-Currently there is no authentication between compute and safekeepers, because this communication layer is under heavy refactoring. After this refactoring support for authentication will be added there too. Now safekeeper supports "hardcoded" token passed via environment variable to be able to use callmemaybe command in pageserver.
+CLI also generates signed token (full access to Pageserver) and saves it in
+the CLI's `config` file under `pageserver.auth_token`.
+Note that pageserver's config does not have any similar parameter.
+CLI is the only component which accesses that token.
+Technically it could generate it from the private key on each run,
+but it does not do that for some reason (_TODO_).
 
-Compute uses token passed via environment variable to communicate to pageserver and in the future to the safekeeper too.
+### Compute
+#### Overview
+Compute is a per-timeline PostgreSQL instance, so it should not have
+any access to data of other tenants.
+All tokens used by a compute are restricted to a specific tenant.
+There is no auth isolation from other timelines of the same tenant,
+but a non-rogue client never accesses another timeline even by an accident:
+timeline IDs are random and hard to guess.
 
-JWT authentication now supports two scopes: tenant and pageserverapi. Tenant scope is intended for use in tenant related api calls, e.g. create_branch. Compute launched for particular tenant also uses this scope. Scope pageserver api is intended to be used by console to manage pageserver. For now we have only one management operation - create tenant.
+#### Incoming connections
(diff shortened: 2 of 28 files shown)


control_plane/src/compute.rs
```diff
@@ -306,6 +306,9 @@ impl PostgresNode {
         conf.append("shared_preload_libraries", "neon");
         conf.append_line("");
         conf.append("neon.pageserver_connstring", &pageserver_connstr);
+        if let AuthType::NeonJWT = auth_type {
+            conf.append("neon.safekeeper_token_env", "$ZENITH_AUTH_TOKEN");
```
yeputons, line 310: This env variable is already set when running the compute node, it's used inside `pageserver_connstring`

petuhovskiy:
Would be good to update https://github.com/neondatabase/neon/blob/main/docs/authentication.md, to follow current state of things better.

IIUC, we currently have 3 storage services operating with Neon JWT:
- compute
- safekeeper
- pageserver

I'd like to have documentation on what tokens are placed in the config of these services, and whether they're optional or not.

In safekeeper and pageserver we have libpq and http interfaces, and we accept queries with tokens on these interfaces. Clarification on what tokens we should use for these interfaces will be also nice.

This is not strictly related to this PR, but you seem to have context for these things which is needed to write a good doc :)
This doc will also make reviewing this PR a little easier.

libs/utils/src/auth.rs
```diff
@@ -23,6 +23,7 @@ const JWT_ALGORITHM: Algorithm = Algorithm::RS256;
 pub enum Scope {
     Tenant,
     PageServerApi,
+    AllSafekeeperData,
```
arssher, line 26: Why All?

libs/utils/src/http/endpoint.rs
```diff
@@ -137,10 +137,14 @@ pub fn auth_middleware<B: hyper::body::HttpBody + Send + Sync + 'static>(
     })
 }
 
-pub fn check_permission(req: &Request<Body>, tenant_id: Option<TenantId>) -> Result<(), ApiError> {
+pub fn check_permission_with<E: Display>(
```
arssher, line 140: E can/should be std::error:Error

control_plane/src/compute.rs
```diff
@@ -324,6 +324,9 @@ impl PostgresNode {
...
```
arssher, line 328: This won't be needed if we make $ZENITH_AUTH_TOKEN the default GUC value -- and better reuse it as a token to both  pageserver and safekeeper (can be in separate pr).

control_plane/src/storage.rs
```diff
@@ -239,6 +240,13 @@ impl PageServerNode {
         let mut filled_cmd = fill_rust_env_vars(cmd.args(&args).arg("--daemonize"));
         filled_cmd = fill_aws_secrets_vars(filled_cmd);
 
+        if self.env.pageserver.auth_type != AuthType::Trust {
```
arssher, line 243: worth a comment 'generate a token to connect from pageserver to safekeeper'

safekeeper/src/wal_service.rs
```diff
@@ -5,25 +5,32 @@
...
 use crate::handler::SafekeeperPostgresHandler;
 use crate::SafeKeeperConf;
 use utils::postgres_backend::{AuthType, PostgresBackend};
 
 /// Accept incoming TCP connections and spawn them into a background thread.
-pub fn thread_main(conf: SafeKeeperConf, listener: TcpListener) -> Result<()> {
+pub fn thread_main(
+    conf: SafeKeeperConf,
+    listener: TcpListener,
+    auth: Option<Arc<JwtAuth>>,
```
arssher, line 21: Let's move public key inside `SafekeeperConf`.

arssher (approved):

libs/utils/src/http/endpoint.rs
yeputons: We use `anyhow::Error` instead. It does not implement `std::error::Error` for some reason. Changed to an explicit type.

safekeeper/src/wal_service.rs
yeputons: Will do together with corresponding Pageserver refactoring in a separate PR.

yeputons:
@arssher @petuhovskiy Would you mind checking if the updated `authentication.md` makes sense? It looks very similar to a file describing data flow in Neon, if we have one.

docs/authentication.md
```diff
@@ -1,30 +1,148 @@
...
+but may be not.
+
+### Safekeeper
+#### Overview
+Safekeeper keeps track of multiple tenants, each having multiple timelines.
+
+#### Incoming connections
+Safekeeper accepts connections from Compute/Pageserver, each
+connection corresponds to a specific timeline and requires
+a corresponding JWT token.
```
arssher, line 130: Let mention here config param which specifies public key to check tokens with.

docs/authentication.md
```diff
@@ -1,30 +1,148 @@
...
+but a non-rogue client never accesses another timeline even by an accident:
+timeline IDs are random and hard to guess.
 
-JWT authentication now supports two scopes: tenant and pageserverapi. Tenant scope is intended for use in tenant related api calls, e.g. create_branch. Compute launched for particular tenant also uses this scope. Scope pageserver api is intended to be used by console to manage pageserver. For now we have only one management operation - create tenant.
+#### Incoming connections
+All incoming connections are from PostgreSQL clients.
+Their authentication is just plain PostgreSQL authentication and out of scope for this document.
 
-Examples for token generation in python:
+There is no administrative API except for those provided by PostgreSQL.
```
arssher, line 64: Word 'for' is redundant here at least.

docs/authentication.md
```diff
@@ -1,30 +1,148 @@
...
+For each timeline, it connects to the corresponding Safekeeper.
+Information about "corresponding Safekeeper" is published by Safekeepers
+in the Etcd, but they do not publish access tokens, otherwise what is
+the point of authentication.
+
+Pageserver keeps a connection to some set of Safekeepers, which
+may or may not correspond to active Computes.
+
+_TODO_: describe specific situations when a Pageserver _needs_
+access to a Safekeeper while the Compute is down.
```
arssher, line 94: I think this is out of scope here, but for example when compute finished earlier than pageserver consumed all WAL stream and it needs to catch up.

docs/authentication.md
```diff
@@ -1,30 +1,148 @@
...
+
+_TODO_: what about pageserver replicas? Is it a thing?
```
arssher, line 96: They are not different from main pageserver WRT auth.

docs/authentication.md
```diff
@@ -1,30 +1,148 @@
...
-tenant_token = jwt.encode({"scope": "tenant", "tenant_id": ps.initial_tenant}, auth_keys.priv, algorithm="RS256")
+#### Incoming connections
+Pageserver listens for connections from computes.
+Each compute should present a token valid for the timeline's tenant.
+
+Pageserver also has HTTP API: some parts are per-tenant,
+some parts are server-wide, these are different scopes.
+
+The `auth_type` configuration variable in Pageserver's config may have
+either of three values:
```
petuhovskiy, line 106: Typo?

```suggestion
either of two values:
```

petuhovskiy:
Great doc, thanks @yeputons! Will review the code after the rebase, but it looks mostly ok.

pageserver/src/bin/pageserver.rs
```diff
@@ -256,6 +258,23 @@ fn start_pageserver(conf: &'static PageServerConf) -> anyhow::Result<()> {
     };
     info!("Using auth: {:#?}", conf.auth_type);
 
+    match var("ZENITH_AUTH_TOKEN") {
```
petuhovskiy, line 261: Random thought, can we rename `ZENITH_AUTH_TOKEN` to `NEON_AUTH_TOKEN`? We don't use it extensively yet, should be easy to rename in all places.

libs/utils/src/auth.rs
```diff
@@ -21,8 +21,15 @@ const JWT_ALGORITHM: Algorithm = Algorithm::RS256;
...
 #[serde(rename_all = "lowercase")]
 pub enum Scope {
+    // Provides access to all data for a specific tenant (specified in `struct Claims` below)
+    // TODO: join these two?
     Tenant,
+    // Provides blanket access to all tenants on the pageserver plus pageserver-wide APIs.
+    // Should only be used e.g. for status check/tenant creation/list.
     PageServerApi,
+    // Provides blanket access to all data on the safekeeper plus safekeeper-wide APIs.
+    // Should only be used e.g. for status check.
```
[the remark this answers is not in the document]
yeputons: Yes, at the moment, adding a comment.

yeputons:
Follow-up PRs and TODOs:

* [ ] `ZENITH_AUTH_TOKEN` --> `NEON_AUTH_TOKEN`
* [ ] Move move public key inside `SafekeeperConf`/`PageserverConf` and load it on start instead of passing a separate argument through chains of calls (`auth_validation_public_key_path` --> `auth_validation_public_key`).
* [ ] `Claims`: make it more structured. E.g. make `Scope::Tenant, tenant_id=None` impossible.
* [ ] Think about generating tokens for debug purposes when SSHed to a staging machine. E.g. `curl http://localhost:12345/some-pageserver-api` is useful

pageserver/src/auth.rs
```diff
@@ -0,0 +1,22 @@
...
+use utils::auth::{Claims, Scope};
+use utils::id::TenantId;
+
+pub fn check_permission(claims: &Claims, tenant_id: Option<TenantId>) -> Result<()> {
+    match (&claims.scope, tenant_id) {
+        (Scope::Tenant, None) => {
+            bail!("Attempt to access management api with tenant scope. Permission denied")
+        }
+        (Scope::Tenant, Some(tenant_id)) => {
+            if claims.tenant_id.unwrap() != tenant_id {
```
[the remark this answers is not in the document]
yeputons: I'd prefer a separate PR: merge `Claims`+`Scope` into a single `enum` so it's impossible to construct incorrect data. Does that work?

control_plane/src/compute.rs
yeputons: Finally decided to keep `NULL` the default as the current default is no-auth.

pageserver/src/auth.rs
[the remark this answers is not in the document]
petuhovskiy: Should work, thanks!

yeputons:
Waiting for #2863 to be reviewed, approved, and merged before this.

yeputons:
Rebased on top of #2863 (which is to be merged as soon as tests pass), no big conflicts with either `main` or the other PR.

Will merge this right PR after #2863, if there are no objections.

Issue #2419, opened 2022-09-09
Add logging with timeline_id that operation failed in some pageserver http handler


ololobus:
There was a project on staging from September 7 created by periodic load https://console.stage.neon.tech/admin/operations?project_id=rough-queen-354759

Operation `create_timeline` failed on first attempt with error `pageserver responded with status 500`, but there were not any related errors in the pageserver log for this timeline during this interval https://observer.zenith.tech/d/8jHyEQOnz/zenith-logs-pageserver?orgId=1&var-environment=staging&var-instance=zenith-us-stage-ps-3.local&var-search=caf8609024a1104786c2ea7ea5e2b2b1&var-exclude=substing%20to%20exclude&from=1662558329375&to=1662558720139

Dmitry Rodionov pointed me to the error with trace [https://observer.zenith.tech/d/8jHyEQOnz/zenith-logs-pageserver?orgId=1&var-environment[…]&var-exclude=gc%20loop&from=1662558329375&to=1662558720139](https://observer.zenith.tech/d/8jHyEQOnz/zenith-logs-pageserver?orgId=1&var-environment=staging&var-instance=zenith-us-stage-ps-3.local&var-search=ERROR%20Error%20processing%20HTTP%20request&var-exclude=gc%20loop&from=1662558329375&to=1662558720139)

But it doesn't contain `timeline_id`, which makes searching for errors more complicated.

Let's either somehow propagate `timeline_id` to http server (which looks a bit as an abstraction leaking) or log about operation failure before that in the pageserver code. Or there are better ideas?

sharnoff:
Closing after discussion with @﻿hlinnaka - this should already be fixed:

While `ApiError::into_response` doesn't log anything...

https://github.com/neondatabase/neon/blob/687ba81366491c84250e4df9c20e31e595b34476/libs/utils/src/http/error.rs#L27-L49

... **but!** the only place it's called is in the nearby `handler` function, which always does:

https://github.com/neondatabase/neon/blob/687ba81366491c84250e4df9c20e31e595b34476/libs/utils/src/http/error.rs#L78-L83

For completion, `handler` is only used when it's attached as the error handler for the router:

https://github.com/neondatabase/neon/blob/687ba81366491c84250e4df9c20e31e595b34476/libs/utils/src/http/endpoint.rs#L57-L62

ololobus:
> but! the only place it's called is in the nearby handler function, which always does:

[...]

PR #2459, opened 2022-09-16
Cleanup/simplify logical size calculation

sharnoff:
Noticed this while looking at sources of errors for #2419.

I believe this should produce identical results; replaces an error case that shouldn't be possible with `expect`. Happy to revert bits of this if that's not correct or if the old style is preferred.


SomeoneToIgnore (approved):
> old style is preferred

I don't think there's such thing and the confusion here comes from my oddities forcing me into avoiding panics at all costs.

I like this code more and I will try to learn to panic more sometimes 🙂

sharnoff:
Ah, makes sense. No worries :)

PR #2467, opened 2022-09-16
[proxy] Pass extra parameters to the console

funbringer:
With this change we now pass additional params to the console's auth methods.

### DIFF
--- a/proxy/src/auth/backend/console.rs
+++ b/proxy/src/auth/backend/console.rs
@@ -1,12 +1,12 @@
 //! Cloud API V2.
 
+use super::ConsoleReqExtra;
 use crate::{
     auth::{self, AuthFlow, ClientCredentials},
     compute::{self, ComputeConnCfg},
     error::{io_error, UserFacingError},
-    scram,
+    http, scram,
     stream::PqStream,
-    url::ApiUrl,
 };
 use serde::{Deserialize, Serialize};
 use std::future::Future;
@@ -120,14 +120,23 @@ pub enum AuthInfo {
 
 #[must_use]
 pub(super) struct Api<'a> {
-    endpoint: &'a ApiUrl,
+    endpoint: &'a http::Endpoint,
+    extra: &'a ConsoleReqExtra<'a>,
     creds: &'a ClientCredentials<'a>,
 }
 
 impl<'a> Api<'a> {
     /// Construct an API object containing the auth parameters.
-    pub(super) fn new(endpoint: &'a ApiUrl, creds: &'a ClientCredentials) -> Self {
-        Self { endpoint, creds }
+    pub(super) fn new(
+        endpoint: &'a http::Endpoint,
+        extra: &'a ConsoleReqExtra<'a>,
+        creds: &'a ClientCredentials,
+    ) -> Self {
+        Self {
+            endpoint,
+            extra,
+            creds,
+        }
     }
 
     /// Authenticate the existing user or throw an error.
@@ -139,16 +148,22 @@ impl<'a> Api<'a> {
     }
 
     async fn get_auth_info(&self) -> Result<AuthInfo, GetAuthInfoError> {
-        let mut url = self.endpoint.clone();
-        url.path_segments_mut().push("proxy_get_role_secret");
-        url.query_pairs_mut()
-            .append_pair("project", self.creds.project().expect("impossible"))
-            .append_pair("role", self.creds.user);
+        let req = self
+            .endpoint
+            .get("proxy_get_role_secret")
+            .header("X-Request-ID", uuid::Uuid::new_v4().to_string())
+            .query(&[("session_id", self.extra.session_id)])
+            .query(&[
+                ("application_name", self.extra.application_name),
+                ("project", Some(self.creds.project().expect("impossible"))),
+                ("role", Some(self.creds.user)),
+            ])
+            .build()?;
 
         // TODO: use a proper logger
-        println!("cplane request: {url}");
+        println!("cplane request: {}", req.url());
 
-        let resp = reqwest::get(url.into_inner()).await?;
+        let resp = self.endpoint.execute(req).await?;
         if !resp.status().is_success() {
             return Err(TransportError::HttpStatus(resp.status()).into());
         }
@@ -162,15 +177,21 @@ impl<'a> Api<'a> {
 
     /// Wake up the compute node and return the corresponding connection info.
     pub(super) async fn wake_compute(&self) -> Result<ComputeConnCfg, WakeComputeError> {
-        let mut url = self.endpoint.clone();
-        url.path_segments_mut().push("proxy_wake_compute");
-        url.query_pairs_mut()
-            .append_pair("project", self.creds.project().expect("impossible"));
+        let req = self
+            .endpoint
+            .get("proxy_wake_compute")
+            .header("X-Request-ID", uuid::Uuid::new_v4().to_string())
+            .query(&[("session_id", self.extra.session_id)])
+            .query(&[
+                ("application_name", self.extra.application_name),
+                ("project", Some(self.creds.project().expect("impossible"))),
+            ])
+            .build()?;
 
         // TODO: use a proper logger
-        println!("cplane request: {url}");
+        println!("cplane request: {}", req.url());
 
-        let resp = reqwest::get(url.into_inner()).await?;
+        let resp = self.endpoint.execute(req).await?;
         if !resp.status().is_success() {
             return Err(TransportError::HttpStatus(resp.status()).into());
         }
--- a/proxy/src/http/server.rs
+++ b/proxy/src/http/server.rs
@@ -1,0 +1,27 @@
+use anyhow::anyhow;
+use hyper::{Body, Request, Response, StatusCode};
+use std::net::TcpListener;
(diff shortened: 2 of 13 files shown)


proxy/src/http/server.rs
```diff
@@ -0,0 +1,27 @@
+use anyhow::anyhow;
```
LizardWizzard, line 1: Looks a bit out of scope according to commit message. Will this check be used by k8s healthcheck system? If so should we make changes to helm charts?

proxy/src/main.rs
```diff
@@ -36,23 +36,6 @@ async fn flatten_err(
     f.map(|r| r.context("join error").and_then(|x| x)).await
 }
 
-/// A proper parser for auth backend parameter.
-impl clap::ValueEnum for auth::BackendType<()> {
```
LizardWizzard, line 40: Was nice to have that in one place instead of specifying manually list of variants and match around them. I understand that it's no longer that simple because of new parameters in backend type variants. Should be possible to solve with additional BackendTypeKind (name is already awkward), but probably it would cause only mental overhead without any benefit, so likely it is not worth it

proxy/src/http/server.rs
```diff
@@ -0,0 +1,27 @@
...
+async fn status_handler(_: Request<Body>) -> Result<Response<Body>, ApiError> {
+    json_response(StatusCode::OK, "")
+}
+
+fn make_router() -> RouterBuilder<hyper::Body, ApiError> {
+    let router = endpoint::make_router();
+    router.get("/v1/status", status_handler)
+}
+
+pub async fn thread_main(http_listener: TcpListener) -> anyhow::Result<()> {
```
LizardWizzard, line 15: Nit: since it is not a thread 
```suggestion
pub async fn task_main(http_listener: TcpListener) -> anyhow::Result<()> {
```

LizardWizzard (approved):
LGTM

Liked your Cow usage pattern

Minor comments

proxy/src/http/server.rs
funbringer: This misnomer is out of scope but I get what you're saying, will change in a follow-up PR.

proxy/src/main.rs
funbringer: Yeah, I don't think it's worth it, especially since I grew tired of current parser impl and will likely rewrite it using something derive-based.

funbringer: Furthermore, I don't believe that verbose impl was better anyway. In a hindsight, the proposed way to define variants by returning a slice of allowed names doesn't look definitively superior. Especially considering the fact that I immediately had to drop it once things got a bit trickier.

proxy/src/http/server.rs
LizardWizzard: My bad, it is existing code that was just moved from previous location and I didnt spot that

proxy/src/auth/backend/console.rs
```diff
@@ -139,16 +148,21 @@ impl<'a> Api<'a> {
...
+        let req = self
+            .endpoint
+            .get("proxy_get_role_secret")
+            .header("X-Request-ID", self.extra.session_id)
+            .query(&[
+                ("application_name", self.extra.application_name),
+                ("project", Some(self.creds.project().expect("impossible"))),
+                ("role", Some(self.creds.user)),
+            ])
+            .build()?;
```
funbringer, line 160: @ololobus do you think this makes sense? I'm specifically interested in your opinion on `X-Request-ID` & `application_name`. See a similar request below in the same file.

ololobus: Strong +1 for request id as console already can digest it and add into context. Not sure that it's properly used everywhere after that, though.

Yet, if it's absent console generated new request id like this: https://github.com/neondatabase/cloud/blob/fa7a5db09a24c5d519bbd75f7ca723dc7bdcd822/goapp/internal/requestid/requestid.go#L19

So I wonder, would it be better if proxy will pass `X-Request-ID` in the same format for consistency? Or we can agree on some different format, but use it everywhere

ololobus: Although, passing `session_id` seems to be important too, as we do two requests per session: get secret and wake up

proxy/src/auth/backend/console.rs
```diff
@@ -139,16 +148,21 @@ impl<'a> Api<'a> {
...
 
         // TODO: use a proper logger
-        println!("cplane request: {url}");
+        println!("cplane request: {}", req.url());
```
ololobus, line 163: Yes, there is a todo, but maybe log this new request context already? I mean project id, request / session id, maybe even application name

funbringer: There's a task for that, so let's do this in a separate PR.

proxy/src/auth/backend/console.rs
ololobus: Ah, missed this part, sorry. Then I even more confused as I thought that `session_id` is what is shown to the user, when one does `psql -h connect.neon.tech`

So this `session_id` passed as `X-Request-ID` in this PR has nothing common with internal proxy session id?

antonyc: It is nice to have a unique Request ID per every outgoing HTTP request to the console. If we reuse Request ID for different requests and one of them has failed, then one who got a Request ID of a failed request from proxy logs, will not have the ability grep console logs by this request id and differ one request from another. IIUC `self.extra.session_id` is used as a Request ID for different HTTP requests. Can we change that to a unique value per request?

funbringer: > So this session_id passed as X-Request-ID in this PR has nothing common with internal proxy session id?

To be clear, there's no such thing as an "internal proxy session id". What we have here is a per-client-connection uuid that's used one time for either one or two http requests. When we accept a new connection, we query auth info and then wake up the node, hence 1 or 2 calls. Mind you, those requests are different, there's never a case where you use the same id for the same kind of request, so the problem you're talking about is impossible.

I call it a session id because that's what it is, but that doesn't mean that we'll use this for 2 or more wakeup calls.

funbringer: Alternatively, we could do as suggested earlier by @ololobus: pass two IDs: one unique (that's `X-Request-ID`) and the other one (idk, `X-Session-ID`?) as it is now.

ololobus: Yep, I think that two ids are better:
- unique `request_id` per each http request to the control-plane
- unique `session_id` per each client session

Ideally, that's what I want to see in the proxy logs:
```
INFO: new client connection, session_id: xxx1
INFO: control-plane get_auth request finished with 200 OK, session_id: xxx1, request_id: yyy1
INFO: control-plane wakeup request finished with 200 OK, session_id: xxx1, request_id: yyy2
INFO: client connected to the compute node, session_id: xxx1
INFO: new client connection, session_id: xxx2
ERROR: control-plane get_auth request finished with 500 InternalServerError, session_id: xxx2, request_id: yyy3, error: I'm tired
ERROR: client connection attempt failed, session: xxx2
```

That way, we can transparently track any particular connection attempt **AND** any particular control-plane request. And this `session_id` should be proxy internal id for the session and it should be logged for each log record corresponding to this session. Name `session_id` totally makes sense to me, BTW.

Does it make sense to you? If yes and you'll show me the issue for proxy logging you mentioned above, I can copy-paste it there :)

--

Sorry, I'm stupid, but what is this **565752694af9abf1** in the `https://console.stage.neon.tech/psql_session/565752694af9abf1`? Is it `session_id`? Does it have the same format for scram-proxy?

funbringer: > Yep, I think that two ids are better:
> * unique request_id per each http request to the control-plane
> * unique session_id per each client session

Fine, let's do this :)

> And this session_id should be proxy internal id for the session and it should be logged for each log record corresponding to this session. 

That's exactly the plan, I just don't want to do everything in one PR, because that'd be much more code.

> Sorry, I'm stupid, but what is this 565752694af9abf1 in the https://console.stage.neon.tech/psql_session/565752694af9abf1? Is it session_id? Does it have the same format for scram-proxy?

No, it's just a random `hex(u64)`. `scram-proxy` doesn't use that, in fact, we have 3 auth backends: console (uses scram exclusively), postgres (to debug scram locally) and link (which uses this u64). This leads me to think that our current naming is not as good is it could be: we should've named those instances `proxy-link` and `proxy-console`.

proxy/src/auth/backend/console.rs
```diff
@@ -139,16 +148,22 @@ impl<'a> Api<'a> {
...
-        let mut url = self.endpoint.clone();
-        url.path_segments_mut().push("proxy_get_role_secret");
-        url.query_pairs_mut()
-            .append_pair("project", self.creds.project().expect("impossible"))
-            .append_pair("role", self.creds.user);
+        let req = self
+            .endpoint
+            .get("proxy_get_role_secret")
+            .header("X-Request-ID", uuid::Uuid::new_v4().to_string())
+            .header("X-Session-ID", self.extra.session_id.to_string())
```
funbringer, line 155: Per discussion above.

funbringer: If you say so. Changed to a query param.

PR #2468, opened 2022-09-16
test_runner: pass password to pgbench via PGPASSWORD

bayandin:
These PR changes approach taken in https://github.com/neondatabase/neon/pull/2434. 
Now we process all files that are used for report creating.

### DIFF
--- a/test_runner/fixtures/neon_fixtures.py
+++ b/test_runner/fixtures/neon_fixtures.py
@@ -287,6 +287,7 @@ class PgProtocol:
         if "dsn" in kwargs:
             result.update(parse_dsn(kwargs["dsn"]))
         result.update(kwargs)
+        result = {k: v for k, v in result.items() if v is not None}
 
         # Individual statement timeout in seconds. 2 minutes should be
         # enough for our tests, but if you need a longer, you can
--- a/test_runner/fixtures/neon_fixtures.py
+++ b/test_runner/fixtures/neon_fixtures.py
@@ -283,6 +283,10 @@ class PgProtocol:
         return str(make_dsn(**self.conn_options(**kwargs)))
 
     def conn_options(self, **kwargs):
+        """
+        Construct a dictionary of connection options from default values and extra parameters.
+        An option can be dropped from the returning dictionary by None-valued extra parameter.
+        """
         result = self.default_options.copy()
         if "dsn" in kwargs:
             result.update(parse_dsn(kwargs["dsn"]))
--- a/test_runner/fixtures/log_helper.py
+++ b/test_runner/fixtures/log_helper.py
@@ -1,6 +1,5 @@
 import logging
 import logging.config
-import re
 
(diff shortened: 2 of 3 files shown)


LizardWizzard (commented):
Do we have an idea of places where passwords still occur? if passwords only occur in command itself i e "executing ..." we can have some bool flag to mask only command, not the output. We can even refactor it so there is a connstr for actual execution, and connstr for showing it in the logs. It can be done with second optional argument, something like "display_cmd". 

Having to scan all the files in the report looks a bit overkill for the task to me.

bayandin:
> Do we have an idea of places where passwords still occur? 

In exception messages.

> Having to scan all the files in the report looks a bit overkill for the task to me.

That was exactly my thought, but unfortunately, there's a possibility that someone will use `subprocess.run` (instead of our wrapper), and it will expose passwords (even before we had a chance to review the PR and suggest not to use it). Another case is that a program we're running with `subprocess_capture` will write a password to stdout/stderr, which we capture and attach to the report.
I don't really have another idea of how not only to fix the issue but also try to prevent it to appear again in the future.

LizardWizzard:
> In exception messages. 

Is it our exception or some third party library? If its ours I think its better to fix it on our side

> there's a possibility that someone will use subprocess.run (instead of our wrapper)

The same is possible if password is emitted in a form that doesnt match current regex.

The proper way IMO is to clean up our API, so it is less possible that someone will misuse it. The problem is always caused by invoking postgres binaries and passing connection string to them. So should be possible to fix it with api changes that are not that big. See my previous suggestions, I can elaborate if needed.

Additionally we can set up a lint, (maybe write our own flake8 plugin or something) and require that lint to pass before we run tests that use outside connection credentials.

IMO best option would be to completely remove sensitive data from command line. Password can be passed via an environment variable and we can wrap all pg binaries we use to avoid this in the future for common cases

bayandin:
Thanks for the feedback @LizardWizzard & @funbringer!
Ok, now the password is passing to pgbench via `PGPASSWORD`

test_runner/fixtures/neon_fixtures.py
```diff
@@ -287,6 +287,7 @@ def conn_options(self, **kwargs):
         if "dsn" in kwargs:
             result.update(parse_dsn(kwargs["dsn"]))
         result.update(kwargs)
+        result = {k: v for k, v in result.items() if v is not None}
```
LizardWizzard, line 290: Curious what can yield None value. Is it password?

LizardWizzard (approved):
Looks good to me, thanks!

test_runner/fixtures/neon_fixtures.py
bayandin: Yep, I use it to drop the password from the connstr (when it's set in default_options).
Worth to add a comment for it.

PR #2477, opened 2022-09-19
Adjust spans around gc and compaction

LizardWizzard:
So compaction and gc loops have their own span to always show tenant id in log messages.


SomeoneToIgnore (approved):
It appears, that we can shorten the span strings, if agree on some kind of a `[{tenant_id}/{timeline_id}]` notation, but not necessary to do now.

LizardWizzard:
> if agree on some kind of a [{tenant_id}/{timeline_id}] notation

In some cases when common per tenant part is detached from per tenant parts we cant create common span, so the other option is to nest them as it is done in this PR, maybe there are better options, but current way looks natural to me

PR #2482, opened 2022-09-20
Regenerate Hakari files, add a CI check for that

SomeoneToIgnore:


hlinnaka (approved):

SomeoneToIgnore:
CI fails, as expected: https://github.com/neondatabase/neon/actions/runs/3087805470/jobs/4993557829#step:15:225

Adding `cargo install cargo-hakari` takes ~4m extra time, so a better approach would be to install this into our `build:pinned` custom image and use it from there, but that requires more things to be done before, IMO we can continue as is for now.

PR #2488, opened 2022-09-21
Use prebuilt image with Hakari for CI style checks

SomeoneToIgnore:
Image PR: https://github.com/neondatabase/build/pull/19

### DIFF
--- a/.github/workflows/codestyle.yml
+++ b/.github/workflows/codestyle.yml
@@ -108,19 +108,31 @@ jobs:
             target
           key: v4-${{ runner.os }}-cargo-${{ hashFiles('./Cargo.lock') }}-rust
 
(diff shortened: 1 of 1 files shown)


.github/workflows/codestyle.yml
```diff
@@ -108,20 +108,32 @@ jobs:
...
 
+  check-rust-dependencies:
+    runs-on: dev
+    container:
+      image: 369495373322.dkr.ecr.eu-central-1.amazonaws.com/rust:pinned
+      options: --init
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
```
bayandin, line 125:
```suggestion
        uses: actions/checkout@v3
```

I don't really know the difference, but sometimes we use v3

bayandin (approved):

PR #2489, opened 2022-09-21
Use uninit mark files during timeline init for atomic creation

SomeoneToIgnore:
Part of https://github.com/neondatabase/neon/pull/2239

Regular, from scratch, timeline creation involves `initdb` to be run in a separate directory, data from this directory to be imported into pageserver and, finally, timeline-related background tasks to start.

This PR ensures we don't leave behind any directories that are not marked as temporary and that pageserver removes such directories on restart, allowing timeline creation to be retried with the same IDs, if needed.

### DIFF
--- a/libs/utils/src/crashsafe_dir.rs
+++ b/libs/utils/src/crashsafe_dir.rs
@@ -12,16 +12,8 @@ pub fn create_dir(path: impl AsRef<Path>) -> io::Result<()> {
     let path = path.as_ref();
 
     fs::create_dir(path)?;
-    File::open(path)?.sync_all()?;
-
-    if let Some(parent) = path.parent() {
-        File::open(parent)?.sync_all()
-    } else {
-        Err(io::Error::new(
-            io::ErrorKind::InvalidInput,
-            "can't find parent",
-        ))
-    }
+    fsync_file_and_parent(path)?;
+    Ok(())
 }
 
 /// Similar to [`std::fs::create_dir_all`], except we fsync all
@@ -65,12 +57,12 @@ pub fn create_dir_all(path: impl AsRef<Path>) -> io::Result<()> {
 
     // Fsync the created directories from child to parent.
     for &path in dirs_to_create.iter() {
-        File::open(path)?.sync_all()?;
+        fsync(path)?;
     }
 
     // If we created any new directories, fsync the parent.
     if !dirs_to_create.is_empty() {
-        File::open(path)?.sync_all()?;
+        fsync(path)?;
     }
 
     Ok(())
@@ -90,6 +82,44 @@ pub fn path_with_suffix_extension(original_path: impl AsRef<Path>, suffix: &str)
     original_path
         .as_ref()
         .with_extension(new_extension.as_ref())
+}
+
+pub fn fsync_file_and_parent(file_path: &Path) -> io::Result<()> {
+    let parent = file_path.parent().ok_or_else(|| {
+        io::Error::new(
+            io::ErrorKind::Other,
+            format!("File {} has no parent", file_path.display()),
+        )
+    })?;
+
+    fsync(file_path)?;
+    fsync(parent)?;
+    Ok(())
+}
+
+pub fn fsync(path: &Path) -> io::Result<()> {
+    File::open(path)
+        .map_err(|e| {
+            io::Error::new(
+                io::ErrorKind::Other,
+                format!("Failed to open the file: {e}"),
+            )
+        })
+        .and_then(|file| {
+            file.sync_all()
+                .map_err(|e| {
+                    io::Error::new(
+                        io::ErrorKind::Other,
+                        format!("Failed to sync file metadata: {e}"),
+                    )
+                })
+                .map_err(|e| {
+                    io::Error::new(
+                        io::ErrorKind::Other,
+                        format!("Failed to fsync file {}: {}", path.display(), e),
+                    )
+                })
+        })
 }
 
 #[cfg(test)]
--- a/libs/utils/src/crashsafe_dir.rs
+++ b/libs/utils/src/crashsafe_dir.rs
@@ -99,25 +99,17 @@ pub fn fsync_file_and_parent(file_path: &Path) -> io::Result<()> {
 
 pub fn fsync(path: &Path) -> io::Result<()> {
     File::open(path)
-        .map_err(|e| {
-            io::Error::new(
-                io::ErrorKind::Other,
-                format!("Failed to open the file {path:?}: {e}"),
-            )
-        })
+        .map_err(|e| io::Error::new(e.kind(), format!("Failed to open the file {path:?}: {e}")))
         .and_then(|file| {
             file.sync_all()
                 .map_err(|e| {
                     io::Error::new(
-                        io::ErrorKind::Other,
+                        e.kind(),
                         format!("Failed to sync file {path:?} data and metadata: {e}"),
                     )
                 })
                 .map_err(|e| {
-                    io::Error::new(
-                        io::ErrorKind::Other,
-                        format!("Failed to fsync file {path:?}: {e}"),
-                    )
+                    io::Error::new(e.kind(), format!("Failed to fsync file {path:?}: {e}"))
                 })
         })
 }
--- a/libs/utils/src/crashsafe_dir.rs
+++ b/libs/utils/src/crashsafe_dir.rs
@@ -88,7 +88,7 @@ pub fn fsync_file_and_parent(file_path: &Path) -> io::Result<()> {
     let parent = file_path.parent().ok_or_else(|| {
         io::Error::new(
             io::ErrorKind::Other,
-            format!("File {} has no parent", file_path.display()),
+            format!("File {file_path:?} has no parent"),
         )
     })?;
 
@@ -102,7 +102,7 @@ pub fn fsync(path: &Path) -> io::Result<()> {
         .map_err(|e| {
             io::Error::new(
                 io::ErrorKind::Other,
-                format!("Failed to open the file: {e}"),
+                format!("Failed to open the file {path:?}: {e}"),
             )
         })
         .and_then(|file| {
@@ -110,13 +110,13 @@ pub fn fsync(path: &Path) -> io::Result<()> {
                 .map_err(|e| {
                     io::Error::new(
                         io::ErrorKind::Other,
-                        format!("Failed to sync file metadata: {e}"),
+                        format!("Failed to sync file {path:?} data and metadata: {e}"),
                     )
                 })
                 .map_err(|e| {
                     io::Error::new(
                         io::ErrorKind::Other,
-                        format!("Failed to fsync file {}: {}", path.display(), e),
+                        format!("Failed to fsync file {path:?}: {e}"),
                     )
                 })
         })
--- a/pageserver/src/config.rs
+++ b/pageserver/src/config.rs
@@ -25,7 +25,7 @@ use crate::tenant_config::{TenantConf, TenantConfOpt};
 
 /// The name of the metadata file pageserver creates per timeline.
 pub const METADATA_FILE_NAME: &str = "metadata";
-pub const TIMELINE_TOMBSTONE_SUFFIX: &str = "___tombstone";
+pub const TIMELINE_UNINIT_MARK_SUFFIX: &str = "___uninit";
 const TENANT_CONFIG_NAME: &str = "config";
 
 pub mod defaults {
@@ -366,14 +366,14 @@ impl PageServerConf {
         self.timelines_path(tenant_id).join(timeline_id.to_string())
     }
 
-    pub fn timeline_tombstone_file_path(
+    pub fn timeline_uninit_mark_file_path(
         &self,
         tenant_id: TenantId,
         timeline_id: TimelineId,
     ) -> PathBuf {
         path_with_suffix_extension(
             self.timeline_path(&timeline_id, &tenant_id),
-            TIMELINE_TOMBSTONE_SUFFIX,
+            TIMELINE_UNINIT_MARK_SUFFIX,
         )
     }
 
--- a/pageserver/src/config.rs
+++ b/pageserver/src/config.rs
@@ -7,7 +7,7 @@
 use anyhow::{anyhow, bail, ensure, Context, Result};
 use remote_storage::RemoteStorageConfig;
 use std::env;
-use utils::crashsafe_dir::path_with_suffix_extension;
+use utils::crashsafe::path_with_suffix_extension;
 
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
--- a/pageserver/src/config.rs
+++ b/pageserver/src/config.rs
@@ -7,6 +7,7 @@
 use anyhow::{anyhow, bail, ensure, Context, Result};
 use remote_storage::RemoteStorageConfig;
 use std::env;
+use utils::crashsafe_dir::path_with_suffix_extension;
 
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
@@ -24,6 +25,7 @@ use crate::tenant_config::{TenantConf, TenantConfOpt};
 
 /// The name of the metadata file pageserver creates per timeline.
 pub const METADATA_FILE_NAME: &str = "metadata";
+pub const TIMELINE_TOMBSTONE_SUFFIX: &str = "___tlitmbstn";
 const TENANT_CONFIG_NAME: &str = "config";
 
 pub mod defaults {
@@ -362,6 +364,17 @@ impl PageServerConf {
 
     pub fn timeline_path(&self, timeline_id: &TimelineId, tenant_id: &TenantId) -> PathBuf {
         self.timelines_path(tenant_id).join(timeline_id.to_string())
+    }
+
+    pub fn timeline_tombstone_file_path(
+        &self,
+        tenant_id: TenantId,
+        timeline_id: TimelineId,
+    ) -> PathBuf {
+        path_with_suffix_extension(
+            self.timeline_path(&timeline_id, &tenant_id),
+            TIMELINE_TOMBSTONE_SUFFIX,
+        )
     }
 
     /// Points to a place in pageserver's local directory,
--- a/pageserver/src/config.rs
+++ b/pageserver/src/config.rs
@@ -25,7 +25,7 @@ use crate::tenant_config::{TenantConf, TenantConfOpt};
 
 /// The name of the metadata file pageserver creates per timeline.
 pub const METADATA_FILE_NAME: &str = "metadata";
-pub const TIMELINE_TOMBSTONE_SUFFIX: &str = "___tlitmbstn";
+pub const TIMELINE_TOMBSTONE_SUFFIX: &str = "___tombstone";
 const TENANT_CONFIG_NAME: &str = "config";
 
 pub mod defaults {
--- a/pageserver/src/tenant.rs
+++ b/pageserver/src/tenant.rs
@@ -122,10 +122,16 @@ pub struct Tenant {
     upload_layers: bool,
 }
 
-pub struct UnInitializedTimeline<'t> {
+#[must_use]
+pub struct UninitializedTimeline<'t> {
     owning_tenant: &'t Tenant,
     timeline_id: TimelineId,
-    raw_timeline: Option<Timeline>,
+    raw_timeline: Option<(Timeline, TimelineTombstone)>,
+}
+
+struct TimelineTombstone {
+    remove_tombstone: bool,
+    tombstone_path: PathBuf,
 }
 
 impl UninitializedTimeline<'_> {
@@ -141,13 +147,15 @@ impl UninitializedTimeline<'_> {
         let timeline_id = self.timeline_id;
         let tenant_id = self.owning_tenant.tenant_id;
 
-        let new_timeline = Arc::new(self.raw_timeline.take().with_context(|| {
+        let (new_timeline, tombstone) = self.raw_timeline.take().with_context(|| {
             format!("No metadata for initalization found for timeline {tenant_id}/{timeline_id}")
-        })?);
+        })?;
+        let new_timeline = Arc::new(new_timeline);
 
         let new_disk_consistent_lsn = new_timeline.get_disk_consistent_lsn();
-        ensure!(new_disk_consistent_lsn.is_valid(),
-            "Timeline {tenant_id}/{timeline_id} has invalid disk_consistent_lsn and cannot be initialized");
+        // TODO it would be good to ensure that, but apparently a lot of our testing is dependend on that at least
+        // ensure!(new_disk_consistent_lsn.is_valid(),
+        //     "Timeline {tenant_id}/{timeline_id} has invalid disk_consistent_lsn and cannot be initialized");
 
         match timelines.entry(timeline_id) {
             hash_map::Entry::Occupied(_) => anyhow::bail!(
@@ -159,8 +167,7 @@ impl UninitializedTimeline<'_> {
                     .context("failed to load layermap")?;
                 new_timeline.launch_wal_receiver()?;
 
-                self.remove_tombstone()
-                    .context("Failed to remove timeline tombstone during init")?;
+                drop(tombstone);
 
                 v.insert(Arc::clone(&new_timeline));
             }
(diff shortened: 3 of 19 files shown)


pageserver/src/tenant.rs
```diff
@@ -131,6 +133,257 @@ pub enum TenantState {
     Broken,
 }
 
+/// A timeline with some of its files on disk, being initialized.
+/// The struct API ensures the atomicity of the timeline init: it's either properly created and inserted into pageserver's memory, or
+/// its local files removed (at the worst case of a crash, left with a mark to get removed on the next restart).
+///
+/// A caller is responsible for proper timeline data filling before the final init.
```
hlinnaka, line 140:
```suggestion
/// A timeline with some of its files on disk, being initialized.
/// The struct API ensures the atomicity of the timeline init: it's either properly created and inserted into pageserver's memory, or
/// its local files removed (at the worst case of a crash, left with a mark to get removed on the next restart).
///
/// A caller is responsible for proper timeline data filling before the final init.
```
```suggestion
/// A timeline with some of its files on disk, being initialized.
/// This struct ensures the atomicity of the timeline init: it's either properly created and inserted into pageserver's memory, or
/// its local files are removed. In the worst case of a crash, a tombstone file is left behind, which causes the directory
/// to be removed on next restart.
///
/// The caller is responsible for proper timeline data filling before the final init.
```

pageserver/src/tenant.rs
```diff
@@ -131,6 +133,257 @@ pub enum TenantState {
...
+            let raw_timeline_dir = &tombstone.timeline_path;
+            let mut should_remove_tombstone = true;
+            match fs::remove_dir_all(raw_timeline_dir) {
+                Ok(()) => info!("Timeline dir removed successfully, removing the tombstone"),
+                Err(e) if e.kind() == io::ErrorKind::NotFound => {
+                    info!("Timeline dir is absent, skipping and removing the tombstone")
+                }
+                Err(e) => {
+                    should_remove_tombstone = false;
+                    error!("Failed to clean up unitialized timeline directory: {e:?}");
```
hlinnaka, line 303:
```suggestion
                    error!("Failed to clean up uninitialized timeline directory: {e:?}");
```

libs/utils/src/crashsafe_dir.rs
```diff
@@ -92,6 +84,44 @@ pub fn path_with_suffix_extension(original_path: impl AsRef<Path>, suffix: &str)
...
+            format!("File {} has no parent", file_path.display()),
+        )
+    })?;
+
+    fsync(file_path)?;
+    fsync(parent)?;
+    Ok(())
+}
+
+pub fn fsync(path: &Path) -> io::Result<()> {
```
hlinnaka, line 100: This seems really verbose. What was wrong with just `File::open(path)?.sync_all()?` ? Don't the io::Errors contain enough context?

libs/utils/src/crashsafe_dir.rs
```diff
@@ -92,6 +84,44 @@ pub fn path_with_suffix_extension(original_path: impl AsRef<Path>, suffix: &str)
...
+                io::ErrorKind::Other,
+                format!("Failed to open the file: {e}"),
+            )
+        })
+        .and_then(|file| {
+            file.sync_all()
+                .map_err(|e| {
+                    io::Error::new(
+                        io::ErrorKind::Other,
+                        format!("Failed to sync file metadata: {e}"),
```
hlinnaka, line 113: `sync_all` flushes data and metadata. We don't know what exactly failed.

pageserver/src/config.rs
```diff
@@ -24,6 +25,7 @@ use crate::tenant_config::{TenantConf, TenantConfOpt};
 
 /// The name of the metadata file pageserver creates per timeline.
 pub const METADATA_FILE_NAME: &str = "metadata";
+pub const TIMELINE_TOMBSTONE_SUFFIX: &str = "___tlitmbstn";
```
hlinnaka, line 28: How about something more readable like `_tombstone` or `_DELETE` ?

pageserver/src/tenant.rs
```diff
@@ -131,6 +133,257 @@ pub enum TenantState {
...
+/// XXX: it's important to create it near the timeline dir, not inside it to ensure timeline dir gets removed first.
+#[must_use]
+struct TimelineTombstone {
+    tombstone_deleted: bool,
+    tombstone_path: PathBuf,
+    timeline_path: PathBuf,
+}
+
+impl UninitializedTimeline<'_> {
+    /// Ensures timeline data is valid, loads it into pageserver's memory and removes all temporary marks on success.
```
hlinnaka, line 160:
```suggestion
    /// Ensures timeline data is valid, loads it into pageserver's memory and removes tombstone file on success.
```

pageserver/src/tenant.rs
```diff
@@ -131,6 +133,257 @@ pub enum TenantState {
...
+
+impl TimelineTombstone {
+    /// Useful for initializing timelines, existing on disk after the restart.
+    pub fn dummy() -> Self {
+        Self {
+            tombstone_deleted: true,
+            tombstone_path: PathBuf::new(),
+            timeline_path: PathBuf::new(),
+        }
+    }
```
hlinnaka, line 324: This feels a bit weird. I think changing the tombstone field in UninitializedTimeline into `Option<TimelineTombstone>` would be more natural, and use None when there is no tombstone file.

hlinnaka (commented):
Left some random comments. I'll continue reviewing this later

pageserver/src/tenant.rs
hlinnaka: Or perhaps have different structs for the different cases:
- `CreateInProgressTimeline` for a newly created branch, with tombstone to delete it on crash
- `LoadInProgressTimeline` for a timeline that exists on disk, and we're loading it. No tombstone file.

pageserver/src/config.rs
SomeoneToIgnore: Have picked `_tombstone`.

libs/utils/src/crashsafe_dir.rs
SomeoneToIgnore:
```rust
// in the upper mod
pub fn old_fsync(path: &Path) -> io::Result<()> {
    let file = File::open(path)?;
    file.sync_all()?;
    Ok(())
}

#[test]
    fn fsync_test() {
        println!("{:#?}", super::fsync(Path::new("bvsbsbs")));
        println!("{:#?}", super::old_fsync(Path::new("bvsbsbs")));
    }
```

and the verbose (and a bit adjusted) version will produce this output:

```
Err(
    Custom {
        kind: Other,
        error: "Failed to open the file \"bvsbsbs\": No such file or directory (os error 2)",
    },
)
Err(
    Os {
        code: 2,
        kind: NotFound,
        message: "No such file or directory",
    },
)
```

I always have issues with some io::Error and ? due to zero clue what exactly is wrong, which file is actually missing, etc.

pageserver/src/tenant.rs
SomeoneToIgnore: >  I think changing the tombstone field in UninitializedTimeline into `Option<TimelineTombstone>` would be more natural, and use None when there is no tombstone file.

I've went with the simplest approach later naturally too, and now the tombstone looks like 
```rust
#[must_use]
pub struct UninitializedTimeline<'t> {
    owning_tenant: &'t Tenant,
    timeline_id: TimelineId,
    raw_timeline: Option<(Timeline, TimelineTombstone)>,
}
```
which, presumably matches the proposal.

SomeoneToIgnore:
@hlinnaka , can you recheck the PR again?

I have found the bug, it's somewhat eerie: `insert_historic` method
https://github.com/neondatabase/neon/blob/14c623b254457f1dd4e1b0a64782cf5f568176bf/pageserver/src/tenant/layer_map.rs#L337-L343

allows to add any amount of duplicate layers (with duplicate file names) and those will be later considered "unique" in GC/compaction: https://github.com/neondatabase/neon/blob/14c623b254457f1dd4e1b0a64782cf5f568176bf/pageserver/src/tenant/timeline.rs#L1541-L1544

In my case, the bug was caused by 
```
        raw_timeline.import_timeline_from_postgres_datadir(pgdata_path, pgdata_lsn)?;
        let timeline = raw_timeline.initialize_with_lock(&mut timelines)?;
        drop(timelines);
```

code, where `raw_timeline.import_timeline_from_postgres_datadir` called `timeline.checkpoint(CheckpointConfig::Forced)` and filled the map with the layer, while `initialize_with_lock` called `load_layer_map` and added another, duplicate historic layer.

In this PR, I've changed this to have a proper sequence of actions for the timeline:
https://github.com/neondatabase/neon/pull/2489/commits/08fe22590cf71c27bdc4a34b53a0c899b337c54e

yet the underlying issue is not really fixed and a person is required to add historic layers very careful.

pageserver/src/tenant.rs
SomeoneToIgnore: Applied the 2nd version.

koivunej (commented):
Few quick comments rather than to keep these as draft while I switch back to the other PR.

libs/utils/src/crashsafe_dir.rs
```diff
@@ -92,6 +84,44 @@ pub fn path_with_suffix_extension(original_path: impl AsRef<Path>, suffix: &str)
...
+                format!("Failed to open the file {path:?}: {e}"),
+            )
+        })
+        .and_then(|file| {
+            file.sync_all()
+                .map_err(|e| {
+                    io::Error::new(
+                        io::ErrorKind::Other,
+                        format!("Failed to sync file {path:?} data and metadata: {e}"),
+                    )
```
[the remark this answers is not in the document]
SomeoneToIgnore: Yes, the point is to chain those as `anyhow` does, with extra semantic information.
Anyhow looses the `ErrorKind::NotFound` case and requires some more dances around it which bloats the code, so I've decided to go with the io error.

pageserver/src/tenant.rs
```diff
@@ -1009,48 +1234,89 @@ impl Tenant {
             TEMP_FILE_SUFFIX,
         );
 
-        // Init temporarily repo to get bootstrap data
+        // a tombstone was placed in the parent method, nothing else can access this timeline files
```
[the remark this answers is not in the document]
SomeoneToIgnore: Yes, I've inlined the code in closer to the end and forgotten to adjust the comment, fixed, thanks.

hlinnaka:
Should we call this something else than "tombstone"? A tombstone normally means that something was live earlier, but is now dead, and will be removed in the future. This usage is a little different. The timeline has a tombstone in the beginning, and it's removed when it's fully initialized.

Not sure what the appropriate term would be. Dead-man's switch? Poison pill?

SomeoneToIgnore:
"uninit mark"?

hlinnaka:
> "uninit mark"?

Works for me

libs/utils/src/crashsafe_dir.rs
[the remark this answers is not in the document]
koivunej: The current version does however always return `e.kind() == ErrorKind::Other`, and the `NotFound` is lost. See: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8515fb3fd0969be56bbc1e30d3fa582d

Perhaps the first `.map_err` is extra left over, or should use the `e.kind()`?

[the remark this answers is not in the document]
SomeoneToIgnore: Nice catch, had used parent's kind everywhere.

pageserver/src/tenant_mgr.rs
```diff
@@ -265,58 +265,98 @@ fn create_tenant_files(
...
+        conf,
+        tenant_conf,
+        tenant_id,
+        &temporary_tenant_dir,
+        &target_tenant_directory,
+    );
+
+    if creation_result.is_err() {
+        error!("Failed to create directory structure for tenant {tenant_id}, cleaning tmp data");
+        if let Err(e) = fs::remove_dir_all(&temporary_tenant_dir) {
```
koivunej, line 286: `crashsafe_dir::create_dir_all` is used above and the renames done by `try_create_target_tenant_dir` after writing the config file are fsynced but the removal is not crashsafe. Should there be at least one fsync here in the err path?

pageserver/src/tenant.rs
```diff
@@ -119,6 +121,234 @@ pub struct Tenant {
...
+                )
+            } else {
+                // unblock later timeline creation attempts
+                warn!(
+                    "Removing intermediate uninit mark file {}",
+                    self.uninit_mark_path.display()
+                );
+                if let Err(e) = fs::remove_file(&self.uninit_mark_path) {
+                    error!("Failed to remove the uninit mark file: {e}")
+                }
```
koivunej, line 346: Similar question about fsync after remove as elsewhere, which can be found on the `TimelineUninitMark::remove_uninit_mark`. I wonder if the same piece of code to delete could be called from both `drop` and `remove_uninit_mark`.

koivunej (commented):
Few more comments, mostly concerning fsync after remove or exists+create_truncate. There's a lot of changes in parts I'm mostly seeing for the first time, so unsure if I should approve.

pageserver/src/tenant.rs
```diff
@@ -1063,34 +1341,77 @@ impl Tenant {
...
+        );
+
+        let timeline_path = self.conf.timeline_path(&timeline_id, &tenant_id);
+        anyhow::ensure!(
+            !timeline_path.exists(),
+            "Timeline {} already exists, cannot create its uninit mark file",
+            timeline_path.display()
+        );
+
+        fs::File::create(&uninit_mark_path)
```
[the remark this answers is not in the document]
SomeoneToIgnore: I was trying to be on the safe here, when one of the pageserver parts exits, leaving the unint mark, but it does not matter with this code, since `fs::File::create(&uninit_mark_path)` will fail if the file exists.
Removed the existence check.

pageserver/src/tenant_mgr.rs
SomeoneToIgnore: It won't hurt, added, thank you.

pageserver/src/tenant.rs
SomeoneToIgnore: Indeed, feels like the most of the deletion could be deduplicated, done.

pageserver/src/tenant.rs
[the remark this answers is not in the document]
koivunej: Note: `std::fs::File` explicitly documents that it doesn't fail if the file exists, it truncates: https://doc.rust-lang.org/stable/std/fs/struct.File.html#method.create

I think you are assuming it to be `OpenOptions::create_new` but it's not.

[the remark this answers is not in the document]
SomeoneToIgnore: So, back to your initial comment, implemented. 
I think it's ok generally, since besides `uninit_mark_path`, we could only leave tmp file and tmp initdb dir created.
The latter gets removed if init is attempted again, the former is overwritten, so should be ok.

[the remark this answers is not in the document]
koivunej: > The latter gets removed if init is attempted again, the former is overwritten, so should be ok.

I am not sure if I can see this right away even from the review, spent some time yesterday looking for it. Resolving as I am agreeing that it most likely doesn't matter with this code.

pageserver/src/tenant.rs
```diff
@@ -1063,34 +1344,71 @@ impl Tenant {
...
             )
         })?;
 
-        let ancestor = new_metadata
-            .ancestor_timeline()
-            .and_then(|ancestor_timeline_id| timelines.get(&ancestor_timeline_id))
-            .cloned();
         let new_timeline = self
-            .initialize_new_timeline(new_timeline_id, new_metadata, ancestor)
+            .create_timeline_data(new_timeline_id, new_metadata, ancestor)
```
LizardWizzard, line 1352: If error appears here, then uninit mark is dropped, and timeline will be considered healthy during restart? Am I correct?

SomeoneToIgnore: > I would imagine that timeline is completely initialized in temporary location (with all initdb loading) and then moved to permatent location with a rename.

That is a great idea and I've tried doing it, but the issue is that we have to write into the tmp directory during initdb result imports. 
That would require is to pass the knowledge about the tmp dir from `Tenant` level to `DeltaLayer` (and others') level: right now, we import the data and do checkpointing. During checkpointing, we use `conf.rs` functions to get the timeline path and derive the layer path to write into out of it. 
Changing that would require a lot of code, unless I'm mistaken somehow.

So, that's why the whole tombstone approach as a "simpler" way of doing things.

SomeoneToIgnore: > If error appears here, then uninit mark is dropped, and timeline will be considered healthy during restart? Am I correct?

I don't think so: with the first step we create the timeline dir, then we would check that in the mark drop logic and keep the mark.
But you're right that we can make it better and at least try to remove the files right away, if fail to create the timeline: if we succeed, we could try to create this timeline before the next restart.

I've moved the `after-timeline-uninit-mark-creation` failpoint to this code and adjusted the corresponding test to not to restart pagerver at all.

libs/utils/src/crashsafe_dir.rs
```diff
@@ -92,6 +84,36 @@ pub fn path_with_suffix_extension(original_path: impl AsRef<Path>, suffix: &str)
         .with_extension(new_extension.as_ref())
 }
 
+pub fn fsync_file_and_parent(file_path: &Path) -> io::Result<()> {
```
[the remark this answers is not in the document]
SomeoneToIgnore: Renamed.

pageserver/src/tenant.rs
```diff
@@ -119,6 +121,237 @@ pub struct Tenant {
...
+                if load_layer_map {
+                    new_timeline
+                        .load_layer_map(new_disk_consistent_lsn)
+                        .with_context(|| {
+                            format!(
+                                "Failed to load layermap for timeline {tenant_id}/{timeline_id}"
+                            )
+                        })?;
+                }
+                new_timeline.launch_wal_receiver().with_context(|| {
```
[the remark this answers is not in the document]
SomeoneToIgnore: Good idea, have moved `launch_wal_receiver` below everything and will wait for your patch to land, thanks for noticing.

pageserver/src/tenant.rs
```diff
@@ -119,6 +121,237 @@ pub struct Tenant {
...
+impl UninitializedTimeline<'_> {
+    /// Ensures timeline data is valid, loads it into pageserver's memory and removes uninit mark file on success.
+    pub fn initialize(self) -> anyhow::Result<Arc<Timeline>> {
+        let mut timelines = self.owning_tenant.timelines.lock().unwrap();
+        self.initialize_with_lock(&mut timelines, true)
+    }
+
+    fn initialize_with_lock(
+        mut self,
+        timelines: &mut MutexGuard<HashMap<TimelineId, Arc<Timeline>>>,
```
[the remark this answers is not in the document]
SomeoneToIgnore: Mostly semantics: adding something into &mut map is more awkward and we have to lock to get that &mut anyway.
But changed that, since no strong opinion here.

libs/utils/src/crashsafe_dir.rs
```diff
@@ -92,6 +84,36 @@ pub fn path_with_suffix_extension(original_path: impl AsRef<Path>, suffix: &str)
...
+                .map_err(|e| {
+                    io::Error::new(
+                        e.kind(),
+                        format!("Failed to sync file {path:?} data and metadata: {e}"),
+                    )
+                })
+                .map_err(|e| {
+                    io::Error::new(e.kind(), format!("Failed to fsync file {path:?}: {e}"))
+                })
+        })
```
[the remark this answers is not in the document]
SomeoneToIgnore: Actually, the latter, thank you for spotting, fixed.

pageserver/src/tenant.rs
LizardWizzard: > then we would check that in the mark drop logic and keep the mark.

You're right. I've missed it. To me tombstone approach is way harder to reason about. Indeed many changes are required to pass around different base directory. Not sure exactly how many, I havent tried it. Hard to say is it worth it to make more changes but make it easier (subjectively) to reason about it

LizardWizzard (commented):
Logic around uninit mark looks correct to me, but I find it quite hard to reason about it.

I wonder is it worth it for me to try implementing another approach with one temp dir? It looks like this approach can be quite generic, something like `ScratchDir` that removes itself on drop, and path to it is passed as a base for tenant creation function and timeline creation with initial checkpoint. First doesnt sound too complicated, but dealing with checkpoint to some temporary location indeed looks problematic, the main problem I see is that we need to adjust path for a layer after final rename to target directory (the `path_or_conf` field). Since you actually tried it most probably it is not worth it to repeat your way

Any other thoughts about that?

pageserver/src/tenant.rs
LizardWizzard: Also coverage thinks that this branch in drop was not covered:
https://neondatabase.github.io/zenith-coverage-data/02d93d71a0ee9b43225ad4dc15613106ff62c4c2/coverage/pageserver/src/tenant.rs.html#L315

Is it coverage problem or test doesnt trigger this case?

SomeoneToIgnore: Neither, it's a special handle: current PR logic actually tries to remove the timeline dir first before removing the mark, and only if the timeline dir removal fails the mark code that you've linked would be triggered.
Not sure how realistic this is, but decided it's safer to keep the mark aware of this.

pageserver/src/tenant.rs
```diff
@@ -119,6 +121,216 @@ pub struct Tenant {
...
+        let timeline_id = self.timeline_id;
+        let tenant_id = self.owning_tenant.tenant_id;
+
+        let (new_timeline, uninit_mark) = self.raw_timeline.take().with_context(|| {
+            format!("No timeline for initalization found for {tenant_id}/{timeline_id}")
+        })?;
+        let new_timeline = Arc::new(new_timeline);
+
+        let new_disk_consistent_lsn = new_timeline.get_disk_consistent_lsn();
+        // TODO it would be good to ensure that, but apparently a lot of our testing is dependend on that at least
```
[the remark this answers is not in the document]
SomeoneToIgnore: All the unit tests in the same module, below.

SomeoneToIgnore:
> Since you actually tried it most probably it is not worth it to repeat your way

No other good thoughts, but I could be wrong about it and somebody else might try to take a stab at it, at least to see if it's all as dire as I see it.

LizardWizzard (approved):
This PR is still an improvement, I'll try to find a simpler way with temporary directory, but dont think it is a good idea to block this PR on that. Not sure when I'll get to that so +1 from me

hlinnaka:
Regression tests are printing new warnings in the logs with this:

```
2022-10-21T08:28:22.955233Z ERROR tenant_create{tenant=4caab9ae07b33d796c4495dff4cb8b37}: Failed to fsync removed temporary tenant directory "/tmp/test_output/test_tenant_creation_fails/repo/tenants/4caab9ae07b33d796c4495dff4cb8b37.___temp": Failed to fsync file "/tmp/test_output/test_tenant_creation_fails/repo/tenants/4caab9ae07b33d796c4495dff4cb8b37.___temp": Failed to open the file "/tmp/test_output/test_tenant_creation_fails/repo/tenants/4caab9ae07b33d796c4495dff4cb8b37.___temp": No such file or directory (os error 2)
```
and
```
2022-10-21T08:20:48.973050Z  WARN timeline_create{tenant=8476a13bcdfa964c806fd34b2c675ac1 new_timeline=None lsn=Some(0/2D469A8) pg_version=None}: Removing intermediate uninit mark file /tmp/test_output/test_branch_creation_before_gc/repo/tenants/8476a13bcdfa964c806fd34b2c675ac1/timelines/ea298ed3e567befaf0ee73611a0e2ed7.___uninit
```

Are these expected?

I'm seeing these now that I rebased PR #2653 over this:
https://neon-github-public-dev.s3.amazonaws.com/reports/pr-2653/debug/3295819192/index.html#suites/ec4311502db344eee91f1354e9dc839b/141e19d82f96d89a/.

SomeoneToIgnore:
I think they are normal, because one way or another trigger timeline creation failure: in this case, we don't preserve the tmp directory and remove the fallpoint.

In the suite linked:

*  `test_branch_creation_before_gc` and `test_branch_behind` fail to create timeline due to `Timeline::check_lsn_is_in_scope` failure and drops the mark which results in the logs
* `test_tenant_creation_fails`, `test_timeline_create_break_after_uninit_mark` and `test_timeline_init_break_before_checkpoint` are the tests that break timelines and tenants, so they expect to produce such logs.

Other couple of test failing in that suite seem to have other error logs, not related to the uninit marks, so all seems to be ok for me.

pageserver/src/tenant.rs
SomeoneToIgnore: Timelines' init tmp dir approach issue: https://github.com/neondatabase/neon/issues/2670

PR #2497, opened 2022-09-22
Fix paths to postgres binaries in the deploy script

lubennikovaav:


lubennikovaav:
Do we ever use `LD_LIBRARY_PATH` that is set in ansible scripts?
I updated it in the commit `217c00af7` to use the default version 14, but this is not entirely correct.

cicdteam (approved):

PR #2503, opened 2022-09-22
Miscellaneous small fixups

sharnoff:
This PR is a small collection of things I've noticed while refactoring elsewhere.

It's broken up into two commits:
* Part A: changes I think are clear improvements
* Part B: changes that are more to do with code style

Part B is mostly included optimistically; I'd rather just merge part A instead if there's any doubt as to whether part B should be included.


hlinnaka (approved):
All looks good to me

PR #2506, opened 2022-09-23
Fix extreme metrics bloat in storage sync

MMeent:
From 78 metrics per (timeline, tenant) pair down to (max) 10 metrics per (timeline, tenant) pair, plus another 117 metrics in a global histogram that replaces the previous per-timeline histogram.

~~Deregistration of per-tenant-and-timeline metrics is not yet handled because it's unclear how that should happen here, as we don't have access to a Timeline object.~~

Deregistration is also fixed, we remove them when the Timeline is dropped.


lassizci (approved):
Looks good to me :+1: but 

Linking for visibility. as there were some good comments from Arthur and Dmitry about future directions in https://github.com/neondatabase/neon/pull/2328

PR #2508, opened 2022-09-23
Update pg_version for old timelines

petuhovskiy:
Some time ago, we had this code for handling ProposerGreeting requests in safekeeper:
https://github.com/neondatabase/neon/blob/87bf7be5370cc2a621cd51d5a4cb3b1ed76e4633/safekeeper/src/safekeeper.rs#L611-L616

It didn't persist `pg_version` to the timeline control file. As a result, we have a lot of old projects with `pg_version == 0`.
This PR bumps control file version and sets `pg_version` to 140005 for all such timelines.

### DIFF
--- a/safekeeper/src/safekeeper.rs
+++ b/safekeeper/src/safekeeper.rs
@@ -25,7 +25,7 @@ use utils::{
 };
 
 pub const SK_MAGIC: u32 = 0xcafeceefu32;
-pub const SK_FORMAT_VERSION: u32 = 6;
+pub const SK_FORMAT_VERSION: u32 = 7;
 const SK_PROTOCOL_VERSION: u32 = 2;
 pub const UNKNOWN_SERVER_VERSION: u32 = 0;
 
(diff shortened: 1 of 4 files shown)


safekeeper/src/timeline.rs
```diff
@@ -314,6 +314,8 @@ impl Timeline {
         ttid: TenantTimelineId,
         wal_backup_launcher_tx: Sender<TenantTimelineId>,
     ) -> Result<Timeline> {
+        let _enter = info_span!("load_timeline", timeline = %ttid.timeline_id).entered();
```
petuhovskiy, line 317: This is needed to have better logs in `upgrade_control_file()`.

safekeeper/src/safekeeper.rs
```diff
@@ -639,7 +639,6 @@ where
 
             let mut state = self.state.clone();
             state.server.system_id = msg.system_id;
-            state.server.wal_seg_size = msg.wal_seg_size;
```
lubennikovaav, line 642: why remove this line?

petuhovskiy: Because we have a check about `wal_seg_size` on the line 625 above, it means that it's the same.

We don't support changing wal_seg_size in safekeepers, and AFAIK we cannot change it in the postgres as well.

arssher (approved):

kelvich (approved):

PR #2515, opened 2022-09-24
Implement custom R-Tree search method traversing layers in LSN decrea…

knizhnik:
…sing order

### DIFF
--- a/pageserver/src/tenant/layer_map.rs
+++ b/pageserver/src/tenant/layer_map.rs
@@ -17,6 +17,7 @@ use crate::tenant::storage_layer::Layer;
 use crate::tenant::storage_layer::{range_eq, range_overlaps};
 use amplify_num::i256;
 use anyhow::Result;
+use itertools::Itertools;
 use num_traits::identities::{One, Zero};
 use num_traits::{Bounded, Num, Signed};
 use rstar::{Envelope, ParentNode, RTree, RTreeNode, RTreeObject, AABB};
@@ -26,7 +27,6 @@ use std::ops::Range;
 use std::ops::{Add, Div, Mul, Neg, Rem, Sub};
 use std::sync::Arc;
 use utils::lsn::Lsn;
-use itertools::Itertools;
 
 ///
 /// LayerMap tracks what layers exist on a timeline.
--- a/pageserver/src/tenant/layer_map.rs
+++ b/pageserver/src/tenant/layer_map.rs
@@ -235,21 +235,11 @@ impl LayerMap {
         let key = IntKey::from(key.to_i128());
         let mut min_lsn = Lsn(0);
         let mut max_lsn = Lsn(0);
-        if let Some(layer) = Self::find_latest_layer_recurs(
-            key,
-            &mut min_lsn,
-            &mut max_lsn,
-            end_lsn,
-            root,
-            image_only,
-        ) {
-            Some(SearchResult {
+        Self::find_latest_layer_recurs(key, &mut min_lsn, &mut max_lsn, end_lsn, root, image_only)
+            .map(|layer| SearchResult {
                 layer,
                 lsn_floor: min_lsn,
             })
-        } else {
-            None
-        }
     }
 
     fn find_latest_layer_recurs(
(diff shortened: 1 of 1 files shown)


knizhnik:
Looks like my assumption concerning where bottleneck is was wrong.
At least for my `test_layer_map.py` test.
I assume that we have a lot of layers with different LSN belonging to the same key range. as as far as currently 'LayerMap::search` uses whole LSN range (starting from zero), then R-Tree lookup for intersecting layer will return a lot of candidates which we have to iterate in `LayerMap::`search`. But is is not true for `test_layer_map.py`.  Because of small  `checkpoint_distance=8192` it produce a lot of small layers, but with different key range. So average number of traverse layers s just 3. So  this is why my optimization with cutting-off nodes with smaller LSN will not help much in this case.

Attached please find FameGraph with query execution in  `test_layer_map.py`.  As you can see, LayerMap::`search` is taken only 15% - sightly more than disk_btree search (12%).
layer_map

knizhnik:
But with compaction we have different story:
compaciton
Here we actually traverse ALL layers. But it is not a problem of index (R-Tree), but our LayerMap code.

knizhnik:
> Do you have new performance profiles for these changes?

In theory such traversal can reduce number of tree nodes (layers) which has to be inspected.
Practically I have tested this PR on three different workloads: one artificial (test_layer_map) and two read (grabbed Kontor layers layout). 

I tried to understand why so much time is spent in layer traversal during compaction in production.
The only real operation I can do without having content of real layers is `time_for_new_image_layer` which is called by   `create_image_layers` to check whether new image layer need to be created. The most of time in `time_for_new_image_layer`  is spent in `LayerMap::image_coverage` which in turn calls `find_latest_image`. So average number of traversed layers with this patch is 15 comparing with 63 for main. But it doesn't have large impact on operation time, 278/432/310 msec vs 255/636/288/msec (min/max/avg). But as far as time is in any case very small, most likely it is not the critical path. 

Results of other benchmarks are more impressive (especially on @bojanserafimov  bench_layer_map  which performs `LayerMap::search` on real layer map):

| test | main | this PR |
| ----- | ------ | ---------- |
| test_layer_map | 8.8827 sec | 6.551 sec | 
| cargo bench --bench bench_layer_map |  203.75 ms | 74.431 ms |

knizhnik:
{{{
-   89,59%     0,17%  background op w  pageserver         [.] pageserver::tenant::timeline::Timeline::create_image_layers                                                                  
   - 89,42% pageserver::tenant::timeline::Timeline::create_image_layers                                                                                                                    
      - 87,71% pageserver::tenant::layer_map::LayerMap::image_coverage                                                                                                                     
         - 53,31% pageserver::tenant::layer_map::LayerMap::find_latest_layer                                                                                                               
            + 44,10% pageserver::tenant::layer_map::LayerMap::find_latest_layer_recurs                                                                                                     
            + 9,03% tracing_core::event::Event::dispatch                                                                                                                                   
         - 29,85% <rstar::algorithm::iterators::SelectionIterator<T,Func> as core::iter::traits::iterator::Iterator>::next                                                                 
              6,91% rstar::point::PointExt::all_component_wise                                                                                                                             
              4,05% rstar::point::PointExt::all_component_wise                                                                                                                             
              3,19% <smallvec::SmallVec<A> as core::iter::traits::collect::Extend<<A as smallvec::Array>::Item>>::extend                                                                   
           1,02% <pageserver::tenant::image_layer::ImageLayer as pageserver::tenant::storage_layer::Layer>::get_lsn_range                                                                  
           0,90% core::ops::range::Range<Idx>::contains                                                                                                                                    
           0,79% <pageserver::tenant::delta_layer::DeltaLayer as pageserver::tenant::storage_layer::Layer>::get_lsn_range                                                                  
      + 1,66% pageserver::tenant::layer_map::LayerMap::count_deltas                                                                                                                        
}}}

knizhnik:
More results for synthetic layer map test added by @bayandin (https://github.com/neondatabase/neon/pull/2979):

| branch | total time | query time |
| --------- | ------------ | -------------- |
| main    | 3m45s     |138.55 ms  |
| this PR| 3m15s     | 242.47 ms |

So when layer map tree is so wide (lrgr key dimension) my  LSN-descending traversal order becomes less efficient.

bayandin:
> More results for synthetic layer map test added by @bayandin (#2979)

By @bojanserafimov I guess 😉

knizhnik:
> > More results for synthetic layer map test added by @bayandin (#2979)
> 
> By @bojanserafimov I guess wink

Yes, sorry - looks like my brain is using LR-1 parser:)

PR #2521, opened 2022-09-26
Preserve task result in TaskHandle by keeping join handle around

LizardWizzard:
The solution is not great, but it should hep to debug staging issue I tried to do it in a least destructive way. TaskHandle used only in one place so it is ok to use something less generic unless we want to extend its usage across the codebase. In its current current form for its single usage place it looks too abstract

Some problems around this code:
1. Task can drop event sender and continue running
2. Task cannot be joined several times (probably not needed, but still, can be surprising)
3. Had to split task event into two types because ahyhow::Error does not implement clone. So TaskContinueEvent derives clone but usual task evend does not. Clone requirement appears because we clone the current value in next_task_event. Taking it by reference is complicated.
4. Split between Init and Started is artificial and comes from watch::channel requirement to have some initial value.

    To summarize from 3 and 4. It may be a better idea to use
    RWLock or a bounded channel instead

### DIFF
--- a/pageserver/src/walreceiver.rs
+++ b/pageserver/src/walreceiver.rs
@@ -31,7 +31,6 @@ use etcd_broker::Client;
 use itertools::Itertools;
 use once_cell::sync::OnceCell;
 use std::future::Future;
-use std::sync::Arc;
 use tokio::sync::watch;
 use tracing::*;
 use url::Url;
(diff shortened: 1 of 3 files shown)


pageserver/src/walreceiver/connection_manager.rs
```diff
@@ -145,18 +145,26 @@ async fn connection_manager_loop_step(
...
-                    TaskEvent::NewEvent(status) => {
-                        if status.has_processed_wal {
-                            // We have advanced last_record_lsn by processing the WAL received
-                            // from this safekeeper. This is good enough to clean unsuccessful
-                            // retries history and allow reconnecting to this safekeeper without
-                            // sleeping for a long time.
-                            walreceiver_state.wal_connection_retries.remove(&wal_connection.sk_id);
+                    TaskEvent::Update(c) => {
+                        match c {
+                            TaskStateUpdate::Init => {},
```
SomeoneToIgnore, line 150: Will write `TaskStateUpdate::Init | TaskStateUpdate::Started => {},`

pageserver/src/walreceiver.rs
```diff
@@ -88,53 +87,80 @@ pub fn is_etcd_client_initialized() -> bool {
...
     /// Aborts current task, waiting for it to finish.
-    pub async fn shutdown(mut self) {
-        self.cancellation.send(()).ok();
-        // wait until the sender is dropped
-        while self.events_receiver.changed().await.is_ok() {}
+    pub async fn shutdown(self) {
+        match self.join_handle {
+            Some(jh) => {
+                self.cancellation.send(()).ok();
+                let _ = jh.await;
```
SomeoneToIgnore, line 161:
```suggestion
                match jh.await {
                    Ok(()) => debug!("Shutdown success"),
                    Err(e) => error!("Shutdown error: {e:?}"),
                }
```

pageserver/src/walreceiver/connection_manager.rs
```diff
@@ -145,18 +145,26 @@ async fn connection_manager_loop_step(
...
+                                }
+                                wal_connection.status = status.to_owned();
+                            }
                         }
-                        wal_connection.status = status;
                     },
-                    TaskEvent::End => {
+                    TaskEvent::End(r) => {
+                        if let Err(e) = r {
+                            error!("wal receiver task finished with an error: {e}")
```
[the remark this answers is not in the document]
SomeoneToIgnore: Will actually write 

```
TaskEvent::End(walreceiver_task_result) => {
      match walreceiver_task_result {
          Ok(()) => debug!("WAL receiving task finished"),
          Err(e) => error!("wal receiver task finished with an error: {e:?}"),
      }
      walreceiver_state.drop_old_connection(false).await;
  },
```

SomeoneToIgnore (approved):
Will push some styling nits (commented the details), LGTM otherwise.

PR #2535, opened 2022-09-30
Bump rustc to 1.61

SomeoneToIgnore:


hlinnaka (approved):
Was there a particular feature or something you wanted from 1.61?

SomeoneToIgnore:
For the old code, I would have used the https://doc.rust-lang.org/std/thread/struct.JoinHandle.html#method.is_finished method to cancel the initial
timeline logical size calculation.

Now that makes less sense due to async rewrite, but generally bumping rustc is beneficial to keep up with the pace and to catch oddities earlier.

PR #2546, opened 2022-10-03
Do not remove empty timelines/ directory for tenants

SomeoneToIgnore:
Part of work on the staging recovery, fixes issues for a few latest clusters, stuck on staging:

```
https://console.stage.neon.tech/admin/projects/steep-voice-412165

zenith-us-stage-ps-4/63387f1aded0790526c9fd84a23364c7/2bbb937d306e72c8ff4ce96e7269d8ec
2022-10-03T11:14:49.631494Z ERROR init_tenant_mgr:local_tenant_timeline_files: Failed to collect tenant files from dir '/storage/pageserver/data/tenants' for entry DirEntry("/storage/pageserver/data/tenants/63387f1aded0790526c9fd84a23364c7"), reason: Failed to list timelines dir entry for tenant 63387f1aded0790526c9fd84a23364c7: No such file or directory (os error 2)

------------------

https://console.stage.neon.tech/admin/projects/solitary-sound-889138

zenith-us-stage-ps-4/bfb91c0842f010f925f8c4c79a57873e/bf54622f59eb105b432c70581997b196

2022-10-03T18:31:19.475359Z ERROR Error processing HTTP request: InternalServerError(cannot create new tenant repo: 'bfb91c0842f010f925f8c4c79a57873e' directory already exists

------------------

https://console.stage.neon.tech/admin/projects/summer-bird-073759

zenith-us-stage-ps-4/b123d21e0a210f11a58bd05bf2d225f4/1cfe16da20ccd7d0f35e417fb39b6d9e

same as above: missing `timelines/` in the tenant dir

------------------

https://console.stage.neon.tech/admin/projects/plain-violet-755403

zenith-us-stage-ps-4/d5ce671c64e125a3c3d4c2766f7906d8/2756c16ea6fb4d3b2ef50b5e34a8c5b1

same as above: missing `timelines/` in the tenant dir
```

-------------

The clusters had their timelines removed entirely and, due the bug in the code, their `timelines/` subdirectory removed as an empty one.
No `timelines/` subdirectory caused the timeline attach logic to fail, as we could not check timeline presence.
I've decided to not restore the `timelines/` directory automatically, since its absence is abnormal, and only hardened the tenant load logic to check for directory existence and fail fast + fixed the bug.

### DIFF
--- a/pageserver/src/tenant_mgr.rs
+++ b/pageserver/src/tenant_mgr.rs
@@ -108,6 +108,10 @@ pub fn init_tenant_mgr(
 /// Attempts to load as many entites as possible: if a certain timeline fails during the load, the tenant is marked as "Broken",
 /// and the load continues.
 ///
+/// For successful tenant attach, it first has to have a `timelines/` subdirectory and a tenant config file that's loaded into memory successfully.
+/// If either of the conditions fails, the tenant will be added to memory with [`TenantState::Broken`] state, otherwise we start to load its timelines.
+/// Alternatively, tenant is considered loaded successfully, if it's already in pageserver's memory (i.e. was loaded already before).
+///
 /// Attach happens on startup and sucessful timeline downloads
 /// (some subset of timeline files, always including its metadata, after which the new one needs to be registered).
 pub fn attach_local_tenants(
@@ -173,16 +177,28 @@ fn load_local_tenant(
         remote_index.clone(),
         conf.remote_storage_config.is_some(),
     ));
-    match Tenant::load_tenant_config(conf, tenant_id) {
-        Ok(tenant_conf) => {
-            tenant.update_tenant_config(tenant_conf);
-            tenant.activate(false);
-        }
-        Err(e) => {
-            error!("Failed to read config for tenant {tenant_id}, disabling tenant: {e:?}");
-            tenant.set_state(TenantState::Broken);
+
+    let tenant_timelines_dir = conf.timelines_path(&tenant_id);
+    if !tenant_timelines_dir.is_dir() {
+        error!(
+            "Tenant {} has no timelines directory at {}",
+            tenant_id,
+            tenant_timelines_dir.display()
+        );
+        tenant.set_state(TenantState::Broken);
+    } else {
+        match Tenant::load_tenant_config(conf, tenant_id) {
+            Ok(tenant_conf) => {
+                tenant.update_tenant_config(tenant_conf);
+                tenant.activate(false);
+            }
+            Err(e) => {
+                error!("Failed to read config for tenant {tenant_id}, disabling tenant: {e:?}");
+                tenant.set_state(TenantState::Broken);
+            }
         }
     }
+
     tenant
 }
 
@@ -630,14 +646,10 @@ fn collect_timelines_for_tenant(
     }
 
     if tenant_timelines.is_empty() {
-        match remove_if_empty(&timelines_dir) {
-            Ok(true) => info!(
-                "Removed empty tenant timelines directory {}",
-                timelines_dir.display()
-            ),
-            Ok(false) => (),
-            Err(e) => error!("Failed to remove empty tenant timelines directory: {e:?}"),
-        }
+        // this is normal, we've removed all broken, empty and temporary timeline dirs
+        // but should allow the tenant to stay functional and allow creating new timelines
+        // on a restart, we require tenants to have the timelines dir, so leave it on disk
+        debug!("Tenant {tenant_id} has no timelines loaded");
     }
 
     Ok((tenant_id, tenant_timelines))
--- a/test_runner/regress/test_tenants.py
+++ b/test_runner/regress/test_tenants.py
@@ -8,8 +8,13 @@ from typing import List
 import pytest
 from fixtures.log_helper import log
 from fixtures.metrics import PAGESERVER_PER_TENANT_METRICS, parse_metrics
-from fixtures.neon_fixtures import NeonEnv, NeonEnvBuilder
-from fixtures.types import Lsn, TenantId
+from fixtures.neon_fixtures import (
+    NeonEnv,
+    NeonEnvBuilder,
+    RemoteStorageKind,
+    available_remote_storages,
+)
+from fixtures.types import Lsn, TenantId, TimelineId
 from prometheus_client.samples import Sample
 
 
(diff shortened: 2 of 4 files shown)


SomeoneToIgnore:
After the change, we've got one extra side effect: tenant attach started to fail, as we're now retaining tenants with 0 timelines in memory.

So
* if pageserver removes empty `timelines/` directory, such tenant will be considered broken after next startup, on local tenant attach (both with the old or the new logic).
* old logic did not load tenants with empty or missing `timelines/` directory into memory, which allowed attaching such tenants, but did not allow to create a timeline for "empty" tenants
* it seems dangerous to remove the rest of the files, if `timelines/` is missing: we could loose the tenant config?

It appears to me, that the best compromise would be to allow attaching tenants (i.e. downloading all timelines for such tenant form the remote storage) with 0 local timelines, which I've pushed as a separate commit.

lubennikovaav:
Do I get it right, that to fix present issue on staging we have to manually recreate `timeline/` directories?

SomeoneToIgnore:
Yes.
That's what I've done already before submitting a PR and the issue had not reappeared so far.

test_runner/regress/test_tenants.py
```diff
@@ -201,3 +202,39 @@ def get_ps_metric_samples_for_tenant(tenant_id: TenantId) -> List[Sample]:
...
         post_detach_samples = set([x.name for x in get_ps_metric_samples_for_tenant(tenant)])
         assert post_detach_samples == set()
+
+
+def test_pageserver_with_empty_tenants(neon_simple_env: NeonEnv):
+    env = neon_simple_env
+    client = env.pageserver.http_client()
+
+    tenant_without_timelines_dir = env.initial_tenant
+    shutil.rmtree(Path(env.repo_dir) / "tenants" / str(tenant_without_timelines_dir) / "timelines")
```
lubennikovaav, line 212: That should never happen in normal pageserver and we manually remove the dir to check the failure scenario, right?
Please, add a comment.

test_runner/regress/test_tenants.py
```diff
@@ -201,3 +202,39 @@ def get_ps_metric_samples_for_tenant(tenant_id: TenantId) -> List[Sample]:
...
+    for timeline_dir_entry in Path.iterdir(
+        Path(env.repo_dir) / "tenants" / str(tenant_with_empty_timelines_dir) / "timelines"
+    ):
+        if timeline_dir_entry.is_dir():
+            shutil.rmtree(timeline_dir_entry)
+        else:
+            timeline_dir_entry.unlink()
+
+    env.postgres.stop_all()
+    for _ in range(0, 3):
```
lubennikovaav, line 224: Why do we need to restart pageserver several times here?

test_runner/regress/test_tenants.py
```diff
@@ -201,3 +202,39 @@ def get_ps_metric_samples_for_tenant(tenant_id: TenantId) -> List[Sample]:
...
+
+    tenant_with_empty_timelines_dir = client.tenant_create()
+    for timeline_dir_entry in Path.iterdir(
+        Path(env.repo_dir) / "tenants" / str(tenant_with_empty_timelines_dir) / "timelines"
+    ):
+        if timeline_dir_entry.is_dir():
+            shutil.rmtree(timeline_dir_entry)
+        else:
+            timeline_dir_entry.unlink()
+
```
lubennikovaav, line 222: Why do we remove these directories manually?
If this is an expected state of pageserver, let's reach it using "legal" commands.

please, add a comment here as well

pageserver/src/tenant_mgr.rs
```diff
@@ -108,6 +108,10 @@ pub fn init_tenant_mgr(
 /// Attempts to load as many entites as possible: if a certain timeline fails during the load, the tenant is marked as "Broken",
 /// and the load continues.
 ///
+/// For successful tenant attach, it first has to have a `timelines/` subdirectory and a tenant config file that's loaded into memory successfully.
+/// If either of the conditions fails, the tenant will be added to memory with [`TenantState::Broken`] state, otherwise we start to load its timelines.
```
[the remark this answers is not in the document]
SomeoneToIgnore: > Do we have any way to fix broken timelines now?

2.5 of them actually:
* the most commonly used one is supposed to be the one you were describing below, from pageserver's remote storage
* if we corrupt / loose the storage somehow, we're able to retract to safepeekers' data, either inmem (for the recent data) or the S3 one (semi manual process)
* nothing prevents us from adding walbackup handling and have another source of restoration, if I understand the mechanism right, hence 0.5 points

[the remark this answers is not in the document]
SomeoneToIgnore: > Will the flow be

I think, now there's no automated flow for this particular situation.
We have 
* "delete timeline" operation that's supposed to remove files from memory, local fs and remote storage, as the name implies
* "detach tenant" that removes the tenant and all of its timelines from the memory and from local FS, not from the remote storage. 
* plan to add "reload" tenant for unloading the data from memory only: https://github.com/neondatabase/neon/issues/2537

Either of the operations could be used to restore timeline data, the simplest restoration source would be remote storage, albeit it's not reliable yet, since has not finished CRC integration.
One could retrieve the correct layer files from the remote storage and try restarting pageserver or reloading its tenant, or do the reattach to use remote storage data.

IMO, we might benefit from per-timeline operations for maintenance, but so far there's no RFC or discussion on the topic.

test_runner/regress/test_tenants.py
SomeoneToIgnore: We don't, changed it to one restart to not to confuse people.

test_runner/regress/test_tenants.py
SomeoneToIgnore: This is actually another good case where I think I'm doing wrong.
Right now, such tenant will be omitted, since we try to collect its timelines and fail, on `fn collect_timelines_for_tenant` level, completely ignoring such tenants on loading, not putting it into memory (as the test shows).

Later, if somebody attempts to create such tenant, it will error, because there are some files on the local disk.
pageserver does not remove such files and does not show anything in the management HTTP API, which I think is wrong.

I would love to change it more later, to actually return such tenants with "Broken" tenant state. I'm not sure if it's worth doing now, wdyt?

test_runner/regress/test_tenants.py
SomeoneToIgnore: Have added logging instead, so if something breaks, we could use logged ids to investigate.

pageserver/src/tenant_mgr.rs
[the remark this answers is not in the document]
SomeoneToIgnore: Looks like SK restoration is more convenient and reliable (since WAL segments have CRC): https://github.com/neondatabase/cloud/wiki/Recovery-from-WAL

lubennikovaav (approved):
LGTM.

I'd suggest to add one more test to ensure that empty tenants handled correctly when s3 is enabled

test_runner/regress/test_tenants.py
lubennikovaav: I think it would be nice to track such tenants somehow.
to help future us who will debug obscure "this tenant already exists" errors.
It's not urgent though. Feel free to create a separate issue

SomeoneToIgnore:
Added a test, that requires a slight update to the way we start remote storages, please raise your concerns if agains.
Will set it to automerge, since the point is to avoid stucking the clusters in the future, will iterate on better broken visibility later.

test_runner/regress/test_tenants.py
SomeoneToIgnore: https://github.com/neondatabase/neon/issues/2552

PR #2554, opened 2022-10-04
[proxy] Improve logging

funbringer:


antonyc:
Not sure if it's present or not in this PR, can we log non-successful console responses? At least URL, status code and time?

funbringer:
> Not sure if it's present or not in this PR, can we log non-successful console responses? At least URL, status code and time?

I think this is the case. For instance, this is what I get for a local run:

```
2022-10-06T10:54:37.100707Z ERROR client{session=a42bee89-87e5-4fde-9cd9-2ed6d679ccf7}: error sending request for url (http://console-staging.local/management/api/v2/proxy_get_role_secret?session_id=a42bee89-87e5-4fde-9cd9-2ed6d679ccf7&application_name=psql&project=foo&role=borat): error trying to connect: dns error: failed to lookup address information: Name or service not known
```

proxy/src/auth/backend/console.rs
```diff
@@ -177,20 +177,19 @@ impl<'a> Api<'a> {
...
             ])
             .build()?;
 
-        // TODO: use a proper logger
-        println!("cplane request: {}", req.url());
-
+        info!(id = request_id, url = req.url().as_str(), "request");
         let resp = self.endpoint.execute(req).await?;
         if !resp.status().is_success() {
             return Err(TransportError::HttpStatus(resp.status()).into());
```
ololobus, line 195: I wonder, is this error logged somewhere? I could not quickly find it in the proxy code. I see info record above, but that'd be nice to see error too

ololobus (approved):
It's hard to say that this amount of logging will be enough, but replacing `println` with a proper logger is a huge step forward and we can add more logs later. Thanks!

proxy/src/auth/backend/console.rs
funbringer: Anton asked the same question essentially. Please see the example above. Every single error like this is propagated to task's root and ends up in `log_error`.

antonyc:
> I think this is the case. For instance, this is what I get for a local run:

@funbringer , it would be great to log request_id for unsuccessful console responses.

PR #2561, opened 2022-10-05
indexpart versioning support

koivunej:
First step in #2456, adding the support for (de)serializing IndexPart documents with optional `version`.

The route taken here is to support the `version` property within the document, which serde calls "internally tagged". Because the field doesn't exist yet, the deserialization part is more complicated than usual. I made the upgrade portion explicit with the custom serialization as well.

Draft for early comments, if this matches the vision at all. This change would be completely backwards compatible, and would support continuing towards:

- [ ] test against all present staging and production `index_part.json`
- [ ] on later PR: add layer sizes in a new version, figure out how to actually migrate


koivunej:
Should make the field `schema_version` or similar and not use the ambiguous name "version" in the likely case we want to add a version for optimistic concurrency control as well -- that could be "timeline_version" or "document_version"?

pageserver/src/storage_sync/index.rs
```diff
@@ -262,19 +262,88 @@ impl RemoteTimeline {
...
+            #[serde(flatten)]
+            inner: &'a I,
+        }
+
+        let envelope = match self {
+            // V2(inner) => Versioned {
+            //     version: IndexPartVersion::V2,
+            //     inner,
+            // },
+            V1 { inner, versioned } if *versioned => Versioned {
```
SomeoneToIgnore, line 308: I'm not sure what `versioned` is supposed to be? An adapter to mitigate the absence of version in the existing format?

From what I understand, the format does not change, but gets the version in the new approach, so it might make sense to serialize it with V1 as basic version, when it's absent.

pageserver/src/storage_sync/index.rs
```diff
@@ -287,47 +356,106 @@ impl IndexPart {
...
+        }
+    }
+
+    pub fn timeline_layers(&self) -> &HashSet<RelativePath> {
+        match self {
+            Self::V1 { inner: v1, .. } => &v1.timeline_layers,
         }
     }
 
+    // FIXME: this should probably be missing_layers()
```
SomeoneToIgnore, line 377: Why not, feel free to update it too.

pageserver/src/storage_sync/index.rs
```diff
@@ -287,47 +356,106 @@ impl IndexPart {
...
+        match self {
+            Self::V1 { inner: v1, .. } => &v1.metadata_bytes,
+        }
+    }
+
+    /// Upgrade the description in-place, if needed.
+    ///
+    /// Returns true when the document was updated and serializing it will produce different
+    /// result than what it was deserialized from.
+    pub fn upgrade(&mut self) -> bool {
```
SomeoneToIgnore, line 400: I have some concerns related to this method usage, I think if we've deserialized the type, we've already made all work on presenting the data in some compatible form and can try to autoupgrade it.

SomeoneToIgnore (commented):
I like the direction in general, not sure how ergonomic it can be for forward-compatible changes, but we can add more commits/branches on top of this and consider next situation: let's add some has into the format, with some "lorem ipsum" inside?

pageserver/src/storage_sync/index.rs
koivunej: Forgot to post this, but I was thinking of not doing automatic version promotion, so that `&serde_json::to_string(serde_json::from_str(&without_version)?)? == without_version`, but otherwise it's not needed.

koivunej:
Looking to demo a v2 with the layer sizes.

pageserver/src/storage_sync/index.rs
SomeoneToIgnore: Yeah, I think we need to discuss that a bit more, are there any downsides with doing so?

PR #2566, opened 2022-10-05
Fix deadlock in safekeeper metrics

petuhovskiy:
Should fix https://github.com/neondatabase/neon/issues/2545

We had a problem where almost all of the threads were waiting on a futex syscall. More specifically:
- `/metrics` handler was inside `TimelineCollector::collect()`, waiting on a mutex for a single Timeline
- This exact timeline was inside `control_file::FileStorage::persist()`, waiting on a mutex for Lazy initialization of `PERSIST_CONTROL_FILE_SECONDS`
- `PERSIST_CONTROL_FILE_SECONDS: Lazy<Histogram>` was blocked on `prometheus::register`
- `prometheus::register` calls `DEFAULT_REGISTRY.write().register()` to take a write lock on Registry and add a new metric
- `DEFAULT_REGISTRY` lock was already taken inside `DEFAULT_REGISTRY.gather()`, which was called by `/metrics` handler to collect all metrics

We wouldn't have this issue if all metrics were initialized right on start without Lazy. But that's inconvenient and it's better not to rely on that.

This PR creates another Registry with a separate lock, to avoid deadlock in a case where `TimelineCollector` triggers registration of new metrics inside default registry.

### DIFF
--- a/libs/metrics/src/lib.rs
+++ b/libs/metrics/src/lib.rs
@@ -3,7 +3,7 @@
 //! Otherwise, we might not see all metrics registered via
 //! a default registry.
 use once_cell::sync::Lazy;
-use prometheus::core::{AtomicU64, GenericGauge, GenericGaugeVec};
+use prometheus::core::{AtomicU64, Collector, GenericGauge, GenericGaugeVec};
 pub use prometheus::opts;
 pub use prometheus::register;
 pub use prometheus::{core, default_registry, proto};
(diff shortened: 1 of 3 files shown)


libs/metrics/src/lib.rs
```diff
@@ -32,13 +33,27 @@ macro_rules! register_uint_gauge_vec {
     }};
 }
 
+/// Special internal registry, to collect metrics independently from the default registry.
+/// Was introduced to fix deadlock with lazy registration of metrics in the default registry.
+static INTERNAL_REGISTRY: Lazy<Registry> = Lazy::new(Registry::new);
+
+/// Register a collector in the internal registry. MUST be called before the first call to `gather()`.
```
arssher, line 40: this can be hardened by using OnceCell instead of Lazy

arssher (approved):
Let's describe in commit message the deadlock.

SomeoneToIgnore (approved):

libs/metrics/src/lib.rs
petuhovskiy: I'm not sure I follow, OnceCell is just a more explicit version of Lazy, which doesn't accept closure in constructor. I'm going to merge this PR, feel free to create another one to replace Lazy with OnceCell.

Issue #2512, opened 2022-09-23
`test_delete_force[True]` may be flaky


sharnoff:
Occurred in CI for #2503, which (a) shouldn't have affected this and (b) didn't fail after rebasing on top of (AFAICT) unrelated changes to `main`

```
2022-09-23T07:13:48.6940476Z =================================== FAILURES ===================================
2022-09-23T07:13:48.6940824Z ___________________________ test_delete_force[True] ____________________________
2022-09-23T07:13:48.6942108Z [gw0] linux -- Python 3.9.2 /github/home/.cache/pypoetry/virtualenvs/neon-_pxWMzVK-py3.9/bin/python
2022-09-23T07:13:48.6942739Z test_runner/regress/test_wal_acceptor.py:1127: in test_delete_force
2022-09-23T07:13:48.6943227Z     assert sk_http.timeline_delete_force(tenant_id, timeline_id_2) == {
2022-09-23T07:13:48.6943791Z E   AssertionError: assert {'dir_existed...active': True} == {'dir_existed...ctive': False}
2022-09-23T07:13:48.6944191Z E     Omitting 1 identical items, use -vv to show
2022-09-23T07:13:48.6944501Z E     Differing items:
2022-09-23T07:13:48.6944942Z E     {'was_active': True} != {'was_active': False}
2022-09-23T07:13:48.6945235Z E     Full diff:
2022-09-23T07:13:48.6945646Z E     - {'dir_existed': True, 'was_active': False}
2022-09-23T07:13:48.6946011Z E     ?                                     ^^^^
2022-09-23T07:13:48.6946332Z E     + {'dir_existed': True, 'was_active': True}
2022-09-23T07:13:48.6946665Z E     ?                                     ^^^
2022-09-23T07:13:48.6947174Z ------------------------------ Captured log setup ------------------------------
2022-09-23T07:13:48.6947769Z 2022-09-23 07:11:37.138 INFO [neon_fixtures.py:2467] get_test_output_dir is /tmp/test_output/test_delete_force[True]
2022-09-23T07:13:48.6948436Z 2022-09-23 07:11:37.139 INFO [neon_fixtures.py:2488] test_output_dir is /tmp/test_output/test_delete_force[True]
2022-09-23T07:13:48.6948970Z ------------------------------ Captured log call -------------------------------
2022-09-23T07:13:48.6949544Z 2022-09-23 07:11:37.141 INFO [neon_fixtures.py:848] Config: 
[...]

PR #2567, opened 2022-10-06
Remove flaky checks in test_delete_force

petuhovskiy:
We can't reliably tell if timeline should be active or not, and activity logic can be changed in the future. This PR disables checks for `was_active` in `test_delete_force`, to fix flaky test fails.


sharnoff:
This resolves #2512

arssher (approved):

PR #2583, opened 2022-10-07
[proxy] Add more context to console requests logging

funbringer:


ololobus (approved):
Had a look on the existing logs, and they got much better, thanks! Changes in this PR look reasonable

PR #2584, opened 2022-10-07
Fix pageserver configuration generation bug

lassizci:
There was an issue in case of changing pageserver's `remote_storage` config values: we would have duplicated the keys in the table, rendering invalid toml.

Instead of patching parts of the configuration, now managing the whole config file. There's also an upside that if we want to remove something from the config, that's dealt with without adding extra task. Managing the config file requires couple things:
  * Install https://galaxy.ansible.com/sivel/toiletwater collection to get filters `from_toml` and `to_toml`
  * Parse the pageserver `id` from the existing config using `from_toml`. This is set up in the initial install, so should always be there and shouldn't change. The id itself is fetched dynamically
   * merge the id with config `pageserver_config_stub` we can manage statically
   * finally render the config from yaml to toml using `to_toml`
   
To make maintaining the configuration as yaml nicer, convert the inventories to yaml instead of ini. An alternative could perhaps have been to separate the variables from inventory. But as there probably isn't a growing demand to keep adding more, I kept those in the inventory file.


lassizci:
Cheers, I'll merge when the pipeline runs.

PR #2594, opened 2022-10-10
Fix POSTGRES_DISTRIB_DIR for benchmarks on ec2 runner

SergeyMelnikov:


.github/workflows/benchmarking.yml
```diff
@@ -72,7 +72,7 @@ jobs:
         echo Poetry
         poetry --version
         echo Pgbench
-        ${POSTGRES_DISTRIB_DIR}/v${DEFAULT_PG_VERSION}/bin/pgbench --version
+        pgbench --version
```
bayandin, line 75: We use path `${POSTGRES_DISTRIB_DIR}/v${DEFAULT_PG_VERSION}` in tests, so this change will only delay the failure.

I've created a symlink `/usr/pgsql/v14` on the runner, so these changes should be enough to fix it:

```diff
     env:
-      POSTGRES_DISTRIB_DIR: /tmp/pg_install
+      POSTGRES_DISTRIB_DIR: /usr/pgsql
       DEFAULT_PG_VERSION: 14
``` 

Before. I've created a symlink in `/tmp`, but it got deleted, having it in `/usr` should be more permanent.

bayandin (approved):

bayandin:
Thanks!

PR #2596, opened 2022-10-10
Add build info metric to pageserver, safekeeper and proxy

lassizci:
It would be nice to be able to show the version of each component in the metric dasboards. So lets create `<component>_build_info` metric for `pageserver`, `safekeeper` and `proxy` with label `revision`, which points to the git commit. `revision` format is the same as we use in logs.

`/metrics` output:
```
pageserver_build_info{revision="git:8c8c603948e5591819d83bd5058fb69f1e3e6aec"}  1.0
```
```
safekeeper_build_info{revision="git:8c8c603948e5591819d83bd5058fb69f1e3e6aec"}  1.0
```
```
proxy_build_info{revision="git:8c8c603948e5591819d83bd5058fb69f1e3e6aec"}  1.0
```

### DIFF
--- a/libs/metrics/src/lib.rs
+++ b/libs/metrics/src/lib.rs
@@ -77,6 +77,16 @@ pub const DISK_WRITE_SECONDS_BUCKETS: &[f64] = &[
     0.000_050, 0.000_100, 0.000_500, 0.001, 0.003, 0.005, 0.01, 0.05, 0.1, 0.3, 0.5,
 ];
 
+pub fn set_build_info_metric(revision: &str) {
+    let metric = register_int_gauge_vec!(
+        "libmetrics_build_info",
+        "Build/version information",
+        &["revision"]
+    )
+    .expect("Failed to register build info metric");
+    metric.with_label_values(&[revision]).set(1);
+}
+
 // Records I/O stats in a "cross-platform" way.
 // Compiles both on macOS and Linux, but current macOS implementation always returns 0 as values for I/O stats.
 // An alternative is to read procfs (`/proc/[pid]/io`) which does not work under macOS at all, hence abandoned.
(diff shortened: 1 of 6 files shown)


petuhovskiy:
Actually, it would be useful to have this for all our storage services (pageserver, safekeeper, proxy), not only for pageserver.
AFAIR we have common metrics in `libs/metrics/src/lib.rs`, and we can add a new one named like `storage_build_info`, `rust_build_info` or `libmetrics_build_info`. WDYT?

lassizci:
> Actually, it would be useful to have this for all our storage services (pageserver, safekeeper, proxy), not only for pageserver. AFAIR we have common metrics in `libs/metrics/src/lib.rs`, and we can add a new one named like `storage_build_info`, `rust_build_info` or `libmetrics_build_info`. WDYT?

Excellent point, we probably want to have the metric for each, I'll add.

Kind of like the very specific ones though: `pageserver_` `safekeeper_` and `proxy_` because otherwise we'd need to carry the component name (they aren't atomically deployed and there can be some issues with deploys etc) in the label, or do relabeling config when we scrape the components and so on. Metric names tend to be a bit more straight forward to use I think. Objections?

lassizci:
@petuhovskiy now implemented for all three. Guess https://docs.rs/prometheus/latest/prometheus/struct.Opts.html#structfield.namespace would be the right way to implement name prefixes for the metrics, but homegrown name templating for now to avoid refactoring :)

petuhovskiy:
Sorry, missed your [previous comment](https://github.com/neondatabase/neon/pull/2596#issuecomment-1273433134). IMO it's better to have a single name (because it's the same metrics across all services) and differentiate using the labels.

We already have a lot of labels, should be enough to query any metric. For example, we have:
- `job = safekeeper`
- `neon_service = safekeeper`
- `instance = zenith-us-stage-sk-6`

My arguments for having the same metric name:
- It's easier to query version for all our services and we still can split it with labels if we want to. If there are different names, it's harder to query all services at once.
- Querying all services together really make sense, because binaries are coming from a single CI build and they have the same SHA commit in the label.

lassizci:
There's now one metric with `libmetrics` prefix. Didn't add component labels as I guess we can rely on scraper relabelings.

PR #2597, opened 2022-10-10
Simplified auth flow for Neon Proxy RFC

funbringer:
[Rendered](https://github.com/neondatabase/neon/blob/funbringer/rfc-proxy-simplified-auth/docs/rfcs/xxx-proxy-simplified-auth.md)

### DIFF
--- a/docs/rfcs/xxx-proxy-simplified-auth.md
+++ b/docs/rfcs/xxx-proxy-simplified-auth.md
@@ -1,0 +1,132 @@
+# Simplified auth flow for Neon Proxy
+
+## Overview
+
+This is how authentication works at the moment.
+
+```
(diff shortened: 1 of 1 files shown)


kelvich:
I think the part about how _exactly_ to do cert auth should be written too. With things like:
* Who stores private keys
* When and how new cert is generated
* Is rotated or not
* How it compares to Kerberos (I guess when we'll go into details of pkey storage and rotation Kerberos may look like an ok alternative. Or not)

It doesn't have to be right from the start, but let's write down options so that we can discuss them in the comments.

funbringer:
IIRC this is no longer needed. @kelvich please reopen if I'm mistaken.

PR #2607, opened 2022-10-11
Don't run perf tests in parallel

bojanserafimov:
Trying to see if measurement noise comes from parallel runs.

Currently we run the tests once a day so there's enough time in a day to run all tests sequentially. When we increase test size we'll have to do parallel again. But that's fine, since I'm just trying to see if the measurement noise disappears. We can revert this.

### DIFF
--- a/.github/workflows/benchmarking.yml
+++ b/.github/workflows/benchmarking.yml
@@ -149,6 +149,7 @@ jobs:
 
     strategy:
       fail-fast: false
+      max-parallel: 1
       matrix:
         # neon-captest-new: Run pgbench in a freshly created project
         # neon-captest-reuse: Same, but reusing existing project


bayandin:
I'm afraid it will make more pgbench tests fail. 
We drop the connection when we deploy the proxy (https://github.com/neondatabase/cloud/issues/1872), and with extended test duration, we increase the probability to get failure because of merged PR triggers deploy of staging and captest.
I've already altered the schedule to have all tests finished before the start of the work day.

hlinnaka (approved):

bojanserafimov:
Is it possible to pick one of these tests and run it manually 5-6 times when nothing else runs (not sure if there's a button somewhere)? That will also answer the question.

bayandin:
> Is it possible to pick one of these tests and run it manually 5-6 times when nothing else runs (not sure if there's a button somewhere)? That will also answer the question.

You can trigger it from this page https://github.com/neondatabase/neon/actions/workflows/benchmarking.yml (Run workflow button), or use can use GitHub CLI: https://cli.github.com/manual/gh_workflow_run

To pick a particular test — just comment out everything else that you don't won't to run and you can choose a branch (either on the Actions page or using CLI)

bojanserafimov:

PR #2609, opened 2022-10-12
Implement corrent truncation of FSM/VM forks on arbitrary position

knizhnik:
refer #2601

### DIFF
--- a/libs/postgres_ffi/src/lib.rs
+++ b/libs/postgres_ffi/src/lib.rs
@@ -178,7 +178,7 @@ pub fn fsm_logical_to_physical(addr: BlockNumber) -> BlockNumber {
         leafno /= pg_constants::SLOTS_PER_FSM_PAGE;
     }
     /* Turn the page count into 0-based block number */
-    return pages - 1;
+    pages - 1;
 }
 
 pub mod waldecoder {
--- a/libs/postgres_ffi/src/lib.rs
+++ b/libs/postgres_ffi/src/lib.rs
@@ -178,7 +178,7 @@ pub fn fsm_logical_to_physical(addr: BlockNumber) -> BlockNumber {
         leafno /= pg_constants::SLOTS_PER_FSM_PAGE;
     }
     /* Turn the page count into 0-based block number */
-    pages - 1;
+    pages - 1
 }
 
 pub mod waldecoder {
--- a/libs/postgres_ffi/src/lib.rs
+++ b/libs/postgres_ffi/src/lib.rs
@@ -163,6 +163,24 @@ pub fn page_set_lsn(pg: &mut [u8], lsn: Lsn) {
     pg[4..8].copy_from_slice(&(lsn.0 as u32).to_le_bytes());
 }
 
+pub fn fsm_logical_to_physical(addr: BlockNumber) -> BlockNumber {
+    let mut leafno = addr;
+    const FSM_TREE_DEPTH: u32 = if pg_constants::SLOTS_PER_FSM_PAGE >= 1626 {
+        3
+    } else {
+        4
+    };
+
+    /* Count upper level nodes required to address the leaf page */
+    let mut pages: BlockNumber = 0;
+    for _l in 0..FSM_TREE_DEPTH {
+        pages += leafno + 1;
+        leafno /= pg_constants::SLOTS_PER_FSM_PAGE;
+    }
+    /* Turn the page count into 0-based block number */
+    return pages - 1;
+}
+
 pub mod waldecoder {
 
     use crate::{v14, v15};
--- a/libs/postgres_ffi/src/lib.rs
+++ b/libs/postgres_ffi/src/lib.rs
@@ -163,6 +163,9 @@ pub fn page_set_lsn(pg: &mut [u8], lsn: Lsn) {
     pg[4..8].copy_from_slice(&(lsn.0 as u32).to_le_bytes());
 }
 
+// This is port of function with the same name from freespace.c.
+// The only difference is that it does not have "level" parameter because XLogRecordPageWithFreeSpace
+// always call it with level=FSM_BOTTOM_LEVEL
 pub fn fsm_logical_to_physical(addr: BlockNumber) -> BlockNumber {
     let mut leafno = addr;
     const FSM_TREE_DEPTH: u32 = if pg_constants::SLOTS_PER_FSM_PAGE >= 1626 {
--- a/pageserver/src/walingest.rs
+++ b/pageserver/src/walingest.rs
@@ -616,7 +616,8 @@ impl<'a> WalIngest<'a> {
             let mut fsm_physical_page_no = fsm_logical_to_physical(fsm_logical_page_no);
             if rec.blkno % pg_constants::SLOTS_PER_FSM_PAGE != 0 {
                 // Tail of last remaining FSM page has to be zeroed.
-                // We are not doing it here: there should be FPI wal-record for this page stored sometimes laters
+                // We are not precise here and instead of digging in FSM bitmap format just clear the whole page.
+                modification.put_rel_page_image(rel, fsm_physical_page_no, ZERO_PAGE.clone())?;
                 fsm_physical_page_no += 1;
             }
             let nblocks = self.get_relsize(rel, modification.lsn)?;
@@ -636,7 +637,8 @@ impl<'a> WalIngest<'a> {
             let mut vm_page_no = rec.blkno / pg_constants::VM_HEAPBLOCKS_PER_PAGE;
             if rec.blkno % pg_constants::VM_HEAPBLOCKS_PER_PAGE != 0 {
                 // Tail of last remaining vm page has to be zeroed.
-                // We are not doing it here: there should be FPI wal-record for this page stored sometimes laters
+                // We are not precise here and instead of digging in VM bitmap format just clear the whole page.
+                modification.put_rel_page_image(rel, vm_page_no, ZERO_PAGE.clone())?;
                 vm_page_no += 1;
             }
             let nblocks = self.get_relsize(rel, modification.lsn)?;
--- a/pageserver/src/walingest.rs
+++ b/pageserver/src/walingest.rs
@@ -24,7 +24,7 @@
 use anyhow::Context;
 use postgres_ffi::v14::nonrelfile_utils::clogpage_precedes;
 use postgres_ffi::v14::nonrelfile_utils::slru_may_delete_clogsegment;
-use postgres_ffi::{page_is_new, page_set_lsn};
+use postgres_ffi::{fsm_logical_to_physical, page_is_new, page_set_lsn};
 
 use anyhow::Result;
 use bytes::{Buf, Bytes, BytesMut};
@@ -612,20 +612,18 @@ impl<'a> WalIngest<'a> {
                 forknum: FSM_FORKNUM,
             };
 
-            // FIXME: 'blkno' stored in the WAL record is the new size of the
-            // heap. The formula for calculating the new size of the FSM is
-            // pretty complicated (see FreeSpaceMapPrepareTruncateRel() in
-            // PostgreSQL), and we should also clear bits in the tail FSM block,
-            // and update the upper level FSM pages. None of that has been
-            // implemented. What we do instead, is always just truncate the FSM
-            // to zero blocks. That's bad for performance, but safe. (The FSM
-            // isn't needed for correctness, so we could also leave garbage in
-            // it. Seems more tidy to zap it away.)
-            if rec.blkno != 0 {
-                info!("Partial truncation of FSM is not supported");
+            let fsm_logical_page_no = rec.blkno / pg_constants::SLOTS_PER_FSM_PAGE;
+            let mut fsm_physical_page_no = fsm_logical_to_physical(fsm_logical_page_no);
+            if rec.blkno % pg_constants::SLOTS_PER_FSM_PAGE != 0 {
+                // Tail of last remaining FSM page has to be zeroed.
+                // We are not doing it here: there should be FPI wal-record for this page stored sometimes laters
+                fsm_physical_page_no += 1;
             }
-            let num_fsm_blocks = 0;
-            self.put_rel_truncation(modification, rel, num_fsm_blocks)?;
+            let nblocks = self.get_relsize(rel, modification.lsn)?;
+            if nblocks > fsm_physical_page_no {
+                // check if something to do: FSM is larger than truncate position
+                self.put_rel_truncation(modification, rel, fsm_physical_page_no)?;
+            }
         }
         if (rec.flags & pg_constants::SMGR_TRUNCATE_VM) != 0 {
             let rel = RelTag {
@@ -635,16 +633,17 @@ impl<'a> WalIngest<'a> {
                 forknum: VISIBILITYMAP_FORKNUM,
             };
 
-            // FIXME: Like with the FSM above, the logic to truncate the VM
-            // correctly has not been implemented. Just zap it away completely,
-            // always. Unlike the FSM, the VM must never have bits incorrectly
-            // set. From a correctness point of view, it's always OK to clear
-            // bits or remove it altogether, though.
-            if rec.blkno != 0 {
-                info!("Partial truncation of VM is not supported");
+            let mut vm_page_no = rec.blkno / pg_constants::VM_HEAPBLOCKS_PER_PAGE;
+            if rec.blkno % pg_constants::VM_HEAPBLOCKS_PER_PAGE != 0 {
+                // Tail of last remaining vm page has to be zeroed.
+                // We are not doing it here: there should be FPI wal-record for this page stored sometimes laters
+                vm_page_no += 1;
             }
-            let num_vm_blocks = 0;
-            self.put_rel_truncation(modification, rel, num_vm_blocks)?;
+            let nblocks = self.get_relsize(rel, modification.lsn)?;
+            if nblocks > vm_page_no {
+                // check if something to do: VM is larger than truncate position
+                self.put_rel_truncation(modification, rel, vm_page_no)?;
+            }
         }
         Ok(())
     }
--- a/test_runner/regress/test_truncate.py
+++ b/test_runner/regress/test_truncate.py
@@ -43,4 +43,5 @@ def test_truncate(neon_env_builder: NeonEnvBuilder, zenbenchmark):
 
     cur.execute("select count(*) from t1")
     res = cur.fetchone()
+    assert res is not None
     assert res[0] == n_records
--- a/test_runner/regress/test_truncate.py
+++ b/test_runner/regress/test_truncate.py
@@ -1,0 +1,46 @@
+import time
+
+from fixtures.neon_fixtures import NeonEnvBuilder
+
+
+#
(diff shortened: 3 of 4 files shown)


hlinnaka:
> I managed to create test reproducing the problem (without this PR).

Great! Please share

> But it cause error in GC which is not currently detected by our test framework.
> It will be done with the patch proposed in https://github.com/neondatabase/neon/pull/2558 (abnormal termination of pageserver built with testing feature in case of errors in GC/compaction threads).

I opened a PR for an alternative approach to that: https://github.com/neondatabase/neon/pull/2653. It will probably take a few iterations through the CI to stabilize that, as with that any unexpected ERRORs will fail tests and I'm sure the CI will find a bunch of unexpected cases. But you could cherry-pick that approach to a single test even before that.

pageserver/src/walingest.rs
```diff
@@ -613,20 +613,18 @@ impl<'a> WalIngest<'a> {
...
-            // it. Seems more tidy to zap it away.)
-            if rec.blkno != 0 {
-                info!("Partial truncation of FSM is not supported");
+            let fsm_logical_page_no = rec.blkno / pg_constants::SLOTS_PER_FSM_PAGE;
+            let mut fsm_physical_page_no = fsm_logical_to_physical(fsm_logical_page_no);
+            if rec.blkno % pg_constants::SLOTS_PER_FSM_PAGE != 0 {
+                // Tail of last remaining FSM page has to be zeroed.
+                // We are not doing it here: there should be FPI wal-record for this page stored sometimes laters
+                fsm_physical_page_no += 1;
+            }
```
arssher, line 622: Shouldn't we also update upper level FSM pages as removed comment said?

knizhnik: We do not update pages at all at this point.
This is what I tried to say in the comment I have added instead of removed.
We are just updating relation size here (it is separate key-value entry in pageserver storage).

All FSM/VM pages are updated t compute nodes and when them are written (when smgrwrite is called for them), 
pagestore_smgr enforces wal-logging of FSM/VM pages so that their content is transferred to pageserver.

pageserver/src/walingest.rs
```diff
@@ -636,16 +634,17 @@ impl<'a> WalIngest<'a> {
...
-            // set. From a correctness point of view, it's always OK to clear
-            // bits or remove it altogether, though.
-            if rec.blkno != 0 {
-                info!("Partial truncation of VM is not supported");
+            let mut vm_page_no = rec.blkno / pg_constants::VM_HEAPBLOCKS_PER_PAGE;
+            if rec.blkno % pg_constants::VM_HEAPBLOCKS_PER_PAGE != 0 {
+                // Tail of last remaining vm page has to be zeroed.
+                // We are not doing it here: there should be FPI wal-record for this page stored sometimes laters
+                vm_page_no += 1;
+            }
```
hlinnaka, line 642: This seems unsafe. Firstly, there's no guarantee that the FPI record follows. PostgreSQL might've crashed before the record was written out. Secondly, you might be doing point-in-time query between the two records.

knizhnik: Content of FSM/VM pages is non critical. It is expected to be lost.
By default changes of FSM/VM are no wal-logged.
So it seems to be normal in Vanilla that after abnormal termination we can have partly updated map.
I do no think here some difference in Neon and Vanilla behavior. Yes, in Neon we are separately updating relation size. But what wrong can happen if length o relation is updated and content - not?

hlinnaka: It's safe to clear bits in the VM, but if a bit is set in the VM, it's important that all the tuples on the pages are visible, or you can get incorrect results in queries. We mustn't leave behind bits that are set when they shouldn't be.

knizhnik: I have to agree with you - it can cause problems.
Actually it is harmless to make some non-zeroed bit in VM map beyond end of the relation.
As far as there are no records o this pages, marking them as all-visible in visibility map should not be a problem.
Moreover - it is even correct: all records on empty page are visible:)

The problem arises when we try to isert some record o this page. Insert WAL record has flag `XLH_INSERT_ALL_VISIBLE_CLEARED`, but it is set only if page was mrked as all-visible.
And new page definitely has no such mark.

But I think that it is not so difficult to fix the problem.
Truncation is performed by `RelationTruncate` which in turn calls `visibilitymap_prepare_truncate`. 
And it contains the following code:
```
		/*
		 * Truncation of a relation is WAL-logged at a higher-level, and we
		 * will be called at WAL replay. But if checksums are enabled, we need
		 * to still write a WAL record to protect against a torn page, if the
		 * page is flushed to disk before the truncation WAL record. We cannot
		 * use MarkBufferDirtyHint here, because that will not dirty the page
		 * during recovery.
		 */
		MarkBufferDirty(mapBuffer);
		if (!InRecovery && RelationNeedsWAL(rel) && XLogHintBitIsNeeded())
			log_newpage_buffer(mapBuffer, false);
```
So if we have `wal_log_hints` set, then we will insert in WAL FPI record with truncated page image  before `XLOG_SMGR_TRUNCATE `. No inconsistency is possible in this case.
But we do not want to switch on `wal_log_hints`. what can we do is just remove `XLogHintBitIsNeeded()` preicate here  or do something like this:
```
	if (!InRecovery && RelationNeedsWAL(rel) && (XLogHintBitIsNeeded() || IsNeonMode)
```

where is `IsNeonMode` (suggest better name) is flag set in neon exntesion.
I am not sure that we need such flag in our Postgres fork. Because I do not see any reasons to use our postgres in "non-Neon" mode. 

Alternative solution is to implement correct zeroing of the tail of last bitmap page in walingest.rs.
Bit if it possible to avoid code duplication (convert C code to Rust), I prefer not to do it.

What do you think?

hlinnaka: > As far as there are no records o this pages, marking them as all-visible in visibility map should not be a problem.
> Moreover - it is even correct: all records on empty page are visible:)

:-). Not quite though: the actual rule is that the if a heap page has `!PD_ALL_VISIBLE`, then the visibility map bit must not be set. A newly-initialized page does not have its `PD_ALL_VISIBLE` flag set.

hlinnaka: > But I think that it is not so difficult to fix the problem. Truncation is performed by `RelationTruncate` which in turn calls `visibilitymap_prepare_truncate`. And it contains the following code:
> 
> ```
> 		/*
> 		 * Truncation of a relation is WAL-logged at a higher-level, and we
> 		 * will be called at WAL replay. But if checksums are enabled, we need
> 		 * to still write a WAL record to protect against a torn page, if the
> 		 * page is flushed to disk before the truncation WAL record. We cannot
> 		 * use MarkBufferDirtyHint here, because that will not dirty the page
> 		 * during recovery.
> 		 */
> 		MarkBufferDirty(mapBuffer);
> 		if (!InRecovery && RelationNeedsWAL(rel) && XLogHintBitIsNeeded())
> 			log_newpage_buffer(mapBuffer, false);
> ```
> 
> So if we have `wal_log_hints` set, then we will insert in WAL FPI record with truncated page image before `XLOG_SMGR_TRUNCATE `. No inconsistency is possible in this case. But we do not want to switch on `wal_log_hints`. what can we do is just remove `XLogHintBitIsNeeded()` preicate here or do something like this:
> 
> ```
> 	if (!InRecovery && RelationNeedsWAL(rel) && (XLogHintBitIsNeeded() || IsNeonMode)
> ```
> 
> where is `IsNeonMode` (suggest better name) is flag set in neon exntesion. I am not sure that we need such flag in our Postgres fork. Because I do not see any reasons to use our postgres in "non-Neon" mode.

Hmm, that works for WAL generated with patched postgres. It will still not do the right thing when digesting WAL from vanilla PostgreSQL, however. It would be good to maintain that compatibility.

> Alternative solution is to implement correct zeroing of the tail of last bitmap page in walingest.rs. Bit if it possible to avoid code duplication (convert C code to Rust), I prefer not to do it.

Yeah, it's a fair amount of hairy code :-(.

Third option would be to overwrite the whole last bitmap page with zeros in walingest.rs. That should be safe. Although that's what we thought about just truncating the VM too, and that turned out to be a bad idea, so I feel a little scared about that.

It would be nice to change this in PostgreSQL. It's weird that we WAL-log the full-page image separately like that. It would seem more natural to include that full-page image in the XLOG_SMGR_TRUNCATE record itself. And when we don't log a full page image, it would seem natural to include the byte and bit ranges that are being zeroed in the XLOG_SMGR_TRUNCATE record, instead of re-doing the math in the redo function.

knizhnik: > It will still not do the right thing when digesting WAL from vanilla PostgreSQL,
 
But our wal-redo in any case will not be able to reply WAL generated by Vanilla because of format difference (cid,...). So I do not think that we loose something by making this change.

> Third option would be to overwrite the whole last bitmap page with zeros in walingest.rs.

Yes, it is possible and will be not difficult. But "aesthetically" I do not like this option: it looks too compromising. I prefer that we either do nothing with bitmap content, either correctly clear it. But here we update bitmap, but not precisely... But if you think that it is the nest solution, then I will implement it: just adding few lines of code.

libs/postgres_ffi/src/lib.rs
```diff
@@ -163,6 +163,24 @@ pub fn page_set_lsn(pg: &mut [u8], lsn: Lsn) {
     pg[4..8].copy_from_slice(&(lsn.0 as u32).to_le_bytes());
 }
 
+pub fn fsm_logical_to_physical(addr: BlockNumber) -> BlockNumber {
```
lubennikovaav, line 166: This function differs from the Postgres one. It lacks this piece of logic:

```
	/*
	 * If the page we were asked for wasn't at the bottom level, subtract the
	 * additional lower level pages we counted above.
	 */
	pages -= addr.level;
```
Is it safe, because caller will never provide non-bottom level page address?
Please, add a comment about it.

pageserver/src/walingest.rs
```diff
@@ -612,20 +612,19 @@ impl<'a> WalIngest<'a> {
...
-            // it. Seems more tidy to zap it away.)
-            if rec.blkno != 0 {
-                info!("Partial truncation of FSM is not supported");
+            let fsm_logical_page_no = rec.blkno / pg_constants::SLOTS_PER_FSM_PAGE;
+            let mut fsm_physical_page_no = fsm_logical_to_physical(fsm_logical_page_no);
+            if rec.blkno % pg_constants::SLOTS_PER_FSM_PAGE != 0 {
+                // Tail of last remaining FSM page has to be zeroed.
+                // We are not precise here and instead of digging in FSM bitmap format just clear the whole page.
+                modification.put_rel_page_image(rel, fsm_physical_page_no, ZERO_PAGE.clone())?;
+                fsm_physical_page_no += 1;
```
lubennikovaav, line 621: Why do we increase `fsm_physical_page_no` here?
If I read the code above right, we only updated already existing last remaining page

lubennikovaav (commented):
Let's add more tests to verify that time travel works well with truncation.
I.e. check that we can see all the data on branch created before VM truncation

test_runner/regress/test_truncate.py
```diff
@@ -0,0 +1,47 @@
...
+# Benchmark searching the layer map, when there are a lot of small layer files.
+#
+def test_truncate(neon_env_builder: NeonEnvBuilder, zenbenchmark):
+
+    env = neon_env_builder.init_start()
+    n_records = 10000
+    n_iter = 10
+
+    # We want to have a lot of lot of layer files to exercise the layer map. Make
+    # gc_horizon and checkpoint_distance very small, so that we get a lot of small layer files.
```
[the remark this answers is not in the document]
knizhnik: Done

test_runner/regress/test_truncate.py
```diff
@@ -0,0 +1,47 @@
...
+    tenant, _ = env.neon_cli.create_tenant(
+        conf={
+            "gc_period": "100 m",
+            "gc_horizon": "1048576",
+            "checkpoint_distance": "1000000",
+            "compaction_period": "1 s",
+            "compaction_threshold": "3",
+            "image_creation_threshold": "1",
+            "compaction_target_size": "1000000",
```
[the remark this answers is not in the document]
knizhnik: I explained them in the comment

test_runner/regress/test_truncate.py
```diff
@@ -0,0 +1,47 @@
...
+        }
+    )
+
+    env.neon_cli.create_timeline("test_truncate", tenant_id=tenant)
+    pg = env.postgres.create_start("test_truncate", tenant_id=tenant)
+    cur = pg.connect().cursor()
+    cur.execute("set statement_timeout=0")
```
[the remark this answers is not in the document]
knizhnik: No, it was needed for large number of records but to managed to reproduce the problem with smaller amount which fits in default statement timeout.

test_runner/regress/test_truncate.py
```diff
@@ -0,0 +1,47 @@
...
+    cur.execute("create table t1(x integer)")
+    cur.execute(f"insert into t1 values (generate_series(1,{n_records}))")
+    cur.execute("vacuum t1")
+    for i in range(n_iter):
+        cur.execute(f"delete from t1 where x>{n_records//2}")
+        cur.execute("vacuum t1")
+        time.sleep(1)
```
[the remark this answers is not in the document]
knizhnik: To make pageserver a chance to create image layer with truncated relation.

libs/postgres_ffi/src/lib.rs
knizhnik: Done

pageserver/src/walingest.rs
knizhnik: `fsm_physical_page_no` is physical number of the page till which relation should be truncated.
If truncation is performed not at page boundary, then we should leave last page and zero remaining items in it.
It is port of  `FreeSpaceMapPrepareTruncateRel`:
```
	/*
	 * Zero out the tail of the last remaining FSM page. If the slot
	 * representing the first removed heap block is at a page boundary, as the
	 * first slot on the FSM page that first_removed_address points to, we can
	 * just truncate that page altogether.
	 */
	if (first_removed_slot > 0)
	{
	     new_nfsmblocks = fsm_logical_to_physical(first_removed_address) + 1;
	}

File: test_runner/regress/test_truncate.py [hunk 9]
```diff
@@ -0,0 +1,47 @@
...
+    env.neon_cli.create_timeline("test_truncate", tenant_id=tenant)
+    pg = env.postgres.create_start("test_truncate", tenant_id=tenant)
+    cur = pg.connect().cursor()
+    cur.execute("set statement_timeout=0")
+    cur.execute("create table t1(x integer)")
+    cur.execute(f"insert into t1 values (generate_series(1,{n_records}))")
+    cur.execute("vacuum t1")
+    for i in range(n_iter):
+        cur.execute(f"delete from t1 where x>{n_records//2}")
+        cur.execute("vacuum t1")
```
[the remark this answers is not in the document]
knizhnik: It is difficult to insert some strict checks here because all operations are non-deterministic.
Order of storing records in relation is not specified, so there is absolutely no warranty that after insertion some number of records and them deleting some some part of them relation size will really decrease.

knizhnik:
> Let's add more tests to verify that time travel works well with truncation. I.e. check that we can see all the data on branch created before VM truncation

The original idea of this test was to reproduce the problem on the unpacked sources. It was very non-trivial. I have spent about week experimenting with different parameters. And still result is non-deterministic. Creation of branch may change workload. I prefer not to do it. Also FSM/VM are just hints. Them do not have direct impact in query execution results.
I.e. if FSM was not truncated or incorrectly truncated, we will not notice it.
we can only notice incorrect truncation of heap relation. But it doesn't have relation to the topic of this ticket.

lubennikovaav:
> The original idea of this test was to reproduce the problem on the unpacked sources. It was very non-trivial. I have spent about week experimenting with different parameters. And still result is non-deterministic.

I've tried to run this test without a fix from this branch and it passed. Is it expected? Do I need to try more iterations?

test_runner/regress/test_truncate.py
```diff
@@ -0,0 +1,46 @@
+import time
+
+from fixtures.neon_fixtures import NeonEnvBuilder
+
+
+#
+# Benchmark searching the layer map, when there are a lot of small layer files.
```
lubennikovaav, line 7:
```suggestion
# Test truncation of FSM and VM forks of a relation
```

test_runner/regress/test_truncate.py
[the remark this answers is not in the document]
lubennikovaav: Please, add this comment in the test

knizhnik:
> I've tried to run this test without a fix from this branch and it passed. Is it expected? Do I need to try more iterations?

At y host the problem is still reproduced with the latest main:
```
E       AssertionError: assert not ['2022-11-22T13:03:55.985492Z  WARN compaction_loop{tenant_id=c47cc0e8d698ac995f2fe810bb97abbc}:compact_timeline{timel...os: could not find data for key 000000067F000032D2000040000200000000 at LSN 0/1B901F8, for request at LSN 0/1DC85B8\n']

test_runner/fixtures/neon_fixtures.py:1838: AssertionError
```
I am using release build.
As I already wrote - the test is not deterministic. So may be the problem is not reproduced at slower (or may be faster) system.
Sorry, but I failed to create some deterministic test.
Using larger number of iterations may help.

But in any case - the problem is clear. Its fix is just porting of postgres code. Certainly I can introduce some bugs during this port. But at least I never saw "key not found"errors with this patch, neither with test_truncate.oy neither with any other test.

Should we add more tests? May be. I do not think that it is necessary, but may just because I too lazy to write more tests:)

test_runner/regress/test_truncate.py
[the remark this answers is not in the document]
knizhnik: Sorry, but TRUNCATE actually doesn't log truncation records in WAL.
It just removes relation files.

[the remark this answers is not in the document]
knizhnik: > So, we don't really test truncation codepath, but only the part that zeroes out the page, right?

Before this PR we truncate FSM/VM fork to zero in this case. And it cause "key not found" error.
So the main goal of this test was to demonstrate that this error is not happening any more.
Yes, used table size is too small to cause truncation of FSM fork. But if we increase it, then it really happen:
```
postgres=# insert into t1 values (generate_series(10000/2+1,1000000));
INSERT 0 995000
postgres=# select pg_relation_size('t1', 'fsm')/8192;
 ?column? 
----------
        4
(1 row)

postgres=# delete from t1 where x>10000/2;
DELETE 1090000
postgres=# vacuum t1;
VACUUM
postgres=# select pg_relation_size('t1', 'fsm')/8192;
 ?column? 
----------
        3
(1 row)
```

lubennikovaav (approved):
Thank you for answering all my concerns)

test_runner/regress/test_truncate.py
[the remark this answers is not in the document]
knizhnik: Done

test_runner/regress/test_truncate.py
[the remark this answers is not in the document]
knizhnik: Removed

PR #2610, opened 2022-10-12
Ensure that local layer metadata is always correct on a type level

SomeoneToIgnore:
This is a proposal for improvements on https://github.com/neondatabase/neon/pull/2582/files#r991166489 discussion.

The code is adjusted to always keep a "either full metadata with file size, or no metadata at all" transition for every layer.

### DIFF
--- a/pageserver/src/storage_sync/index.rs
+++ b/pageserver/src/storage_sync/index.rs
@@ -212,8 +212,8 @@ impl RemoteTimelineIndex {
 /// Restored index part data about the timeline, stored in the remote index.
 #[derive(Debug, Clone)]
 pub struct RemoteTimeline {
-    timeline_layers: HashMap<PathBuf, LayerFileMetadata>,
-    missing_layers: HashMap<PathBuf, LayerFileMetadata>,
+    timeline_layers: HashMap<PathBuf, Option<LayerFileMetadata>>,
+    missing_layers: HashMap<PathBuf, Option<LayerFileMetadata>>,
 
     pub metadata: TimelineMetadata,
     pub awaits_download: bool,
@@ -233,14 +233,16 @@ impl RemoteTimeline {
         &mut self,
         new_layers: impl IntoIterator<Item = (PathBuf, LayerFileMetadata)>,
     ) {
-        self.timeline_layers.extend(new_layers);
+        self.timeline_layers
+            .extend(new_layers.into_iter().map(|(p, m)| (p, Some(m))));
     }
 
     pub fn add_upload_failures(
         &mut self,
         upload_failures: impl IntoIterator<Item = (PathBuf, LayerFileMetadata)>,
     ) {
-        self.missing_layers.extend(upload_failures);
+        self.missing_layers
+            .extend(upload_failures.into_iter().map(|(p, m)| (p, Some(m))));
     }
 
     pub fn remove_layers(&mut self, layers_to_remove: &HashSet<PathBuf>) {
@@ -257,14 +259,12 @@ impl RemoteTimeline {
 
     pub fn from_index_part(timeline_path: &Path, index_part: IndexPart) -> anyhow::Result<Self> {
         let metadata = TimelineMetadata::from_bytes(&index_part.metadata_bytes)?;
-        let default_metadata = &IndexLayerMetadata::default();
 
-        let find_metadata = |key: &RelativePath| -> LayerFileMetadata {
-            index_part
-                .layer_metadata
-                .get(key)
-                .unwrap_or(default_metadata)
-                .into()
+        let find_metadata = |key: &RelativePath| -> Option<LayerFileMetadata> {
+            let index_metadata = index_part.layer_metadata.get(key)?;
+            Some(LayerFileMetadata {
+                file_size: index_metadata.file_size?,
+            })
         };
 
         Ok(Self {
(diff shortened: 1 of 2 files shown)


pageserver/src/storage_sync/index.rs
```diff
@@ -448,22 +427,26 @@ pub struct IndexLayerMetadata {
...
-    input: &HashMap<PathBuf, LayerFileMetadata>,
+    input: &HashMap<PathBuf, Option<LayerFileMetadata>>,
     output: &mut HashSet<RelativePath>,
     layer_metadata: &mut HashMap<RelativePath, IndexLayerMetadata>,
 ) -> anyhow::Result<()> {
-    for (path, metadata) in input {
+    for (path, local_layer_metadata) in input {
         let rel_path = RelativePath::new(timeline_path, path)?;
-        let metadata = IndexLayerMetadata::from(metadata);
+        // TODO kb wrong transition
```
SomeoneToIgnore, line 443: Note: this code has this and lacks the way of "upgrading" the old files with no metadata: we can do this, since download the entire timeline locally.

pageserver/src/storage_sync/index.rs
```diff
@@ -257,14 +259,12 @@ impl RemoteTimeline {
...
-            index_part
-                .layer_metadata
-                .get(key)
-                .unwrap_or(default_metadata)
-                .into()
+
+        let find_metadata = |key: &RelativePath| -> Option<LayerFileMetadata> {
+            let index_metadata = index_part.layer_metadata.get(key)?;
+            Some(LayerFileMetadata {
+                file_size: index_metadata.file_size?,
```
koivunej, line 266: The reason I dismissed this design is what will happen after adding:

```patch
 pub struct LayerFileMetadata {
     file_size: u64,
+    file_hash: FileHash,
 }
```

When upgrading from this size-only metadata to size+hash, would we be ready to let go of the existing size-only metadata in that case as well? It feels like it would be nice to have an additional state for being-loaded or being-downloaded `RemoteIndex` which is when this upgrade could take place, but this is only starting to dawn upon me now that I've wrestled the download changes [1](https://github.com/neondatabase/neon/pull/2582/commits/b9eee63dba0937693d0ed64d12f14479800e286b), [2](https://github.com/neondatabase/neon/pull/2582/commits/08c7cb38fd01ea7c23043ab7a817feb0a99b21cd) in.

SomeoneToIgnore: >  would we be ready to let go of the existing size-only metadata in that case as well?

Maybe this is something that I'm not understanding and most probably due to the fact we still have no place where we propagate old layer metadata to the new one.

In my imagination, we still have the layer file at some point, where we could calculate both.
As a middle ground, we can use something else instead of `Option`:
```rust
enum LayerMetadata {
    Complete(ProperLayerMetadata),
    Partial(IndexLayerMetadata),
    None,
}
```
but maybe that's an overcomplication already, need to see the upgrade place first...

SomeoneToIgnore:
The PR already has conflicts, I have posted links to this example PR, closing it to not to cause confusion.

pageserver/src/storage_sync/index.rs
koivunej: Just recapping our yesterday chat here, probably with some extras.

> Maybe this is something that I'm not understanding and most probably due to the fact we still have no place where we propagate old layer metadata to the new one.

During initial download is the place where we combine or merge the possibly old metadata into the current version: https://github.com/neondatabase/neon/pull/2582/commits/08c7cb38fd01ea7c23043ab7a817feb0a99b21cd -- To facilitate removing the `Option` from `LayerFileMetadata` during download, the type of `LayerFileMetadata` must be closer to `IndexLayerMetadata`.

My best guess for *how* this could happen is to complicate the type by which we are doing the download by, which I intend to investigate today. I was earlier thinking that `RemoteIndex` needs to change but perhaps the `Option<LayerFileMetadata>` as might work as well. However after attach we don't really need the Option<LayerFileMetadata>` either.

koivunej: Adding two variants for `RemoteIndex`:

```
pub enum RemoteTimeline {
    AwaitingDownload {
        timeline_layers: HashMap<PathBuf, IndexLayerMetadata>,
        missing_layers: HashMap<PathBuf, IndexLayerMetadata>,
        metadata: TimelineMetadata,
    },
    Ready {
        timeline_layers: HashMap<PathBuf, LayerFileMetadata>,
        missing_layers: HashMap<PathBuf, LayerFileMetadata>,
        metadata: TimelineMetadata,
    },
}
```

Then in `compare_local_and_remote_timeline`:
- `RemoteTimeline::AwaitingDownload` gets turned into `RemoteTimeline::Ready` iff all of the metadata matches up, and all files are present
    - Metadata "matches up" is a new addition compared to previous
    - Assumes that we can read all metadata during startup, which is an easy assumption with the layer file length
- After a successful download a new operation will have to do the similar "upgrade" from RemoteTimeline::AwaitingDownload

At later "runtime" after the tenant has become ready, this does pose some issues however, because we will still have to continually manage the "which RemoteTimeline variant is it". I can imagine two solutions for that:

1. Complicate `RemoteIndex::AwaitingDownload` by separating the "local files" and "remote not yet downloaded files" (which is also managed in `LayerDownload`
2. Move the RemoteTimeline enumerization split higher up, by having different types for "being attached timeline" and "attached timeline"

I have a feeling the "needing to juggle two types at runtime" can only be pulled upwards until we have complete separation for completely ready tenants and possibly incomplete tenants. Doing that seems counterproductive in light of already conflicting #2575 and the future direction of #2595. I haven't yet read either of these but I assume at least in #2595 runtime juggling of different states is requirement.

SomeoneToIgnore: Feels like we're on the right track: we now have a notion of remote and local timelines.

I'm thinking, that with on-demand downloads we start to abandon the holistic timeline approach: there will be `RemoteLayer` in https://github.com/neondatabase/neon/pull/2595, so timeline would not necessary be "either with all layers locally or will all layers remotely" as in the enum above, it will have a mix of both.

In our case, a layer could have full metadata (and that metadata will be in index.json) or could have incomplete/empty metadata that require downloading the layer, calculating and updating the info in the index and remotely.

I'm not sure that you have to invest heavily into this approach unless you see the way forward now: your version has some good roots in it and we will improve it later.

koivunej: To summarize: At least I cannot see a way out of needing to have `Option<u64>` at the `LayerFileMetadata`. If not there, that kind of "type state" needs to exist as long as we have a central place to store tenants in different states. Or perhaps it will be discovered the instant #2582 is merged.

At this point #2582 still needs some actual usage of `LayerFileMetadata`, namely in the `compare_local_and_remote_timeline` to decide to download truncated layer files for example. I'll continue towards that and we can circle back to this topic if good ideas come up.

PR #2611, opened 2022-10-12
Reimplement explicit timeline creation on safekeepers.

arssher:
With the ability to pass commit_lsn. This allows to perform project WAL recovery through different (from the original) set of safekeepers (or under different ttid) by
1) moving WAL files to s3 under proper ttid;
2) explicitly creating timeline on safekeepers, setting commit_lsn to the latest point;
3) putting the lastest .parital file to the timeline directory on safekeepers, if desired.

Extend test_s3_wal_replay to exersise this behaviour.

Also extends timeline_status endpoint to return postgres information.

### DIFF
--- a/safekeeper/src/http/routes.rs
+++ b/safekeeper/src/http/routes.rs
@@ -1,8 +1,8 @@
-use anyhow::anyhow;
 use hyper::{Body, Request, Response, StatusCode, Uri};
 
 use anyhow::Context;
 use once_cell::sync::Lazy;
+use postgres_ffi::WAL_SEGMENT_SIZE;
 use serde::Serialize;
 use serde::Serializer;
 use std::collections::{HashMap, HashSet};
@@ -10,6 +10,7 @@ use std::fmt::Display;
 use std::sync::Arc;
 use tokio::task::JoinError;
 
+use crate::safekeeper::ServerInfo;
 use crate::safekeeper::Term;
 use crate::safekeeper::TermHistory;
 
@@ -77,6 +78,7 @@ struct TimelineStatus {
     #[serde(serialize_with = "display_serialize")]
     timeline_id: TimelineId,
     acceptor_state: AcceptorStateStatus,
+    pg_info: ServerInfo,
     #[serde(serialize_with = "display_serialize")]
     flush_lsn: Lsn,
     #[serde(serialize_with = "display_serialize")]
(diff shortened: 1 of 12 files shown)


safekeeper/src/http/routes.rs
```diff
@@ -136,12 +138,29 @@ async fn timeline_create_handler(mut request: Request<Body>) -> Result<Response<
...
+        tenant_id: request_data.tenant_id,
         timeline_id: request_data.timeline_id,
     };
     check_permission(&request, Some(ttid.tenant_id))?;
 
-    Err(ApiError::BadRequest(anyhow!("not implemented")))
+    let server_info = ServerInfo {
+        pg_version: request_data.pg_version,
+        system_id: request_data.system_id.unwrap_or(0),
+        wal_seg_size: request_data.wal_seg_size.unwrap_or(16 * (1 << 20)),
```
petuhovskiy, line 149: We have a const with default SEGMENT_SIZE somewhere, better to use it here.

safekeeper/src/timeline.rs
```diff
@@ -339,10 +339,12 @@ impl Timeline {
         ttid: TenantTimelineId,
         wal_backup_launcher_tx: Sender<TenantTimelineId>,
         server_info: ServerInfo,
+        commit_lsn: Lsn,
+        local_start_lsn: Lsn,
     ) -> Result<Timeline> {
```
petuhovskiy, line 344: Since we now can accept arbitrary `commit_lsn` and `local_start_lsn`, it's better to check validity (`commit_lsn >= local_start_lsn`) of them, I'd do that in `SharedState::create_new`.

safekeeper/src/timelines_global_map.rs
```diff
@@ -190,6 +198,9 @@ impl GlobalTimelines {
             Ok(_) => {
                 // We are done with bootstrap, release the lock, return the timeline.
                 drop(shared_state);
+                timeline
```
petuhovskiy, line 201: It's probably better to send wal_backup notififcation right in the `bootstrap()`.

safekeeper/src/safekeeper.rs
```diff
@@ -233,10 +235,10 @@ impl SafeKeeperState {
             server: server_info,
             proposer_uuid: [0; 16],
             timeline_start_lsn: Lsn(0),
-            local_start_lsn: Lsn(0),
-            commit_lsn: Lsn(0),
-            backup_lsn: Lsn::INVALID,
-            peer_horizon_lsn: Lsn(0),
+            local_start_lsn,
+            commit_lsn,
+            backup_lsn: commit_lsn,
```
petuhovskiy, line 240: `local_start_lsn` is probably a better fit here and for `peer_horizon_lsn`.

petuhovskiy (approved):
`timeline_start_lsn` will be incorrect, but I think it's ok for manual tests

PR #2618, opened 2022-10-14
Bump regular dependencies

SomeoneToIgnore:
* `etcd-client` is not updated, since we plan to replace it with another client and the new version errors with some missing `prost` library error
* `clap` had released another major update that requires changing every CLI declaration again, deserves a separate PR

### DIFF
--- a/workspace_hack/Cargo.toml
+++ b/workspace_hack/Cargo.toml
@@ -16,18 +16,16 @@ publish = false
 [dependencies]
 ahash = { version = "0.7", features = ["std"] }
 anyhow = { version = "1", features = ["backtrace", "std"] }
-bstr = { version = "0.2", features = ["lazy_static", "regex-automata", "serde", "serde1", "serde1-nostd", "std", "unicode"] }
 bytes = { version = "1", features = ["serde", "std"] }
-chrono = { version = "0.4", features = ["clock", "libc", "oldtime", "serde", "std", "time", "winapi"] }
+chrono = { version = "0.4", features = ["clock", "iana-time-zone", "js-sys", "oldtime", "serde", "std", "time", "wasm-bindgen", "wasmbind", "winapi"] }
+clap = { version = "3", features = ["atty", "color", "std", "strsim", "suggestions", "termcolor"] }
 crossbeam-utils = { version = "0.8", features = ["once_cell", "std"] }
 either = { version = "1", features = ["use_std"] }
 fail = { version = "0.5", default-features = false, features = ["failpoints"] }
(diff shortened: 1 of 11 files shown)


SomeoneToIgnore:
This bump removes a few dependencies, but what bugs me is a bunch of `wasm` dependencies stuck in Cargo.lock such as `web-sys`, `js-sys` and others: I don't see any reason for them to be in our dependency tree.

koivunej:
> `wasm` dependencies stuck in Cargo.lock

I think they are there due to some crate having a feature requiring them, but sadly these are not locked to "only enabled features." Haven't figured out a way to get rid of them at least.

workspace_hack/Cargo.toml
```diff
@@ -16,26 +16,23 @@ publish = false
 [dependencies]
 ahash = { version = "0.7", features = ["std"] }
 anyhow = { version = "1", features = ["backtrace", "std"] }
-bstr = { version = "0.2", features = ["lazy_static", "regex-automata", "serde", "serde1", "serde1-nostd", "std", "unicode"] }
 bytes = { version = "1", features = ["serde", "std"] }
-chrono = { version = "0.4", features = ["clock", "libc", "oldtime", "serde", "std", "time", "winapi"] }
+chrono = { version = "0.4", features = ["clock", "iana-time-zone", "js-sys", "oldtime", "serde", "std", "time", "wasm-bindgen", "wasmbind", "winapi"] }
```
koivunej, line 20: I am not entirely up to speed what do workspace hacks do but having `wasm-bindgen`, `wasmbind` and `winapi` show up as features does seem weird. Was this expected?

koivunej (approved):
I didn't go through the whole Cargo.lock file, but otherwise looks as expected. Commented on the wasm dependencies above, but then found them in the workspace hack, which I am not sure if it was intended.

SomeoneToIgnore:
I've run `cargo build --timings` to see that those are not picked locally, and generally it seems that wasm ones are from `criterion`, so it might be ok for now.

workspace_hack/Cargo.toml
SomeoneToIgnore: https://crates.io/crates/cargo-hakari guess this would be the best explanation

As far as I understand it, it tries to unit all crates across all project under minimal common feature subset, so that the same crate does not get recompiled multiple times due to different features used.

koivunej: Resolved this already but seems that they are just the defaults:

https://github.com/chronotope/chrono/blob/f60893a0dce544ac6013038b6def0bddd562b1ca/Cargo.toml#L19 

which means it comes from pageserver and compute_tools.

Issue #2394, opened 2022-09-05
Epic: replace etcd with custom grpc based broker


arssher:
## Motivation

Custom broker would be much more scalable: https://neondb.slack.com/archives/C036U0GRMRB/p1661011605069469

## DoD

etcd is replaced, all current tests pass.

## Tasks
- [x] [publish the RFC](https://github.com/neondatabase/neon/pull/2466)
- [x] Add routing to the prototype https://github.com/arssher/tonic-bench
- [x] Replace s3 offloader election in safekeepers 
- [x] Add binary https://github.com/neondatabase/neon/pull/2829
- [x] Write helm chart for it https://github.com/neondatabase/helm-charts/pull/15
- [x] Deploy it https://github.com/neondatabase/neon/pull/2922
- [x] Replace etcd in cloud repo tests https://github.com/neondatabase/cloud/pull/3058
- [x] Actually replace https://github.com/neondatabase/neon/pull/2733

stepashka:
https://observer.zenith.tech/d/kcUnPx37z/etcd?orgId=1&refresh=1m — we should provide a replacement for this too :)

arssher:
pr https://github.com/neondatabase/neon/pull/2733

PR #2621, opened 2022-10-16
Determine safekeeper for offloading WAL without etcd election API.

arssher:
This API is rather pointless (as this commit --stat shows), since sane choice anyway requires knowledge of peers status and leaders lifetime in any case can intersect, which is fine for us -- so manual elections are straightforward. Here, we deterministically choose among the reasonably caught up safekeepers, shifting by timeline id to spread the load.

A step towards custom broker https://github.com/neondatabase/neon/issues/2394

### DIFF
--- a/safekeeper/src/wal_backup.rs
+++ b/safekeeper/src/wal_backup.rs
@@ -1,8 +1,7 @@
 use anyhow::{Context, Result};
-use etcd_broker::subscription_key::{
-    NodeKind, OperationKind, SkOperationKind, SubscriptionKey, SubscriptionKind,
-};
+
 use tokio::task::JoinHandle;
+use utils::id::NodeId;
 
 use std::cmp::min;
 use std::collections::HashMap;
@@ -26,13 +25,10 @@ use tracing::*;
 
 use utils::{id::TenantTimelineId, lsn::Lsn};
 
-use crate::broker::{Election, ElectionLeader};
-use crate::timeline::Timeline;
-use crate::{broker, GlobalTimelines, SafeKeeperConf};
+use crate::timeline::{PeerInfo, Timeline};
+use crate::{GlobalTimelines, SafeKeeperConf};
 
 use once_cell::sync::OnceCell;
-
-const BROKER_CONNECTION_RETRY_DELAY_MS: u64 = 1000;
 
 const UPLOAD_FAILURE_RETRY_MIN_MS: u64 = 10;
 const UPLOAD_FAILURE_RETRY_MAX_MS: u64 = 5000;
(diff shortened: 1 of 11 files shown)


safekeeper/src/wal_backup.rs
```diff
@@ -70,47 +66,100 @@ struct WalBackupTimelineEntry {
...
+                .unwrap_or(Lsn(0));
+            let mut caughtup_peers = capable_peers
+                .clone()
+                .filter(|p| p.commit_lsn >= threshold)
+                .collect::<Vec<_>>();
+            caughtup_peers.sort_by(|p1, p2| p1.sk_id.cmp(&p2.sk_id));
+
+            // To distribute the load, shift by timeline_id.
+            let offloader = caughtup_peers
+                [(u128::from(ttid.timeline_id) % caughtup_peers.len() as u128) as usize]
```
petuhovskiy, line 110: `caughtup_peers.len()` can be unstable in some cases and this can lead to electing different safekeepers at the same time, electing no one, or flapping between elected and not.

Possible causes:
- PeerInfo is sent from each safekeeper with the same interval (each second), but (phase?) shift is random. This means that PeerInfo from two different peers can have info a second apart. In the current logic we have `threshold` on `commit_lsn` and that means we can have problems if WAL is written faster than `128MB/s`.
- After timeline is just loaded, we have empty `alive_peers` and then new peers will come one by one in random order.
- PeerInfo is broadcasted from safekeeper only when it's active, it can also affect the offloader.


This should be ok for S3 upload, but we can think of how to make it better.

safekeeper/src/wal_backup.rs
```diff
@@ -70,47 +66,100 @@ struct WalBackupTimelineEntry {
...
+                .sk_id;
+
+            (
+                Some(offloader),
+                format!(
+                    "elected {} among {:?} peers, with {} of them being caughtup",
+                    offloader,
+                    capable_peers
```
petuhovskiy, line 118: Would be good to sort this array, to make reading logs easier.

safekeeper/src/wal_backup.rs
```diff
@@ -229,117 +265,75 @@ async fn backup_task_main(
...
+            // Perhaps peers advanced the position, check shmem value.
+            backup_lsn = self.timeline.get_wal_backup_lsn();
+            if backup_lsn.segment_number(self.wal_seg_size)
+                >= commit_lsn.segment_number(self.wal_seg_size)
+            {
+                continue;
+            }
 
-                        retry_attempt = min(retry_attempt + 1, u32::MAX);
+            match backup_lsn_range(
```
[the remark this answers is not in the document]
arssher: no, we just always PUT the segment

petuhovskiy (approved):

PR #2623, opened 2022-10-17
Bump clap and other minor dependencies

SomeoneToIgnore:
Follow-up of https://github.com/neondatabase/neon/pull/2618, that bumps `clap` and the rest of the dependencies.

### DIFF
--- a/proxy/src/main.rs
+++ b/proxy/src/main.rs
@@ -45,98 +45,43 @@ async fn main() -> anyhow::Result<()> {
         .with_target(false)
         .init();
 
-    let arg_matches = clap::App::new("Neon proxy/router")
-        .version(GIT_VERSION)
-        .arg(
-            Arg::new("proxy")
-                .short('p')
-                .long("proxy")
-                .takes_value(true)
-                .help("listen for incoming client connections on ip:port")
-                .default_value("127.0.0.1:4432"),
-        )
-        .arg(
-            Arg::new("auth-backend")
-                .long("auth-backend")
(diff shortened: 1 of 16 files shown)


proxy/src/main.rs
```diff
@@ -45,52 +45,122 @@ async fn main() -> anyhow::Result<()> {
...
             Arg::new("mgmt")
                 .short('m')
                 .long("mgmt")
-                .takes_value(true)
                 .help("listen for management callback connection on ip:port")
                 .default_value("127.0.0.1:7000"),
         )
         .arg(
             Arg::new("http")
-                .short('h')
```
SomeoneToIgnore, line 75: cc @funbringer 

I've removed this parameter due to the new `cli` test started failing below.
It complained, that this `-h` conflicts with the short form of `--help`.

LMK if there's anything else you want to do about this.

koivunej (approved):
Cannot see anything obviously wrong, but the clap changes with the refactorings are a lot to look at. But as the test suite works, those must be good.

proxy/src/main.rs
funbringer: Sure! To be honest, I'd drop al shorthands.

funbringer (approved):

proxy/src/main.rs
SomeoneToIgnore: I've reached my breaking change limit recently and prefer to do less of those for a while, so will leave the others as is so far, thank you for confirming 🙂

PR #2626, opened 2022-10-17
Add data format backward compatibility tests

bayandin:
This PR adds two tests:
- `test_prepare_snapshot`: 
  - Creates a new project with `pg_bench -i` and then creates some activity in it with `pg_bench` for 1 minute.
  - Dumps the project with `pg_dumpall`
  - Does s a timeline checkpoint
  - Puts the project dir and the dump file into a special directory 

In a CI workflow, we upload this snapshot to S3 for `release` branch PRs.

Before running tests on a PR, we download the previous snapshot from S3. 

- `test_backward_compatibility`
  - Starts a project from the downloaded snapshot directory via `neon_local` CLI
  - Dumps a project with `pg_dumpall` and check that there's no difference in the dump (but commented lines, maybe)
  - Tries to perform a recovery from WAL (based on https://github.com/neondatabase/cloud/wiki/Recovery-from-WAL) 

There's a way to allow breaking changes to pass — add a `breaking changes` label to the PR, it will be converted into `ALLOW_BREAKING_CHANGES=true` env variable, and it will allow the project to fail to start from the snapshot dir (the logical dump won't be checked in this case).

I've tested  that`test_backward_compatibility` catches breaking changes like https://github.com/neondatabase/neon/commit/1fb3d081854a31f9afd1f4e5161fa4cbf9738299:
```bash
git revert 1fb3d081854a31f9afd1f4e5161fa4cbf9738299
CARGO_BUILD_FLAGS="--features=testing" make -j$(nproc)
./scripts/pytest test_runner/regress/test_compatibility.py::test_prepare_snapshot
git revert @
CARGO_BUILD_FLAGS="--features=testing" make -j$(nproc)
COMPATIBILITY_SNAPSHOT_DIR=test_output/test_prepare_snapshot/compatibility_snapshot_pg14 ./scripts/pytest test_runner/regress/test_compatibility.py::test_backward_compatibility
# it fails successfully
ALLOW_BREAKING_CHANGES=true COMPATIBILITY_SNAPSHOT_DIR=test_output/test_prepare_snapshot/compatibility_snapshot_pg14 ./scripts/pytest test_runner/regress/test_compatibility.py::test_backward_compatibility
# it xfails successfully, ie it allows a test suite to pass
```

The current version of the snapshot I've created on the top of the last release we had: https://github.com/neondatabase/neon/tree/979ad60c1985966fa0452d50c46f7b705d338609

### DIFF
--- a/.github/actions/run-python-test-set/action.yml
+++ b/.github/actions/run-python-test-set/action.yml
@@ -167,7 +167,7 @@ runs:
       if: github.ref_name == 'release'
       uses: ./.github/actions/upload
       with:
-        name: compatibility-snapshot-${{ inputs.build_type }}-pg14
+        name: compatibility-snapshot-${{ inputs.build_type }}-pg14-${{ github.run_id }}
         # The path includes a test name (test_prepare_snapshot) and directory that the test creates (compatibility_snapshot_pg14), keep the path in sync with the test
         path: /tmp/test_output/test_prepare_snapshot/compatibility_snapshot_pg14/
         prefix: latest
--- a/.github/actions/run-python-test-set/action.yml
+++ b/.github/actions/run-python-test-set/action.yml
@@ -73,6 +73,13 @@ runs:
       shell: bash -euxo pipefail {0}
       run: ./scripts/pysync
 
+    - name: Download compatibility snapshot for Postgres 14
+      uses: ./.github/actions/download
+      with:
+        name: compatibility-snapshot-${{ inputs.build_type }}-pg14
+        path: /tmp/compatibility_snapshot_pg14
+        prefix: latest
+
     - name: Run pytest
       env:
         NEON_BIN: /tmp/neon/bin
@@ -80,6 +87,8 @@ runs:
         BUILD_TYPE: ${{ inputs.build_type }}
         AWS_ACCESS_KEY_ID: ${{ inputs.real_s3_access_key_id }}
         AWS_SECRET_ACCESS_KEY: ${{ inputs.real_s3_secret_access_key }}
+        COMPATIBILITY_SNAPSHOT_DIR: /tmp/compatibility_snapshot_pg14
+        ALLOW_BREAKING_CHANGES: contains(github.event.pull_request.labels.*.name, 'breaking changes')
       shell: bash -euxo pipefail {0}
       run: |
         # PLATFORM will be embedded in the perf test report
@@ -153,6 +162,15 @@ runs:
           export REPORT_TO="$PLATFORM"
           scripts/generate_and_push_perf_report.sh
         fi
+
+    - name: Upload compatibility snapshot for Postgres 14
+      if: github.ref_name == 'release'
+      uses: ./.github/actions/upload
+      with:
+        name: compatibility-snapshot-${{ inputs.build_type }}-pg14
+        # The path includes a test name (test_prepare_snapshot) and directory that the test creates (compatibility_snapshot_pg14), keep the path in sync with the test
+        path: /tmp/test_output/test_prepare_snapshot/compatibility_snapshot_pg14/
+        prefix: latest
 
     - name: Create Allure report
       if: always()
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -844,7 +844,7 @@ jobs:
           submodules: true
           fetch-depth: 0
 
-      - name: Configure environment 
+      - name: Configure environment
         run: |
           helm repo add neondatabase https://neondatabase.github.io/helm-charts
           aws --region us-east-2 eks update-kubeconfig --name dev-us-east-2-beta --role-arn arn:aws:iam::369495373322:role/github-runner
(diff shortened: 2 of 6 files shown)


.github/actions/run-python-test-set/action.yml
```diff
@@ -73,13 +73,22 @@ runs:
       shell: bash -euxo pipefail {0}
       run: ./scripts/pysync
 
+    - name: Download compatibility snapshot for Postgres 14
+      uses: ./.github/actions/download
+      with:
+        name: compatibility-snapshot-${{ inputs.build_type }}-pg14
```
SomeoneToIgnore, line 79: We would add pg15 eventually, might be worth to extract it to a parameter with the default value already.
Seems that there are more similar places in the workflow yml, so might be a good separate task.

.github/workflows/build_and_test.yml
```diff
@@ -819,3 +819,24 @@ jobs:
...
+
+  promote-compatibility-test-snapshot:
+    runs-on: dev
+    container:
+      image: 369495373322.dkr.ecr.eu-central-1.amazonaws.com/rust:pinned
+      options: --init
+    needs: [ deploy, deploy-proxy ]
+    if: github.ref_name == 'release' && github.event_name != 'workflow_dispatch'
+    steps:
+      - name: Promote comaptibility snapshot for the release
```
SomeoneToIgnore, line 831:
```suggestion
      - name: Promote compatibility snapshot for the release
```

SomeoneToIgnore (commented):
This looks great.
I would allow somebody else to approve that, since my Python skills are not very nice.

A few ideas:
* we could store Safekeepers' WAL segments and do the restoration via them: they store segments on the remote storage and locally (might worth to switch to 1 SK to be sure we have all its data easily?) https://github.com/neondatabase/cloud/wiki/Recovery-from-WAL
* could we check (manually) how the SQL diff looks like? I'm a bit suspicious that it could get unreadable quite fast.

SomeoneToIgnore:
> Does it make sense?

Totally, as long as we ensure that the data restored is the same 🙂 
Thank you for implementing this!

SomeoneToIgnore (approved):

test_runner/regress/test_compatibility.py
```diff
@@ -0,0 +1,181 @@
...
+
+    shutil.copytree(compatibility_snapshot_dir / "repo", repo_dir)
+
+    # Remove old logs to avoid confusion in test artifacts
+    for logfile in repo_dir.glob("**/*.log"):
+        logfile.unlink()
+
+    # Update paths in configs
+    pageserver_config = (repo_dir / "pageserver.toml").read_text()
+    pageserver_config = pageserver_config.replace(
```
LizardWizzard, line 63: Hmm, which field contains this path? Cannot see it in typical test config. Maybe worth a comment or a toml parsing instead of `.replace`

test_runner/regress/test_compatibility.py
```diff
@@ -0,0 +1,181 @@
...
+        "/test_prepare_snapshot/", "/test_backward_compatibility/compatibility_snapshot/"
+    )
+    (repo_dir / "pageserver.toml").write_text(pageserver_config)
+
+    for postmaster_opts in repo_dir.glob("**/postmaster.opts"):
```
LizardWizzard, line 68: Tried to search where do we get this file from and it seems that it originates from computes in local pgdatadirs. Not sure if we need to keep this data around. Compute is stateless so should be fine to start a new one. WDYT?

test_runner/regress/test_compatibility.py
```diff
@@ -0,0 +1,181 @@
...
+        postmaster_opts_content = postmaster_opts.read_text()
+        postmaster_opts_content = postmaster_opts_content.replace(
+            "/test_prepare_snapshot/", "/test_backward_compatibility/compatibility_snapshot/"
+        )
+        postmaster_opts.write_text(postmaster_opts_content)
+
+    snapshot_config = toml.load(repo_dir / "config")
+
+    # NeonEnv stub to make NeonCli happy
```
LizardWizzard, line 77: Were there some roadblocks when you tried to use default env? In my understanding using env manually without calling to init should be fine. WDYT?

LizardWizzard: Also since real env is a context manager you dont have to manually add finalizers. Env will properly shutdown itself

LizardWizzard (commented):
Nice!

Added some comments

test_runner/regress/test_compatibility.py
bayandin: Do you mean `NeonEnvBuilder` / `neon_env_builder` fixture?
That's a nice idea, I've tried to use it, but `start` doesn't start without `init`, unfortunately:
https://github.com/neondatabase/neon/blob/f5e9d3010bfb6d0fc5cfa2aaba971b668ae9d69a/test_runner/fixtures/neon_fixtures.py#L556-L563
And cooking up env manually (or rather from a config) will require some refactoring. I'd like to do it, but probably in a separate PR and later.

LizardWizzard: At a quick glance doesnt look too wordy, but up to you

koivunej (approved):
Been reading this over a few times already, makes sense and looks good to me.

PR #2627, opened 2022-10-17
GRANT CREATE ON SCHEMA public TO web_access

lubennikovaav:


kelvich:
What will happen with 14 during this grant?

compute_tools/src/spec.rs
```diff
@@ -380,6 +380,10 @@ pub fn handle_grants(node: &ComputeNode, client: &mut Client) -> Result<()> {
         info!("grant query {}", &query);
 
         client.execute(query.as_str(), &[])?;
+
+        // Explicitly grant CREATE ON SCHEMA PUBLIC to the web_access user.
+        // This is needed since postgres 15, where this privilege is removed by default.
+        client.execute("GRANT CREATE ON SCHEMA public TO web_access", &[])?;
```
ololobus, line 386: I think, user may hit the same problem if they create a new non-db-owner role. So it won't be able to create objects in the public schema in non-owned databases.

The solution could be to grant the same to every role after db creation

lubennikovaav: This is considered postgres best-practice (and mentioned in almost all pg 15 announcements so far)...

It's ok  to grant access back to `web_access`, because in our model it acts like a database owner.
But I'd keep new default for other postgres users.

Database owners may still run `GRANT CREATE ON SCHEMA public TO new_user` to explicitly grant access to another user.  I've checked and it works now.

Any objections?

ololobus: > Database owners may still run GRANT CREATE ON SCHEMA public TO new_user to explicitly grant access to another user. I've checked and it works now.

That's nice, then it's probably ok to go with it. Did you test it with `neon` instance?

> Do we ever delete web_access role?

Nope, I was worrying about other non-db-owner roles users can create.

But this will be probably solved automatically by `reassign_owned_objects()`. We do `REASSIGN OWNED` to the database owner before deleting the role, then `DROP OWNED BY`, so it seems to be enough

ololobus (approved):

compute_tools/src/spec.rs
lubennikovaav: > That's nice, then it's probably ok to go with it. Did you test it with neon instance?

Yep. I tested it on my staging instance with psql connections from different users.

PR #2629, opened 2022-10-17
Improve pageserver IO metrics

bojanserafimov:
I'm trying to measure whether slow pgbench init runs write more data. This metric will help.

### DIFF
--- a/pageserver/src/metrics.rs
+++ b/pageserver/src/metrics.rs
@@ -425,6 +425,8 @@ impl Drop for TimelineMetrics {
         let _ = WAIT_LSN_TIME.remove_label_values(&[tenant_id, timeline_id]);
         let _ = CURRENT_PHYSICAL_SIZE.remove_label_values(&[tenant_id, timeline_id]);
         let _ = CURRENT_LOGICAL_SIZE.remove_label_values(&[tenant_id, timeline_id]);
+        let _ = NUM_PERSISTENT_FILES_CREATED.remove_label_values(&[tenant_id, timeline_id]);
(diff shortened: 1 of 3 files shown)


koivunej (approved):
PR changes two global counters to be timeline specific (tenant_id, timeline_id). I'm guessing the metrics are not needed for anything other, so this is looking good.

koivunej:
Reran failed because that was a familiar-looking `test_tenant_tasks` failure in regression tests.

pageserver/src/metrics.rs
```diff
@@ -386,8 +388,12 @@ impl TimelineMetrics {
...
             .get_metric_with_label_values(&[&tenant_id, &timeline_id])
             .unwrap();
-        let num_persistent_files_created = NUM_PERSISTENT_FILES_CREATED.clone();
-        let persistent_bytes_written = PERSISTENT_BYTES_WRITTEN.clone();
+        let num_persistent_files_created = NUM_PERSISTENT_FILES_CREATED
+            .get_metric_with_label_values(&[&tenant_id, &timeline_id])
+            .unwrap();
+        let persistent_bytes_written = PERSISTENT_BYTES_WRITTEN
+            .get_metric_with_label_values(&[&tenant_id, &timeline_id])
+            .unwrap();
```
koivunej, line 396: Took some re-checking but it seems that the two metrics had always been used as if they were tenant/timeline specific but were just cloned from the single global one previously, so nothing at the metric updating code needs to change :+1:

bojanserafimov:
> I'm guessing the metrics are not needed for anything other

Yes we only use them for pytest benchmarks right now, to measure write amplification

PR #2630, opened 2022-10-17
Remove stale todos

bojanserafimov:
The "global flag" solution suggested in the FIXME is already implemented.


lubennikovaav (approved):
CI failure is non-related flaky test.
https://github.com/neondatabase/neon/issues/2587

PR #2631, opened 2022-10-17
remove redundant expect_tenant_to_download_timeline

LizardWizzard:
Not sure what happened there, probably some merge conflict. 

There were two copies of `expect_tenant_to_download_timeline`, one added by @SomeoneToIgnore and the second one by @hlinnaka. From the usage I believe existing `assert_no_in_progress_downloads_for_tenant` should work the same way

Let me know if I've missed something


SomeoneToIgnore (approved):
Thank you.

koivunej (approved):

PR #2643, opened 2022-10-18
Deploy scram proxies into new account

SergeyMelnikov:


.github/workflows/build_and_test.yml
```diff
@@ -825,3 +825,34 @@ jobs:
...
+    # Compute image isn't strictly required for proxy deploy, but let's still wait for it to run all deploy jobs consistently.
+    needs: [ push-docker-hub, calculate-deploy-targets, tag, regress-tests ]
+    if: |
+      (github.ref_name == 'main' || github.ref_name == 'release') &&
+      github.event_name != 'workflow_dispatch'
+    defaults:
+      run:
+        shell: bash
+    env:
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_DEV }}
```
zoete, line 841: Does asible need the explicit AWS_ACCESS in env?

zoete (approved):

.github/workflows/build_and_test.yml
zoete: As container: 369495373322.dkr.ecr.eu-central-1.amazonaws.com/ansible:pinned will inherit the instance profile from the github runner it is running on

.github/workflows/build_and_test.yml
```diff
@@ -825,3 +825,31 @@ jobs:
...
           helm upgrade ${{ matrix.proxy_job }}-scram neondatabase/neon-proxy --namespace neon-proxy --install -f .github/helm-values/${{ matrix.proxy_config }}-scram.yaml --set image.tag=${DOCKER_TAG} --wait --timeout 15m0s
+
+  deploy-proxy-new:
+    runs-on: dev
+    container: 369495373322.dkr.ecr.eu-central-1.amazonaws.com/ansible:pinned
+    # Compute image isn't strictly required for proxy deploy, but let's still wait for it to run all deploy jobs consistently.
+    needs: [ push-docker-hub, calculate-deploy-targets, tag, regress-tests ]
+    if: |
+      (github.ref_name == 'main' || github.ref_name == 'release') &&
+      github.event_name != 'workflow_dispatch'
```
bayandin, line 836: @SergeyMelnikov for `deploy-new` job we have a bit different if-clause:
https://github.com/neondatabase/neon/blob/830dc928a880d72c7395428a23f5903a9308b1e0/.github/workflows/build_and_test.yml#L747-L755

I believe we want them synchronised, right?

PR #2652, opened 2022-10-19
Use postgres::config to properly escape database names

ololobus:
We've got a user in production that cannot create a database with a trailing space in the name.

This happens because we use `url` crate for manipulating the DATABASE_URL, but it follows a standard that doesn't fit really well with Postgres. For example, it trims all trailing spaces from the path:
  > Remove any leading and trailing C0 control or space from input.
  > https://url.spec.whatwg.org/#url-parsing

But we use path for database name and it's totally valid to have trailing spaces in the database name in Postgres.

Thus, bring back `urlencoding` crate to additionally pre-escape identifiers before using as a path in the DATABASE_URL.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -557,6 +557,7 @@ dependencies = [
  "tokio",
  "tokio-postgres",
  "url",
+ "urlencoding",
  "workspace_hack",
 ]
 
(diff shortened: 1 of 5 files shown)


kelvich (approved):

lubennikovaav (approved):

LizardWizzard (approved):
LGTM

PR #2654, opened 2022-10-19
Fix profiling check fixture

bojanserafimov:
We changed how we show whether profiling was enabled but didn't update this test. So we've been skipping the profiling test even when pageserver is compiled with the profiling feature.


SomeoneToIgnore (approved):
LGTM

Maybe let's pass https://github.com/neondatabase/neon/pull/2636 first and rebase on top?
I can do the work.

SomeoneToIgnore:
This PR conflicts, I've reapplied the code logic on top in https://github.com/neondatabase/neon/pull/2659 thank you for spotting it.

Issue #2616, opened 2022-10-13
EOF in the pageserver http response, when the wrong postgres version is passed


ololobus:
## Steps to reproduce

Try creating a timeline with some wrong version like `0`

## Expected result

Proper error and http code

## Actual result

Console just got EOF and in the pageserver logs there is a panic:
```log
2022-10-13T16:41:59.782867Z  INFO timeline_create{tenant=cd788604d9513179a47676e608450b8a new_timeline=Some(e895b76522ee876aad1362193d381912) lsn=None pg_version=Some(0)}: running initdb in /data/tenants/cd788604d9513179a47676e608450b8a/timelines/basebackup-e895b76522ee876aad1362193d381912.___temp... 
thread 'mgmt request worker' panicked at 'Unsupported postgres version: 0', pageserver/src/config.rs:391:18
```

In the console it looks like
```log
could not execute operation: could not create timeline: Post "http://172.30.42.11:9898/v1/tenant/cd788604d9513179a47676e608450b8a/timeline/": EOF
```

## Environment

PR now, later staging and other envs

## Logs, links
- https://github.com/neondatabase/cloud/pull/2469

ololobus:
cc @lubennikovaav

PR #2664, opened 2022-10-20
#2616 handle 'Unsupported pg_version' error properly

lubennikovaav:


hlinnaka (approved):

ololobus (commented):
Cool, that it won't panic now. I'm not sure, what will be the HTTP error code in the response? 500? The more correct way will be to respond with 400 or 422

Yet, probably it's too hard to achieve with the current http server implementation. And we definitely can live with 500 here as console will highly unlikely pass anything beside 14 or 15, so +1 for merging it now

libs/postgres_ffi/wal_craft/src/lib.rs
```diff
@@ -37,35 +37,35 @@ pub static REQUIRED_POSTGRES_CONFIG: Lazy<Vec<&'static str>> = Lazy::new(|| {
 });
 
 impl Conf {
-    pub fn pg_distrib_dir(&self) -> PathBuf {
+    pub fn pg_distrib_dir(&self) -> Result<PathBuf> {
```
SomeoneToIgnore, line 40: NIT: let's not shadow `Result` from the implicitly imported stdlib prelude and use `anyhow::Result` instead, for readability.
Don't change it now if you don't want to, but please consider it in the future.

SomeoneToIgnore (approved):
A follow-up idea to consider later:

we're kind of hardcoding pg version via numbers 14 and 15 in a few places in config.rs, so I wonder, how dynamic this version would be in the future?

If not much (as now), we could turn `pg_version: u32` into `pg_version: PgVersion` with 
```rust
pub enum PgVersion {
    V14,
    V15,
}
```

this way, the compiler would ensure there are no other variants and we can avoid adding extra `Result`s everywhere, even in such "late" code places as `walredo`.

Issue #1943, opened 2022-06-16
Docker Compose demo setup


moltar:
Would be great to have a Docker Compose made from the official Docker images (See #1942). Can also combine it with Minio for local S3 compatible storage. This would be a great opportunity to test workloads locally and for integration tests.

mikecaat:
Can we close the issue?

SomeoneToIgnore:
Closed in https://github.com/neondatabase/neon/pull/2666

PR #2666, opened 2022-10-21
Add a docker-compose example file (#1943)

mikecaat:
Hi, thanks for developing interesting database. 

As a beginner, a docker-compose example file is very useful because it took me a while to find the way to create a neon cluster on multiple servers.

The docker-compose example file is useful for local testing and beginner tutorials. It creates the following containers, and it supports taking backup pages and WAL to S3 compatible storage, MinIO.

  - etcd x 3
  - pageserver x 1
  - safekeeper x 3
  - compute x 1
  - MinIO x 1        # This is Amazon S3 compatible object storage

Please feel free to comments.


docs/docker.md
```diff
@@ -18,3 +18,63 @@ We build all images after a successful `release` tests run and push automaticall
...
+2022-10-21 06:15:48.757 GMT [56] LOG:  connection authorized: user=cloud_admin database=postgres application_name=psql
+2022-10-21 06:17:00.307 GMT [56] LOG:  [NEON_SMGR] libpagestore: connected to 'host=pageserver port=6400'
+(...omit...)
+```
+
+4. If you want to see durable data in MinIO which is s3 compatible storage
+
+Access http://localhost:9001 and sign in.
+
+- Username: `minio`
```
SomeoneToIgnore, line 77: Can we somehow explicitly set those in the compose file? 
If we plan to keep it as a documentation, users might find it simpler to play with the compose file right away, and creds there would help to discover it sooner.

It also seems that the whole new doc could be placed into the compose file, but I might be wrong :slightly_smiling_face:

docker-compose/compute/shell/compute.sh
```diff
@@ -0,0 +1,46 @@
...
+sed "s/TENANT_ID/${tenant_id}/" ${SPEC_FILE_ORG} > ${SPEC_FILE}
+sed -i "s/TIMELINE_ID/${timeline_id}/" ${SPEC_FILE}
+
+cat ${SPEC_FILE}
+
+echo "Start compute node"
+/usr/local/bin/compute_ctl --pgdata /var/db/postgres/compute \
+     -C "postgresql://cloud_admin@localhost:55433/postgres"  \
+     -b /usr/local/bin/postgres                              \
+     -S ${SPEC_FILE}
```
SomeoneToIgnore, line 46: We usually prefer to have newlines in the end of the files, as least seems that a few maintainers have this setting enabled.
If/When we merge this PR, let's have them added everywhere.

docker-compose/docker-compose.yml
```diff
@@ -0,0 +1,209 @@
...
+                                                        bucket_region='eu-north-1',
+                                                        prefix_in_bucket='/pageserver/'}\""
+    depends_on:
+      - etcd1
+      - etcd2
+      - etcd3
+      - miniocreatebuckets
+
+  safekeeper1:
+    image: neondatabase/neon:2221
```
SomeoneToIgnore, line 99: Those tags are all hardcoded, and not promised to be permanent.
I think, what is expected here is to have a `build` declaration with a few params to create Docker images from it

docker-compose/image/compute/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM neondatabase/compute-node:2221
```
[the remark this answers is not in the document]
mikecaat: Oh, thanks. I'll change to use `Dockerfile.compute-node-v15`

docs/docker.md
mikecaat: Does the following is not enough? Or is it better to replace environment variables?
```
  miniocreatebuckets:
    image: minio/mc
    environment:
      - MINIO_ROOT_USER=minio
      - MINIO_ROOT_PASSWORD=password
```

docker-compose/compute/shell/compute.sh
mikecaat: OK, I'll fix it.

docker-compose/docker-compose.yml
mikecaat: OK. I'll try it.

docker-compose/image/compute/Dockerfile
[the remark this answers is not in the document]
SomeoneToIgnore: Just to be explicit: all `image: neondatabase/neon:2221` in docker-compose has to be pointed to the corresponding Docker files and their builds configured, otherwise we're still relying on some hardcoded tag that's not guaranteed to be stored for long.

docs/docker.md
SomeoneToIgnore: Oh, apologies, I must have overlooked that part.

docker-compose/docker-compose.yml
```diff
@@ -0,0 +1,209 @@
+version: '3'
+
+services:
+  etcd1:
```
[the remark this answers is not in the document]
mikecaat: I used the above configuration.

docker-compose/image/compute/Dockerfile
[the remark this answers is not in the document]
mikecaat: OK, thanks. I understood. I changed to use `compute-node-v14`  as default image.
But, I found my docker-compose doesn't work with `compute-node-v15` . I'll check the reason.
```
2022-10-24 10:04:38.477 GMT [39] FATAL:  database files are incompatible with server
2022-10-24 10:04:38.477 GMT [39] DETAIL:  The data directory was initialized by PostgreSQL version 14, which is not compatible with this version 15.0.
```

docker-compose/compute/shell/compute.sh
mikecaat: I fixed.

docker-compose/docker-compose.yml
mikecaat: I used a environment variable instead to use `build` declaration. Is it enough now?

docker-compose/docker-compose.yml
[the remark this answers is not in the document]
SomeoneToIgnore: Great, let's update the docs (if you plan to leave them): https://github.com/neondatabase/neon/pull/2666/files#diff-ae5611658dcd443d7a30575bebcdf15a71b7a0f5abebb18b62d00aa0373cae74R26

docker-compose/docker-compose.yml
SomeoneToIgnore: Looks nice, might be good to mention TAG in the docs, however obvious that is for us.

SomeoneToIgnore (commented):
Thank you.

One last thing I'd love to see in any for is the automated verification of the whole set-up.

There's still quite an amount of hardcode used, which is ok for now per se, but we would need to detect when things change and this hardcode does not work anymore, so that someone can come and fix it.

No worries if it does not work in the CI, I can take care of it, but please let's make some kind of a script or a python test (similar to what we have in the project already) that works for you locally and can be reproduced by me (so that I could work on its CI runs later)

docker-compose/docker-compose.yml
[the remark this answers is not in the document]
mikecaat: fixed.

docker-compose/image/compute/Dockerfile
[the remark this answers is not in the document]
mikecaat: fixed to specify `pg_version` when to create new timeline.

docker-compose/docker-compose.yml
mikecaat: OK. I updated the doc.

mikecaat:
Thanks for your kind help.

I fixed your review comments and so on.

* Update docs
* Support proxy argument when to build docker image
* Support v15 of postgres version
* Add a container to check if the compute is ready or not in docker-compose.yml, and add a test script ( /scripts/docker-compose_test.sh ) which fails if the compute does not ready after 60s.

Please let me know if I forgot anything.
Regards,

SomeoneToIgnore (approved):
Thank you, I've tested the script and it seems to work.
I would have expected some SQL (similar to what you have in the docs) being run in that test script, do you think it's simple to add one in the tests?

Either way, I think this is a good increment for the first time and now we need to think on the whole set-up better, identify things to improve and work on that.
Also, the CI checks for this are vital to get into, that should be the next step.

I've started the CI for the PR and will leave it to hang by the end of the day if somebody else comes to comment it.
If nothing extraordinary happens, I'll merge the PR before the end of tomorrow.

Thank you again for bringing it up.

mikecaat:
> I would have expected some SQL (similar to what you have in the docs) being run in that test script, do you think it's simple to add one in the tests?

Yes, I agreed. I made a small patch to do so.
Please check if the script works in your environment.

> Either way, I think this is a good increment for the first time and now we need to think on the whole set-up better, identify things to improve and work on that.

Yes, I think so. As a beginner, it's useful document for users to create a neon cluster. 
Thanks a lot for your help!

PR #2676, opened 2022-10-21
[RFC] pageserver s3 coordination

LizardWizzard:
During discussion with @SomeoneToIgnore about toggling background activity PR (https://github.com/neondatabase/neon/pull/2665) decided to step back a little bit and explore alternatives. 

Approach with avoiding metadata entirely looks promising to me. 
Not really proposing to implement one of the approaches, rather wanted to brainstorm given possible options

[Rendered](https://github.com/neondatabase/neon/blob/dkr/pageserver-s3-coordination/docs/rfcs/020-pageserver-s3-coordination.md)


docs/rfcs/020-pageserver-s3-coordination.md
```diff
@@ -0,0 +1,75 @@
...
+
+One possible solution for relocation case is to orchestrate background jobs from outside. The oracle who runs migration can turn off background jobs on PS1 before migration and then run migration -> enable them on PS2. The problem comes if migration fails. In this case in order to resume background jobs oracle needs to guarantee that PS2 doesnt run background jobs and if it doesnt respond then PS1 is stuck unable to run compaction/gc. This cannot be solved without human ensuring that no upload from PS2 can happen. In order to be able to resolve this automatically CAS is required on S3 side so pageserver can avoid overwriting index part if it is no longer the leading one
+
+Note that flag that disables background jobs needs to be persistent, because otherwise pageserver restart will clean it
+
+### Avoid index_part.json
+
+Index part consists of two parts, list of layers and metadata. List of layers can be easily obtained by \`ListObjects\` S3 API method. But what to do with metadata? Create metadata instance for each checkpoint and add some counter to the file name?
+
+Back to potentially long s3 ls.
```
SomeoneToIgnore, line 61: In addition, this set-up would require us to rehash the approach to checksums and file sizes that we have now.

We plan to add more data into the index_part.json file : https://github.com/neondatabase/neon/issues/2456 and https://github.com/neondatabase/neon/issues/2683

While current it's possible to emulate both tasks via S3 commands, that requires more code to write and more issues to solve with hyper (to pass our own checksums to S3).

SomeoneToIgnore: https://docs.aws.amazon.com/cli/v1/userguide/cli-usage-pagination.html

> By default, the AWS CLI uses a page size determined by the individual service and retrieves all available items. For example, Amazon S3 has a default page size of 1000. If you run aws s3api list-objects on an Amazon S3 bucket that contains 3,500 objects, the AWS CLI automatically makes four calls to Amazon S3, handling the service-specific pagination logic for you in the background and returning all 3,500 objects in the final output.

S3 ls is capable of displaying sizes and checksums in S3 LS: https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html#API_ListObjectsV2_ResponseSyntax
but not capable of displaying extra blob metadata: that would require a HEAD request per blob or some other calls.

If we plant to use more data later (layer format version?) the S3 LS approach would again be required to invent some storage for this data that's easy to access.

docs/rfcs/020-pageserver-s3-coordination.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Coordination based approach
+
+Do it like safekeepers chose leader for WAL upload. Ping each other and decide based on some heuristics e g smallest node id. During relocation PS1 sends "resign" ping message so others can start election without waiting for a timeout.
+
+This still leaves metadata question open and non deterministic layers are a problem as well
```
SomeoneToIgnore, line 67: I like this approach more, since we do something similar with SKs already and SKs could benefit from the information pageservers post to etcd (shared dashboard).

In addition, it appears to me that etcd allows to solve the metadata sync problem:
> If one of the pageservers does not ingest wal (and just picks up layer updates) then it lags behind and cannot really answer queries in the same pace as the primary one

could be easily solved if the timeline leader pushes index_part.json contents as a ping message: this way every other PS that's interested in the timeline but not a leader, can subscribe and get the fresh data.

On leader reelection, we can reread the index file and presumably also keep the most up-to-date info about the remote layers on the node side that writes to S3.

docs/rfcs/020-pageserver-s3-coordination.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Avoid metadata file
+
+One way to eliminate metadata file is to store it in layer files under some special key. This may resonate with intention to keep all relation sizes in some special segment to avoid initial download during size calculation. Maybe with that we can even store pre calculated value.
+
+As a downside each checkpoint gets 512 bytes larger.
+
+If we entirely avoid metadata file this opens up many approaches
```
SomeoneToIgnore, line 75: I'm not sure how to do that properly. We need `disk_consistent_lsn` and `latest_gc_cutoff_lsn` to be updated periodically on the local and remote side.
All other properties are more or less static and less important, except for "metadata version" field.


While the `disk_consistent_lsn` is possible to embed into the layer name, presumably, since the layer files are being flushed only once and cause `disk_consistent_lsn` to be increased due to that.
But `latest_gc_cutoff_lsn` changes unpredictably and does not produce new layers: instead, it removes the old ones and demands that the `latest_gc_cutoff_lsn` for the new layer set is persisted (if we use the older, lesser value on the new layer set there might be issues).
How to store that without `metadata`?

More complex issues are related to "metadata version" since certain PS versions won't be compatible with certain layer formats and we need a good way to detect that before downloading layers to disk and checking their bytes.
Might be better to work on layer format more before seriously discussing this.

Generally, the biggest questions is how to access this data with the new approach? If we start to store layer size inside of each layer and we will have, say, 1000 new layers on a remote storage, what would be the good way to collect the size information from every layer without downloading them?

SomeoneToIgnore (commented):
I lean towards the general model, where PS nodes are able to read "small" (orders of magnitude smaller than the layer size) metadata about all available layers and decide, what to do with them.

For that, the broker approach seems to be the most promising to me among what's proposed.
I'd try doing that since we already have etcd and leader election implemented and tested on SKs.

docs/rfcs/020-pageserver-s3-coordination.md
LizardWizzard: > could be easily solved if the timeline leader pushes index_part.json contents as a ping message

I think this is not the case. When layer is generated is it already too late. By default layer is generated after each 256 megabytes of WAL. If we wait for timeout it is even worse.

docs/rfcs/020-pageserver-s3-coordination.md
LizardWizzard: > Generally, the biggest questions is how to access this data with the new approach? If we start to store layer size inside of each layer and we will have, say, 1000 new layers on a remote storage, what would be the good way to collect the size information from every layer without downloading them?

This is the same problem as with relation sizes that we need to collect on startup to populate logical size. If we go with s3 ls instead of index part we can pick layer created by latest checkpoint and extract this data from it

> How to store that without metadata?

As a kv pair in layer created on each checkpoint for example

> and we need a good way to detect that before downloading layers to disk and checking their bytes.

We can have a header in each layer file with such metadata (like it is for usual metadata) and start download sort of speculatively and abort it if we detected something unexpected.

LizardWizzard:
> For that, the broker approach seems to be the most promising to me among what's proposed.

With my current understanding broker approach cannot solve all mentioned problems. It can coordinate certain actions between pageservers but it cannot solve mutable metadata problem

docs/rfcs/020-pageserver-s3-coordination.md
SomeoneToIgnore: Ah, so the issue is that we don't share the non-flushed data with the other pageservers that serve the same get_page requests for the same timeline?

Similar to SK nodes, we could coordinate the connection strings in our blackboard messages and proxy the get_page requests to other nodes, if they have the Lsn the current node does not have yet.

koivunej: > and proxy the get_page requests to other nodes, if they have the Lsn the current node does not have yet.

This makes sense.

In general we would then have the problem of not having those newer WAL segments in case the leader disappears, but we already have the safekeepers which would be able to assist with the recovery from the latest flushed layer file onwards.

LizardWizzard: > Ah, so the issue is that we don't share the non-flushed data with the other pageservers that serve the same get_page requests for the same timeline?

Correct

> proxy the get_page requests to other nodes

I see two downsides there. 

1. Failover process
2. Uneven load distribution 

Let me go into details for each point:
1. If primary dies than replica wont be able to forward such requests. In order to continue operations it needs to detect that primary has failed (2 ping timeouts or something). Then it needs to start replication, and then finally ingest WAL needed to answer `get_page` request. If both pageservers ingest WAL most probably they do that in the same pace, so  in this case there are no actions needed for failover from pageserver side. Compute can detect failed pageserver and route queries to the healthy one. Or as aurora does request pages from both and pick the first reply
2. With this forwarding primary will be always responsible for both, ingesting wal (writes) an for reads of the recent data. I think may not help to scale reads unless significant portion of queries goes into old data (this may be the case for read replica postgres), because replica wont have data for tip of the timeline so for such queries it will only add forwarding latency without any performance boost

docs/rfcs/020-pageserver-s3-coordination.md
SomeoneToIgnore: Oh, that looks very much similar to the homegrown compression approach with blob headers and their streaming downloads we've made, yet now it's per layer, not per timeline.

If it's indeed possible to stream a small header part of the layer, to get that kv data, we're automatically getting its S3 metadata (sha, file size, custom metadata strings that we set additionally).

-------------
I wonder, would throttling be an issue or not in such case?

https://aws.amazon.com/premiumsupport/knowledge-center/s3-request-limit-avoid-throttling/ says that 

> You can send [3,500 PUT/COPY/POST/DELETE or 5,500 GET/HEAD requests](https://docs.aws.amazon.com/AmazonS3/latest/dev/optimizing-performance.html) per second per prefix in an Amazon S3 bucket. There are no limits to the number of prefixes that you can have in your bucket.

I have failed to find the numbers for AWS IAM we're using to avoid storing actual credentials, but that was hundreds requests per second, before it started failing.

I've checked ps-2 staging and there are tenants with timelines that have 178 layers, with `mean` ~13 layers /timeline, but I'm afraid that's due to mostly inactive projects.

I like the integrity of the approach (if we're truly able to stream the layer header fast), but concerned about AWS limitations we might hit.

docs/rfcs/020-pageserver-s3-coordination.md
SomeoneToIgnore: For 1., I would imagine that we keep all nodes as up-to-date as possible for the timeline, potentially evicting some _old_ (LRU?) layers based on some heuristics. 

Hence,

> so in this case there are no actions needed for failover from pageserver side

imagined that would be mostly the case.

koivunej (commented):
Just writing down "have read" review. Good writeup of the relocation and (hot?) standby issues. I have still so many questionmarks around `latest_gc_cutoff_lsn` and didn't even realize the wal ingestion wasn't deterministic so I cannot really comment more now.

docs/rfcs/020-pageserver-s3-coordination.md
LizardWizzard: 1. Didnt understand your point. With this "ping on upload" we'll have 256+ mb lag. Eviction especially for old layers shouldnt interfere with this. So the problem persists.
2. If layer generation is deterministic then there is no need to perform s3 ls other than during startup. Replica will ingest wal, create local layers and on eviction just drop them so they're still in layer map as remote ones

SomeoneToIgnore: 1. Oh, no, no "ping on upload" but rather ping every N (1) seconds as SK do, just update the data between pings, if needed.
So, we would have a more or less up-to-date vision on PS "cluster" state around the timeline and who's the leader.

LizardWizzard: 1. Do not understand how it can solve lag problem. You mean replicating timeline data through broker?

But otherwise yes. SK-like model is applicable to sync cluster state

docs/rfcs/020-pageserver-s3-coordination.md
LizardWizzard: > I wonder, would throttling be an issue or not in such case?

Shouldnt be much different from current usage pattern. If we can determine latest layer from `s3 ls` output then we download only one file with all the metadata

I think throttling will anyway become a problem with ondemand download/eviction being in place

docs/rfcs/020-pageserver-s3-coordination.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Manual orchestration
+
+One possible solution for relocation case is to orchestrate background jobs from outside. The oracle who runs migration can turn off background jobs on PS1 before migration and then run migration -> enable them on PS2. The problem comes if migration fails. In this case in order to resume background jobs oracle needs to guarantee that PS2 doesnt run background jobs and if it doesnt respond then PS1 is stuck unable to run compaction/gc. This cannot be solved without human ensuring that no upload from PS2 can happen. In order to be able to resolve this automatically CAS is required on S3 side so pageserver can avoid overwriting index part if it is no longer the leading one
+
+Note that flag that disables background jobs needs to be persistent, because otherwise pageserver restart will clean it
+
+### Avoid index_part.json
+
+Index part consists of two parts, list of layers and metadata. List of layers can be easily obtained by \`ListObjects\` S3 API method. But what to do with metadata? Create metadata instance for each checkpoint and add some counter to the file name?
```
tychoish, line 59: Could index_part.json be not mutable and split across a number of files, therefore operating something like a cache rather than a signleton?

docs/rfcs/020-pageserver-s3-coordination.md
```diff
@@ -0,0 +1,75 @@
...
+
+Back to potentially long s3 ls.
+
+### Coordination based approach
+
+Do it like safekeepers chose leader for WAL upload. Ping each other and decide based on some heuristics e g smallest node id. During relocation PS1 sends "resign" ping message so others can start election without waiting for a timeout.
```
tychoish, line 65: The overhead of having a consensus process per-tennat seems quite complex and could put an unnecessary practical limit on the number of nodes that could be in a set.

Inverting the liveness check (e.g. resign) means that nodes that fail can never resign (except via operator intervention,) which I think is a potential risk.

I'd also like to see some target expectations about the latency targets of doing a fail-over, relative to a cold-restart of a page server.

docs/rfcs/020-pageserver-s3-coordination.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Observations
+
+- If both pageservers ingest wal then their layer set diverges, because layer file generation is not deterministic
+- If one of the pageservers does not ingest wal (and just picks up layer updates) then it lags behind and cannot really answer queries in the same pace as the primary one
+- Can compaction help make layers deterministic? E g we do not upload level zero layers and construction of higher levels should be deterministic. This way we can guarantee that layer creation by timeout wont mess things up. This way one pageserver uploads data and second one can just ingest it. But we still need some form of election
+
+## Solutions
+
+### Manual orchestration
```
tychoish, line 51: Is there a way to view this not as "manual" orchestration but as "external orchestration," which is to say that the pageservers not be responsible for maintaining the their own tenant ownership (say) and instead giving that responsibility to something outside, which could simplify things, (particularly in combination with one of the solutions below that removes the (potential) for shared metadata state.

docs/rfcs/020-pageserver-s3-coordination.md
```diff
@@ -0,0 +1,75 @@
...
+
+Do it like safekeepers chose leader for WAL upload. Ping each other and decide based on some heuristics e g smallest node id. During relocation PS1 sends "resign" ping message so others can start election without waiting for a timeout.
+
+This still leaves metadata question open and non deterministic layers are a problem as well
+
+### Avoid metadata file
+
+One way to eliminate metadata file is to store it in layer files under some special key. This may resonate with intention to keep all relation sizes in some special segment to avoid initial download during size calculation. Maybe with that we can even store pre calculated value.
+
+As a downside each checkpoint gets 512 bytes larger.
```
tychoish, line 73: I might lack context on this but this doesn't seem like a large downside, how large of an impact is this?

docs/rfcs/020-pageserver-s3-coordination.md
```diff
@@ -0,0 +1,75 @@
...
+    note over S3: Index now L2, L3
+    PS2->>S3: Uploads new layer L4 <br/> (added to previous view of the index)
+    note over S3: Index now L1, L2, L4
+```
+
+At this point it is not possible to restore from index, it contains L2 which is no longer available in s3 and doesnt contain L3 added by compaction by the first pageserver. So if any of the pageservers restart initial sync will fail (or in on-demand world it will fail a bit later during page request from missing layer)
+
+### Standby pageserver
+
+Another related case is standby pageserver. In this case second pageserver can be used as a replica to scale reads and serve as a failover target in case first one fails.
```
tychoish, line 39: This seems like a cool feature, however, my question is generally:
- are pageservers read-bound, and if so what amount of read scaling are we expecting to get out of this. I'd be interested in weighing the cost of multiplying the cost of pageservers against the relative benefit.
- do the read and write workloads interact or impact each other?

You could imagine alternate solutions whereby you have more smaller pageservers with fewer numbers of tenants that you then scale up and down (vertically) as needed to accommodate load.

docs/rfcs/020-pageserver-s3-coordination.md
LizardWizzard: What do you mean by splitting? Split contents of a single file or create a separate file for each new version? The latter is indeed an option, but it gives up on original index file purpose to speed up listing of all files in a prefix

docs/rfcs/020-pageserver-s3-coordination.md
LizardWizzard: Yeah, thats how migration works currently. And the oracle who knows thigh level picture and is allowed to change things is control plane.

And thats a possible option to give the steering wheel to control plane even more. If we can avoid possible corruption when state is overwritten by concurrent writer then such external orchestration is a good way forward. 

From the other side in this case control plane needs to monitor pageservers and perform a failover if needed

docs/rfcs/020-pageserver-s3-coordination.md
LizardWizzard: Yep, it doesnt. Probably shoul've put it a little bit softer. As long as metadata is small there is no problem in that

docs/rfcs/020-pageserver-s3-coordination.md
LizardWizzard: Yeah, that needs a lot of experiments. I think this is rather out of scope for the RFC. It is more about how to allow multiple pageservers look into the same s3 prefix without corrupting things underneath.

> are pageservers read-bound

In read heavy workloads I think they are. This way reads can scale horizontally, so theoretically two pageservers can give twice the throughput. But then comes work from the compute side, how to properly schedule requests, ongoing pre-fetching work, etc etc. So thats hard to guess what will be the real scaling factor

> do the read and write workloads interact or impact each other

Since pageserver is multitenant there can be some skewed picture overall if one tenant's workload is read heavy, and for the other one it is write heavy

> smaller pageservers with fewer numbers of tenants

Especially interesting to consider edge deployments, where you have one writer and many readers across the globe. For this case having even small pageservers but close to the client looks like a good way to cut latency

docs/rfcs/020-pageserver-s3-coordination.md
tychoish: This makes a lot of sense to me. 

I was thinking about splitting the contents of a single file in some way, but I definitely didn't understand the purpose of this file when I made this content. 

I think it might be interesting to think about other ways of manging this data (splitting data up, versioning each file, caching the list of files in s3 in a database/etc.) as a way to make this problem easier to solve.

arssher:
So I there are two loosely related problems -- coordination of pageserver work with s3 and remote index maintenance. The latter is only needed to deal with slow s3 listings + update remote_consistent_lsn and latest_gc_cutoff_lsn.

Imagine for now s3 listings are fast -- we don't have index files. How should work with s3 look like, taking into account multiple pageservers?

- Obviously we should strive to have only one pageserver should upload to (and clean) s3, otherwise it is redundant work.
- However, s3 doesn't provide CAS API, which means we can never reliably linearize s3 operations -- s3 side can't tell an old uploading pageserver request from a newer uploader request -- they can interleave. Well, we could rely on physical time to ensure uploaders don't intersect, but that's fragile. External coordination can't improve on this. Consequently, _any approach writing different contents to the same key (like current index file) is broken_ (unless we operate manually).
- Fortunately, generally there is nothing bad with concurrent uploaders -- we might get intersection of data in layers, but redo is (or should be) able to cope with that; and data won't be corrupted.

That said, I think the following would work:
- Clearly we need some way to determine the uploading pageserver, but without strict requirements on non-intersecting with previous uploader. Something similar to how safekeepers approach this would do -- just deterministically calculate based on data exchanged through the broker. Just reminding that it might sound surprising, but IMV specialized API which etcd provides here doesn't really save the efforts -- it doesn't help to distinguish old & new leaders, and without it (generational numbers) leader election per se is trivial.
- This uploading pageserver lists remote contents and then maintains it -- uploads new layers, removes compacted and gc'ed ones. There might be some concurrent activity, but it won't harm. If pageserver maintains local cache of s3 contents, it might get out of date: 1) layer fetch might fail because it was compacted by a competing pageserver -- then it should relist s3 contents to find newer version. 2) also gc should sometimes delete layers uploaded by others (again relist required), but this can be very rare.
- Other pageservers, if they accept stream of WAL, process it, compact and gc locally (creating another independent set of layers), but don't upload. If they evict unused layers and thus want to be able to fetch them back later, they need to relist s3 contents. We might at some point broadcast s3 operations through the broker to speed up the process of remote index syncing, but not sure.
- Current index file also stores remote_consistent_lsn and latest_gc_cutoff_lsn. These can be written to s3 under (unique) uploading pageserver id; read would collect max() of all such values.


Getting back to slow s3 listing, the following approaches are possible:
1. Actually do nothing, i.e. drop index files. It would be nice to check/write down somewhere how slow s3 listing actually is (and how many layers a db might have), perhaps now it is not really required. metadata LSNs can be put under different prefixes, as said above.
2. Have a separate storage for the s3 index -- this seems to be a popular approach; PG would do, e.g. control plane database. So, on each layer creation, it is first uploaded and them inserted into the db. On each deletion, it is first deleted from the db and then from s3. Works fine, but the obvious cons is one more involved entity.
3. We can try to stay with current index file, but spread it to avoid overwrites (like in rfc's diagram example): similar to LSNs, each pageserver writes index under his id prefix. Creations are obvious; if pageserver A deletes (compacts) layer created by pageserver B, it writes 'deleted' op for it. To list files we need to load all indexes and union them, layers with 'delete' ops are removed. Such entries can be deleted themselves once they go behind gc horizon.

LizardWizzard:
> The latter is only needed to deal with slow s3 listings + update remote_consistent_lsn and latest_gc_cutoff_lsn.

To be more precise. Currently index file contains snapshot of `TimelineMetadata` file at the moment of upload.

Its structure is the following:

```rust
struct TimelineMetadataBodyV2 {
    disk_consistent_lsn: Lsn,
    // This is only set if we know it. We track it in memory when the page
    // server is running, but we only track the value corresponding to
    // 'last_record_lsn', not 'disk_consistent_lsn' which can lag behind by a
    // lot. We only store it in the metadata file when we flush *all* the
    // in-memory data so that 'last_record_lsn' is the same as
    // 'disk_consistent_lsn'.  That's OK, because after page server restart, as
    // soon as we reprocess at least one record, we will have a valid
    // 'prev_record_lsn' value in memory again. This is only really needed when
    // doing a clean shutdown, so that there is no more WAL beyond
    // 'disk_consistent_lsn'
    prev_record_lsn: Option<Lsn>,
    ancestor_timeline: Option<TimelineId>,
    ancestor_lsn: Lsn,
    latest_gc_cutoff_lsn: Lsn,
    initdb_lsn: Lsn,
    pg_version: u32,
}
```

`remote_consistent_lsn` is the `disk_consistent_lsn` from this metadata snapshot.

> Consequently, any approach writing different contents to the same key (like current index file) is broken (unless we operate manually).

Agreed. Even with manual orchestration it is not so obvious. E g on startup pageserver needs to ask control plane whether its still the leader (at least it is how I see it, maybe there is a workaround, check the updated rfc for the note + diagram)

>  1) layer fetch might fail because it was compacted by a competing pageserver -- then it should relist s3 contents to find newer version. 2) also gc should sometimes delete layers uploaded by others (again relist required), but this can be very rare.

Depending on local cache size (basically the eviction frequency), compaction can cause frequent deletions which will cause higher latency for on demand downloads on the follower since it needs to relist things frequently and do that on the `get_page` request path.

I think coordination through broker is the way.

> Have a separate storage for the s3 index -- this seems to be a popular approach; PG would do, e.g. control plane database.

That creates chicken or the egg problem which would be nice to avoid. Options remind me about ClickHouse way which used to store info about partitions in ZooKeeper so everyone knows what exists where

> We can try to stay with current index file, but spread it to avoid overwrites

I like this idea because it will allow to continue using index file to store file sizes and checksums for layers

***

I updated the RFC, please let me know if I missed something

SomeoneToIgnore (approved):
Feels like everybody is fine with the approach selected, it's documented and we can merge it?

LizardWizzard:
Merging since no objections were raised. Next step is working on the implementation. Progress is tracked here: https://github.com/neondatabase/neon/issues/2739

PR #2679, opened 2022-10-22
Fix division by zero in determine_offloader

petuhovskiy:
I found a panic in the staging logs:
```
thread 'wal backup launcher thread' panicked at 'attempt to calculate the remainder with a divisor of zero', safekeeper/src/wal_backup.rs:112:18
```

It points to this code: `(u128::from(ttid.timeline_id) % caughtup_peers.len() as u128)`

I'm not sure how this can happen and how to write a test for it yet, but here is a fix.

PR #2680, opened 2022-10-22
Fix division by zero panic in determine_offloader.

arssher:


petuhovskiy (approved):
Should fix the issue, but I'd still add a check for zero before the division, to eliminate any possibility of a panic, like in https://github.com/neondatabase/neon/pull/2679

arssher:
Well, close reading of these 5-10 nearby lines should say this is impossible. I mean, we could add a check, but then if it fails it would be reasonable to do assert/panic to attract attention.

petuhovskiy:
> Well, close reading of these 5-10 nearby lines should say this is impossible.

Yes, but complete verification is a bit harder than reading 5-10 nearby lines. To verify that panic is impossible, you should also look into `Lsn` type definition, whether it's signed or not, check if `Lsn(0) <= anything` is always true, look into what `checked_sub` does, and check that actual types of all used variables (like `conf.max_offloader_lag_bytes` and `p.commit_lsn`) match your expectations.

Actually, I'm more worried about future changes (for example, additional logic to filtering) that can cause unexpected behavior in the future. I'd rely less on nearby code and add more checks, because it's easy to overlook a change in a single line at a 5-10 line distance.

> I mean, we could add a check, but then if it fails it would be reasonable to do assert/panic to attract attention.

Attracting attention is good, but assert/panic can cause a lot of damage (which is much worse than the benefits of attracting attention). For example, in the current issue, we had a panic that:
1. failed the whole `wal backup launcher thread` (backup is not working for all timelines after that)
2. created an entry about the panic in the logs (but it didn't attract attention)
3. `wal_backup_launcher_rx` was dropped and `_tx` became closed (and we have code relying on `wal_backup_launcher_tx` is never closed)
4. computes couldn't connect to safekeeper, because we send notifications to `wal_backup_launcher_tx` when compute is connecting
5. projects didn't work and it was not obvious where the bug is

What we should do IMO:
- prefer errors to panics (1)
- try to localize the errors, so that only a single job for a single timeline fails, without affecting other timelines (1)
- create an alert for new panics in the logs (2)
- use something to attract attention in unexpected scenarios (to start, we can just write more logs about unexpected cases) (2)
- monitor failures and restart background threads (1)
- don't rely on non-essential dependencies (e.g. keep working even if backup channel is closed) (3,4)

arssher:
True.

> create an alert for new panics in the logs
@yeputons was going to add it.

> e.g. keep working even if backup channel is closed
that's easy actually, can do that

Issue #2585, opened 2022-10-07
Duplicate requests from prefetcher


bojanserafimov:
## Steps to reproduce
Run the init part of our pgbench test. Look at the pagestream read request pattern

## Problem
After visualizing the read request trace at the pageserver, I saw that the init workload consists of 4 sequential scans. One of those scans makes two almost consecutive (18 requests apart in the trace) requests for each page. So half of the requests in this scan are unnecessary. In other words, 20% of all requests during the init stage are unnecessary.

More precisely, the request pattern looks like this:
```
...
1663/13134/16396 8980 0/83C0B10 <- notice this query for 8980
1663/13134/16396 8972 0/83C0B10
1663/13134/16396 8981 0/83C0B10
1663/13134/16396 8973 0/83C0B10
1663/13134/16396 8982 0/83C0B10
1663/13134/16396 8974 0/83C0B10
1663/13134/16396 8983 0/83C0B10
1663/13134/16396 8975 0/83C0B10
1663/13134/16396 8984 0/83C0B10
1663/13134/16396 8976 0/83C0B10
1663/13134/16396 8985 0/83C0B10
1663/13134/16396 8977 0/83C0B10
1663/13134/16396 8986 0/83C0B10
1663/13134/16396 8978 0/83C0B10
1663/13134/16396 8987 0/83C0B10
1663/13134/16396 8979 0/83C0B10
1663/13134/16396 8988 0/83C0B10
1663/13134/16396 8980 0/83C0B10 <- here it is again
...
```

I'm not sure exactly what the problem is, but disabling prefetching by adding `false &&` to [this](https://github.com/neondatabase/neon/blob/ps-trace/pgxn/neon/pagestore_smgr.c#L1150) condition removes duplicate requests and decreases total traffic by 20%. My best guess is that we prefetch the correct pages, but instead of using the prefetched result we query again.

Off topic: Also maybe it's a problem that this code runs at all when prefetching is disabled (I didn't do anything to enable it).

## Environment
Local

relevant: https://github.com/neondatabase/neon/issues/2544

bojanserafimov:
Can't reproduce it now. Seems fixed.

PR #2687, opened 2022-10-24
Update prefetch mechanism:

MMeent:
Prefetch requests and responses are stored in a ringbuffer instead of a queue, which means we can utilize prefetches of many relations concurrently -- page reads of un-prefetched relations now don't imply dropping prefetches.

In a future iteration, this will detect sequential scans based on the read behavior of sequential scans, and will dynamically prefetch buffers for such relations as needed.

See also: #2544

### DIFF
--- a/pgxn/neon/pagestore_smgr.c
+++ b/pgxn/neon/pagestore_smgr.c
@@ -115,15 +115,31 @@ static UnloggedBuildPhase unlogged_build_phase = UNLOGGED_BUILD_NOT_IN_PROGRESS;
  * Prefetch implementation:
  * 
  * Prefetch is performed locally by each backend.
- * 
- * 
- * There can be up to MAX_PREFETCH_REQUESTS registered using smgr_prefetch
- * before smgr_read. All this requests are appended to primary smgr_read request.
- * It is assumed that pages will be requested in prefetch order.
- * Reading of prefetch responses is delayed until them are actually needed (smgr_read).
- * It make it possible to parallelize processing and receiving of prefetched pages.
- * In case of prefetch miss or any other SMGR request other than smgr_read,
- * all prefetch responses has to be consumed.
+ *
+ * There can be up to READ_BUFFER_SIZE active IO requests registered at any
+ * time. Requests using smgr_prefetch are sent to the pageserver, but we don't
+ * wait on the response. Requests using smgr_read are either read from the
+ * buffer, or (if that's not possible) we wait on the response to arrive -
+ * this also will allow us to receive other prefetched pages. 
+ * Each request is immediately written to the output buffer of the pageserver
+ * connection, but may not be flushed if smgr_prefetch is used: pageserver
+ * flushes sent requests on manual flush, or every neon.flush_output_after
+ * unflushed requests; which is not necessarily always and all the time.
+ *
+ * Once we have received a response, this value will be stored in the response
+ * buffer, indexed in a hash table. This allows us to retain our buffered
+ * prefetch responses even when we have cache misses.
+ *
+ * Reading of prefetch responses is delayed until them are actually needed
+ * (smgr_read). In case of prefetch miss or any other SMGR request other than
+ * smgr_read, all prefetch responses in the pipeline will need to be read from
+ * the connection; the responses are stored for later use.
+ *
+ * NOTE: The current implementation of the prefetch system implements a ring
+ * buffer of up to READ_BUFFER_SIZE requests. If there are more _read and
+ * _prefetch requests between the initial _prefetch and the _read of a buffer,
+ * the prefetch request will have been dropped from this prefetch buffer, and
+ * your prefetch was wasted.
  */
 
 /* Max amount of tracked buffer reads */
@@ -134,8 +150,8 @@ typedef enum PrefetchStatus {
 	PRFS_REQUESTED,		/* request was written to the sendbuffer to PS, but not
 						 * necessarily flushed.
 						 * all fields except response valid */
-	PRFS_RECEIVED,		/* all fields valid, response contains data */
-	PRFS_TAG_REMAINS,	/* only buftag, *OfRel are still valid */
+	PRFS_RECEIVED,		/* all fields valid */
+	PRFS_TAG_REMAINS,	/* only buftag and my_ring_index are still valid */
 } PrefetchStatus;
 
 typedef struct PrefetchRequest {
@@ -205,23 +221,12 @@ typedef struct PrefetchState {
 	PrefetchRequest prf_buffer[READ_BUFFER_SIZE]; /* prefetch buffers */
 } PrefetchState;
 
-/*
- * Prefetch implementation:
- * Prefetch is performed locally by each backend.
- * There can be up to MAX_PREFETCH_REQUESTS registered using smgr_prefetch
- * before smgr_read. All this requests are appended to primary smgr_read request.
- * It is assumed that pages will be requested in prefetch order.
- * Reading of prefetch responses is delayed until them are actually needed (smgr_read).
- * It make it possible to parallelize processing and receiving of prefetched pages.
- * In case of prefetch miss or any other SMGR request other than smgr_read,
- * all prefetch responses has to be consumed.
- */
-
 PrefetchState *MyPState;
 
-int			n_prefetch_hits;
-int			n_prefetch_misses;
-int			n_prefetch_dupes;
+int			n_prefetch_hits = 0;
+int			n_prefetch_misses = 0;
+int			n_prefetch_missed_caches = 0;
+int			n_prefetch_dupes = 0;
 
 XLogRecPtr	prefetch_lsn = 0;
 
@@ -235,9 +240,6 @@ static inline void prefetch_set_unused(uint64 ring_index, bool hash_cleanup);
 
 static XLogRecPtr neon_get_request_lsn(bool *latest, RelFileNode rnode,
 									   ForkNumber forknum, BlockNumber blkno);
-
-ps_disconnect_handle inner_ps_disconnect_hook = NULL;
-void prefetch_on_ps_disconnect(void);
 
 
 /*
@@ -269,7 +271,7 @@ prefetch_cleanup(void)
 	}
 }
 
-/* 
+/*
  * Wait for slot of ring_index to have received its response.
  * The caller is responsible for making sure the request buffer is flushed.
  */
@@ -330,9 +332,6 @@ prefetch_read(PrefetchRequest *slot)
 void
 prefetch_on_ps_disconnect(void)
 {
-	if (inner_ps_disconnect_hook != NULL)
-		inner_ps_disconnect_hook();
-
 	for (; MyPState->ring_receive < MyPState->ring_unused; MyPState->ring_receive++)
 	{
 		PrefetchRequest *slot;
@@ -430,11 +429,11 @@ prefetch_do_request(PrefetchRequest *slot, bool *force_latest, XLogRecPtr *force
 		 * have been in cache and evicted between those LSN values, which
 		 * then would have had to result in a larger request LSN for this page.
 		 * 
-		 * It is possible that a concurrent backend loads the page, modifies it
-		 * and then evicts it again, but the LSN of that eviction cannot be
-		 * smaller than the current redo pointer, which is already larger
-		 * than this prefetch_lsn. So in any case, that would invalidate this
-		 * cache.
+		 * It is possible that a concurrent backend loads the page, modifies
+		 * it and then evicts it again, but the LSN of that eviction cannot be
+		 * smaller than the current WAL insert/redo pointer, which is already
+		 * larger than this prefetch_lsn. So in any case, that would
+		 * invalidate this cache.
 		 * 
 		 * The best LSN to use for effective_request_lsn would be
 		 * XLogCtl->Insert.RedoRecPtr, but that's expensive to access.
@@ -450,6 +449,7 @@ prefetch_do_request(PrefetchRequest *slot, bool *force_latest, XLogRecPtr *force
 
 	/* update prefetch state */
 	MyPState->n_requests_inflight += 1;
+	MyPState->n_unused -= 1;
 	MyPState->ring_unused += 1;
 
 	/* update slot state */
@@ -517,8 +517,8 @@ prefetch_register_buffer(BufferTag tag, bool *force_latest, XLogRecPtr *force_ls
 	}
 
 	/*
-	 * If we prefetch queue is full, we need to make room by clearing the
-	 * oldest slot. If the oldest slot  holds a buffer that was already
+	 * If the prefetch queue is full, we need to make room by clearing the
+	 * oldest slot. If the oldest slot holds a buffer that was already
 	 * received, we can just throw it away; we fetched the page unnecessarily
 	 * in that case. If the oldest slot holds a request that we haven't
 	 * received a response for yet, we have to wait for the response to that
@@ -553,11 +553,8 @@ prefetch_register_buffer(BufferTag tag, bool *force_latest, XLogRecPtr *force_ls
 	}
 
 	/*
-	 * We could not use the requested buffer (if we had any), so any current
-	 * buffer has been released. We now have to aquire a new request buffer.
-	 * 
-	 * We made sure that the next buffer was empty, so we can now start the
-	 * insertion process.
+	 * The next buffer pointed to by `ring_unused` is now unused, so we can insert
+	 * the new request to it.
 	 */
 	ring_index = MyPState->ring_unused;
 	index = (ring_index % READ_BUFFER_SIZE);
@@ -568,8 +565,8 @@ prefetch_register_buffer(BufferTag tag, bool *force_latest, XLogRecPtr *force_ls
 	Assert(slot->status == PRFS_UNUSED);
 
 	/*
-	 * We must update the slot data before insertion, because otherwise the
-	 * hash would be incorrect.
+	 * We must update the slot data before insertion, because the hash
+	 * function reads the buffer tag from the slot.
 	 */
 	slot->buftag = tag;
 	slot->my_ring_index = ring_index;
@@ -1061,6 +1058,8 @@ neon_init(void)
 		return;
 
 	MyPState = MemoryContextAllocZero(TopMemoryContext, sizeof(PrefetchState));
+	
+	MyPState->n_unused = READ_BUFFER_SIZE;
 
 	MyPState->bufctx = SlabContextCreate(TopMemoryContext,
 										 "NeonSMGR/prefetch",
@@ -1072,8 +1071,6 @@ neon_init(void)
 	MyPState->hashctx = AllocSetContextCreate(TopMemoryContext,
 											  "NeonSMGR/prefetch",
 											  ALLOCSET_DEFAULT_SIZES);
-	inner_ps_disconnect_hook = ps_disconnect_hook;
-	ps_disconnect_hook = prefetch_on_ps_disconnect;
 
 	info.keysize = sizeof(BufferTag);
 	info.entrysize = sizeof(uint64);
@@ -1588,6 +1585,7 @@ neon_read_at_lsn(RelFileNode rnode, ForkNumber forkNum, BlockNumber blkno,
 			}
 			/* drop caches */
 			prefetch_set_unused(entry->slot->my_ring_index, true);
+			n_prefetch_missed_caches += 1;
 			/* make it look like a prefetch cache miss */
 			entry = NULL;
 		}
--- a/pgxn/neon/libpagestore.c
+++ b/pgxn/neon/libpagestore.c
@@ -47,9 +47,6 @@ int			flush_every_n_requests = 8;
 
 static void pageserver_flush(void);
 
-ps_connect_handle ps_connect_hook = NULL;
-ps_disconnect_handle ps_disconnect_hook = NULL;
-
 static void
 pageserver_connect()
 {
@@ -114,9 +111,6 @@ pageserver_connect()
 	neon_log(LOG, "libpagestore: connected to '%s'", page_server_connstring_raw);
 
 	connected = true;
-
-	if (ps_connect_hook)
-		ps_connect_hook();
 }
 
 /*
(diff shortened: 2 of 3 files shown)


pgxn/neon/pagestore_smgr.c
```diff
@@ -1011,7 +1282,7 @@ neon_close(SMgrRelation reln, ForkNumber forknum)
 void
 neon_reset_prefetch(SMgrRelation reln)
 {
-	n_prefetch_requests = 0;
+	Assert(false);
 }
```
hlinnaka, line 1286: Can we just remove this?

pgxn/neon/pagestore_smgr.c
```diff
@@ -127,35 +129,291 @@ static UnloggedBuildPhase unlogged_build_phase = UNLOGGED_BUILD_NOT_IN_PROGRESS;
...
+static void
+prefetch_register_buffer(BufferTag tag)
+{
+	int		index,
+			step = 1;
+	uint64	ring_index,
+			lastofrelfork = 0;
+	PrefetchRequest *slot;
+
+	/* Check that there is no prefetch in the buffers */
```
hlinnaka, line 330:
```suggestion
	/* Check that there is no prefetch for the requested buffer already */
```

hlinnaka (commented):
+1 on the PrefetchRequest and PrefetchState structs. 

I don't quite understand the `nextOfRel`and `prevOfRel` ring stuff. Within the array of prefetch buffers, we form a linked list of all requests belonging to the same relation? How critical is it that _all_ requests belonging to the same relation are in the same list? If you have two sequential scans happening on the same relation at the same time but starting at different positions, would they be part of the same list?

ISTM the only place where the nextOfRel/prevOfRel are used is in `prefetch_register_buffer`, to check if there already is a prefetch request queued for the page. Or to speed up that search, to be precise?  How about having a hash table, to map from BufferTag to prefetch request, instead of nextOfRel/prevOfRel? That would seem simpler.

I don't quit understand the `ring_unused`, `ring_receive` and `ring_last` fields either. Are they related to the linked lists that nextOfRel/prevOfRel form, or something completely different?

A comment with a high-level description of how this works, at top of `pagestore_smgr.c`, for example, would be very helpful.

hlinnaka:
>In a future iteration, this will detect sequential scans based on the read behavior of sequential scans, and will dynamically prefetch buffers for such relations as needed.

Oh is that what the 'nextOfRel`/`prevOfRel` are for?

pgxn/neon/pagestore_fetchbuf.h
```diff
@@ -0,0 +1,79 @@
...
+ * ring_unused points to the first unused slot of the buffer
+ * ring_receive is the next request that is to be received
+ * ring_last is the oldest received entry in the buffer
+ * 
+ * Apart from being an entry in the ring buffer of prefetch requests, each
+ * PrefetchRequest is linked to the next and previous PrefetchRequest of its
+ * RelNodeFork through the nextOfRel and prevOfRel relative pointers into the
+ * ring buffer. This provides a linked list for each relations' fork, which
+ * will allow us to detect sequential scans; eventually removing (or reducing)
+ * the need for core modifications in the heap AM for prefetching buffers.
```
hlinnaka, line 60: This makes sense. For now, let's keep it simple though. If we don't need the ring stuff yet, let's remove it for now, and introduce it later when we actually try to detect sequential scans.

pgxn/neon/pagestore_fetchbuf.h
```diff
@@ -0,0 +1,79 @@
...
+/* Max amount of tracked buffer reads */
+#define READ_BUFFER_SIZE 128
+
+typedef enum PrefetchStatus {
+	PRFS_EMPTY = 0,
+	PRFS_REQUESTED,
+	PRFS_RECEIVED,
+	PRFS_TAG_REMAINS,
+} PrefetchStatus;
+
```
hlinnaka, line 30: A comment to explain what each value means would be nice. All the other ones are quite self-explanatory actually, but I cannot guess what PRFS_TAG_REMAINS means (haven't read the rest of the patch yet).

pgxn/neon/pagestore_smgr.c
```diff
@@ -126,35 +128,338 @@ static UnloggedBuildPhase unlogged_build_phase = UNLOGGED_BUILD_NOT_IN_PROGRESS;
...
+	);
+
+	MyPState->n_requestes_inflight++;
+
+	page_server->send((NeonRequest *) &request);
+	slot->status = PRFS_REQUESTED;
+}
+
+static void
+prefetch_move_buffers(uint64 to, uint64 from, int count)
```
hlinnaka, line 330: this function is unused

pgxn/neon/pagestore_smgr.c
```diff
@@ -126,35 +128,338 @@ static UnloggedBuildPhase unlogged_build_phase = UNLOGGED_BUILD_NOT_IN_PROGRESS;
...
+									   ForkNumber forknum, BlockNumber blkno);
+
+/*
+ * Make sure that there are no responses still in the buffer.
+ */
 static void
 consume_prefetch_responses(void)
 {
-	for (int i = n_prefetched_buffers; i < n_prefetch_responses; i++)
+	MemoryContext old = MemoryContextSwitchTo(MyPState->context);
```
hlinnaka, line 159: Took me a while to find where we do allocations where this takes effect. It's in `prefetch_read`, it calls `page_server->receive()`, which returns a palloc'd struct.

I'd recommend moving this `MemoryContextSwitchTo` into `prefetch_read` itself. As it is, it's a subtle requirement that the caller must've have done that. That's easy to miss if this gets refactored around in the future, and it could appear to work at first if the calling memory context is fairly long-lived.

pgxn/neon/pagestore_smgr.c
```diff
@@ -617,7 +921,32 @@ neon_wallog_page(SMgrRelation reln, ForkNumber forknum, BlockNumber blocknum, ch
...
+	HASHCTL info;
+
+	if (MyPState != NULL)
+		return;
+
+	MyPState = palloc0(sizeof(PrefetchState));
+	
+	MyPState->bufctx = SlabContextCreate(TopMemoryContext,
+										 "NeonSMGR/prefetch",
+										 SLAB_DEFAULT_BLOCK_SIZE * 17,
```
hlinnaka, line 933: Where does the "17" come from?

pgxn/neon/pagestore_smgr.c
```diff
@@ -126,35 +127,335 @@ static UnloggedBuildPhase unlogged_build_phase = UNLOGGED_BUILD_NOT_IN_PROGRESS;
...
+		PrefetchRequest *slot;
+		int		index = MyPState->ring_receive % READ_BUFFER_SIZE;
+
+		slot = &MyPState->prf_buffer[index];
+		Assert(slot->status == PRFS_REQUESTED);
+		Assert(slot->my_ring_index == MyPState->ring_receive);
+
+		/* clean up the request */
+		slot->status = PRFS_TAG_REMAINS;
+		prefetch_set_unused(MyPState->ring_receive);
```
hlinnaka, line 282: This `PRFS_TAG_REMAINS` status is very short-lived now, `prefetch_set_unused` will immediately change the status to PRFS_UNUSED. Let's simplify and remove `PRFS_TAG_REMAINS` altogether.

pgxn/neon/pagestore_smgr.c
```diff
@@ -126,35 +127,335 @@ static UnloggedBuildPhase unlogged_build_phase = UNLOGGED_BUILD_NOT_IN_PROGRESS;
...
+	}
+}
+
+/*
+ * prefetch_set_unused() - clear a received prefetch slot
+ *
+ * The slot at ring_index must be a current member of the ring buffer,
+ * and may not be in the PRFS_REQUESTED state.
```
hlinnaka, line 290: Took me a moment to understand why it must not be in `PRFS_REQUESTED` state: it's because that would make the requests go "out of sync" with the responses that we expect from the pageserver.

If we lose the compute <-> pageserver connection, though, we would want to flush all the buffers, and we would want to call this with `PRFS_REQUESTED`. Right?

pgxn/neon/pagestore_smgr.c
```diff
@@ -126,35 +127,335 @@ static UnloggedBuildPhase unlogged_build_phase = UNLOGGED_BUILD_NOT_IN_PROGRESS;
...
+	MemSet(slot, 0, sizeof(PrefetchRequest));
+	slot->status = PRFS_UNUSED;
+
+	/* run cleanup if we're holding back ring_last */
+	if (MyPState->ring_last == ring_index)
+		prefetch_cleanup();
+}
+
+static void
+prefetch_do_request_ext(PrefetchRequest *slot, bool *force_latest, XLogRecPtr *force_lsn)
```
hlinnaka, line 329:
```suggestion
prefetch_do_request(PrefetchRequest *slot, bool *force_latest, XLogRecPtr *force_lsn)
```
No need for the `ext` AFAICS

pgxn/neon/pagestore_smgr.c
```diff
@@ -126,35 +127,335 @@ static UnloggedBuildPhase unlogged_build_phase = UNLOGGED_BUILD_NOT_IN_PROGRESS;
...
+/*
+ * prefetch_register_buffer() - register and prefetch buffer
+ *
+ * Register that we may want the contents of BufferTag in the near future.
+ */
+static uint64
+prefetch_register_buffer(BufferTag tag)
+{
+	return prefetch_register_buffer_ext(tag, NULL, NULL);
+}
```
hlinnaka, line 372: This wrapper function seems a bit unnecessary, I'd suggest to remove this function, rename `prefetch_register_buffer_ext` to just `prefetch_register_buffer`, and call that directly

pgxn/neon/pagestore_smgr.c
```diff
@@ -126,35 +127,335 @@ static UnloggedBuildPhase unlogged_build_phase = UNLOGGED_BUILD_NOT_IN_PROGRESS;
...
+
+static uint64
+prefetch_register_buffer_ext(BufferTag tag, bool *force_latest, XLogRecPtr *force_lsn)
```
hlinnaka, line 375: What does `force_latest` and `force_lsn` mean? What does NULL mean? A comment would be nice

pgxn/neon/pagestore_smgr.c
```diff
@@ -126,35 +127,335 @@ static UnloggedBuildPhase unlogged_build_phase = UNLOGGED_BUILD_NOT_IN_PROGRESS;
...
+		}
 	}
-	n_prefetched_buffers = 0;
-	n_prefetch_responses = 0;
+
+	ring_index = MyPState->ring_unused;
+	index = (ring_index % READ_BUFFER_SIZE);
+	slot = &MyPState->prf_buffer[index];
+
+	if (MyPState->ring_last == (MyPState->ring_unused - READ_BUFFER_SIZE))
```
hlinnaka, line 411:
```suggestion
	/*
	 * If we prefetch queue is full, we need to make room by clearing the oldest slot. If the oldest slot
	  * holds a buffer that was already received, we have to throw it away; we fetched the page
	  * unnecessarily in that case. If the oldest slot holds a request that we haven't received a response for
	  * yet, we have to wait for the response to that before we can continue. We might not have even flushed
	  * the request to the pageserver yet, it might be just sitting in the output buffer. In that case, we flush it
	  * and wait for the response. (We could decide not to send it, but it's hard to abort when the request is
	  * already in the output buffer)
	  */
	if (MyPState->ring_last == (MyPState->ring_unused - READ_BUFFER_SIZE))
```
Did I understand this correctly? Maybe we should not call `page_server->send()` on the requests until just before we flush them? That way, we could decide to just drop the unsent request here. It probably doesn't happen very often to matter much, though.

pgxn/neon/pagestore_smgr.c
```diff
@@ -126,35 +127,335 @@ static UnloggedBuildPhase unlogged_build_phase = UNLOGGED_BUILD_NOT_IN_PROGRESS;
...
+static uint64
+prefetch_register_buffer_ext(BufferTag tag, bool *force_latest, XLogRecPtr *force_lsn)
+{
+	int		index;
+	bool	found;
+	uint64	ring_index;
+	PrefetchRequest *slot;
+	PrfHashEntry *entry;
+
+	entry = prfh_lookup(MyPState->prf_hash, (PrefetchRequest *) &tag);
```
hlinnaka, line 383: This function does a `prfh_lookup`, and later `prfh_insert`. You could save some cycles by doing `prfh_insert` directly, and checking the return code for whether it already existed. I'm not sure if there would be problems if the `prefetch_set_unused` calls below remove entries from the hash table, I don't remember if it's safe to keep accessing an entry returned by `prfh_insert`, if you modify the hash table at the same time.

If we don't dare to do that, an easier win would be calculate the hash key only once, and call `prfh_lookup_hash` and `prfh_insert_hash`.

pgxn/neon/pagestore_fetchbuf.h
```diff
@@ -0,0 +1,77 @@
...
+	MemoryContext hashctx; /* context for prf_buffer */
+
+	/* buffer indexes */
+	uint64	ring_unused;		/* first unused slot */
+	uint64	ring_receive;		/* next slot that is to receive a response */
+	uint64	ring_last;			/* min slot with a response value */
+
+	/* metrics / statistics  */
+	int		n_responses_buffered;	/* count of PS responses not yet in buffers */
+	int		n_requestes_inflight;	/* count of PS requests considered in flight */
```
hlinnaka, line 69:
```suggestion
	int		n_requests_inflight;	/* count of PS requests considered in flight */
```

hlinnaka (approved):
I left a lot of comments on the details, but at high level this looks good to me.

pgxn/neon/pagestore_smgr.c
hlinnaka: and I don't know if this makes any difference at all in practice, I don't have a feel for how performance-critical this is. :-)

pgxn/neon/pagestore_fetchbuf.h
```diff
@@ -0,0 +1,77 @@
...
+typedef enum PrefetchStatus {
+	PRFS_UNUSED = 0, /* unused slot */
+	PRFS_REQUESTED, /* request is sent to PS, all fields except response valid */
+	PRFS_RECEIVED, /* all fields valid, response contains data */
+	PRFS_TAG_REMAINS, /* only buftag, *OfRel are still valid */
+} PrefetchStatus;
+
+typedef struct PrefetchRequest {
+	BufferTag	buftag; /* must be first entry in the struct */
+	XLogRecPtr	effective_request_lsn;
```
hlinnaka, line 33: This is never read. I think we should check this in `neon_read_at_lsn`, before we accept a prefetched response. It's possible that we issue a prefetch request with LSN 1000, and receive a response, but before we consume the response, another backend reads the same page, modifies it with LSN 2000, and it gets evicted again. In that case, we must not use the old prefetched page version at LSN 1000.

This is assigned from `prefetch_lsn`, in `prefetch_do_request_ext`, but `prefetch_lsn` is never set.

pgxn/neon/pagestore_smgr.c
```diff
@@ -126,35 +127,335 @@ static UnloggedBuildPhase unlogged_build_phase = UNLOGGED_BUILD_NOT_IN_PROGRESS;
...
-BufferTag	prefetch_requests[MAX_PREFETCH_REQUESTS];
-BufferTag	prefetch_responses[MAX_PREFETCH_REQUESTS];
-int			n_prefetch_requests;
-int			n_prefetch_responses;
-int			n_prefetched_buffers;
 int			n_prefetch_hits;
 int			n_prefetch_misses;
+int			n_prefetch_dupes;
+
 XLogRecPtr	prefetch_lsn;
```
hlinnaka, line 136:
```suggestion
```

pgxn/neon/pagestore_smgr.c
```diff
@@ -126,35 +127,335 @@ static UnloggedBuildPhase unlogged_build_phase = UNLOGGED_BUILD_NOT_IN_PROGRESS;
...
+{
+	NeonGetPageRequest request = {
+		.req.tag = T_NeonGetPageRequest,
+		.req.latest = false,
+		.req.lsn = 0,
+		.rnode = slot->buftag.rnode,
+		.forknum = slot->buftag.forkNum,
+		.blkno = slot->buftag.blockNum,
+	};
+	slot->effective_request_lsn = prefetch_lsn;
```
hlinnaka, line 339: `prefetch_lsn` is never set. Suggest to move this after the if-else, and do:
```
	slot->effective_request_lsn = request.req.lsn;
```

pgxn/neon/pagestore_smgr.c
hlinnaka: ok

pgxn/neon/pagestore_smgr.c
```diff
@@ -113,6 +111,99 @@ typedef enum
...
+ * 
+ * 
+ * There can be up to MAX_PREFETCH_REQUESTS registered using smgr_prefetch
+ * before smgr_read. All this requests are appended to primary smgr_read request.
+ * It is assumed that pages will be requested in prefetch order.
+ * Reading of prefetch responses is delayed until them are actually needed (smgr_read).
+ * It make it possible to parallelize processing and receiving of prefetched pages.
+ * In case of prefetch miss or any other SMGR request other than smgr_read,
+ * all prefetch responses has to be consumed.
+ */
```
hlinnaka, line 127: This comment needs some love.

pgxn/neon/pagestore_smgr.c
```diff
@@ -126,35 +217,379 @@ static UnloggedBuildPhase unlogged_build_phase = UNLOGGED_BUILD_NOT_IN_PROGRESS;
...
+		 * LSN, but still correct:
+		 * 
+		 * We know there are no changes between the actual requested LSN and
+		 * the value of effective_request_lsn: If there were, the page would
+		 * have been in cache and evicted between those LSN values, which
+		 * then would have had to result in a larger request LSN for this page.
+		 * 
+		 * It is possible that a concurrent backend loads the page, modifies it
+		 * and then evicts it again, but the LSN of that eviction cannot be
+		 * smaller than the current redo pointer, which is already larger
```
hlinnaka, line 435: "current redo pointer"? What does that refer to?

pgxn/neon/pagestore_smgr.c
```diff
@@ -126,35 +217,379 @@ static UnloggedBuildPhase unlogged_build_phase = UNLOGGED_BUILD_NOT_IN_PROGRESS;
...
+			n_prefetch_dupes++;
+			return ring_index;
+		}
 	}
-	n_prefetched_buffers = 0;
-	n_prefetch_responses = 0;
+
+	/*
+	 * If we prefetch queue is full, we need to make room by clearing the
+	 * oldest slot. If the oldest slot  holds a buffer that was already
```
hlinnaka, line 521:
```suggestion
	 * If the prefetch queue is full, we need to make room by clearing the
	 * oldest slot. If the oldest slot holds a buffer that was already
```

pgxn/neon/pagestore_smgr.c
```diff
@@ -126,35 +217,379 @@ static UnloggedBuildPhase unlogged_build_phase = UNLOGGED_BUILD_NOT_IN_PROGRESS;
...
+				pg_unreachable();
+		}
+	}
+
+	/*
+	 * We could not use the requested buffer (if we had any), so any current
+	 * buffer has been released. We now have to aquire a new request buffer.
+	 * 
+	 * We made sure that the next buffer was empty, so we can now start the
+	 * insertion process.
```
hlinnaka, line 560:
```suggestion
	 * The next buffer pointed to by `ring_unused` is now unused, so we can insert
	 * the new request to it.
```

pgxn/neon/pagestore_smgr.c
hlinnaka: Hmm, I think `PS_GETPAGERESPONSE_SIZE * 16` would be more understandable then.

I wonder if using the slab allocator is just a premature optimization. Compared to copying 8kB of data to the buffer and all the other processing we do on each response, the overhead of one malloc+free is not that big. Did you see malloc/free in profiles?

Or we could redefine `pageserver_receive` so that the caller passes a StringInfo, and the function writes the response to that, enlarging it if needed. Then we could keep using the same buffer, skipping the palloc+free overhead altogether. That's a slightly larger refactoring though. If we're going to do that, then it would probably make sense to try to avoid the memcpy we currently do in `nm_unpack_response`. So probably best to leave that for another PR...

pgxn/neon/pagestore_smgr.c
```diff
@@ -1094,81 +1550,71 @@ neon_read_at_lsn(RelFileNode rnode, ForkNumber forkNum, BlockNumber blkno,
...
-				memcpy(buffer, page, BLCKSZ);
-				pfree(resp);
-				return;
+				page_server->flush();
+				prefetch_wait_for(entry->slot->my_ring_index);
 			}
+			/* drop caches */
+			prefetch_set_unused(entry->slot->my_ring_index, true);
+			/* make it look like a prefetch cache miss */
+			entry = NULL;
```
hlinnaka, line 1592: Would be good to have a counter to track how often this happens. I would expect it to be very rare, so if that you see that counter going very high in some workload, that would indicate that we have a bug or mistuned system.

hlinnaka:
@bojanserafimov can you reproduce https://github.com/neondatabase/neon/issues/2585 with this PR?

bojanserafimov:
> @bojanserafimov can you reproduce #2585 with this PR?

I can check but merging my very hacky branch that discovered the bug with main is not worth the effort. I'll try to reproduce it after https://github.com/neondatabase/neon/pull/2762 and I'll close the issue if I can't

hlinnaka:
The `neon.flush_output_after` GUC doesn't work as I expected. I started a sequential scan and looked at the traffic with Wireshark. This was with `enable_seqscan_prefetch=on`, `seqscan_prefetch_buffers = 100`, `enable_indexonlyscan=off`, `max_parallel_workers_per_gather=0`, and `neon_flush_output_after=8`. After the initial burst of prefetch requests at the start of the scan, I'm seeing a separate prefetch request being sent after each received response. See screenshot.

image

hlinnaka:
> The neon.flush_output_after GUC doesn't work as I expected.

Opened https://github.com/neondatabase/neon/pull/2771 to fix that

Issue #2569, opened 2022-10-06
walproposer hangs if single safekeepers hangs


petuhovskiy:
## Steps to reproduce

Run a single safekeeper that accepts new connections and accepts new messages, but doesn't reply with any messages. Other 2 safekeepers should be healthy. Run `postgres --sync-safekeepers` and it will stuck.

For example, this happened in https://github.com/neondatabase/cloud/issues/2426 because we had a deadlock (#2566) and timelines became locked when they tried to persist something to disk. This made safekeeper read new messages without replying back.

## Expected result

walproposer requires only 2 working safekeepers to finish `--sync-safekeepers`. It also should drop connection to safekeeper if it doesn't receive new messages for some time (~5 seconds).

## Actual result

`postgres --sync-safekeepers` doesn't exit

andresrsanchez:
Hi @petuhovskiy, so i did a little test on my computer trying to replicate that scenario:

- Manually provoke a deadlock on [PhysicalStorage](https://github.com/neondatabase/neon/blob/main/safekeeper/src/wal_storage.rs#L270) to avoid writing to disk in only one safekeeper. 
- Modify [test_sync_safekeepers](https://github.com/neondatabase/neon/blob/main/test_runner/regress/test_wal_acceptor.py#L688) to do `append_logical_message` in background just to avoid the exception because of the previous deadlock.
- Run the test and i see that hangs on `sync_safekeepers`, great!
- With gdb i can check that one safekeeper is stuck in `epoll_wait` coming from [WaitEventSetWait](https://github.com/neondatabase/neon/blob/main/pgxn/neon/walproposer.c#L344) with `SS_HANDSHAKE_RECV` state, and this state cannot move forward.

So at the end i modified `WalProposerPoll`:
```
for (int i = 0; i < n_safekeepers; i++)
{
	Safekeeper *sk = &safekeeper[i];

	if ((sk->state == SS_CONNECTING_WRITE ||
		 sk->state == SS_CONNECTING_READ ||
                  sk->state == SS_HANDSHAKE_RECV ) && /* <------------------ I added this line */
		TimestampDifferenceExceeds(sk->startedConnAt, now, wal_acceptor_connect_timeout))
	{
		elog(WARNING, "failed to connect to node '%s:%s': exceeded connection timeout %dms",  sk->host, sk->port, wal_acceptor_connect_timeout);
		ShutdownConnection(sk);
	}
}
```
And now the conn gets dropped and the test does not get stuck.

[...]

PR #2694, opened 2022-10-26
Avoid hanging on sync-safekeeper with one SK stuck

andresrsanchez:
refer #2569 
I created the PR as draft to see if this is the correct approach.

- The `startedConnAt` is not strictly necessary now, we can go back to it with the combination of `SS_CONNECTING_WRITE` and  `SS_CONNECTING_READ`.
- @petuhovskiy when you said that `we have a logic to send a message every second between compute and safekeeper` this keep alive is when the SK is in `SS_ACTIVE` state if i'm not wrong, I updated the `latestMsgReceivedAt` in `AsyncReadMessage` to get the most updated data possible, with this we cover other states than `SS_ACTIVE`.

Let me know what you think please, thanks!!!


petuhovskiy:
The code looks good! I'm going to do some additional testing and will try to get it merged this week.

>  this keep alive is when the SK is in SS_ACTIVE state if i'm not wrong

Yes, and when SK is not in SS_ACTIVE state, we expect it to work instantly, so we wait only for latency between the compute and the safekeepers in most states.

andresrsanchez:
Thanks! Ping me if u need anything

petuhovskiy:
Force pushed your changes to properly retrigger tests

arssher (approved):

arssher:
Wups, I forgot to squash before merge =(

PR #2701, opened 2022-10-27
Enable etcd compaction in neon_local.

arssher:


knizhnik (approved):
May be we should explain why we have added this option: to avoid "etcdserver: mvcc: database space exceeded" error

PR #2707, opened 2022-10-27
test_backward_compatibility: friendlier error message

bayandin:
Make the error message friendlier for local runs when the compatibility snapshot directory is not set or doesn't exist.


alexstanovoy (approved):
Now it works perfectly locally. Thanks!

PR #2714, opened 2022-10-28
Add initial tenant sizing model and a http route to query it

koivunej:
~Pricing~ tenant size information is gathered by using existing parts of `Tenant::gc_iteration` which are now separated as `Tenant::refresh_gc_info`. `Tenant::refresh_gc_info` collects branch points, and invokes `Timeline::update_gc_info`; nothing was supposed to be changed there. The gathered branch points (through Timeline's `GcInfo::retain_lsns`), `GcInfo::horizon_cutoff`, and `GcInfo::pitr_cutoff` are used to build up a Vec of updates fed into the `libs/tenant_size_model` to calculate the history size.

The gathered information is now exposed using `GET /v1/tenant/{tenant_id}/size`, which which will respond with the actual calculated size. Initially the idea was to have this delivered as tenant background task and exported via metric, but it might be too computationally expensive to run it periodically as we don't yet know if the returned are any good.

Adds one new metric:
- pageserver_storage_operations_seconds with label `logical_size`
    - separating from original `init_logical_size`

Adds a pageserver wide configuration variable:
- `concurrent_tenant_size_logical_size_queries` with default 1

TODOs:
- [x] pricing model inputs: they possibly need to only be gathered up to `gc_cutoff_lsn`
- [x] pageserver configurable semaphore permits for concurrent logical_sizes
- [x] python test
- [x] get tests to pass
- [x] remove extra python test


pageserver/src/tenant.rs
```diff
@@ -120,6 +126,10 @@ pub struct Tenant {
 
     /// Makes every timeline to backup their files to remote storage.
     upload_layers: bool,
+
+    /// Cached logical sizes updated updated on each [`Tenant::collect_pricing_operations`].
+    // TODO: Lsn: Hash
+    cached_logical_sizes: tokio::sync::Mutex<HashMap<(TimelineId, u64), u64>>,
```
koivunej, line 132: I don't see why Lsn shouldn't implement Hash, but didn't yet search any archives for a reason.

pageserver/src/tenant.rs
```diff
@@ -1446,6 +1481,21 @@ impl Tenant {
...
+        use pricing::gather_inputs;
+
+        // TODO: get the size from self.conf, which could also be used to disable if zero?
+        let logical_sizes_at_once =
+            LOGICAL_SIZES_AT_ONCE.get_or_init(|| Arc::new(tokio::sync::Semaphore::new(8)));
+
+        let mut shared_cache = self.cached_logical_sizes.lock().await;
+
+        gather_inputs(self, logical_sizes_at_once, &mut *shared_cache).await
+    }
```
koivunej, line 1498: Should rename these as well towards history size.

koivunej:
Decided to drop the periodic task and metric and only do http request support for now.

pageserver/src/tenant.rs
koivunej: Derived this.

pageserver/src/tenant.rs
koivunej: Did not remain these, but we can revisit this idea.

koivunej:
Re: 1/2 of tasks done: I am still fuzzy on the actual pricing model inputs: following off-github conversations, currently they are calculated by taking all of the branch points and then an additional state for `next_gc_cutoff_lsn`, should the `gc_cutoff_lsn` be after the branch point.

Alternative one would be to include the branch points only if the `gc_cutoff_lsn` comes after the branch point, leaving out all branches happening after it.

koivunej:
"Pricing model" is not a great name and this is one of the confusion points. Alternatives:
- history_size
- retained_history_size
- ideal_history_size

koivunej:
Just logging progress in case you happen yo review right now. Creating a simple test reveals this doesn't currently work with empty database.

libs/pricing_model/src/lib.rs
```diff
@@ -0,0 +1,338 @@
+use std::borrow::Cow;
+use std::collections::HashMap;
+
+/// Maintains latest [`Segment`] for each of the branches, faciliates the pricing model
+/// calculations.
```
hlinnaka, line 5: The `segments` vector maintains *all* segments, not just the latest of each branch. This comment refers just to the 'branches` field, not the whole struct.

libs/pricing_model/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+pub struct Storage<K: 'static> {
+    segments: Vec<Segment>,
+    branches: HashMap<K, usize>,
```
hlinnaka, line 8: What are the key and value here?

Comparing with the original hacky version of this that I wrote, the key was a branch name, as a String. Perhaps it should be TimelineId here?

And the value is an index into the `segments` array. Needs a comment

libs/pricing_model/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+}
+
+/// State of a branch.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Segment {
+    parent: Option<usize>,
+
+    /// Description of how did we get to this state.
+    pub op: Cow<'static, str>,
```
hlinnaka, line 17: Not sure we can really get that information from a live system

libs/pricing_model/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+/// Maintains latest [`Segment`] for each of the branches, faciliates the pricing model
+/// calculations.
+pub struct Storage<K: 'static> {
+    segments: Vec<Segment>,
+    branches: HashMap<K, usize>,
+}
+
+/// State of a branch.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Segment {
```
hlinnaka, line 13: It's a bit confusing to store the logical size as part of the segment, if a segment represents a range of LSNs. I was confused by that myself even though I wrote this, but in hindsight, perhaps it would be more clear to have Points and Segments, so that a Point represents a particular LSN on a timeline, and and a Segment represents a range of LSNs, between two Points.

(If you got this working as it is, no need to start changing it now, but maybe in the future..)

pageserver/src/tenant/pricing.rs
```diff
@@ -0,0 +1,458 @@
...
+#[derive(Debug, serde::Serialize)]
+pub struct ModelInputs {
+    updates: Vec<Update>,
+    retention_period: u64,
+    #[serde_as(as = "HashMap<serde_with::DisplayFromStr, _>")]
+    timeline_inputs: HashMap<TimelineId, TimelineInputs>,
+}
+
+/// Collect all relevant LSNs to the inputs, even though they will only be helpful in the
+/// serialized form.
```
hlinnaka, line 26: I don't understand this comment. How are they only helpful in the serialized form?

libs/pricing_model/src/lib.rs
hlinnaka: (after reading the PR further): I see you actually did something like that with the `updates` and `Command` stuff. Makes sense! You could move that here, I think.

hlinnaka (approved):
I think this is in pretty good shape. Lots of new code, and I'm not sure we got all the details right. But since this is a new feature, it doesn't need to be perfect in the first version, we can fix and improve it incrementally.

pageserver/src/tenant/pricing.rs
```diff
@@ -0,0 +1,458 @@
...
+                    lsn: *lsn,
+                    timeline_id: timeline.timeline_id,
+                    command: Command::Update(*size),
+                });
+
+                needed_cache.insert((timeline.timeline_id, *lsn));
+            } else {
+                let timeline = Arc::clone(&timeline);
+                let parallel_size_calcs = Arc::clone(limit);
+                joinset.spawn(calculate_logical_size(parallel_size_calcs, timeline, *lsn));
```
SomeoneToIgnore, line 152: I'm not sure we have to fix it right now, but generally I'm against such approach: 
* every timeline could be decomposed as [initdb_lsn, ..., interesting_lsn_1, ..., interesting_lsn_i, ..., disk_consistent_lsn, ...] — only data after `disk_consistent_lsn` changes, the rest of the data is always constant, so it does not make any sense to recalculate the start of the timeline multiple times
* `interesting_lsns` consists of branch points, which are usually placed on top of some data due to nature of the branching: we create them to experiment on some data without spoiling it and I would imagine that users would create those branches close to their latest data, meaning that all those calculations will be heavy 
* Plans like https://neondb.slack.com/archives/C026T7K2YP9/p1667293260574949 could lead us to many-many branches created, which will slow the entire process down, since we have to wait for them all to be calculated

So, the whole process gets slower and slower the more branches we'll create of a timeline, even if timeline's data is not changing at all: that is very odd.
And then, there's an initial logical size computation background task, that's not even considered as part of these calculations, even though it does the same essentially.

pageserver/src/tenant/pricing.rs
```diff
@@ -0,0 +1,458 @@
...
+    horizon_cutoff: Lsn,
+    #[serde_as(as = "serde_with::DisplayFromStr")]
+    pitr_cutoff: Lsn,
+    #[serde_as(as = "serde_with::DisplayFromStr")]
+    next_gc_cutoff: Lsn,
+}
+
+pub(super) async fn gather_inputs(
+    tenant: &Tenant,
+    limit: &Arc<Semaphore>,
```
SomeoneToIgnore, line 44: I don't see any `acquire` usages of this, is that ok?

pageserver/src/tenant/pricing.rs
```diff
@@ -0,0 +1,458 @@
...
+                    "failed to calculate logical size at {lsn}: {error:#}"
+                );
+                have_any_error = true;
+            }
+            Err(join_error) if join_error.is_cancelled() => {
+                unreachable!("we are not cancelling any of the futures, nor should be");
+            }
+            Err(join_error) => {
+                // cannot really do anything, as this panic is likely a bug
+                warn!("logical size query panicked, trying again later: {join_error:#}");
```
SomeoneToIgnore, line 200: Might be worth `error!`'ing such cases, after all, we `bail!` afterwards.

pageserver/src/http/routes.rs
```diff
@@ -571,6 +571,47 @@ async fn tenant_status(request: Request<Body>) -> Result<Response<Body>, ApiErro
...
+    ///
+    /// The type is described with `id` and `history_size` in the openapi_spec file, but the
+    /// `inputs` is intentionally left out. The type resides in the pageserver not to expose
+    /// `ModelInputs`.
+    #[serde_with::serde_as]
+    #[derive(serde::Serialize)]
+    struct TenantHistorySize {
+        #[serde_as(as = "serde_with::DisplayFromStr")]
+        id: TenantId,
+        history_size: u64,
```
SomeoneToIgnore, line 601: What are the units of this size?

pageserver/src/tenant/pricing.rs
```diff
@@ -0,0 +1,458 @@
...
+                Command::Update(sz) => {
+                    let latest = last_state.get_mut(timeline_id).ok_or_else(|| {
+                        anyhow::anyhow!(
+                        "ordering-mismatch: there must had been a previous state for {timeline_id}"
+                    )
+                    })?;
+
+                    let lsn_bytes = {
+                        let now: u64 = lsn.0;
+                        let prev: u64 = latest.0 .0;
```
SomeoneToIgnore, line 266: That looks scary. Could we use non-shortened identifiers (`sz`) or maybe desctucture the `latest` in place?

pageserver/src/tenant/pricing.rs
```diff
@@ -0,0 +1,458 @@
...
+}
+
+#[derive(Debug, Clone, Copy)]
+enum LsnKind {
+    BranchPoint,
+    GcCutOff,
+}
+
+/// Newtype around the tuple that carries the timeline at lsn logical size calculation.
+struct TimelineAtLsnSizeResult(
```
SomeoneToIgnore, line 325: NIT: Feels beneficial to add some names to the fields, but up to you.

SomeoneToIgnore (commented):
I'm not sure I've verified the branchpoint selection and recursive algorithms good enough, but had accumulated some notes already, might worth to publish them.
There might be many messages, sorry for the noise, those are mostly NITs though, so feel free to ignore whatever you don't agree with (would be cool to see a comment to that in this case though).

I have 4 major issues with the PR:

* bad naming: `p1`, `p2`, `sz`, `x`, etc. + some structs without the fields + some `usize` types in the same structs, that are used for both price and segment IDs. Not sure if worth paying attention to right now.

Most important bad naming implication is the "pricing" term itself, that gets freely converted into "size" and something else. Let's give a definition, what is that pricing and what is that size and what are their *units*. 
Might help us to understand how to name it better.

> let history_size = inputs.price().map_err(ApiError::InternalServerError)?;

Gives some hint what this "price" could be, but I'm still not quite sure what's going on.

* the semaphore to limit the calculations seems to be unused, am I wrong here?

* potential panics in the price calculation code: we're lucky to have this in isolated places of pageserver such as HTTP handler, but not sure why do we want to do that.

* non-incremental calculation of all sizes/prices

pageserver/src/tenant.rs
```diff
@@ -1446,6 +1476,20 @@ impl Tenant {
...
+    /// Gathers inputs from all of the timelines to produce a pricing model input.
+    ///
+    /// Future is cancellation safe. Only one calculation can be running at once per tenant.
+    #[instrument(skip_all, fields(tenant_id=%self.tenant_id))]
+    pub async fn gather_pricing_inputs(&self) -> anyhow::Result<pricing::ModelInputs> {
+        use pricing::gather_inputs;
+
+        let logical_sizes_at_once = self.conf.concurrent_pricing_logical_size_queries.inner();
+
+        let mut shared_cache = self.cached_logical_sizes.lock().await;
```
SomeoneToIgnore, line 1489: That leaves a bad taste, since presumably every other HTTP request has to wait on other (potentially lengthy) calculation to finish.

Again, feels like incremental size computation could help with that.

pageserver/src/tenant/pricing.rs
```diff
@@ -0,0 +1,458 @@
...
+            command: Command::BranchFrom(Some(ids[0])),
+            timeline_id: ids[1],
+        },
+    ];
+
+    let expected = r#"[{"lsn":"0/0","command":{"branch_from":null},"timeline_id":"7ff1edab8182025f15ae33482edb590a"},{"lsn":"0/10E49380","command":{"update":42164224},"timeline_id":"7ff1edab8182025f15ae33482edb590a"},{"lsn":"0/10E49380","command":{"branch_from":"7ff1edab8182025f15ae33482edb590a"},"timeline_id":"b68d6691c895ad0a70809470020929ef"}]"#;
+
+    let actual = serde_json::to_string(&updates).unwrap();
+
+    assert_eq!(expected, actual);
```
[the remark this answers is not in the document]
koivunej: Good point, this is now extra. Was previewing it this way and didn't consider to remove ut after many customizations were not needed.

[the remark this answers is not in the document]
koivunej: Removed in 7fe5bb6befb0d4f5c483d80f16ec6efe9d2f0616.

libs/pricing_model/src/main.rs
```diff
@@ -0,0 +1,256 @@
...
+        (scenario_2, 9000),
+        (scenario_3, 13000),
+        (scenario_4, 16000),
+        (scenario_5, 17000),
+        (scenario_6, 333_792_000),
+    ];
+
+    for (scenario, expected) in truths {
+        let (_, price) = scenario();
+        assert_eq!(*expected, price.total_children());
```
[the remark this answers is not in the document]
koivunej: Can we turn fn pointers to symbol names which they point to?

Usually when debugging these sorts of tests I add a `line!()` to the `truths` in this case as in 33abedaa.

test_runner/fixtures/neon_fixtures.py
```diff
@@ -2766,12 +2766,12 @@ def wait_for_last_record_lsn(
     tenant: TenantId,
     timeline: TimelineId,
     lsn: Lsn,
-):
+) -> Lsn:
     """waits for pageserver to catch up to a certain lsn"""
```
koivunej, line 2770: Should note the return value here, in: 5c52677b

pageserver/src/tenant/pricing.rs
koivunej: Handled in next push: ff61fcc3

pageserver/src/tenant/pricing.rs
koivunej: Fixed in 0738945d

pageserver/src/tenant/pricing.rs
```diff
@@ -0,0 +1,458 @@
...
+    logical_size_cache: &mut HashMap<(TimelineId, Lsn), u64>,
+) -> anyhow::Result<ModelInputs> {
+    // with joinset, on drop, all of the tasks will just be de-scheduled, which we can use to
+    // our advantage with `?` error handling.
+    let mut joinset = tokio::task::JoinSet::new();
+
+    // TODO: spawn_blocking, this will update latest_gc_cutoff_lsn, which can take a
+    // long time.
+    let timelines = tenant
+        .refresh_gc_info(None)
```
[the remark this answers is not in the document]
koivunej: Aah yes, misunderstood you at first. Good point. Fixed in 17ab7942

pageserver/src/tenant/pricing.rs
koivunej: It's acquired (and awaited) within the task which is spawned in the `joinset`, see `calculate_logical_size` later in this file.

pageserver/src/tenant/pricing.rs
koivunej: Yes, I was hinting towards manual inspection happening later on the returned document from the http api, but this comment is not great either, need to think about this...

pageserver/src/tenant.rs
```diff
@@ -1446,6 +1476,20 @@ impl Tenant {
 
         Ok(())
     }
+
+    /// Gathers inputs from all of the timelines to produce a pricing model input.
+    ///
+    /// Future is cancellation safe. Only one calculation can be running at once per tenant.
+    #[instrument(skip_all, fields(tenant_id=%self.tenant_id))]
+    pub async fn gather_pricing_inputs(&self) -> anyhow::Result<pricing::ModelInputs> {
+        use pricing::gather_inputs;
```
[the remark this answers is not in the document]
koivunej: Good point, fixed in d1104a23

pageserver/src/http/openapi_spec.yml
```diff
@@ -354,6 +354,51 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                type: object
+                required:
+                  - id
+                  - history_size
+                properties:
+                  tenant_id:
+                    type: string
```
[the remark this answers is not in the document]
koivunej: Good catch! Fixed in cd175450

pageserver/src/config.rs
```diff
@@ -476,6 +499,12 @@ impl PageServerConf {
                 "log_format" => builder.log_format(
                     LogFormat::from_config(&parse_toml_string(key, item)?)?
                 ),
+                "concurrent_pricing_logical_size_queries" => builder.concurrent_pricing_logical_size_queries({
+                    let s = parse_toml_string(key, item)?;
+                    let u = s.parse::<usize>().context("expected a number of initial permits, not {s:?}")?;
```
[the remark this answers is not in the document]
koivunej: Renamed these in: 08101c0e

libs/pricing_model/Makefile
```diff
@@ -0,0 +1,13 @@
+all: 1.svg 2.svg 3.svg 4.svg 1.png 2.png 3.png 4.png
```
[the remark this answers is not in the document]
koivunej: Aah yes, broke it when renaming.

Would probably be easier to define the binary as phony target and always invoke it through `cargo run` instead of querying where is the target directory.

libs/pricing_model/README.md
```diff
@@ -0,0 +1,9 @@
+# Logical size + WAL pricing
+
+This is a simulator to calculate the storage price of different
+scenarios, using the "Logical size + WAL" method. It was used to
+create the graphs in this presentation:
+
+https://docs.google.com/presentation/d/1OapE4k11xmcwMh7I7YvNWGC63yCRLh6udO9bXZ-fZmo/edit?usp=sharing
```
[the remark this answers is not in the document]
koivunej: Had missed this; I think I will remove it, describe that graphviz images can be created with the binary and were used to explore the history_size cost.

libs/pricing_model/src/lib.rs
koivunej: Extracted `K` to support strings in the original scenarios, `Option<TimelineId>` when used from pageserver.

I didn't want to use TimelineId not to change the existing scenarios. Key being generic comes at the cost of a bit of boilerplate, but it seems quite minimal.

libs/pricing_model/src/lib.rs
koivunej: I am quite fuzzy on how well it works as discussed. Testing this in integration tests is quite the chore, as it's LSN dependent and those might be unstable so there cannot really be any direct comparisons. Well, should try that still with CI. Failed to get unstable results locally, but it's likely I failed to apply the correct TZ and locale settings.

libs/pricing_model/src/lib.rs
koivunej: Adding a comment on what methods use this, and using `const NO_OP = Cow::Borrowed("");` in `scenario_6` in an attempt to clarify this.

libs/pricing_model/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+    parent: Option<usize>,
```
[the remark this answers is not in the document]
koivunej: Changing the comment to "Snapshot of a branch" for `Segment`, as for this `usize` it is an index into `Storage::segments`, commenting that as well.

Wouldn't want to do the newtyping. There are no removals from the indexed data so I am not too worried about them mismatching.

libs/pricing_model/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+    /// LSN at this state
+    pub end_lsn: u64,
+
+    /// Logical size before this state
+    start_size: u64,
+
+    /// Logical size at this state
+    pub end_size: u64,
+
+    children_after: Vec<usize>,
```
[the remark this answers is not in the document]
koivunej: I will comment that this only includes segments of the same branch, not other branches, if that'd clarify. Also commenting the indexing into `Storage::segments`..

Hmm, looking around, I think this has 0 or 1 elements always. Will do that refactoring in a separate commit, so it means a successor/next segment or snapshot of the branch happening later.

libs/pricing_model/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+
+    fn price_from_snapshot_later(&self, seg_id: usize) -> SegmentPrice {
+        // If this is needed, then it's time to do the snapshot and continue
+        // with wal method.
+        let seg = &self.segments[seg_id];
+        //eprintln!("snap: seg{}: {} needed: {}", seg_id, seg.children_after.len(), seg.needed);
+        if seg.needed {
+            let mut children = Vec::new();
+
+            for child_id in seg.children_after.iter() {
```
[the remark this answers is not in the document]
koivunej: Fixed this.

libs/pricing_model/Makefile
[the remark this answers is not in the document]
koivunej: Using `cargo metadata` would be the best way, but the `../../target/debug/pricing_model` works as well. We have other dependencies towards `target` being in the checkout so not inclined to make this anymore complicated.

libs/pricing_model/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+        let newseg_id = self.segments.len();
+        let lastseg = &mut self.segments[lastseg_id];
+
+        let newseg = Segment {
+            op,
+            parent: Some(lastseg_id),
+            start_lsn: lastseg.end_lsn,
+            end_lsn: lastseg.end_lsn + lsn_bytes,
+            start_size: lastseg.end_size,
+            end_size: (lastseg.end_size as i64 + size_bytes) as u64,
```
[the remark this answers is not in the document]
koivunej: > `self.segments[lastseg_id]` could panic

With regards to the indices my view is that panics are still safe, and at least these seem to be quite sealed within and not even described outside. But agreed that it's a bit unfortunate use of unwrap here, which could be rewritten quite easily.

I'm currently looking why are not the ci workflows starting though first at least.

libs/pricing_model/src/lib.rs
koivunej: Enhanced this in [32322b9](https://github.com/neondatabase/neon/pull/2714/commits/32322b9086a8708b2c6b39393fa575a9c0505aa5)

pageserver/src/tenant/pricing.rs
koivunej: They would make the match arms where this is deconstructed multiline so I'd hope to steer away from that.

libs/pricing_model/README.md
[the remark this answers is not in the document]
koivunej: Well, maybe it's useful to have the link here anyways, but just noting it's private in 6f6bfb3a.

libs/pricing_model/src/main.rs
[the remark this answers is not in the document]
SomeoneToIgnore: Not sure and I think we can just hardcode the scenario names, but line could work too, I guess.

libs/pricing_model/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "pricing_model"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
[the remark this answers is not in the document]
koivunej: I'd rather go with the repository license, so resolving this, we have plenty of others :)

libs/pricing_model/Makefile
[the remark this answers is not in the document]
koivunej: Fixed in https://github.com/neondatabase/neon/pull/2714/commits/ae5d84010b2fe99c7c0c17f8adf62a6305c25df4 if you'd like to retry?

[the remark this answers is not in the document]
SomeoneToIgnore: Works like a charm now.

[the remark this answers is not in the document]
SomeoneToIgnore: One thing we could consider doing is to add some gitignore for the new files:
```
work/neon/neon (6f6bfb3) * ❯ git s
## HEAD (no branch)
?? libs/pricing_model/1.png
?? libs/pricing_model/1.svg
?? libs/pricing_model/2.png
?? libs/pricing_model/2.svg
?? libs/pricing_model/3.png
?? libs/pricing_model/3.svg
?? libs/pricing_model/4.png
?? libs/pricing_model/4.svg
```

maybe, put them into a subdirectory and ignore that?

pageserver/src/http/routes.rs
koivunej: I am pretty sure it's bytes given it's mix of WAL and logical size.

libs/pricing_model/Makefile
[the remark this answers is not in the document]
koivunej: Would you be ok with local git ignore for `*.{png,svg,dot}`? Have to admit I don't know an elegant makefile solution to move them all in a subdirectory.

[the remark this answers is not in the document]
SomeoneToIgnore: I wonder, if we really need a makefile instead of a script then, but up to you.
If we don't plan to generate those images frequently, might be ok to leave everything as is already, I imagine.

pageserver/src/tenant/timeline.rs
```diff
@@ -979,9 +979,13 @@ impl Timeline {
     /// Calculate the logical size of the database at the latest LSN.
     ///
     /// NOTE: counted incrementally, includes ancestors, this can be a slow operation.
-    fn calculate_logical_size(&self, up_to_lsn: Lsn) -> anyhow::Result<u64> {
+    pub fn calculate_logical_size(&self, up_to_lsn: Lsn) -> anyhow::Result<u64> {
         info!("Calculating logical size for timeline {}", self.timeline_id);
-        let timer = self.metrics.init_logical_size_histo.start_timer();
+        let timer = if up_to_lsn == self.initdb_lsn {
```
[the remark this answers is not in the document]
koivunej: > Is it important to distinguish such cases? 

Distinguishing: I was thinking of it from the perspective of the metrics we've gathered 
and assume that the later LSN queries done will be much more expensive.

> that are created but never updated that would have `up_to_lsn == self.initdb_lsn` condition always true

I don't think the `initdb_lsn` is currently being queried because I crudely filter out any `gc_cutoff` in the parent branch, which might require fixing later but right now it's difficult to say if it's right or wrong. I didn't test this out at `initdb_lsn` because that would be below gc_horizon I guess in all usual configurations..

Even if this was true, it would be calculated one extra time per tenant only.

Idea of using the `Timeline::current_logical_size` as cache if it has already been calculated makes sense though. Should it be done on this PR?

libs/pricing_model/src/main.rs
```diff
@@ -0,0 +1,256 @@
...
+    storage.insert("c", 4000);
+    storage.insert("a", 2000);
+
+    let price = storage.price(5000);
+
+    (storage.into_segments(), price)
+}
+
+fn scenario_6() -> (Vec<Segment>, SegmentPrice) {
+    let ids = [
```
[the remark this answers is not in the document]
koivunej: Yes they are, ids was like the `timeline_ids` but just shorter to write. Renamed this to `branches`.

libs/pricing_model/src/main.rs
```diff
@@ -0,0 +1,256 @@
+use pricing_model::{Segment, SegmentPrice, Storage};
+
+// Main branch only. Some updates on it.
+fn scenario_1() -> (Vec<Segment>, SegmentPrice) {
+    // Create main branch
+    let mut storage = Storage::new("main");
+
+    // Bulk load 1 GB of data to it
+    storage.insert("main", 5_000);
```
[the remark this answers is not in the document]
SomeoneToIgnore: Which is also funny since GB in our repo mainly means GiB: https://github.com/neondatabase/neon/pull/534

But never mind, it's not that important.

libs/pricing_model/src/main.rs
```diff
@@ -0,0 +1,256 @@
+use pricing_model::{Segment, SegmentPrice, Storage};
```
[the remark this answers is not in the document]
koivunej: Added a module comment in [32322b9](https://github.com/neondatabase/neon/pull/2714/commits/32322b9086a8708b2c6b39393fa575a9c0505aa5)

pageserver/src/tenant/timeline.rs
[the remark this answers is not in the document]
SomeoneToIgnore: Depends how simple it is to do 🙂 
I still insist on the incremental, segment-based calculations for timeline sizes later, so this could be implemented later together.

pageserver/src/tenant/pricing.rs
```diff
@@ -0,0 +1,458 @@
...
+                next_gc_cutoff,
+            )
+        };
+
+        // update this to have a retention_period later for the pricing_model
+        // it is currently unsure what this should be.
+        if let Some(cutoff_distance) = last_record_lsn.checked_sub(next_gc_cutoff) {
+            match min_max_cutoff_distance.as_mut() {
+                Some((min, max)) => {
+                    *min = std::cmp::min(*min, cutoff_distance);
```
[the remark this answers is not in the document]
koivunej: Yes, I do however prefer the `std::cmp::{max,min}`. This is now only gathering the `max`, but it is still most likely off and I am unsure if it should be fixed on this PR and how.

pageserver/src/tenant/pricing.rs
koivunej: Overall this is a good point, and I had not noticed that discussion/idea.

Right now however this will only be provided via manually used http api so the danger side is low, however going back the original idea of a metric calculated periodically this size calculation has to become much more efficient overall, in addition to likely big fixes around where the data is actually gathered, it is likely to be wrong here (see how `refresh_gc_info` returns only the timelines which have grown past the gc horizon).

pageserver/src/tenant/pricing.rs
```diff
@@ -0,0 +1,458 @@
...
+    Ok(ModelInputs {
+        updates,
+        retention_period,
+        timeline_inputs,
+    })
+}
+
+impl ModelInputs {
+    pub fn price(&self) -> anyhow::Result<u64> {
+        let mut storage = pricing_model::Storage::<Option<TimelineId>>::new(None);
```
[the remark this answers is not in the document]
koivunej: Been trying to link the rationale for using `Option<TimelineId>` but it's where `Timeline::get_ancestor_timeline_id()` is used (also Option<TimelineId>) but I should repeat or move the description here, so that it can be referred from elsewhere.

pageserver/src/tenant/timeline.rs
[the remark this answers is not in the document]
koivunej: Had to rebase, but I think it's simple enough: 36f035d4f8804f7922da435baa7ef8f3ce58915b, what do you think?

pageserver/src/tenant/pricing.rs
```diff
@@ -0,0 +1,458 @@
...
+                .iter()
+                .inspect(|&&x| {
+                    trace!(
+                        timeline_id=%timeline.timeline_id,
+                        "retained lsn: {x:?}, is_before_ancestor_lsn={}",
+                        x < timeline.get_ancestor_lsn()
+                    )
+                })
+                .filter(|&&lsn| lsn > timeline.get_ancestor_lsn())
+                .cloned()
```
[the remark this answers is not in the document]
koivunej: Handled this in 84eb7acd (renaming x to lsn, cloned -> copied)

libs/pricing_model/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+    Skipped,
+}
+
+use SegmentMethod::*;
+
+impl<K: std::hash::Hash + Eq + 'static> Storage<K> {
+    /// Creates a new storage with the given default branch name.
+    pub fn new(initial_branch: K) -> Storage<K> {
+        let init_segment = Segment {
+            op: "".into(),
```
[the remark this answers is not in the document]
koivunej: However non-empty strings here will be used in the png's. I think the cost is minimal and I did try not to make unnecessary modifications to `libs/pricing_model` overall to keep it malleable for coming iterations.

test_runner/regress/test_history_size.py
```diff
@@ -0,0 +1,344 @@
...
+    assert layer_removed
+
+    env.pageserver.start()
+    main_pg.start()
+    first_branch_pg.start()
+
+    # was expecting that this timeline would be broken
+    with first_branch_pg.cursor() as cur:
+        with pytest.raises(Exception, match="""relation "t1" does not exist"""):
+            cur.execute("SELECT count(1) FROM t1")
```
koivunej, line 341: Included this test case for discussion and confirming that this is the expected behaviour. Removing the branched off timelines' layer file doesn't cause any failure, other than not being able to find the data when querying postgres.

I was suprised to find that there's no check for layer files being around up to disk_consistent_lsn, but then again, this is a contrived non-remote_storage test case, which should probably be removed?

pageserver/src/http/routes.rs
koivunej: After the `pricing_model` => `tenant_size` rename, this name changed to `size` I added a comment on it but now looking back, maybe you were thinking of encoding it to the name?

koivunej:
What a chance to win two flaky tests in both regression tests. Trying to fix that test.

koivunej:
Had to do another rebasing, added tests consume free ports.

pageserver/src/http/routes.rs
SomeoneToIgnore: > encoding it to the name

I think we've done it somewhere in the code already, but might be that the comment is also enough.

pageserver/src/tenant.rs
```diff
@@ -977,6 +984,67 @@ impl Tenant {
...
+
+            let result = timeline.gc()?;
+            totals += result;
+        }
+
+        totals.elapsed = now.elapsed();
+        Ok(totals)
+    }
+
+    /// Refreshes the Timeline::gc_info for all timelines or only the given timeline, returning the
```
SomeoneToIgnore, line 1025: > for all timelines or only the given timeline

Not true anymore.

test_runner/regress/test_tenant_size.py
```diff
@@ -0,0 +1,342 @@
...
+
+    with env.postgres.create_start("main", tenant_id=tenant_id) as pg:
+        with pg.cursor() as cur:
+            cur.execute("SELECT 1")
+            row = cur.fetchone()
+            assert row is not None
+            assert row[0] == 1
+        # running compute should not change the size
+        size = get_tenant_size(http_client, tenant_id)[0]
+        assert size == 0
```
SomeoneToIgnore, line 36: NIT

```suggestion
        assert size == 0, running compute should not change the size
```

Both Rust and Python provide a way to embed such comments into the assertions: if we keep them sane, it becomes a bit easier to search for the test that failed.

test_runner/regress/test_tenant_size.py
```diff
@@ -0,0 +1,342 @@
...
+
+    # after compute has been shut down, the size should be the same, until we
+    # increase the size over the gc_horizon
+    size = get_tenant_size(http_client, tenant_id)[0]
+    assert size == 0
+
+
+def get_tenant_size(http_client: PageserverHttpClient, tenant_id: TenantId) -> Tuple[int, Any]:
```
SomeoneToIgnore, line 44: We can move it into `PageserverHttpClient` directly, can't we?

test_runner/regress/test_tenant_size.py
```diff
@@ -0,0 +1,342 @@
...
+    for lsn, size in collected_responses:
+        log.info(f"collected: {lsn}, {size}")
+
+    env.pageserver.stop()
+    env.pageserver.start()
+
+    size_after = get_tenant_size(http_client, tenant_id)[0]
+
+    assert (
+        size_after == collected_responses[-1][1]
```
SomeoneToIgnore, line 159: NIT: feels more readable to extract `collected_responses[-1][1]` into a separate variable with the name.

test_runner/regress/test_tenant_size.py
```diff
@@ -0,0 +1,342 @@
...
+
+    wait_for_last_flush_lsn(env, main_pg, tenant_id, main_timeline_id)
+    size_at_branch = get_tenant_size(http_client, tenant_id)[0]
+    assert size_at_branch > 0
+
+    first_branch_timeline_id = env.neon_cli.create_branch(
+        "first-branch", main_branch_name, tenant_id
+    )
+
+    # unsure why this happens
```
SomeoneToIgnore, line 194: That would be interesting to know indeed, we have crated two branches, so we can check the size differences and guess better. Maybe the difference is not even a constant?

test_runner/regress/test_tenant_size.py
```diff
@@ -0,0 +1,342 @@
...
+            raise
+
+    assert deleted
+
+    size_after_deleting_first = get_tenant_size(http_client, tenant_id)[0]
+    assert size_after_deleting_first < size_after_thinning_branch
+
+    http_client.timeline_delete(tenant_id, second_branch_timeline_id)
+    size_after_deleting_second = get_tenant_size(http_client, tenant_id)[0]
+    assert size_after_deleting_second < size_after_deleting_first
```
SomeoneToIgnore, line 275: What would be the "reverted initial" size? Presumably, not 0 anymore, since `gc_cutoff_lsn` is not negative anymore?

test_runner/regress/test_tenant_size.py
```diff
@@ -0,0 +1,342 @@
...
+    main_pg.stop()
+
+    http_client.timeline_checkpoint(tenant_id, first_branch_timeline_id)
+
+    env.pageserver.stop()
+
+    layer_removed = False
+    for path in Path.iterdir(env.timeline_dir(tenant_id, first_branch_timeline_id)):
+        log.info(f"iterating timeline dir: {path.name}")
+        if not layer_removed and path.name.startswith("00000"):
```
SomeoneToIgnore, line 326: Turns out to be quite popular operation, we might consider adding an utility function for that (maybe later)

pageserver/src/http/openapi_spec.yml
```diff
@@ -354,6 +354,52 @@ paths:
...
+    parameters:
+      - name: tenant_id
+        in: path
+        required: true
+        schema:
+          type: string
+          format: hex
+    get:
+      description: |
+        Calculate tenant's size.
```
SomeoneToIgnore, line 367: Let's also mention what units the size has.

test_runner/regress/test_tenant_size.py
SomeoneToIgnore: Got interested myself

```
diff --git a/test_runner/regress/test_tenant_size.py b/test_runner/regress/test_tenant_size.py
index 25b0f478..d0690c5c 100644
--- a/test_runner/regress/test_tenant_size.py
+++ b/test_runner/regress/test_tenant_size.py
@@ -194,6 +194,7 @@ def test_get_tenant_size_with_multiple_branches(neon_env_builder: NeonEnvBuilder
     # unsure why this happens
     size_after_branch = get_tenant_size(http_client, tenant_id)[0]
     assert size_after_branch > size_at_branch
+    log.error(f'size_after_branch: {size_after_branch}, size_at_branch: {size_at_branch}, first diff: {size_after_branch - size_at_branch}')
 
     first_branch_pg = env.postgres.create_start("first-branch", tenant_id=tenant_id)
 
@@ -220,6 +221,7 @@ def test_get_tenant_size_with_multiple_branches(neon_env_builder: NeonEnvBuilder
     )
     size_after_branch = get_tenant_size(http_client, tenant_id)[0]
     assert size_after_branch > size_after_continuing_on_main
+    log.error(f'size_after_branch: {size_after_branch}, size_after_continuing_on_main: {size_after_continuing_on_main}, second diff: {size_after_branch - size_after_continuing_on_main}')
 
     second_branch_pg = env.postgres.create_start("second-branch", tenant_id=tenant_id)
 ```

and that shows the result

```
2022-11-02 22:01:16.430 ERROR [test_tenant_size.py:197] size_after_branch: 24556288, size_at_branch: 24425216, first diff: 131072
2022-11-02 22:01:19.218 ERROR [test_tenant_size.py:224] size_after_branch: 26719448, size_after_continuing_on_main: 26580104, second diff: 139344
```

the diff seems close, but odd that it's a bit different.
Also, I seem to have used proper variables to find the diff, is that `131072` _bytes_? Seems to be quite a difference..

File: test_runner/regress/test_tenant_size.py (hunk 44 above)
Comment (koivunej) [reply test_runner/regress/test_tenant_size.py:326]: Yes, but I probably need to remove this test because it doesn't fail right now, the timeline doesn't become broken. It just loses data.

File: test_runner/regress/test_tenant_size.py (hunk 42 above)
Comment (koivunej) [reply test_runner/regress/test_tenant_size.py:194]: 131072 with one of these tests sounds familiar, might be the gc horizon, or maybe there's some alignment going on with the wal? I didn't yet go through yesterdays successful testing run, which should have the outputs for single branch testing to compare.

File: test_runner/regress/test_tenant_size.py [hunk 46]
```diff
@@ -0,0 +1,342 @@
...
+
+        current_lsn = wait_for_last_flush_lsn(env, pg, tenant_id, timeline_id)
+
+        size, _ = get_tenant_size(http_client, tenant_id)
+
+        assert size > collected_responses[-1][1]
+        collected_responses.append((current_lsn, size))
+
+    for lsn, size in collected_responses:
+        log.info(f"collected: {lsn}, {size}")
```
koivunej, line 151: Both locally and remotely the values are stable in the environment, but the values between local and remote differ:

```
--- local-collected
+++ remote-collected
@@ -1,10 +1,10 @@
-0/16C1198, 23728536
-0/16C2DF8, 23735800
-0/16CAB48, 23767880
-0/16D2AC0, 23800512
-0/16D5838, 23812152
-0/16D97F0, 23828464
-0/16DB268, 23835240
-0/16DCCE8, 23842024
-0/16DE628, 23848488
-0/16E2C00, 23866368
+0/16BC518, 23708952
+0/16BE178, 23716216
+0/16C5EB0, 23748272
+0/16CDE28, 23789176
+0/16D2C08, 23800840
+0/16D6BC0, 23817152
+0/16D8638, 23823928
+0/16DA0B8, 23830712
+0/16DB9E0, 23837152
+0/16DDF68, 23846760
```

test_runner/regress/test_tenant_size.py
koivunej: The value matches the gc_horizon. I'll push a commit with assertion, lets see if it's stable.

test_runner/regress/test_tenant_size.py
koivunej: 23866368 - 23846760 = 19608; adding a comment on the log.info's about this.

test_runner/regress/test_tenant_size.py
koivunej: Did this, and will push next the `last_size` => `prev` as well.

test_runner/regress/test_tenant_size.py
koivunej: Moved in [670f8ec](https://github.com/neondatabase/neon/pull/2714/commits/670f8ece986288fbdfdd105a6751ed9138ee67ae)

test_runner/regress/test_tenant_size.py
koivunej: It did pass on both debug and release. Something to think about in future.

test_runner/regress/test_tenant_size.py
koivunej: Added assertions after refactoring not to reuse variables in the test: [90f9f38](https://github.com/neondatabase/neon/pull/2714/commits/90f9f385ad68e0a2df68e8f79287359c4ba55be5)

test_runner/regress/test_tenant_size.py
koivunej: Moved most of the comments as assertion messages.

pageserver/src/http/openapi_spec.yml
koivunej: In the next push, elaborated on the description and the actual field.

pageserver/src/tenant.rs
koivunej: Good catch! In the next push.

libs/tenant_size_model/src/lib.rs
```diff
@@ -0,0 +1,349 @@
+use std::borrow::Cow;
+use std::collections::HashMap;
+
+/// Pricing model or history size builder.
```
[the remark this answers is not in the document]
koivunej: https://github.com/neondatabase/neon/pull/2714/commits/f082c1090ee892666fa36813b6e7b3bae6217c34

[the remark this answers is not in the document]
koivunej: Still remains the `pricing` in libs/tenant_size_model/README.md header... I am inclined to keep it, because it should be revealed somewhere that this is related to pricing attempts.

pageserver/src/tenant.rs
koivunej: I'll record a `TODO:` above this, once I find your other comment about the incremental computation.

pageserver/src/tenant/pricing.rs
koivunej: Elaborated on the manual inspection hint.

libs/pricing_model/src/lib.rs
koivunej: Ended up with bunch of comparisons in the tests without any actual data, since CI and my local results differ.

libs/pricing_model/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+    pub method: SegmentMethod,
+
+    this_price: u64,
+
+    pub children: Vec<SegmentPrice>,
+}
+
+impl SegmentPrice {
+    fn total(&self) -> u64 {
+        self.this_price + self.children.iter().fold(0, |acc, x| acc + x.total())
```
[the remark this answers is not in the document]
koivunej: Opting not to touch the model yet. Apologies on repeating this comment all over.

libs/pricing_model/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+// 2. D+C+a+b
+// 3. D+A+B
+
+/// [`Segment`] which has had it's price calculated.
+pub struct SegmentPrice {
+    pub seg_id: usize,
+
+    pub method: SegmentMethod,
+
+    this_price: u64,
```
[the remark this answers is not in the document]
koivunej: Opting not to touch the model yet. Apologies on repeating this comment all over. This field is now `this_size` but that doesn't really answer the original point.

libs/pricing_model/src/lib.rs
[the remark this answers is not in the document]
koivunej: Added a comment about the size, but didn't touch this in favor of expediting merging.

libs/pricing_model/src/lib.rs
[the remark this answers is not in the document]
koivunej: So, workflows didn't start while there were conflicts but bunch of those workflows suffered from worker shortage yesterday.

> Also interested why this particular method is needed to be pub

This is because of using these from `main` which I think has "external" linkage, didn't try to hide the methods.

> could panic

This is unfortunate, but I'd rather not touch this further until we have some panic on our hands from staging.

libs/pricing_model/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+
+        // Phase 2: For each oldest segment in a chain that needs to be retained,
+        // calculate if we should store snapshot or WAL
+        self.price_from_snapshot_later(0)
+    }
+
+    fn price_from_wal(&self, seg_id: usize) -> SegmentPrice {
+        let seg = &self.segments[seg_id];
+
+        let this_price = seg.end_lsn - seg.start_lsn;
```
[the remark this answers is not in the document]
koivunej: Collecting this comment to the TODO about incremental computation above the single mutex.

pageserver/src/tenant.rs
```diff
@@ -72,6 +72,8 @@ pub mod storage_layer;
 
 mod timeline;
 
+pub mod pricing;
```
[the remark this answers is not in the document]
koivunej: Ended up renaming this all to `tenant_size`, `this_size`, `Storage::calculate` after internal discussion.

test_runner/regress/test_tenant_size.py
SomeoneToIgnore: I am not sure if that test should fail, actually.
Generally, the question about broken branches and their size tracking is interesting. 

I think we could discuss it better separately and the test could stay as a reminder, but no strong opinion.

pageserver/src/tenant/pricing.rs
koivunej: Initial size computetation is now used as a cache as well, optimistically, in https://github.com/neondatabase/neon/pull/2714/commits/2fdef1aa5887fecebe14fbece64b930e8e14d3af if we decide to keep it. Other thread where this was discussed more: https://github.com/neondatabase/neon/pull/2714#discussion_r1011694080

Note that the initdb_lsn is unlikely to be used here however.

Related note for anyone coming spelunking after, in tests we noted that the act of branching increases the ancestor timeline lsn, moving the branch point away (in tests amount of gc_horizon -- might be the miniscule gc_horizon only) off where it was thought to be, leading to new size recalculation.

test_runner/regress/test_history_size.py
koivunej: Removed, additional discussion: https://github.com/neondatabase/neon/pull/2714#discussion_r1012741091

test_runner/regress/test_tenant_size.py
SomeoneToIgnore: So, every branch that we create, increments the total `tenant_size` by `gc_horizon` value?
I would not expect this behavior, so seems like a bug? If you prefer to consider it later, let's add an issue or a TODO around.

----------
Maybe, due to `pricing` concept introduced initially, I'm more critical to such errors than usual: since it's the end user that get billed for these errors.
If I'm wrong with my assumptions, let's move on then.

pageserver/src/tenant.rs
koivunej: In next push, waiting for the e2e test run.

pageserver/src/tenant.rs
```diff
@@ -120,6 +122,9 @@ pub struct Tenant {
 
     /// Makes every timeline to backup their files to remote storage.
     upload_layers: bool,
+
+    /// Cached logical sizes updated updated on each [`Tenant::collect_pricing_operations`].
+    cached_logical_sizes: tokio::sync::Mutex<HashMap<(TimelineId, Lsn), u64>>,
```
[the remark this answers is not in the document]
koivunej: Included in the TODO comment, incrementalizing both the model update and size calcuations, on the site of the mutex locking. On the next push, after e2e tests pass.

test_runner/regress/test_tenant_size.py
koivunej: Noted in the TODO issue.

SomeoneToIgnore (approved):

PR #2719, opened 2022-10-28
Fix test_prepare_snapshot.

arssher:
It should checkpoint pageserver after waiting for all data arrival, not before.

reported by
https://github.com/neondatabase/neon/actions/runs/3346541946/jobs/5543616053


kelvich (approved):

bayandin:
Thanks @arssher

Issue #2697, opened 2022-10-26
Add CI tests for docker-compose example in Neon repo


SomeoneToIgnore:
https://github.com/neondatabase/neon/pull/2666 brought an initial version of the whole storage set-up in docker-compose.
To avoid breaking or accidentally obsoleting the set-up, we would need a set of CI tests. The PR already contains a script which we can adjust if needed and add to the pipeline to test the images that are built for every PR.

PR #2721, opened 2022-10-30
fix: path in scripts/docker-compose_test.sh missed '-'

NikeNano:
AWESOME with serverless postgres! 

When taking a look at https://github.com/neondatabase/neon/issues/2697 and playing around with neon to check it out I realised that the path in  docker-compose_test.sh where wrong.

### DIFF
--- a/scripts/docker-compose_test.sh
+++ b/scripts/docker-compose_test.sh
@@ -6,7 +6,7 @@ set -eux -o pipefail
 SCRIPT_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
 COMPOSE_FILE=$SCRIPT_DIR/../docker-compose/docker-compose.yml
 
-COMPUTE_CONTAINER_NAME=dockercompose_compute_1
+COMPUTE_CONTAINER_NAME=docker-compose_compute_1
 SQL="CREATE TABLE t(key int primary key, value text); insert into t values(1,1); select * from t;"
 PSQL_OPTION="-h localhost -U cloud_admin -p 55433 -c '$SQL' postgres"
 


NikeNano:
FYI: @ SomeoneToIgnore

SomeoneToIgnore:
Thank you for pointing out, I've noticed that but got distracted and never got to fix all things for tests there.
My fixes are here: https://github.com/neondatabase/neon/compare/kb/compose-tests?expand=1 and the container name is different from yours.

Please also note that `restart: always` is needed for compute, since otherwise it does not recover from any errors.

Yet, this is the container name I get on the linux machine based on the CI one, so I don't think your version is going to help us much, alas. 
I'll close this PR and use my branch to fix things eventually.

NikeNano:
> https://github.com/neondatabase/neon/compare/kb/compose-tests?expand=1

I think there is a difference between mac and Linux in this case for me I still get: 

```
Creating docker-compose_minio_1 ... done
Creating docker-compose_etcd_1  ... done
Creating docker-compose_minio_create_buckets_1 ... done
Creating docker-compose_safekeeper2_1          ... done
Creating docker-compose_safekeeper3_1          ... done
Creating docker-compose_safekeeper1_1          ... done
Creating docker-compose_pageserver_1           ... done
Creating docker-compose_compute_1              ... done
Creating docker-compose_compute_is_ready_1     ... done
```

I need to check further but might be that `--compatibility` [will fix it](https://github.com/docker/for-mac/issues/6035#issuecomment-970083140) or we state clearly that the expectation is Linux.

SomeoneToIgnore:
I think there's more differences, might be that `docker compose` is different from `docker-compose` in that regard.
We don't have the latter and will not install it because there's a `docker compose` one already present.

Here's what I see on the linux machine mentioned above, on my branch:
```
[+] Running 10/9
 ⠿ Container docker-compose-compute_is_ready-1      Removed                                                                                                                                                      0.0s
 ⠿ Container docker-compose-compute-1               Removed                                                                                                                                                     10.4s
 ⠿ Container docker-compose-safekeeper2-1           Removed                                                                                                                                                     10.6s
 ⠿ Container docker-compose-safekeeper1-1           Removed                                                                                                                                                     10.6s
 ⠿ Container docker-compose-safekeeper3-1           Removed                                                                                                                                                     10.6s
 ⠿ Container docker-compose-pageserver-1            Removed                                                                                                                                                     10.7s
 ⠿ Container docker-compose-minio_create_buckets-1  Removed                                                                                                                                                      0.0s
 ⠿ Container docker-compose-etcd-1                  Removed                                                                                                                                                      0.3s
 ⠿ Container docker-compose-minio-1                 Removed                                                                                                                                                      0.7s
 ⠿ Network docker-compose_default                   Removed                                                                                                                                                      0.1s
+ break



$ uname -a
Linux ip-172-31-9-105 5.10.0-14-cloud-amd64 #1 SMP Debian 5.10.113-1 (2022-04-29) x86_64 GNU/Linux
```

Mac naming seems to match that, might be also due to the same `docker compose` used there, but M1 mac test does not work for me at all, unfortunately:
```
+ docker exec -it docker-compose-compute-1 /bin/bash -c 'psql -h localhost -U cloud_admin -p 55433 -c '\''CREATE TABLE t(key int primary key, value text); insert into t values(1,1); select * from t;'\'' postgres'
ERROR:  could not read block 4 in rel 1663/12972/2704.0 from page server at lsn 0/016B8F40
LINE 1: CREATE TABLE t(key int primary key, value text); insert into...
```

that's caused by an error in pageserver:
```
docker-compose-pageserver-1  | 2022-10-31T11:01:55.290509Z ERROR handle_pagerequests{tenant_id=e316221f1ad84ced22ff6bf531638db8 timeline_id=bb83877422e1c6676fd95600ca27303a}:handle_get_page_at_lsn_request{rel=1663/12972/3455 blkno=4 req_lsn=0/16BC208}: wal-redo-postgres: 2022-10-31 11:01:55.289 UTC [332] FATAL:  seccomp: could not load test trap

```


We also leave all the containers running if any errors appears, btw, that also has to be fixed somehow..
But for now, as long as the tests are running on Linux for starters, we're good, mac is another issue.

NikeNano:
> We also leave all the containers running if any errors appears, btw, that also has to be fixed somehow.. But for now, as long as the tests are running on Linux for starters, we're good, mac is another issue.

Make sense :)

PR #2723, opened 2022-10-31
Nightly Benchmarks: increase timeout for perf tests

bayandin:
Suddenly `test_hot_page[remote].write` and `test_hot_table[remote].write` started to take ~ 115 minutes (instead of 75 minutes), which caused our `bench` job to fail. This PR increases a timeout to let these tests pass and collect results.

We have a discussion on internal slack: https://neondb.slack.com/archives/C033RQ5SPDH/p1667226785865769 (I will create a new issue, if we won't figure it out fast)


bayandin:
The job has recovered: https://github.com/neondatabase/neon/actions/runs/3366657622/jobs/5583365208

PR #2724, opened 2022-10-31
Refactoring: replace `utils::connstring` with `Url`-based APIs

funbringer:


pageserver/src/walreceiver/connection_manager.rs
```diff
@@ -836,12 +836,13 @@ fn wal_stream_connection_string(
...
-        .parse::<postgres::config::Config>()
-        .with_context(|| {
-            format!("Failed to parse pageserver connection string '{sk_connstr}' as a postgres one")
-        })?;
-    let (host, port) = utils::connstring::connection_host_port(&me_conf);
+    let sk_connurl: url::Url = sk_connstr
+        .parse()
+        .with_context(|| format!("Failed to parse pageserver connection URL '{sk_connstr}'"))?;
+
+    let host = sk_connurl.host_str().context("host is missing")?;
```
SomeoneToIgnore, line 843: NIT: could add more context and print the connection string that caused the error in the message too, since we do not print it anywhere else (as it was done before the refactoring):

>  format!("Failed to parse pageserver connection string '{sk_connstr}' as a postgres one")

SomeoneToIgnore (approved):

pageserver/src/walreceiver/connection_manager.rs
funbringer: Done, thanks.

PR #2730, opened 2022-11-01
pageserver: make wal_source_connstring: String a separate type

yeputons:
* Now it is `wal_source: tokio_postgres::Config`.
* In the future the connection string will include password, which should not be dumped into logs. Using a structured type allows such customization: `tokio_postgres::Config` does not implement `Display`, and its `Debug` implementation hides the password.
* Performance hits are not measured, but we rarely manipulate configs.

Kind of continuing #2724

### DIFF
--- a/pageserver/src/tenant/timeline.rs
+++ b/pageserver/src/tenant/timeline.rs
@@ -48,6 +48,7 @@ use utils::{
     simple_rcu::{Rcu, RcuReadGuard},
 };
 
+use crate::connection_string::ConnectionString;
 use crate::repository::GcResult;
 use crate::repository::{Key, Value};
 use crate::task_mgr;
@@ -288,7 +289,7 @@ impl LogicalSize {
 }
 
 pub struct WalReceiverInfo {
-    pub wal_source_connstr: String,
+    pub wal_source_connstr: ConnectionString,
     pub last_received_msg_lsn: Lsn,
     pub last_received_msg_ts: u128,
 }
--- a/pageserver/src/walreceiver/connection_manager.rs
+++ b/pageserver/src/walreceiver/connection_manager.rs
@@ -31,6 +31,7 @@ use pageserver_api::models::TimelineState;
 use tokio::{select, sync::watch};
 use tracing::*;
 
+use crate::connection_string::ConnectionString;
 use crate::{
     exponential_backoff, walreceiver::get_etcd_client, DEFAULT_BASE_BACKOFF_SECONDS,
     DEFAULT_MAX_BACKOFF_SECONDS,
@@ -425,7 +426,11 @@ impl WalreceiverState {
     }
 
     /// Shuts down the current connection (if any) and immediately starts another one with the given connection string.
-    async fn change_connection(&mut self, new_sk_id: NodeId, new_wal_source_connstr: String) {
+    async fn change_connection(
+        &mut self,
+        new_sk_id: NodeId,
+        new_wal_source_connstr: ConnectionString,
+    ) {
         self.drop_old_connection(true).await;
 
         let id = self.id;
@@ -726,7 +731,7 @@ impl WalreceiverState {
     fn select_connection_candidate(
         &self,
         node_to_omit: Option<NodeId>,
-    ) -> Option<(NodeId, &SkTimelineInfo, String)> {
+    ) -> Option<(NodeId, &SkTimelineInfo, ConnectionString)> {
         self.applicable_connection_candidates()
             .filter(|&(sk_id, _, _)| Some(sk_id) != node_to_omit)
             .max_by_key(|(_, info, _)| info.commit_lsn)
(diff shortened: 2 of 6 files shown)


pageserver/src/tenant/timeline.rs
```diff
@@ -275,7 +275,7 @@ impl LogicalSize {
 }
 
 pub struct WalReceiverInfo {
-    pub wal_source_connstr: String,
+    pub wal_source: tokio_postgres::Config,
```
funbringer, line 278: To be honest, I don't think this is a good substitute. The downsides of this change are:

* The struct itself is bloated (it has multiple vectors) and contains many things not needed in such a simple use case.
* It does some magic that's uncalled for. For instance, it will set a default port 5432 if you don't provide one. This is fine when you're connecting to a regular postgres, but it's anything but useful when you're connecting to something bespoke, because who in their right mind would use a default port in that case?
* It gives you more flexibility than needed, like the ability to set multiple hosts and ports. **It will no longer be possible to extract a connection string** in a straightforward manner (we used to have a hack for that which [I eliminated yesterday](https://github.com/neondatabase/neon/commit/0df3467146f9368d67a145c59949e7f0ee6c01c9)). Furthermore, it will be possible to abuse this and create broken configurations with e.g. 2 hosts and 3 ports which the library can't handle.
* This is an implementation detail that would be hard to get rid of if we ever decide to replace tokio-postgres with anything else (this is not as hypothetical as you might think).

---

In my opinion, it'd be better to create a newtype like this:

```rust
struct ConnectionUrl(url: url::Url);

impl ConnectionUrl {
  fn as_str(&self) -> &str { self.url.as_str() }
}

// Do the validation here.
impl std::str::FromStr for ConnectionUrl { ... }
```

You could reuse [this](https://github.com/neondatabase/neon/blob/0df3467146f9368d67a145c59949e7f0ee6c01c9/control_plane/src/connection.rs#L4).

yeputons:
After some consideration I concluded that I don't like `Url` either. Reasons are mostly the same: `Url` is hard to validate, it can contain much more that our connection string actually needs in this case. But the main problem is postgres connection string looks awful in URL.

As we don't manipulate connection strings in Pageserver much (only one place for creation and one place for consumption), I'd continue the trend with simple strings.

@funbringer WDYT?

funbringer:
I agree that whether `Url` is suitable or not hugely depends on the context. But the point is, I didn't suggest to use bare `Url`: the key idea is to create your own type and be in full control of what's inside. If you like strings for storing the underlying info: go ahead and use them, I don't mind! Just make sure they're wrapped in a 1st class type with proper parsing, validation and accessors.

yeputons:
> PS: it just so happens -- at least in my practice -- that strings fall short as soon as you try to add more sophisticated accessors. 

Agree. I think in this specific case we don't need any accessors at all. Create _something_ in one place (from structured data), alter it slightly and use immediately in another place.

The only reason for this PR is to sanitize logs. Making the next step and actually storing connection parameters separately is not worth it at this moment, I think, we only have two places that handle them: one creates connection string, and one uses. A proper (de)serializer will take another 50-100 lines of code (including tests), and it won't make the code easier. At least now.

pageserver/src/walreceiver/connection_manager.rs
```diff
@@ -843,11 +850,15 @@ fn wal_stream_connection_string(
...
+            let no_secrets = format!(
                 "host={host} \
                  port={port} \
                  options='-c timeline_id={timeline_id} tenant_id={tenant_id}'"
-            ))
+            );
+            Ok(ConnectionString {
+                no_secrets: no_secrets.clone(),
+                with_secrets: no_secrets,
+            })
```
funbringer, line 861: It looks like at the moment there's no real difference between `no_secrets` and `with_secrets`. I'd advise to stick with YAGNI and add another field once we introduce at least one secret that shouldn't be printed. Until then, you could add helper TODO/XXX/NOTE comments with the reasoning.

LizardWizzard (commented):
To me having something other than just a string is a big plus. Url is less error prone because it will fail validation in more obvious cases when tokio_postgres::Config will simply accept wrong input. As @funbringer discovered it will accept hostname with a whitespace in it for example. So to me plain string is definitely a step back from Url guarantees, yet plain string looks nicer in logs.

>A proper (de)serializer will take another 50-100 lines of code (including tests), and it won't make the code easier.

Having a small newtype wrapper in self contained module wont make it harder, I think the opposite. With it you dont have to guess what the structure is, there will be less room for confusion and unexpected errors. IMO newtype paterrn fits really well here. Internal Url (or something else in the future) is private and combination of parsing and accessor functions ensure that it stays well-formed and wont break ones expectations. Proper debug/display impls can make it both credentials-free and human friendly.

IMO Price of 100-150 lines is low enough compared to safety benefits it brings. 

I would say this should build upon existing PgConnectionConfig. Having more than one way to describe connection strings in our codebase wont make things easier. (Actually one can still use definitions from other crates, but still with this PR we end up with two definitions in our code, PgConnectionConfig and ConnectionString)

PR #2733, opened 2022-11-02
Replace etcd with custom neon_broker.

arssher:
It is a simple and fast pub-sub message bus based on tonic. In this patch only safekeeper message is supported, but others can be easily added.

Compilation now requires protoc to be installed. Installing protobuf-compiler package is fine for Debian/Ubuntu.


arssher:
Core is there and suitable for review, tests should pass, but some ironing/todos are still in order:
- [ ] decent logging in broker;
- [ ] basic metrics in broker;
- [ ] sort out deps tree;
- [ ] update readmes/docs

broker/proto/neon_broker.rs
```diff
@@ -0,0 +1,383 @@
...
+pub struct TenantTimelineId {
+    #[prost(bytes="vec", tag="1")]
+    pub tenant_id: ::prost::alloc::vec::Vec<u8>,
+    #[prost(bytes="vec", tag="2")]
+    pub timeline_id: ::prost::alloc::vec::Vec<u8>,
+}
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct Empty {
+}
+/// Generated client implementations.
```
SomeoneToIgnore, line 57: Curious, how it gets [re]generated? Could add some module docs for that.

broker/src/bench.rs
```diff
@@ -0,0 +1,175 @@
...
+            }
+            yield info;
+            time::sleep(Duration::from_millis(1000)).await;
+        }
+    };
+    let response = client.publish_safekeeper_info(Request::new(outbound)).await;
+    println!("pub response is {:?}", response);
+}
+
+#[tokio::main]
```
SomeoneToIgnore, line 149: Feels it could be a separate `bin/` module?
Or we could try to move it under `benches/` directory and exclude from the regular compilation cycle entirely (since we do not plan running it regularly)

pageserver/src/walreceiver.rs
```diff
@@ -27,56 +27,55 @@ use crate::config::PageServerConf;
...
-        .context("Failed to connect to etcd")?;
-
-    // FIXME: Should we still allow the pageserver to start, if etcd
-    // doesn't work? It could still serve GetPage requests, with the
-    // data it has locally and from what it can download from remote
-    // storage
-    if ETCD_CLIENT.set(etcd_client).is_err() {
-        panic!("etcd already initialized");
+        .context(format!(
+            "Failed to connect to broker at {}",
```
SomeoneToIgnore, line 55: That's not true anymore, since we connect lazily?

Also, is it really beneficial to connect lazily? This way we at least fail fast and bring down the entire pageserver with us, while with the new scenario, we could start and pretend to be working and then start to fail in walreceiver, which is harder to test and understand (I would imagine, the client would see some WAL timeout error, as usual, if this happens?)

broker/src/broker.rs
```diff
@@ -0,0 +1,515 @@
...
+        let sub_key = SubscriptionKey::from_proto_subscription_key(proto_key)?;
+        let mut subscriber = self.registry.register_subscriber(sub_key).await;
+
+        // transform rx into stream with item = Result, as method result demands
+        let output = async_stream::try_stream! {
+            while let Some(info) = subscriber.sub_rx.recv().await {
+                    yield info
+                }
+
+            // internal generator
```
SomeoneToIgnore, line 449: ?

broker/src/broker.rs
```diff
@@ -0,0 +1,515 @@
...
+
+#[derive(Parser, Debug)]
+#[command(version, about = "Broker for neon storage nodes communication", long_about = None)]
+struct Args {
+    /// Endpoint to listen on.
+    #[arg(short, long, default_value = DEFAULT_LISTEN_ADDR)]
+    listen_addr: SocketAddr,
+}
+
+#[tokio::main]
```
SomeoneToIgnore, line 487: If this is a binary crate, let's move it under `src/bin/`?

broker/src/broker.rs
```diff
@@ -0,0 +1,515 @@
...
+            }
+            SubscriptionKey::Timeline(ref ttid) => {
+                remove_sub(&mut self.subs_to_timelines, ttid, sub_id);
+                Some(SubAnnounce::RemoveTimeline(*ttid, sub_id))
+            }
+        };
+        announce.map(|a| (self.pub_txs.values().cloned().collect(), a))
+    }
+
+    pub fn report(&mut self) {
```
SomeoneToIgnore, line 203: Should we use some logging here instead of just prints?
Also, plenty of prints in this module are commented out, that's odd.

broker/src/broker.rs
```diff
@@ -0,0 +1,515 @@
...
+        })),
+    };
+    let neon_broker_service = NeonBrokerService {
+        registry: registry.clone(),
+    };
+
+    tokio::spawn(async move { registry.report().await });
+
+    Server::builder()
+        .http2_keepalive_interval(Some(Duration::from_millis(5000)))
```
SomeoneToIgnore, line 509: I propose to put both report and this timeouts/intervals and the queue size into `Args`.

broker/src/broker.rs
```diff
@@ -0,0 +1,515 @@
...
+            subs_to_all: HashMap::new(),
+            subs_to_timelines: HashMap::new(),
+            next_sub_id: 0,
+        })),
+    };
+    let neon_broker_service = NeonBrokerService {
+        registry: registry.clone(),
+    };
+
+    tokio::spawn(async move { registry.report().await });
```
SomeoneToIgnore, line 506: Technically, this could fail (panic) and we would never know about it now. 
Should we handle the `Handle` too, in `select!` with the server?

broker/src/broker.rs
```diff
@@ -0,0 +1,515 @@
...
+        let sub_id = self.next_sub_id;
+        self.next_sub_id += 1;
+        let announce = match sub_key {
+            SubscriptionKey::All => {
+                assert!(!self.subs_to_all.contains_key(&sub_id));
+                self.subs_to_all.insert(sub_id, sub_tx.clone());
+                SubAnnounce::AddAll(sub_tx)
+            }
+            SubscriptionKey::Timeline(ttid) => {
+                match self.subs_to_timelines.entry(ttid) {
```
SomeoneToIgnore, line 151: NIT: feels like `.or_default().push(` could work here instead of `match`

broker/src/broker.rs
```diff
@@ -0,0 +1,515 @@
...
+    }
+
+    // Register new subscriber in shared state.
+    pub async fn register_subscriber(&self, sub_key: SubscriptionKey) -> Subscriber {
+        let (tx, rx) = mpsc::channel(CHAN_SIZE);
+        let id;
+        let mut pub_txs;
+        let announce;
+        {
+            let mut ss = self.shared_state.lock().unwrap();
```
SomeoneToIgnore, line 255: Does it actually make any sense to use `Mutex` from `std`? I see plenty of async code around it, feels like the `tokio` one could be more suitable?

SomeoneToIgnore (commented):
Not entirely sure I've understood `broker.rs` properly, it's a bit tangled, but as long as it works, I have plenty of opportunities to do that later.

SomeoneToIgnore:
We now have a docker-compose file with etcd in `neon` repo after https://github.com/neondatabase/neon/pull/2666
It needs a patch to run properly: https://github.com/neondatabase/neon/pull/2721#issuecomment-1296832372
but otherwise is pretty functional.

Could you also adjust it to use the new binary?

broker/build.rs
```diff
@@ -0,0 +1,7 @@
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // tonic_build::compile_protos("proto/broker.proto")?;
```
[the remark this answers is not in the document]
arssher: True, I only mentioned that in commit msg so far.

control_plane/src/local_env.rs
```diff
@@ -78,67 +78,26 @@ pub struct LocalEnv {
...
-    #[serde_as(as = "Vec<DisplayFromStr>")]
-    pub broker_endpoints: Vec<Url>,
-
-    /// Etcd binary path to use.
-    #[serde(default)]
-    pub etcd_binary_path: PathBuf,
+#[serde(default)]
+pub struct NeonBroker {
+    /// Broker listen address for storage nodes coordination, e.g. '127.0.0.1:50051'.
+    pub listen_addr: String,
```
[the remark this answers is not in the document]
arssher: I thought of that, but specifying URL as listen addr is slightly weird -- meaning of schema is unclear. I don't have strong opinion, if you think that is better, I will do that.

control_plane/src/local_env.rs
```diff
@@ -78,67 +78,26 @@ pub struct LocalEnv {
...
-                }
-            })
-            .fold(String::new(), |mut comma_separated_urls, url| {
-                if !comma_separated_urls.is_empty() {
-                    comma_separated_urls.push(',');
-                }
-                comma_separated_urls.push_str(url);
-                comma_separated_urls
-            })
+// Dummy, for automatic Deserialize impl.
```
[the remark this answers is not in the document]
arssher: Well, I also think we don't really need it (that's why it is dummy), but `Deserialize` derive doesn't work unless Default impl is provided, and we do need Deserialize.

broker/proto/neon_broker.rs
arssher: Ouch, I forgot to add this file to .gitignore :) build.rs generates it.

safekeeper/Cargo.toml
```diff
@@ -11,7 +11,7 @@ hyper = "0.14"
 fs2 = "0.4.3"
 serde_json = "1"
 tracing = "0.1.27"
-clap = "4.0"
+clap = { version = "4.0", features = ["derive"] }
```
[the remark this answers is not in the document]
arssher: broker uses it, and it is better to compile crate once for the project instead of twice.

broker/src/broker.rs
arssher: Sure, that's in todo list.

broker/src/broker.rs
arssher: There is no await inside, so std is fine.

safekeeper/Cargo.toml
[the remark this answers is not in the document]
SomeoneToIgnore: I suggest trying the workspace hack thing (`hakari`), this is exactly what is is for: you can leave the "regular" clap definitions in every crate and hakari will pick the least common subset needed to compile the crate only once.

broker/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+futures-core = "0.3"
+futures-util = "0.3"
+tonic = "0.8"
+prost = "0.11"
+tokio = { version = "1.0", features = ["macros", "rt-multi-thread"] }
+# for exploring with tokio-console
+# tokio = { version = "1", features = ["full", "tracing"] }
+# console-subscriber = "0.1.8"
+tokio-stream = "0.1"
+clap = { version = "4.0", features = ["derive", "string"] }
```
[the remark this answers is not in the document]
arssher: TIL, thanks.

broker/src/broker.rs
arssher: I used report() for debugging, but I think normally it should be replaced with metrics or some status requests, so improving report itself is not very valuable.

pageserver/src/walreceiver/connection_manager.rs
```diff
@@ -393,10 +344,8 @@ struct RetryInfo {
 
 /// Data about the timeline to connect to, received from etcd.
 #[derive(Debug)]
-struct EtcdSkTimeline {
-    timeline: SkTimelineInfo,
-    /// Etcd generation, the bigger it is, the more up to date the timeline data is.
-    etcd_version: i64,
```
SomeoneToIgnore, line 399: Almost forgot to ask: before, this allowed us to get the latest timeline state out of the queue.
I did not notice anything similar in the broker code, so I wonder, how realistic is the situation when we start to "lag" on the receiving side (presumably, could happen since we're distracting on processing other signals from walreceiver)?
If that could happen, we could consider the node to be alive more than it actually is and get more latency.

control_plane/src/local_env.rs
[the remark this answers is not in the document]
funbringer: `url` could be a good start, but we could as well use any newtype implementing a strict parser. The point is to leave less room for interpretation and subsequent errors.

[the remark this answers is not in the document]
arssher: > Actually, I think SocketAddr could be the best fit here.

It is definitely better than String, yes. But then it is different from client endpoints, which must include http:// , as Kirill pointed out....

pageserver/src/walreceiver.rs
arssher: We actually always connected lazily. Initial client creation only checked that endpoints are well formed -- and this is still the case, I believe. And I do think that's the right way -- requirement for broker to be available when pageserver starts is obscure and not needed. For example, currently you won't be even able to run tests, as we start pageserver first and broker second.

libs/safekeeper_api/src/models.rs
```diff
@@ -22,3 +22,36 @@ pub struct TimelineCreateRequest {
     // If not passed, it is assigned to the beginning of commit_lsn segment.
     pub local_start_lsn: Option<Lsn>,
 }
+
+/// Data about safekeeper's timeline. Fields made optional for easy migrations.
+#[serde_as]
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct SkTimelineInfo {
+    /// Term of the last entry.
+    pub last_log_term: Option<u64>,
```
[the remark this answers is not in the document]
arssher: Aligned them. Not sure about moving to the broker module; from one hand it is indeed mirrioring proto struct, on the other this is safekeeper-specific endpoints with narrow (manual intervention) scope...

broker/src/bench.rs
arssher: It is already excluded, it is under feature flag which must be explicitly set to get it built. Not sure whether common `benches/` is better...

broker/src/broker.rs
arssher: I used this for benchmarking subscribers only, so that on subscribe() endless stream is generated in the broker. It is trivial to add back, removed it.

pageserver/src/walreceiver.rs
```diff
@@ -27,56 +27,55 @@ use crate::config::PageServerConf;
...
 ///
-pub async fn init_etcd_client(conf: &'static PageServerConf) -> anyhow::Result<()> {
-    let etcd_endpoints = conf.broker_endpoints.clone();
+pub async fn init_broker_client(conf: &'static PageServerConf) -> anyhow::Result<()> {
+    let broker_endpoints = conf.broker_endpoints.clone();
     ensure!(
-        !etcd_endpoints.is_empty(),
-        "Cannot start wal receiver: etcd endpoints are empty"
+        !broker_endpoints.is_empty(),
+        "Cannot start wal receiver: broker endpoints are empty"
```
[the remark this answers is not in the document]
SomeoneToIgnore: Trying to understand what's up with the "empty string pattern" and why cannot we avoid it?
So far, my understanding is we're not going to work without broker, since no wall arrives and we do not have the readonly replicas still.
When we will, they might need some broker sync either, as we've discussed in a recent storage sync RFC.

So, feels like we don't really need to turn the broker off and there should be a hard error if it's missing? If so, then an empty string is not really nice IMO and we should always have some url and try to reconnect it endlessly.

If it's not true and we might want to start storage binaries without the broker, we can alter the toml parsing code and make the `broker_url(s)` non-mandatory? So we parse it into some `Option<Url>` or `None` and react accordingly.

If that's not working for some reason and we have to parse/pass some value related to the endpoint into config, I'd vote for keeping an array then, since we could pass an empty array instead of an empty string.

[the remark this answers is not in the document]
arssher: > my understanding is we're not going to work without broker, since no wall arrives and we do not have the readonly replicas still.

Generally true, but you might imagine a pageserver serving read only replica at fixed LSN, then WAL stream is not needed.

> non-mandatory? So we parse it into some Option<Url> or None and react accordingly.

Yes, but in this case default value localhost:50051 seems to me more useful than None -- but there is no way to explicitly pass None except as empty string (in which case you don't need Option).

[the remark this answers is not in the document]
SomeoneToIgnore: > but there is no way to explicitly pass None except as empty string

I was imagining, that if we don't pass anything, that would be the real `None`, but I think I am spending your time on this minor thing for too much and not understanding something.

So, if my proposals don't work, let's move on 🙂

petuhovskiy (commented):
Looked through the broker code, looks mostly good, but not finished with review yet.

Logging should be improved as written in [follow-ups], I think logging all [start/finish]x[subscribe/publish] events should be enough for the initial logging support.

Building docker image is 100% required, otherwise e2e tests from cloud repo will stop working.

We definitely should think how to deploy it with minimal downtime. I'm not sure if it's a good idea to break compatibility in a single commit, a lot of things can go wrong.

I liked the proposal at the meeting to split PR into:
1) PR with the code for the broker service
2) PR for everything else

Let's do that? Would be good to deploy broker service in all our environments first, and do the switch in the next release, when we verified the broker is working and accessible from storage instances.

petuhovskiy:
`neon_broker.rs` implementation with channels actually looks a bit more complicated than I expected. I would have written it without channels and would store all subscribers in `RwMutex`, something like this:
```
subs_to_all: RwLock<HashMap<SubId, Sender<SafekeeperTimelineInfo>>>,
subs_to_timelines: RwLock<HashMap<TenantTimelineId, Vec<SubSender>>>,
```

When there are no connects and disconnects, all threads are not blocked on anything and there is no overhead.

I'll do several more iterations on reading `neon_broker.rs`, maybe it will become less complicated after that.

arssher:
> `neon_broker.rs` implementation with channels actually looks a bit more complicated than I expected. I would have written it without channels and would store all subscribers in `RwMutex`, something like this:
> 
> ```
> subs_to_all: RwLock<HashMap<SubId, Sender<SafekeeperTimelineInfo>>>,
> subs_to_timelines: RwLock<HashMap<TenantTimelineId, Vec<SubSender>>>,
> ```

Hmm, indeed, that would be simpler and hardly worse in terms of concurrency.

SomeoneToIgnore:
Prepared compose example tests: https://github.com/neondatabase/neon/pull/2767
Now they work and only will need the new binary adjustments.

broker/proto/broker.proto
```diff
@@ -0,0 +1,45 @@
...
+    // A connection string to use for WAL receiving.
+    string safekeeper_connstr = 10;
+}
+
+message TenantTimelineId {
+    bytes tenant_id = 1;
+    bytes timeline_id = 2;
+}
+
+message Empty {
```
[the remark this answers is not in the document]
arssher: nice, switched to it

broker/src/bin/neon_broker.rs
```diff
@@ -0,0 +1,503 @@
...
+        })),
+    };
+    let neon_broker_service = NeonBrokerService {
+        registry: registry.clone(),
+    };
+
+    tokio::spawn(async move { registry.report().await });
+
+    Server::builder()
+        .http2_keepalive_interval(Some(Duration::from_millis(5000)))
```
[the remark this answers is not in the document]
arssher: oversight, thanks

broker/proto/broker.proto
```diff
@@ -0,0 +1,45 @@
+syntax = "proto3";
+
+package neon_broker;
+
+service NeonBroker {
+    // Subscribe to safekeeper updates.
+    rpc SubscribeSafekeeperInfo(SubscribeSafekeeperInfoRequest) returns (stream SafekeeperTimelineInfo) {};
+
+    // Publish safekeeper updates.
+    rpc PublishSafekeeperInfo(stream SafekeeperTimelineInfo) returns (Empty) {};
```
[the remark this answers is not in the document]
arssher: It would wait for answer before .await on sending message resolves then.

broker/proto/broker.proto
```diff
@@ -0,0 +1,45 @@
...
+    uint64 local_start_lsn = 9;
+    // A connection string to use for WAL receiving.
+    string safekeeper_connstr = 10;
+}
+
+message TenantTimelineId {
+    bytes tenant_id = 1;
+    bytes timeline_id = 2;
+}
+
```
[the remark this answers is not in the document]
arssher: Yes. Also, likely these types would still be different from our internal `Lsn` and stuff (though we can add Into/From, yes), which is unpleasant.

broker/proto/broker.proto
```diff
@@ -0,0 +1,45 @@
+syntax = "proto3";
+
+package neon_broker;
+
+service NeonBroker {
+    // Subscribe to safekeeper updates.
+    rpc SubscribeSafekeeperInfo(SubscribeSafekeeperInfoRequest) returns (stream SafekeeperTimelineInfo) {};
```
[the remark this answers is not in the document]
arssher: > This doesn't scale well, because it's not convenient to keep N (number of types) active subscription streams for different types,

It seems okay to me given that underlying code is the same (so need to replace SafekeeperTimelineInfo with template argument), not too much boilerplate and have type safety. But what you are describing is also possible.

petuhovskiy:
> I liked the proposal at the meeting to split PR into:
> 
>     PR with the code for the broker service
>     PR for everything else
> 
> Let's do that? Would be good to deploy broker service in all our environments first, and do the switch in the next release, when we verified the broker is working and accessible from storage instances.

So, what do you think about splitting this PR? I would not want to break anything until we deploy the broker to all regions. That means that we can postpone merging code that removes etcd support to main, by extracting it to another PR.

arssher:
> That means that we can postpone merging code that removes etcd support to main, by extracting it to another PR.

Yes, most likely that would be the way.

Right now I'm trying to add metrics on the same port, it is not, well, straightforward.

arssher:
Added metrics on the same port.

broker/src/bin/neon_broker.rs
```diff
@@ -0,0 +1,563 @@
...
+        let subscriber = Subscriber {
+            id,
+            key: sub_key,
+            sub_rx: rx,
+            registry: self.clone(),
+        };
+        // Notify existing publishers about new subscriber.
+        for pub_tx in pub_txs.iter_mut() {
+            // Closed channel is fine; it means publisher has gone.
+            pub_tx.send(announce.clone()).await.ok();
```
SomeoneToIgnore, line 296: This and next `.await`'s are done in a loop on bounded (256 messages is the default size) channel that is being emptied by the receiver (if I've got that right, the receiver end goes into `while let Some(info) = subscriber.sub_rx.recv().await {`).


Since we don't guarantee what happens on the client side, there could be situations where a client opens connection and holds it more than 256 seconds due to some bug: maybe, it's slowly receiving WAL and spends resources there.

Then, every such await will wait for some time, not allowing the rest of the `pub_tx` to receive their announcements faster.

Instead, I think it's better to submit such tasks into `tokio::task::JoinSet` or `FuturesUnordered`/`FuturesOrdered` and .await on that concurrently.

-----

This is somewhat related to https://github.com/neondatabase/neon/pull/2733#discussion_r1012156020
since we don't really have to hold that many messages per channel for our tasks, IMO: we use `etcd` as a "blackboard", where we saw the latest metadata on the topic when we requested it, not necessarily seeing every update to the metadata.
In the old code, we did skip towards the latest version of the message per timeline in walreceiver, and only then applied some logic on top of the latest message.

broker/src/bin/neon_broker.rs
```diff
@@ -0,0 +1,563 @@
...
+// SharedState wrapper for post-locking operations (sending to pub_tx chans).
+#[derive(Clone)]
+struct Registry {
+    shared_state: Arc<Mutex<SharedState>>,
+}
+
+impl Registry {
+    // Register new publisher in shared state.
+    pub fn register_publisher(&self) -> Publisher {
+        let (announce_tx, announce_rx) = mpsc::channel(128);
```
SomeoneToIgnore, line 247: > 128

Could that be configured too? Trying to understand what does it limit and can we somehow hit that limit during our regular work, as with the 256 one above.

SomeoneToIgnore (commented):
I'm still somewhat concerned about the subscription semantics change https://github.com/neondatabase/neon/pull/2733#discussion_r1012156020
as noted something related to the channel sizes in this review.

Otherwise, the PR became cleaner, the only surprising thing was the code that misused the `broker_endpoint=[]` strings and seemed not to break any tests? See docker compose yml file for example, Docker tests seem to work despite that?

Interesting, that we don't have `error!` or `warn!` logs in the whole project. Are there really zero bad edge cases?

broker/src/bin/bench.rs
```diff
@@ -0,0 +1,175 @@
...
+    };
+    let response = client.publish_safekeeper_info(Request::new(outbound)).await;
+    println!("pub response is {:?}", response);
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let args = Args::parse();
+
+    let mut counters = Vec::with_capacity(args.num_subs as usize);
```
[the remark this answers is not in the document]
sharnoff: I don't think that's equivalent: as written, each counter is distinct, but `vec![]` uses `Clone` to repeat the expression.

pageserver/src/walreceiver/connection_manager.rs
arssher: In theory it is possible, but in practice I don't think so. The cure for that could be comparing timestamps with sender, but this needs synchronized clocks and more code. However, reading the queue is cheap; so in such situation I think considering nodes too alive is a minor problem compared to general node overload.

pageserver/Cargo.toml
```diff
@@ -50,6 +50,7 @@ signal-hook = "0.3.10"
 svg_fmt = "0.4.1"
 tar = "0.4.33"
 thiserror = "1.0"
+tonic = "0.8"
```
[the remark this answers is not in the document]
arssher: good idea, will try

libs/safekeeper_api/src/models.rs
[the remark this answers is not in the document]
arssher: I think so. I mean, we either should switch everything to optional or leave as is; but to me current situation when default value is clearly not valid is fine and removes sensible amount of boilerplate.

broker/src/bin/bench.rs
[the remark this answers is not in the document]
arssher: I stepped on that actually =)

broker/src/bin/neon_broker.rs
arssher: This particular send doesn't carry data, it just notifies publishers about the new subscriber (happens only on new subscription event).

The whole thing is designed not to block on overflow. If publisher can't send to subscriber, it drops the message. Add of new subs and drop of old ones is a rare operation, but anyway publisher is expected to consume these notifications quickly as this requires nothing but a bit of CPU.

pageserver/src/walreceiver/connection_manager.rs
SomeoneToIgnore: > comparing timestamps

Curious, why similar `broker_version: u64` could not be used? For our needs, that was very simple to use at least.
But sure if that's hard to do now, maybe not needed.

My general concern after all recent discussions is that we might get these corner cases in theory, but do not make them very visible?
For example, as I've learned [recently](https://github.com/neondatabase/neon/pull/2733#discussion_r1022586630) 

> If publisher can't send to subscriber, it drops the message. 

that could worsen the case of receiving some old data about the service from the broker subscription: our pageserver walreceiver heuristics would consider nodes alive more than needed, let's at least adjust those `trace!` messages to `warn!` (`error!`?) ones to be scrapped by default.

I think it would be also nice metrics on that and more logic around the `etcd_version: i64` and "skipping" somewhere, as it does not seem hard to do compared to the rest of the heavy-lifting 🙂 
But as long as we can notice such abnormalities, it's a good start already.

arssher: > those trace! messages to warn!

I was somewhat barred by possibility of overflowing logs, but ok, let's do first version with that.

> why similar broker_version: u64 could not be used?

I don't see how it is useful here.

SomeoneToIgnore: I am thinking that walreceiver benefits from getting the freshmost updates to every SK node timeline data, while new model forces us to loop through all old messages and drops latest updates instead of the ones before: older data is obsolete for walreceiver.
If the node stops broadcasting broker updates, retaining its last messages also feels more natural: since messages are sent per second now, there could be big LSN increments between the skipped messages.

> possibility of overflowing logs

I think, we have log deduplication now, or at least the work was done on that front, and if I understand that right, ~ 85seconds (256 / 3) of some subscriber not reading the messages is also very odd and deserves logging.

arssher:
To make sure broker is running before the merge of replacement, extracted the PR with the binary:
https://github.com/neondatabase/neon/pull/2829
it is not changed otherwise.

Actually there are going to be 3 PRs:
1) the binary
2) deploy of helm chart
3) the replacement

petuhovskiy:
We also need to prepare a PR for the cloud repo. Here is a list of all files with "etcd" in it: https://github.com/neondatabase/cloud/search?q=etcd

The most important one is `docker-compose.yml` which has etcd and also specifies etcd address for safekeepers and pageservers. Tests use that to get up storage, and it doesn't work currently.

Dockerfile
```diff
@@ -79,7 +79,7 @@ COPY --from=pg-build /home/nonroot/postgres_install.tar.gz /data/
 RUN mkdir -p /data/.neon/ && chown -R neon:neon /data/.neon/ \
     && /usr/local/bin/pageserver -D /data/.neon/ --init \
        -c "id=1234" \
-       -c "broker_endpoints=['http://etcd:2379']" \
+       -c "broker_endpoint='http://127.0.0.1:50051'" \
```
petuhovskiy, line 82:
```suggestion
       -c "broker_endpoint='http://storage_broker:50051'" \
```

Should be slightly better for running in docker.

storage_broker/src/lib.rs
```diff
@@ -21,8 +21,9 @@ pub use tonic::Request;
 pub use tonic::Streaming;
 
 pub const DEFAULT_LISTEN_ADDR: &str = "127.0.0.1:50051";
+pub const DEFAULT_ENDPOINT: &str = const_format::formatcp!("127.0.0.1:{DEFAULT_LISTEN_ADDR}");
```
petuhovskiy, line 24: `127.0.0.1:127.0.0.1:50051` ?

pageserver/src/walreceiver/connection_manager.rs
```diff
@@ -182,22 +167,16 @@ async fn connection_manager_loop_step(
...
-                            (&mut broker_subscription.watcher_handle).await,
-                            walreceiver_state,
-                        );
+                    Ok(Some(broker_update)) => walreceiver_state.register_timeline_update(broker_update),
+                    Err(e) => {
+                        error!("broker subscription failed: {e}");
+                        return ControlFlow::Continue(());
+                    }
+                    Ok(None) => {
+                        error!("broker subscription stream ended"); // can't happen
```
petuhovskiy, line 179: Why "can't happen"? gRPC request with streaming can break at any time, for example if there are issues with network connection or the server decided to shut down the request. We should consider this and write code for retries, the etcd version was doing it fine.

Also, it seems that we don't have tests with broker unavailability/restarts, such a test would be good for testing this retry logic.

arssher: > Why "can't happen"?

We never end the stream. In case of connection failure, it would be
```
2022-11-28T06:45:12.889282Z ERROR wal_connection_manager{tenant=de2dfd66cce92d8c4e5266e5e42725e7 timeline=692a8f4f575b0ce8bc4801d601b8abaf}: broker subscription failed: status: Unknown, message: "error reading a body from connection: broken pipe", details: [], metadata: MetadataMap { headers: {} }
```
neighbour arm. But we still do retries regardless whether it was failure or end of stream.


Yes, adding tests would be nice. I'd probably do that after adding to sk endpoint listing current walsender connections, it would make test more interesting (terminate broker and sk which is currently connected, ensure things still work).

petuhovskiy: Hmm, it seems that I misunderstood the scope of the `return ControlFlow::Continue(());` and got triggered by it. Now it looks ok, but I'll try to restart broker in the docker-compose setup to see what will happen.

petuhovskiy (approved):
LGTM, I'm ok with merging it once tests are green

arssher:
Failure of test_forward|backward_compatibility is expected; though there is no data format changes, broker configuration changed almost everywhere, so I'm adding labels allowing the failure.

I checked locally that these tests pass when run against this PR itself though.

arssher:
@bayandin FYI `ALLOW_BACKWARD_COMPATIBILITY_BREAKAGE` doesn't cover `prepare_snapshot`, which raises if neon_local configuration incompatibly changes, I'm going to extend it here. (same with forward compat)

bayandin:
> Failure of test_forward|backward_compatibility is expected; though there is no data format changes, broker configuration changed almost everywhere, so I'm adding labels allowing the failure.

I think we need to teach `prepare_snapshot` to handle both configurations (and migrate the old configuration format to the new one if needed).
Otherwise, it will require adding `* compatibility breakage` to all PRs until the release (and collection of new snapshots with the new configuration), if I understand it correctly.

> @bayandin FYI ALLOW_BACKWARD_COMPATIBILITY_BREAKAGE doesn't cover prepare_snapshot, which raises if neon_local configuration incompatibly changes, I'm going to extend it here. (same with forward compact)

See the message above, probably don't need to do it (or I'd try to avoid it if possible).

Please let me know if you need help with this.

arssher:
> Otherwise, it will require adding * compatibility breakage to all PRs until the release

Oh, that's indeed unfortunate. Generally this doesn't make sense; apparently such label should trigger snapshot recreation to avoid that.

In this particular case manual upgrade is indeed not hard, though it doesn't seem very useful to me and I thought that putting label is faster (and won't require keeping/maintaining upgrade code in the repo).

bayandin:
> apparently such label should trigger snapshot recreation to avoid that.

Yep, that's a good idea, but it could not be so trivial to implement. We need to trigger the rebuilding of a snapshot in the main branch after a PR with this label has been merged into it.
Another inconvenience is that other PRs without merged main will start to fail (without changes in a code on their side).
So probably, for now, adding such handling to the code is not the worst case if we want to check that changes in PRs are compatible. 

> (and won't require keeping/maintaining upgrade code in the repo).

You're totally correct. We can delete this code right after the release.

arssher:
> Otherwise, it will require adding * compatibility breakage to all PRs until the release

On the second look, I'm not sure this is the case. Looking through action.yml, AFAIU PRs to main are compared with latest commit on main; PRs to release are compared with latest release. So we'd need to suppress it only twice -- in this PR and in the PR releasing it, no?

bayandin:
>  AFAIU PRs to main are compared with latest commit on main; PRs to release are compared with latest release.

Not really; we do collect and upload data snapshots only for the release branch, and then all PRs start to use these snapshots
- https://github.com/neondatabase/neon/blob/main/.github/actions/run-python-test-set/action.yml#L184
- https://github.com/neondatabase/neon/blob/main/.github/workflows/build_and_test.yml#L977

Files itself stored on S3: https://s3.console.aws.amazon.com/s3/buckets/neon-github-public-dev?region=eu-central-1&prefix=artifacts/latest/&showversions=false

Dockerfile
SomeoneToIgnore: I would agree: in terms of Docker, localhost access feels odd, as not that much extra processes are started fluently to listen something.

control_plane/src/broker.rs
```diff
@@ -0,0 +1,36 @@
...
+
+    background_process::start_process(
+        "storage_broker",
+        &env.base_data_dir,
+        &env.storage_broker_bin(),
+        &args,
+        [],
+        background_process::InitialPidFile::Create(&storage_broker_pid_file_path(env)),
+        || {
+            Ok(true) // TODO
```
SomeoneToIgnore, line 23: We have a `GET /metrics` endpoint now and this module has `requwest` as a dependency, let's wait for that properly?
Should be simple, since that's how we essentially tested `etcd` before:

```rust
let client = reqwest::blocking::Client::new();

//////

|| {
    for broker_endpoint in &etcd_broker.broker_endpoints {
        let request = broker_endpoint
            .join("health")
            .with_context(|| {
                format!(
                    "Failed to append /health path to broker endopint {}",
                    broker_endpoint
                )
            })
            .and_then(|url| {
                client.get(&url.to_string()).build().with_context(|| {
                    format!("Failed to construct request to etcd endpoint {url}")
                })
            })?;
        if client.execute(request).is_ok() {
            return Ok(true);
        }
    }

    Ok(false)
},
```

pageserver/src/walreceiver.rs
```diff
@@ -27,56 +27,54 @@ use crate::config::PageServerConf;
 use crate::task_mgr::WALRECEIVER_RUNTIME;
```
SomeoneToIgnore, line 27: NIT: ~4 more etcd words are in storage, on top of this file is one of them

> keys in etcd broker

SomeoneToIgnore (approved):
A few nits, LGTM otherwise.
We need to fix the tests, apparently?

pageserver/src/walreceiver.rs
arssher: removed

arssher:
> We need to fix the tests, apparently?

Yep, there was an issue was backward|forward compat test, see above. Should be fixed now, but let's see.

control_plane/src/broker.rs
arssher: > Should be simple, since that's how we essentially tested etcd before:

That code was actually broken, I rewrote it.

bayandin:
@arssher you may want to remove labels now

.github/workflows/build_and_test.yml
```diff
@@ -927,7 +927,7 @@ jobs:
       - name: Deploy storage-broker
         run:
           DOCKER_TAG=${{ needs.tag.outputs.build-tag }}
```
arssher, line 929: No strong opinion, but shouldn't we pass that in helm values as other configs?

.github/workflows/build_and_test.yml
```diff
@@ -927,7 +927,7 @@ jobs:
...
-          helm upgrade neon-storage-broker neondatabase/neon-storage-broker --namespace ${{ matrix.storage_broker_ns }} --create-namespace --install -f .github/helm-values/${{ matrix.storage_broker_config }}.yaml --set image.tag=${DOCKER_TAG} --wait --timeout 15m0s
+          helm upgrade neon-storage-broker neondatabase/neon-storage-broker --namespace ${{ matrix.storage_broker_ns }} --create-namespace --install -f .github/helm-values/${{ matrix.storage_broker_config }}.yaml --set image.tag=${DOCKER_TAG} --set settings.sentryUrl=${{ secrets.SENTRY_URL_BROKER }} --wait --timeout 15m0s
```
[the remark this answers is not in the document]
arssher: ok

PR #2736, opened 2022-11-02
Tenant guard object

alexstanovoy:
Tenant guard and active guard objects to enable RAII-like control over when tenants are dropped.

### DIFF
--- a/pageserver/src/http/routes.rs
+++ b/pageserver/src/http/routes.rs
@@ -189,27 +189,32 @@ async fn timeline_create_handler(mut request: Request<Body>) -> Result<Response<
 
     let state = get_state(&request);
 
-    let tenant = tenant_mgr::get_tenant(tenant_id, true).map_err(ApiError::NotFound)?;
-    let new_timeline_info = async {
-        match tenant.create_timeline(
-            request_data.new_timeline_id.map(TimelineId::from),
-            request_data.ancestor_timeline_id.map(TimelineId::from),
-            request_data.ancestor_start_lsn,
-            request_data.pg_version.unwrap_or(crate::DEFAULT_PG_VERSION)
-        ).await {
+    let new_timeline_info =
+        match tenant_mgr::with_tenant_async(tenant_id, move |tenant| async move {
+            tenant
+                .create_timeline(
+                    request_data.new_timeline_id.map(TimelineId::from),
+                    request_data.ancestor_timeline_id.map(TimelineId::from),
+                    request_data.ancestor_start_lsn,
+                    request_data.pg_version.unwrap_or(crate::DEFAULT_PG_VERSION),
+                )
+                .await
+        })
+        .await
+        .map_err(ApiError::NotFound)?
+        {
             Ok(Some(new_timeline)) => {
                 // Created. Construct a TimelineInfo for it.
                 let timeline_info = build_timeline_info(state, &new_timeline, false, false)
                     .await
                     .map_err(ApiError::InternalServerError)?;
-                Ok(Some(timeline_info))
+                Some(timeline_info)
             }
-            Ok(None) => Ok(None), // timeline already exists
-            Err(err) => Err(ApiError::InternalServerError(err)),
-        }
-    }
-    .instrument(info_span!("timeline_create", tenant = %tenant_id, new_timeline = ?request_data.new_timeline_id, lsn=?request_data.ancestor_start_lsn, pg_version=?request_data.pg_version))
-        .await?;
+            Ok(None) => None, // timeline already exists
+            Err(err) => {
+                return Err(ApiError::InternalServerError(err));
+            }
+        };
 
     Ok(match new_timeline_info {
         Some(info) => json_response(StatusCode::CREATED, info)?,
@@ -228,8 +233,8 @@ async fn timeline_list_handler(request: Request<Body>) -> Result<Response<Body>,
     let state = get_state(&request);
 
     let timelines = info_span!("timeline_list", tenant = %tenant_id).in_scope(|| {
-        let tenant = tenant_mgr::get_tenant(tenant_id, true).map_err(ApiError::NotFound)?;
-        Ok(tenant.list_timelines())
+        tenant_mgr::with_tenant(tenant_id, |tenant| Ok(tenant.list_timelines()))
+            .map_err(ApiError::NotFound)?
     })?;
 
     let mut response_data = Vec::with_capacity(timelines.len());
@@ -293,12 +298,11 @@ async fn timeline_detail_handler(request: Request<Body>) -> Result<Response<Body
 
     let timeline_info = async {
         let timeline = tokio::task::spawn_blocking(move || {
-            tenant_mgr::get_tenant(tenant_id, true)?.get_timeline(timeline_id, false)
+            tenant_mgr::with_tenant(tenant_id, |tenant| tenant.get_timeline(timeline_id, false))?
         })
         .await
-        .map_err(|e: JoinError| ApiError::InternalServerError(e.into()))?;
-
-        let timeline = timeline.map_err(ApiError::NotFound)?;
+        .map_err(|err| ApiError::InternalServerError(err.into()))?
+        .map_err(|err| ApiError::NotFound(err.into()))?;
 
         let timeline_info = build_timeline_info(
             state,
@@ -329,9 +333,10 @@ async fn get_lsn_by_timestamp_handler(request: Request<Body>) -> Result<Response
         .map_err(ApiError::BadRequest)?;
     let timestamp_pg = postgres_ffi::to_pg_timestamp(timestamp);
 
-    let timeline = tenant_mgr::get_tenant(tenant_id, true)
-        .and_then(|tenant| tenant.get_timeline(timeline_id, true))
-        .map_err(ApiError::NotFound)?;
+    let timeline =
+        tenant_mgr::with_tenant(tenant_id, |tenant| tenant.get_timeline(timeline_id, true))
+            .map_err(ApiError::NotFound)?
+            .map_err(ApiError::NotFound)?; // What should we return if there's no timeline?
     let result = match timeline
         .find_lsn_for_timestamp(timestamp_pg)
         .map_err(ApiError::InternalServerError)?
@@ -351,8 +356,8 @@ async fn tenant_attach_handler(request: Request<Body>) -> Result<Response<Body>,
 
     info!("Handling tenant attach {tenant_id}");
 
-    tokio::task::spawn_blocking(move || match tenant_mgr::get_tenant(tenant_id, false) {
-        Ok(tenant) => {
+    tokio::task::spawn_blocking(move || {
+        tenant_mgr::with_tenant(tenant_id, |tenant| {
             if tenant.list_timelines().is_empty() {
                 info!("Attaching to tenant {tenant_id} with zero timelines");
                 Ok(())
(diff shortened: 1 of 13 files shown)


pageserver/src/tenant_guard.rs
```diff
@@ -0,0 +1,122 @@
...
+            // which references ephemeral files which are deleted on drop. So if we keep these references,
+            // we will attempt to remove files which no longer exist. This can be fixed by having shutdown
+            // mechanism for tenant that will clean temporary data to avoid any references to ephemeral files
+            let local_tenant_directory = conf.tenant_path(&tenant_id);
+            fs::remove_dir_all(&local_tenant_directory).with_context(|| {
+                format!(
+                    "Failed to remove local tenant directory '{}'",
+                    local_tenant_directory.display()
+                )
+            })?;
```
koivunej, line 50: Adding a review comment as we discussed this. Removal of local files or the equivalent of `detach_tenant` is not wanted in the normal case, as far as I remember from the RFC.

Instead I think there should be way to wait for "having zero active tenant guards" in `tenant_detach` and only then remove the files. One way which I suggested was tokio's broadcast channel. All it would need is a way to subscribe to it, and in this `Drop`, send out a message via that channel.

If there's no message being sent through the channel (it just gets closed) -- then something needs to be done for the semantics of that ... Anyway, it might be a good way to communicate for example any failure to checkpoint which happened.

koivunej: Alternative discussed: having some mutable state to enable the detaching on drop. Could be that the channel would be extra, I just think it would be nice to await for such thing to complete by default :) Could of course be that tenant detach must await for all cleanup to happen.

LizardWizzard: +1 And one note on detach semantics. Currently detach is expected to return ok back to the client only when everything is successfully stopped. Otherwise it gives an error so client can retry the call. Doing detach via spawning another task without waiting for completion may lead to loss of the detach operation  in case when we return ok to client and then pageserver is restarted losing the background task

pageserver/src/http/routes.rs
```diff
@@ -523,9 +520,7 @@ async fn tenant_status(request: Request<Body>) -> Result<Response<Body>, ApiErro
     check_permission(&request, Some(tenant_id))?;
 
     // if tenant is in progress of downloading it can be absent in global tenant map
-    let tenant = tokio::task::spawn_blocking(move || tenant_mgr::get_tenant(tenant_id, false))
-        .await
-        .map_err(|e: JoinError| ApiError::InternalServerError(e.into()))?;
+    let tenant = tenant_mgr::get_alive_tenant(tenant_id);
```
SomeoneToIgnore, line 523: We should return any status (which is state, actually), even if the timeline is broken, not mentioning other non-active states.

pageserver/src/tenant_mgr.rs
```diff
@@ -407,25 +388,35 @@ pub fn update_tenant_config(
...
+    let guard = m
         .get(&tenant_id)
         .with_context(|| format!("Tenant {tenant_id} not found in the local state"))?;
-    if active_only && !tenant.is_active() {
-        anyhow::bail!("Tenant {tenant_id} is not active")
-    } else {
-        Ok(Arc::clone(tenant))
+    match guard.upgrade() {
+        Some(tenant) if tenant.is_active() => Ok(tenant),
+        _ => bail!("Tenant {tenant_id} is not active"),
```
SomeoneToIgnore, line 404: NIT: could separate alive/active error messages here too, as in `with_context` below.

pageserver/src/tenant_guard.rs
```diff
@@ -0,0 +1,117 @@
...
+        match tenant.tenant.as_ref().unwrap().current_state() {
+            TenantState::Active { .. } | TenantState::Paused => Some(ActiveTenantGuard(tenant)),
+            TenantState::Broken | TenantState::Stopping => None,
+        }
+    }
+
+    pub fn current_state(&self) -> TenantState {
+        match self.upgrade() {
+            Some(tenant) => tenant.current_state(),
+            None => TenantState::Paused,
```
SomeoneToIgnore, line 95: This is not always true, since `TenantState::Broken` state results in `None` branch.

pageserver/src/tenant_guard.rs
```diff
@@ -0,0 +1,117 @@
...
+        }
+    }
+
+    pub fn is_active(&self) -> bool {
```
SomeoneToIgnore, line 99: Not used anywhere?

pageserver/src/tenant_guard.rs
```diff
@@ -0,0 +1,117 @@
...
+    pub fn upgrade(&self) -> Option<ActiveTenantGuard> {
+        let tenant = self.0.upgrade()?;
+        match tenant.tenant.as_ref().unwrap().current_state() {
+            TenantState::Active { .. } | TenantState::Paused => Some(ActiveTenantGuard(tenant)),
+            TenantState::Broken | TenantState::Stopping => None,
+        }
+    }
+
+    pub fn current_state(&self) -> TenantState {
+        match self.upgrade() {
```
SomeoneToIgnore, line 93: NIT: Seems somewhat wasteful to do all the Arc upgrade magic compared to keeping a Receiver subscription on the status: we do care about the Sender end drop, so can keep as many subscriptions as possible.

I cannot really tell what's better though, so not insisting.

pageserver/src/tenant_guard.rs
```diff
@@ -0,0 +1,117 @@
...
+        };
+        BACKGROUND_RUNTIME.spawn(future);
+
+        (Self(weak_inner), ActiveTenantGuard(arc_inner))
+    }
+
+    pub fn upgrade(&self) -> Option<ActiveTenantGuard> {
+        let tenant = self.0.upgrade()?;
+        match tenant.tenant.as_ref().unwrap().current_state() {
+            TenantState::Active { .. } | TenantState::Paused => Some(ActiveTenantGuard(tenant)),
```
SomeoneToIgnore, line 87: I'm not entirely sure this is correct: `active` and `Paused` do not match well, so at least the name would need some change.

pageserver/src/tenant_guard.rs
```diff
@@ -0,0 +1,117 @@
...
+
+        self.drop_watcher.send_replace(());
+    }
+}
+
+#[derive(Clone)]
+pub struct TenantGuard(Weak<TenantGuardInner>);
+
+impl TenantGuard {
+    pub fn new(tenant: Tenant) -> (Self, ActiveTenantGuard) {
```
SomeoneToIgnore, line 60: `ActiveTenantGuard` is used in all cases only to update its state, so should we just add a `set_state` method on `TenantGuard` that would try to update the tenant and set a state instead? 

This way, we expose one less `Arc` around the tenant and don't let anybody to store it.

pageserver/src/tenant_guard.rs
```diff
@@ -0,0 +1,117 @@
...
+    pub fn is_active(&self) -> bool {
+        self.upgrade().is_some()
+    }
+
+    pub fn subscribe_for_tenant_shutdown(&self) -> Option<watch::Receiver<()>> {
+        Some(self.upgrade()?.0.drop_watcher.subscribe())
+    }
+}
+
+#[derive(Clone)]
```
SomeoneToIgnore, line 108: Not used and I'm not sure what's the point of allowing to clone such things.

SomeoneToIgnore (commented):
I like `pub struct TenantGuard(Weak<TenantGuardInner>);` structure and not fond of `pub struct ActiveTenantGuard(Arc<TenantGuardInner>);`, since it allows anyone to store this struct in `async move loop {}` context and never let go.

This is not very notable in `Tenant` case, but here's a `Timeline` example where it matters: https://github.com/neondatabase/neon/blob/51121429976da628f3118db9130d3171a8e29dab/pageserver/src/walreceiver/walreceiver_connection.rs#L58

Walreceiver keeps on working in some loop, while timeline could be stopped at any moment instead.
For now, it works because we run `task_mgr::shutdown_tasks` correctly, same as in this PR, but what happens if we apply current design to Timeline? 
One could simply pass `ActiveTenantGuard` there without noticing, breaking all our defence.

So it's more or less the same situation with `Arc<Tenant>` in this PR, with more ceremonies to make an error, but it's still possible.
I think it's not good and that's exactly why the guard should exist: to never allow such errors, check statuses and accommodate nuisances of using `Weak<Tenant>` under the hood.
We call the same `task_mgr::shutdown_tasks` in this PR, while ideally the tasks should stop next time they try to access a tenant that's `Paused`, `Broken` or `Stopping`.

Yet it leaves the question how to design such API which I wanted to explore..
Do you think it's possible and feasible? If not, might it be better to fix the races and leave the guard out of the scope of the PR?

SomeoneToIgnore (commented):
So, our current compromises are:
* `pub struct TenantAccessor(Arc<TenantGuardInner>);` still could be stored in some struct field and defer the tenant drop.

But current API makes it relatively hard and the only alternative to turn it into `&Tenant` now is to use boxing (https://neondb.slack.com/archives/C0277TKAJCA/p1665762397772679)

* `async move {`

But seems that all we have to move in as arguments do not require `clone` and generally quite good.


On the positive side, we have a bit better status situation now and are one step closer to drop & react-on-status-change set-up.
I think it's a good step forward and we could propagate it on `Timeline` and see how it applies there, but first let's the discussions we have now open, might be that we find something more.

pageserver/src/tenant_guard.rs
```diff
@@ -0,0 +1,138 @@
...
+        matches!(self.current_state(), TenantState::Active { .. })
+    }
+
+    pub fn subscribe_for_tenant_shutdown(&self) -> Option<watch::Receiver<()>> {
+        Some(self.0.upgrade()?.drop_watcher.subscribe())
+    }
+
+    pub fn with_upgrade<F, T>(&self, f: F) -> T
+    where
+        F: FnOnce(anyhow::Result<TenantAccessor>) -> T,
```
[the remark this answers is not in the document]
alexstanovoy: Instead, I think it's better to return `Result<O>` from `with_upgrade` (changed to `with_tenant` from the last time). This shows whether the closure was executed, and I think that proved easy to use when I refactored the code later. What about this approach?

IMO there's one problem currently: we can't distinguish between "no tenant locally" and "tenant present but is dead" without checking the error itself. But looks like it will only add complexity to the code without benefits.

pageserver/src/tenant_mgr.rs
```diff
@@ -147,62 +147,68 @@ pub fn attach_local_tenants(
...
-                        info!("successfully loaded local timelines for tenant {tenant_id}");
-                        tenant.activate(has_timelines);
-                    }
-                    Err(e) => {
-                        error!("Failed to attach tenant timelines: {e:?}");
-                        tenant.set_state(TenantState::Broken);
+        tenant.with_upgrade(|tenant| {
+            let tenant = tenant.expect("tenant must be alive after creation");
+            match local_timelines {
+                TenantAttachData::Broken(e) => warn!("{}", e),
```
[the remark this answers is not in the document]
alexstanovoy: I think it won't affect performance, but since `TenantAccessor` stores state, this state could be changed somehow by operations that are currently inside the closure. This could lead to missed changes in the state by the accessor and break the invariants.

Probably we should forbid creating tenant without TenantAccessor, i.e private it?

alexstanovoy:
Should we prevent anything to launch in `with_upgrade` on the tenant that is in the Stopping state?

PR #2737, opened 2022-11-02
Move pq_proto to its own crate

funbringer:


libs/pq_proto/src/lib.rs
```diff
@@ -0,0 +1,1081 @@
+//! Postgres protocol messages serialization-deserialization. See
```
SomeoneToIgnore, line 1: Interesting, that this file was not marked as moved, similar to `sync.rs` below.
Were there any special changes besides moving for this file?

SomeoneToIgnore (approved):

libs/pq_proto/src/lib.rs
funbringer: A bit, I had to patch some imports and add `mod sync`.

PR #2741, opened 2022-11-02
Impl serialize for pagestream FeMessage

bojanserafimov:
Also generalize `parse` to work with `Read`, so that we can also parse a list of consecutive messages without knowing lengths up front.

Ideally we'd use BeSer, but the API isn't exactly what's needed so it's not trivial.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2190,6 +2190,7 @@ name = "pageserver_api"
 version = "0.1.0"
 dependencies = [
  "anyhow",
+ "byteorder",
  "bytes",
  "const_format",
  "postgres_ffi",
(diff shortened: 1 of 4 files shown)


SomeoneToIgnore (approved):
Should we add a test to do serialize -> deserialize for every message kind, to ensure we don't mess up the field order?

Also curious to know where this change leads in general, is it related to benchmarking tasks?

bojanserafimov:
> Should we add a test to do serialize -> deserialize for every message kind, to ensure we don't mess up the field order?
> 
> Also curious to know where this change leads in general, is it related to benchmarking tasks?

I'll add a test.

Next steps:
- add pagestream client
- (in parallel) add option for read request logging. It's a lot of data so it should be packed tightly in binary
- use pagestream client + read traces to write various perf tests
- (in parallel) read request visualization + validation. Do we have duplicate read requests that don't make sense (yes), etc.

I have a hacky implementation of this and I'm tidying up and spllitting it into small PRs like this.

bojanserafimov:
opened a separate PR for tests https://github.com/neondatabase/neon/pull/2746

PR #2742, opened 2022-11-02
Fix clippy errors on main

bojanserafimov:
The errors probably appear because we pick up new clippy versions with new lints


SomeoneToIgnore (approved):
Odd, our CI did not fail on the recent rustc bump PR, but the change looks very clippy-ish indeed, thank you.

PR #2743, opened 2022-11-02
fix: use different port for temporary postgres

koivunej:
`test_tenant_relocation` starts up a temporary postgres instance with a fixed port. PR makes the port configurable at `scripts/export_import_between_pageservers.py` and uses that in `test_tenant_relocation`. Found earlier in #2714 in both test runs at the same time, here's [one](https://neon-github-public-dev.s3.amazonaws.com/reports/pr-2714/debug/3378582321/index.html#categories/784f2376295054d7afc63716a0689c32/3e59af69806f30af/?attachment=f65e01b937fe0ef5).

### DIFF
--- a/scripts/reformat
+++ b/scripts/reformat
@@ -6,6 +6,6 @@ set -euox pipefail
 echo 'Reformatting Rust code'
 cargo fmt
 echo 'Reformatting Python code'
-poetry run isort test_runner
-poetry run flake8 test_runner
-poetry run black test_runner
+poetry run isort test_runner scripts
+poetry run flake8 test_runner scripts
+poetry run black test_runner scripts
--- a/scripts/export_import_between_pageservers.py
+++ b/scripts/export_import_between_pageservers.py
@@ -378,7 +378,7 @@ def reconstruct_paths(log_dir, pg_bin, base_tar, port: int):
 
         # Start a vanilla postgres from the given datadir and query it to find
         # what relfiles should exist, but possibly don't.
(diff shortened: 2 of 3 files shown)


scripts/reformat
```diff
@@ -6,6 +6,6 @@ set -euox pipefail
 echo 'Reformatting Rust code'
 cargo fmt
 echo 'Reformatting Python code'
-poetry run isort test_runner
-poetry run flake8 test_runner
-poetry run black test_runner
+poetry run isort test_runner scripts
+poetry run flake8 test_runner scripts
+poetry run black test_runner scripts
```
bayandin, line 11: Nice!
Didn't know about this script

bayandin (approved):

koivunej:
Hmm interesting mypy error. I did get that while fixing changing the port but interesting why did it error now...

bojanserafimov (approved):
Thanks, looks good.

Another solution would be to delete `add_missing_rels`. It's been a while since that migration and we no longer need to deal with missing empty relations.

scripts/reformat
koivunej: Linked to be today by @SomeoneToIgnore :)

SomeoneToIgnore: It's a bit of an odd name given that one of the tools fails on the non-formatted imports.
So I think we could rename it to `style_check` (?), add run_clippy.sh and a mypy check, but not sure yet, since those two are slower.

PR #2745, opened 2022-11-02
export_import_between_pageserver: remove add_missing_rels

koivunej:
Following https://github.com/neondatabase/neon/pull/2743#pullrequestreview-1165654659, it seemed feasible to clean up the `scripts/export_import_between_pageservers.py`. Removes the `add_missing_rels` and all that follows. Might be a good win for test running perf, if these old pageservers no longer exist anywhere.

There is still a verification in tests that all relations are now accessible.

### DIFF
--- a/pageserver/src/bin/pageserver.rs
+++ b/pageserver/src/bin/pageserver.rs
@@ -205,7 +205,7 @@ fn start_pageserver(conf: &'static PageServerConf) -> anyhow::Result<()> {
             new_lock_contents,
             file,
         } => {
-            info!("Created lock file at {lock_file_path:?} with contenst {new_lock_contents}");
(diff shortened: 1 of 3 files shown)


bojanserafimov:
The old pageservers have all been migrated a while ago. The changes look good.

I'll leave it to @hlinnaka to confirm that the "return size 0 if rel doesn't exist" hack is **fully** removed in the current pageservers. If not, we might still need this reconstruction logic.

koivunej:
e2e failure seems unrelated: https://github.com/neondatabase/cloud/actions/runs/3380891853/jobs/5614383237#step:34:405

koivunej:
This just needs ... rebasing, and review. Drafting.

PR #2746, opened 2022-11-03
Write message serialization test

bojanserafimov:
Just something to prevent obvious bugs until we transition serialization logic to serde

### DIFF
--- a/libs/pageserver_api/src/models.rs
+++ b/libs/pageserver_api/src/models.rs
@@ -226,7 +226,7 @@ pub struct TimelineGcRequest {
 }
 
 // Wrapped in libpq CopyData
-#[derive(PartialEq)]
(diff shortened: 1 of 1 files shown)


libs/pageserver_api/src/models.rs
```diff
@@ -242,29 +243,29 @@ pub enum PagestreamBeMessage {
     DbSize(PagestreamDbSizeResponse),
 }
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
```
SomeoneToIgnore, line 246: I think, clippy used to complain that no `Eq` is derived, but maybe it will do that in newer rustc versions.

SomeoneToIgnore (approved):

Issue #2722, opened 2022-10-31
Large number of threads in a safekeeper


petuhovskiy:
In one of our environments, safekeeper processes have reached 3000 active threads. After restart, most of the threads are gone, but there was still one safekeeper with >2000 threads left.

I collected some numbers from that environment and a particular safekeeeper with a large number of threads:
- 97 active projects (in env)
- 108 compute nodes running (in env)
- 195 active timelines in safekeepers
- 1035 connections from pageservers to safekeeper
  - there are 2 active pageserver, connections between them are split evenly (about ~500 from each)
- 2070 threads in safekeeper
  - half of the threads are `HotStandbyFeedback thread`
  - other half is corresponding threads for WAL streaming

The first issue derived from these numbers is non-matching numbers on different levels:
- 108 computes
- 195 active timelines
- 1035 replication connections

We usually expect a number of computes to match the number of active timelines and active replication streams from pageserver. Currently the picture is different, and we have ~x10 of replication streams.
The most probable cause of this issue is that we don't break connections when timeline becomes inactive in safekeeper, but I'm not sure.
- [x] Stop START_REPLICATION connections after compute stops

Another issue is a large number of threads. Once we will hit a number of 4096 active threads, we may start getting errors from one of our used Rust library, which doesn't support more than 4096 threads. We can completely fix the threads issue when we switch to async.
- [ ] Switch all safekeeper code to use async

In theory, we can also exhaust IP ports when we will have ~32000 connections, but we're not there yet.

antonyc:
> but there was still one safekeeper with >2000 threads left.

I did not touch this safekeeper (sk-2), because it did not hit the threshold.

arssher:
Thanks for collecting data.

> Stop START_REPLICATION connections after compute stops

Indeed, it was broken.

> Once we will hit a number of 4096 active threads, we may start getting errors 

I'd raise that limit actually anyway.

PR #2747, opened 2022-11-03
Fix sk->ps walsender shutdown on sk side on caughtup.

arssher:
This will fix many threads issue, but code around still awfully wants improvement.

https://github.com/neondatabase/neon/issues/2722

### DIFF
--- a/safekeeper/src/timeline.rs
+++ b/safekeeper/src/timeline.rs
@@ -555,13 +555,20 @@ impl Timeline {
         if self.is_cancelled() {
             return true;
         }
-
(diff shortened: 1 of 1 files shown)


petuhovskiy (approved):
Not sure if this will fix the issue and also don't like the code, but let's deploy this to staging to see what will happen.

Also I think we need to use `is_active` and be more careful with `Lsn::INVALID`, `Lsn::MAX`, `Lsn(0)` and similar cases, because right now it seems like we have a bug and something bad can happen with such Lsn's.

PR #2750, opened 2022-11-03
[proxy] Propagate more console API errors to the user

funbringer:


proxy/src/auth/backend/console.rs
```diff
@@ -37,61 +55,89 @@ pub enum GetAuthInfoError {
...
+/// This is essentially a [`Result`] but with a different
+/// [`Deserialize`] impl that makes more sense in our case.
+#[derive(Deserialize, Debug)]
+// Note that order of variants in untagged enums is important.
+#[serde(untagged)]
+enum ConsoleResponse<T> {
+    /// Nested payload.
+    Ok(T),
+    /// Console API error as text.
+    Err { error: Box<str> },
```
funbringer, line 119: @ololobus I suppose it's fine to show this text to user, but I'd like to hear a confirmation.

LizardWizzard (approved):
Looks good to me

proxy/src/auth/backend/console.rs
funbringer: @tychoish wdyt?

ololobus: > I suppose it's fine to show this text to user, but I'd like to hear a confirmation.

I **thought** so too, but I've just had a look on the management API server code. In the public v2 we have a middleware that only renders messages for the known semantic error, but I don't see something similar in the management API (I was looking on `ChiErrorHandler`). Thus, I assume it can throw away any internal error text (it also responds with `StatusBadRequest`, so I don't know how to distinguish errors), cc @antonyc correct me if I'm wrong

funbringer: In that case the only thing I could show user is `Internal error`, because I can't differentiate between "user doesn't exist", "not enough monies" or "warning: very secret error message" etc.

antonyc: > it can throw away any internal error text (it also responds with StatusBadRequest, so I don't know how to distinguish errors)
You are correct. Management API can expose any error to the user. We can add support for semantic errors with almost zero effort (just copy-paste some code).

ololobus: I think we need it, so that proxy was able to distinguish what it should present to the user (likely all non-500s) and what to hide (500s)

funbringer: Discussed this with @antonyc in private. I'm going to rework some parts.

ololobus:
@nikitakalyanov pointed one possible problem with propagating too detailed errors from the management API to the user: https://github.com/neondatabase/cloud/pull/3142#discussion_r1038278144

For example (if we propagate these errors):
- console returns 404 "role not found", which means endpoint exists, but no such role
- console returns secret, but it doesn't match with user-provided password, we already return "password doesn't match" from the proxy

That way, one can probe both endpoint id, then roles inside it, then finally password. Probably that's fine, since endpoint ids are not really easy to guess and default role names are unique.

Another way would be to just show generic error like "password authentication failed" unless password authentication passed in proxy. Yet, that's not really user-friendly

funbringer:
Those are very valid concerns. I've already implemented a proper fix in this PR, though.

https://github.com/neondatabase/neon/pull/2750/files#diff-aff3f6820f3e51b92663b3492db14f8bf5cc1c0ba0ed85b830020d1a90349e25R289-R295

test_runner/regress/test_proxy.py
```diff
@@ -28,61 +28,58 @@ def test_password_hack(static_proxy: NeonProxy):
...
-
-    log.info("sending session activation message")
-    psql = await PSQL(host=link_proxy.host, port=link_proxy.mgmt_port).run(db_info)
-    out = (await psql.stdout.read()).decode("utf-8").strip()
-    assert out == "ok"
-
-
 @pytest.mark.asyncio
 async def test_psql_session_id(vanilla_pg: VanillaPostgres, link_proxy: NeonProxy):
+    def get_session_id(uri_prefix, uri_line):
```
LizardWizzard, line 33: Not sure I get the idea of moving these helper functions inside the test function. Could you elaborate?

LizardWizzard (commented):
Looks good to me, left minor comments

proxy/src/auth/backend/console.rs
```diff
@@ -5,26 +5,68 @@ use crate::{
...
+}
 
-impl UserFacingError for TransportError {}
+impl UserFacingError for ApiError {
+    fn to_string_client(&self) -> String {
+        // Ask @neondatabase/control-plane for review before adding more.
+        let harmless = [
+            HttpStatusCode::LOCKED,                // 423
+            HttpStatusCode::NOT_FOUND,             // 404
+            HttpStatusCode::INTERNAL_SERVER_ERROR, // 500
```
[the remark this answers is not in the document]
funbringer: I'm still not sure about that one, too.

test_runner/regress/test_proxy.py
funbringer: I think it's a matter of taste, and to me that looks better.

proxy/src/auth/backend/console.rs
```diff
@@ -249,6 +330,30 @@ where
...
+        info!("request succeeded, processing the body");
+        return Ok(response.json().await?);
+    }
+
+    // Don't throw an error here because it's not as important
+    // as the fact that the request itself has failed.
+    let body = response.json().await.unwrap_or_else(|e| {
+        warn!("failed to parse error body: {e}");
+        ConsoleError {
+            error: "Broken error message".into(),
```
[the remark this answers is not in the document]
funbringer: Changed. Feel free to propose more options.

proxy/src/auth/backend/console.rs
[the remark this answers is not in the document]
ololobus: > Can there be arbitrary text in 500 messages?

We shouldn't leak it, but I think we can now. Something like https://github.com/neondatabase/cloud/pull/3195 should fix that.

Generally, the flow is following:
- every generic Go error is rendered as vague 'internal error' or 'unknown internal error' (this depends on API, I think we can make it universal) with code 500 automatically by errors middleware
- also we have semantic errors, which are widely used if we have something to say to the user. They are created by methods like `semerr.WrapWithInvalidInput(err, msg)`, where `msg` is what will be rendered to the client. Each semantic error has a corresponding HTTP code
- yet, there is a small intersection of these two cases, we have also `semerr.WrapWithInternal()`, so that we can render some custom text on 500s too. Usually, this is not needed, but for some flows like OAuth we usually cannot say who is responsible for error, but we can hint that there is a probably some missconfiguration in the flow. I see misuse of this in the code, mostly in the v1, but I think it could be useful

I hope that this clarifies a bit the question about errors.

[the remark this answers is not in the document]
funbringer: Since https://github.com/neondatabase/cloud/pull/3195 is merged, I guess we're done here. @ololobus concerns?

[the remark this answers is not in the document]
ololobus: > it'd be ridiculous if there was nothing in the console (idk, defensive programming?) to prevent leaks from happening in user-facing APIs

And this actually brings my previous question back https://github.com/neondatabase/neon/pull/2750#issuecomment-1335501487. Management API that proxy uses for auth isn't something user-facing, obviously, as you get role role secrets out of it.

So I turned detailed errors in it for now as 'defensive programming' measure to do not accidentally leak internal state through proxy. But that was actually an intended behavior (there was even a test for that), since we want to see detailed errors in the management API (it's internal thing after all).

That said, in the long-term run this is a proxy responsibility to decide what to show to the user and what to don't. The same as we receive a raw error from pageserver API in the control-plane, but only save / log it internally, without showing to the user. One option to implement that could be to include both `error` and `message` in the management API response, so that proxy can show the content of `message`, but log `error`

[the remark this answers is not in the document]
ololobus: TL;DR I don't think that sending everything that we receive through some internal API to the client is a viable long term solution.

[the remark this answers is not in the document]
funbringer: > That said, in the long-term run this is a proxy responsibility to decide what to show to the user and what to don't. 

Sorry, but I can't agree with that. If we're talking about `{"error": "..."}` alone, proxy can't possibly know severity and importance of this message, because it's completely unconstrained. I'd go as far as to say that we should never implicitly assume that our APIs are private, because that's a very dangerous assumption! Consider this: why do we give an error message if it should be kept a secret? We could just log it in the console and return nothing but the status code; problem solved.

There are very specific cases when the proxy knows better, like when there's no secret, but that's enough of a burden.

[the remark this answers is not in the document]
ololobus: > Consider this: why do we give an error message if it should be kept a secret? We could just log it in the console and return nothing but the status code; problem solved.

Because when you set up new pageserver or do tenant relocation, you usually want to see the detailed error message in the internal management API response without going through logs. I don't say that leaking secrets in the logs and error responses is the right way to do, that's generally bad.

kelvich:
Discussion here feels a bit abstract. It would be easier to reason about it having some list of errors that proxy can return to the user that would be useful to the user. I can see the following types of error that we can propagate to the user:

* endpoint not found
* auth failed, both for wrong user and password we should return `password authentication failed for user "foouser"` as is https://github.com/neondatabase/neon/issues/1685
* auth is ok, but some compute limits are reached. We don't have such functionality now, but we''l have it soon.
* maintenance mode could treated as a previous "limits are reached" error or as a separate one.
* internal error whatever it would be.

Any other error message that could be useful to the user?

With that list we can use corresponding http codes: `404 Not Found`, `401 Unauthorized`, `429 Limit Exceeded`, `500 Internal error`. Maintenance mode could be either `503 Service Unavailable` or `429 Limit Exceeded`.

So if that is the full list I'd rather fix responses in proxy and log errors. Also matching postgres auth failed message is important and proxy seems to be the right place for emitting that message.

kelvich:
> Any error that happens before user's connection passes the auth process should be rendered as password authentication failed for user "foouser". The motivation is primarily based on the concerns listed in the comment [proxy] Propagate more console API errors to the user #2750 (comment), i.e. we return 404 from console for get role secret if either endpoint or such are missing, and that doesn't look like something that random person on the internet should be able to probe.

Right. I was thinking about intentionally allowing endpoint probing -- I myself stepped few times on a situation when I tried connecting to deleted projects, got some generic error, and it took some time to understand what is going on. Seems that we will gain more in usability compared to what we will potentially lose in security (if any).

However, with absent user / wrong password we'd better show generic "auth failed" error like in postgres. And user probing is a bigger leakage -- one can probe for github usernames and deduce owner or use popular framework/app-induced database names to gain some info about app.

> Once the auth is passed, then yes, there are several cases <...>

So my list was:
a1. auth failed
a2. limits reached
a3. maintenance mode
a4. internal error
a5. endpoint not found

Your list is:
b1. auth failed
b2. limits reached / payment required
b3. maintenance mode
b4. internal error
b5. endpoint locked
b6. endpoint disabled

Seems that we have first 4 common. `b5` and `b6` seems to be reasonable, +1. I'd also keep a5 for reasons mentioned in first half of the message

ololobus:
> I'd also keep a5 for reasons mentioned in first half of the message

Yeah, probably it's ok to have it. We will anyway eventually need some protection from brute forcing the connection in proxy, I guess. I.e. having dozens of connection attempts with different creds from the same ip during short period could be suspicious (hi folks behind NAT), but not sure

kelvich:
We've chatted with @funbringer and plan in last message seems reasonable. We don't have a way to distinguish between a1 and a5. Let's try to merge this with auth failed for missing endpoint, add new http code in console and then split error in two if @ololobus is okay with that.

Also @funbringer noted that we can probably avoid psql non-tls reconnect if we throw fatal instead of error.

LizardWizzard (approved):
LGTM

kelvich (approved):

PR #2762, opened 2022-11-05
Trace read requests

bojanserafimov:
Add tenant option to record all read requests and write the trace to a file. The intended use case is to run a test, record a trace, and then either:
- inspect the trace for bugs (duplicate requests, etc)
- visualize the trace
- replay the trace in various ways (serial, parallel, real-time, etc)

I write the traces in binary format to save space. They can be big.


pageserver/src/config.rs
```diff
@@ -414,6 +414,22 @@ impl PageServerConf {
...
 
+    pub fn traces_path(&self) -> PathBuf {
+        self.workdir.join("traces")
+    }
+
+    pub fn trace_path(
+        &self,
+        tenant_id: &TenantId,
+        timeline_id: &TimelineId,
+        connection_id: &TimelineId, // TODO make a new type
```
bojanserafimov, line 425: will address this before merging

test_runner/performance/test_trace_replay.py
```diff
@@ -0,0 +1,37 @@
...
+    tenant, _ = env.neon_cli.create_tenant(
+        conf={
+            "trace_read_requests": "true",
+        }
+    )
+
+    # TODO This doesn't work because I haven't updated tenant_config_handler
+    # env.neon_cli.config_tenant(tenant, conf={
+    #     "trace_read_requests": "true",
+    # })
```
bojanserafimov, line 21: will address this before merging

test_runner/performance/test_trace_replay.py
```diff
@@ -0,0 +1,31 @@
+from contextlib import closing
+
+from fixtures.neon_fixtures import NeonEnvBuilder
+
+
+# This test demonstrates how to collect a read trace. It's useful until
+# it gets replaced by a test that actually does stuff with the trace.
+def test_read_request_tracing(neon_env_builder: NeonEnvBuilder):
```
knizhnik, line 8: Test is called trace replay, bit it seems to be not replaying any traces.

knizhnik (approved):
As far as trace format is binary we definitely need some utility for dumping traces.

Issue #2729, opened 2022-11-01
Create a test for gc that detects lost history when gc misses info about certain timelines


LizardWizzard:
## Steps to reproduce

See https://github.com/neondatabase/neon/pull/2726
We need a test that can detect this

## Expected result

Tests passed

## Actual result

Tests should fail on such changes

## Logs, links
-

andresrsanchez:
Hi @LizardWizzard i was looking to do this test but i have one question with these steps i did:

- Create one branch and request to pageserver the details about it, receiving a `state: Active`.
- Stop the compute to trigger `on_compute_disconnect` setting the timeline state to `active=false` in SK
- Wait a little bit and again request to pageserver the details about it, receiving again a `state: Active` 

Why is that? Is it correct? 

I was looking to set a few timelines non active and then trigger a gc to test the not deletion of data needed by a child timeline.
Thanks!

LizardWizzard:
Hi! @andresrsanchez 

> Wait a little bit and again request to pageserver the details about it, receiving again a state: Active

Yeah, this indeed feels wrong. There is an option in pageserver called `checkpoint_timeout`. Checkpoints are performed by wal receiver either when enough size gets accumulated (`checkpoint_distance` option) or when `checkpoint_timeout` is reached. And currently state is not changed when we reach this timeout which is by default 10 minutes. I think it is a separate issue. Thought about opening a new one, but it turns out there is an existing one: https://github.com/neondatabase/neon/issues/2027

For this issue I think it may be simpler to create a unit test, but it may require additional glue code for setting up gc stuff there

PR #2763, opened 2022-11-06
Add test for getting branchpoints from an inactive timeline

andresrsanchez:
refer #2729 

I did the unit test as u said @LizardWizzard but i feel that checking for branchpoints it's not enough, let's see what you think.

Thanks!

### DIFF
--- a/pageserver/src/tenant.rs
+++ b/pageserver/src/tenant.rs
@@ -2068,6 +2068,35 @@ mod tests {
      */
 
     #[tokio::test]
+    async fn test_get_branchpoints_from_an_inactive_timeline() -> anyhow::Result<()> {
+        let tenant =
+            TenantHarness::create("test_get_branchpoints_from_an_inactive_timeline")?.load();
+        let tline = tenant
+            .create_empty_timeline(TIMELINE_ID, Lsn(0), DEFAULT_PG_VERSION)?
+            .initialize()?;
+        tline.set_state(TimelineState::Paused);
+        make_some_layers(tline.as_ref(), Lsn(0x20)).await?;
+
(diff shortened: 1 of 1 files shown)


LizardWizzard:
Thanks for the patch @andresrsanchez! I think your test is good. It should catch original issue. 

Originally I thought about doing something like this: (looking at [this comment](https://github.com/neondatabase/neon/blob/main/pageserver/src/tenant/timeline.rs#L2147), trying to trigger a scenario that checks everything except third point)
1. Insert some data into parent timeline
2. Branch new timeline
3. Overwrite inserted data with larger lsn value in parent timeline
4. Move cutoff + PITR beyond corresponding horizon
5. Check that this data is still accessible in child timeline (can be done with `Timeline::get_rel_page_at_lsn`)

And now I realized that there are somewhat similar tests: `test_retain_data_in_parent_which_is_needed_for_child` and `test_parent_keeps_data_forever_after_branching` but apparently they miss something and didnt catch the original issue

So I think next steps can be:
 a) extend your test with checks getting actual data using `Timeline::get` API like it is done in tests above and merge it
 b) I think it's worth poking into existing tests why they didnt catch original issue

andresrsanchez:
Thanks for the comment @LizardWizzard !!!

I executed `test_retain_data_in_parent_which_is_needed_for_child` with and without the [removed lines](https://github.com/neondatabase/neon/pull/2726/files) from the original issue and the logs were identical, the thing is that the `TimelineState` is always `Active` due to [initialize_with_lock](https://github.com/neondatabase/neon/blob/main/pageserver/src/tenant.rs#L166) so the `filter` is not going to do any filtering so any info is actually missed.

Ping me if you need me to dig deeper, thanks!

LizardWizzard:
Merged with #3801

PR #2767, opened 2022-11-07
Test storage Docker images

SomeoneToIgnore:
Closes https://github.com/neondatabase/neon/issues/2697
Example: https://github.com/neondatabase/neon/actions/runs/3416774593/jobs/5688394855

Adds a set of tests on the storage Docker images before they are pushed to the public registries:
* tests that pageserver binary has the correct version string (other binaries are built with the same library, so it should be enough to test one)
* tests that the compose file set-up works and all components are able to start and perform a single SQL query (CREATE TABLE)


hlinnaka (commented):
This runs after the "promote-images" step, but before "push-docker-hub". Is it too late? If the test fails, we'll have the new version in ECR but not in dockerhub.

SomeoneToIgnore:
I think, this way we're able to debug things better: if anyone wants to check what's wrong with the compose setup, he might want to avoid rebuilding everything (mac users might not be able to 100% reproduce the build) and instead pull the images that failed during testing.

ololobus (commented):
I have two questions:

1) The same as Heikki's. IIUC, the only thing promotion does is that it tags the image built in this workflow as `latest`. But you only want to test images built in this workflow, and you already have them with a tag same as workflow id, e.g. `3416774593`. So you can safely put your test before promote, and image artifacts will be already in the ECR for debuggability, isn't it?

2) Unrelated to this PR, why we ever promote images built in PRs to the `latest`?! It doesn't make any sense to me. There could be concurrent PRs / jobs and you never know where this `latest` points. What is worse, it could be concurrently updated, so it can go backward? c @zoete

zoete:
Image promotion was designed to be the task that will only run once you merge into main/release, and as a result only run on a production runner. However keeping everything backwards compatible with our old account, plus the requirement that we also keep docker hub fully in sync (else cloud repo can't pull) kind of broke the design I had in mind.

SomeoneToIgnore:
I have updated the workflow to test things before promotion, and I think 

> An ideal flow would be

is closer now, since we're now triggering e2e from `neon` in the `cloud` repo using the tag of the image built and pushed into `neondatabase/neon:${SOMETHING}`. 
We still need to work on the `GITHUB_RUN_ID` and maybe earlier access to images (before they are pushed to the public registry), but that feels dependent on a standalone task of moving all `cloud` CI to custom runners.

ololobus (approved):
LGTM

PR #2769, opened 2022-11-07
walredo process management: handle every error on the kill() and drop path

problame:
If we're not calling kill() before dropping the PostgresRedoProcess, we currently leak it.
That's most likely the root cause for #2761.
This patch
1. adds an error log message for that case and
2. adds error handling for all errors on the kill() path. If we're a `testing` build, we panic. Otherwise, we log an error and leak the process.

The error handling changes (2) are necessary to conclusively state that the root cause for #2761 is indeed (1). If we didn't have them, the root cause could be missing error handling instead.

To make the log messages useful, I've added tracing::instrument attributes that log the tenant_id and PID. That helps mapping back the PID of `defunct` processes to pageserver log messages. Note that a defunct process's `/proc/$PID/` directory isn't very useful. We have left little more than its PID.

Once we have validated the root cause, we'll find a fix, but that's still an ongoing discussion.

refs https://github.com/neondatabase/neon/issues/2761

### DIFF
--- a/pageserver/src/walredo.rs
+++ b/pageserver/src/walredo.rs
@@ -697,7 +697,7 @@ impl PostgresRedoProcess {
             ($file:ident) => {{
                 let res = set_nonblock($file.as_raw_fd());
                 if let Err(e) = &res {
-                    error!(error = %e, concat!("set_nonblock failed on ", stringify!($file)));
+                    error!(error = %e, file = stringify!($file), pid = child.id(), "set_nonblock failed");
                 }
                 res
             }};
--- a/pageserver/src/walredo.rs
+++ b/pageserver/src/walredo.rs
@@ -689,17 +689,22 @@ impl PostgresRedoProcess {
             datadir.display()
         );
 
-        // NB: don't return an Err() from here on without kill()ing and wait()ing
-        //     for the child. Once we return it inside the Ok(...), it's the callers
-        //     responsibility to call Self::kill().
-
         let stdin = child.stdin.take().unwrap();
         let stdout = child.stdout.take().unwrap();
         let stderr = child.stderr.take().unwrap();
 
-        set_nonblock(stdin.as_raw_fd()).unwrap();
-        set_nonblock(stdout.as_raw_fd()).unwrap();
-        set_nonblock(stderr.as_raw_fd()).unwrap();
+        macro_rules! set_nonblock_or_log_err {
+            ($file:ident) => {{
+                let res = set_nonblock($file.as_raw_fd());
+                if let Err(e) = &res {
+                    error!(error = %e, concat!("set_nonblock failed on ", stringify!($file)));
+                }
+                res
+            }};
+        }
+        set_nonblock_or_log_err!(stdin)?;
+        set_nonblock_or_log_err!(stdout)?;
+        set_nonblock_or_log_err!(stderr)?;
 
         Ok(PostgresRedoProcess {
             tenant_id,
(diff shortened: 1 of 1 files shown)


hlinnaka (approved):
Seems OK as a way to find the root cause, but in the long run we should not require the callers to always call `kill`

pageserver/src/walredo.rs
```diff
@@ -838,6 +894,14 @@ impl PostgresRedoProcess {
     }
 }
 
+impl Drop for PostgresRedoProcess {
+    fn drop(&mut self) {
+        if !self.called_kill {
+            error!(tenant_id=%self.tenant_id, pid = %self.child.id(), "dropping PostgresRedoProcess that wasn't killed, likely causing defunct postgres process");
+        }
```
koivunej, line 901: We could do the kill + try_wait.

koivunej (changes requested):

pageserver/src/walredo.rs
problame: We'll do the fix in a separate change, once we know this is the root cause.

pageserver/src/walredo.rs
```diff
@@ -681,30 +684,82 @@ impl PostgresRedoProcess {
...
+        self.called_kill = true;
+
+        // If false, we will panic upon errors in this function.
+        // If there are any wal-redo processes, they will be cleaned up like so:
+        // 1. Our write-side pipe FDs will be closed by kernel.
+        // 2. In response, walredo process will eventually hit SIGPIPE, either
+        //    when writing a result to stdout or reading a new command from stdin.
+        // 3. walredo process will exit and the supervisor process (systemd) will reap it.
+        let allow_leaks = cfg!(feature = "testing");
+
```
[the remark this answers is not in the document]
problame: We can think about this when we get to the actual fix.
Right now, I don't want to change the behavior except logging, so that we can confirm the root cause.

In general, I don't think it matters because we send a SIGKILL before .wait()ing. What tokio does is to avoid stupid mistakes like

```
process = spawn process with piped stdin
process.stdin.as_mut_ref().write(...)
# forget to close process.stdin
process.wait()
```

If the child process is a simple loop that reads stdin until EOF, this will cause the parent to wait forever.
Tokio papers over the error by closing stdin.

pageserver/src/walredo.rs
```diff
@@ -689,22 +693,59 @@ impl PostgresRedoProcess {
...
+                let res = set_nonblock($file.as_raw_fd());
+                if let Err(e) = &res {
+                    error!(error = %e, concat!("set_nonblock failed on ", stringify!($file)));
+                }
+                res
+            }};
+        }
+        set_nonblock_or_log_err!(stdin)?;
+        set_nonblock_or_log_err!(stdout)?;
+        set_nonblock_or_log_err!(stderr)?;
```
koivunej, line 707: no need to change on this PR but would much rather have the logging happen only at the site of handling the error.

one could do this instead for the same effect:

```suggestion
        macro_rules! set_nonblock_with_target {
            ($file:ident) => {{
                set_nonblock($file.as_raw_fd())
                  .context(concat!("set_nonblock failed on ", stringify!($file)))
            }};
        }
        }
        set_nonblock_with_target!(stdin)?;
        set_nonblock_with_target!(stdout)?;
        set_nonblock_with_target!(stderr)?;
```

But as we discussed, most of the comments root cause was the panic!=abort confusion. This could even just be left as it was.

pageserver/src/walredo.rs
koivunej: Recording some of the discussion we had; it's unlikely that if a kill was really needed the `try_wait` would "see" it right away, would have to have some sleep+try_wait loop, or just be able to wait for the pid non-blockingly like with tokio.

But yeah resolving this.

koivunej (approved):

pageserver/src/walredo.rs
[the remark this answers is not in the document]
koivunej: Ack, just recording that I thought it was SIGTERM instead but we checked it should be SIGKILL. Also, luckily we don't target windows where I've ran into this deadlock :)

pageserver/src/walredo.rs
problame: I generally agree, but then we lose the tracing span context, don't we?
I.e., 
```
 #[instrument(skip_all,fields(tenant_id=%tenant_id, pg_version=pg_version))]
```

That being said, I guess we should also be logging the PID. Will push another fixup.

koivunej: Oki, good point about the additional context. Maybe it is better here, in case we see that many defunct processes end up being created because of this handling...

PR #2777, opened 2022-11-08
Add perf test instructions

bojanserafimov:
We should all know these things by now but IMO it's useful for onboarding.

Also should help people outside Neon, who might run a debug build and wonder why it's slow

Resolves https://github.com/neondatabase/neon/issues/2577

### DIFF
--- a/test_runner/performance/README.md
+++ b/test_runner/performance/README.md
@@ -1,3 +1,19 @@
+# Running locally
+
+First make a release build. The profiling flag is optional, used only for tests that
+generate flamge graphs. The `-s` flag just silences a lot of output, and makes it
+easier to see if you have compile errors without scrolling up.
+`BUILD_TYPE=release CARGO_BUILD_FLAGS="--features=testing,profiling" make -s -j8`
+
+Then run the tests
+`NEON_BIN=./target/release poetry run pytest test_runner/performance"`
+
+Some handy pytest flags for local development:
+- `-x` tells pytest to stop on first error
+- `-s` shows test output
+- `-k` selects a test to run
+- `--timeout=0` disables our default timeout of 300s (see `setup.cfg`)
+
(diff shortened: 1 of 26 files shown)


hlinnaka (approved):

test_runner/performance/README.md
```diff
@@ -1,3 +1,19 @@
+# Running locally
+
+First make a release build. The profiling flag is optional, used only for tests that
+generate flamge graphs. The `-s` flag just silences a lot of output, and makes it
+easier to see if you have compile errors without scrolling up.
+`BUILD_TYPE=release CARGO_BUILD_FLAGS="--features=testing,profiling" make -s -j8`
```
alexstanovoy, line 6: This doesn't work in my case. Even `cargo build --features profiling` with the following error:
https://pastebin.com/d1GWtSKx
Haven't investigated that, will try in the morning.

alexstanovoy (changes requested):

test_runner/performance/README.md
bojanserafimov: Looks like the libc crate exports different types on linux (`timer_t`) vs osx (`time_t`), and we use `timer_t` in pprof https://github.com/neondatabase/pprof-rs/blob/wallclock-profiling/src/timer.rs#L24

bojanserafimov: I added a note explaining this.

test_runner/performance/README.md
```diff
@@ -1,3 +1,22 @@
+# Running locally
+
+First make a release build. The profiling flag is optional, used only for tests that
+generate flamge graphs. The `-s` flag just silences a lot of output, and makes it
```
LizardWizzard, line 4: typo
```suggestion
generate flame graphs. The `-s` flag just silences a lot of output, and makes it
```

PR #2797, opened 2022-11-10
👀 what happens

kelvich:
Not intended for commit, just to run tests with logical wal level

### DIFF
--- a/control_plane/src/compute.rs
+++ b/control_plane/src/compute.rs
@@ -287,7 +287,7 @@ impl PostgresNode {
         conf.append("shared_buffers", "1MB");
         conf.append("fsync", "off");
         conf.append("max_connections", "100");
-        conf.append("wal_level", "replica");
+        conf.append("wal_level", "logical");
         // wal_sender_timeout is the maximum time to wait for WAL replication.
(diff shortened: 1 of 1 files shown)


knizhnik:
Looks like this failures are caused by "redundant" assertion in walingest.rs:
```
--- a/pageserver/src/walingest.rs
+++ b/pageserver/src/walingest.rs
@@ -379,7 +379,7 @@ impl<'a> WalIngest<'a> {
                 }
             } else if info == pg_constants::XLOG_HEAP_DELETE {
                 let xlrec = XlHeapDelete::decode(buf);
-                assert_eq!(0, buf.remaining());
+                //assert_eq!(0, buf.remaining());
                 if (xlrec.flags & pg_constants::XLH_DELETE_ALL_VISIBLE_CLEARED) != 0 {
                     new_heap_blkno = Some(decoded.blocks[0].blkno);
                 }
```
Delete WAL record contains more data in case of logical replication.

knizhnik:
Isolation tests is still not passed with `wal_level=logical`.
But it seems to be related with existed VM/FSM truncation problem.
It is passed in https://github.com/neondatabase/neon/pull/2609 (which is still not merged).

Issue #2683, opened 2022-10-24
Epic: Layer checksums


koivunej:
Follow-up to #2456, which started collecting index file metadata (just the size). Related to #987 but this isn't about page checksums but file verification with hashing.

> * file checksums (at least, as a placeholder to do the format changes once) to later fill with the values, compatible with what we have in AWS S3

_Originally posted by @SomeoneToIgnore in https://github.com/neondatabase/neon/issues/2456#issuecomment-1248384538_

My ideas so far:

- collect $hash during startup for all files and post-for created files (l0, compacted)
- use $hash with upload
    - add handling for upload failing for hash mismatch
- investigate if $hash can be read during download (or, an AsyncWrite wrapper) from remote metadata
    - if can be read, LocalFS will need to start storing these
- refactor LayerFileMetadata matching
    - currently this is done in three places, should de-duplicate
    - de-duplication needed even before solving `Option` problem in `IndexFileMetadata`

S3 supports SHA2-256 which we already have as a dependency through `sha2`.

andresrsanchez:
Hey @koivunej are you with this task or can i try to tackle it? 

Thanks!

koivunej:
@andresrsanchez feel free to work on this. If there are any questions, feel free to ping me!

shanyp:
@koivunej still relevant ?

koivunej:
This is very much relevant, but not easy.

LizardWizzard:
Worth checking if s3 sdk now supports new generation of checksum passing API. This allows to cheaply verify checksums without reading file twice. See https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html the `Using trailing checksums` paragraph

LizardWizzard:
Some more context: https://github.com/awslabs/aws-sdk-rust/discussions/539#discussioncomment-2784356

shanyp:
@LizardWizzard once this is implemented by the sdk, could we just toggle the check for checksum validation?

koivunej:
https://docs.rs/aws-sdk-s3/latest/aws_sdk_s3/operation/put_object/builders/struct.PutObjectFluentBuilder.html#method.checksum_sha256 exists already in the most recent (which may or may not yet use).

Looking at the docs for get_object it's not entirely clear for me how to request the hashes.

What we'd need in pageserver is then:
1. record the sha2-256 during creating of DeltaLayer and ImageLayer
[...]

PR #2802, opened 2022-11-14
Introduce aws-sdk-rust as rusoto S3 replacement

andresrsanchez:
refer #2683 
- First step to introduce checksum metadata leveraging the checksum calculation to S3.
- `aws-smithy-http`: Needed because of `SdkBody` see https://github.com/awslabs/smithy-rs/issues/1759
- `aws-types`: Needed because of `SharedCredentialsProvider`, the recommended way from aws is something like `aws_config::from_env().region("us-east-1").load().await` but that is problematic because of:

    - `sync -> async ` in the creation of S3Client and i don't want to change the signature of any method in this class.
    - We do not need the four default steps in https://github.com/awslabs/aws-sdk-rust/blob/main/sdk/aws-config/src/default_provider/credentials.rs#L235

- `Hyper`: Similar to what's currently doing Rusoto in https://github.com/rusoto/rusoto/blob/master/rusoto/signature/src/signature.rs#L59 to stream the body, see also https://github.com/awslabs/aws-sdk-rust/discussions/361 

Need to confirm all tests work in the pipeline and check compilation times outside my dev environment.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -170,6 +170,333 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
+name = "aws-config"
+version = "0.51.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "56a636c44c77fa18bdba56126a34d30cfe5538fe88f7d34988fa731fee143ddd"
+dependencies = [
+ "aws-http",
+ "aws-sdk-sso",
+ "aws-sdk-sts",
+ "aws-smithy-async",
+ "aws-smithy-client",
+ "aws-smithy-http",
+ "aws-smithy-http-tower",
+ "aws-smithy-json",
+ "aws-smithy-types",
+ "aws-types",
+ "bytes",
+ "hex",
+ "http",
+ "hyper",
+ "ring",
+ "time 0.3.15",
+ "tokio",
+ "tower",
+ "tracing",
+ "zeroize",
+]
+
+[[package]]
+name = "aws-endpoint"
+version = "0.51.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6ca8f374874f6459aaa88dc861d7f5d834ca1ff97668eae190e97266b5f6c3fb"
+dependencies = [
+ "aws-smithy-http",
+ "aws-smithy-types",
+ "aws-types",
+ "http",
+ "regex",
+ "tracing",
+]
+
+[[package]]
+name = "aws-http"
+version = "0.51.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "78d41e19e779b73463f5f0c21b3aacc995f4ba783ab13a7ae9f5dfb159a551b4"
+dependencies = [
+ "aws-smithy-http",
+ "aws-smithy-types",
+ "aws-types",
+ "bytes",
+ "http",
+ "http-body",
+ "lazy_static",
+ "percent-encoding",
+ "pin-project-lite",
+ "tracing",
+]
+
+[[package]]
+name = "aws-sdk-s3"
+version = "0.21.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a9f08665c8e03aca8cb092ef01e617436ebfa977fddc1240e1b062488ab5d48a"
+dependencies = [
+ "aws-endpoint",
+ "aws-http",
+ "aws-sig-auth",
+ "aws-sigv4",
+ "aws-smithy-async",
+ "aws-smithy-checksums",
+ "aws-smithy-client",
+ "aws-smithy-eventstream",
+ "aws-smithy-http",
+ "aws-smithy-http-tower",
+ "aws-smithy-types",
+ "aws-smithy-xml",
+ "aws-types",
+ "bytes",
+ "bytes-utils",
+ "http",
+ "http-body",
+ "tokio-stream",
+ "tower",
+ "tracing",
+]
+
+[[package]]
+name = "aws-sdk-sso"
+version = "0.21.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "86dcb1cb71aa8763b327542ead410424515cff0cde5b753eedd2917e09c63734"
+dependencies = [
+ "aws-endpoint",
+ "aws-http",
+ "aws-sig-auth",
+ "aws-smithy-async",
+ "aws-smithy-client",
+ "aws-smithy-http",
+ "aws-smithy-http-tower",
+ "aws-smithy-json",
+ "aws-smithy-types",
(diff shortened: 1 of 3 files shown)


SomeoneToIgnore:
Had tested it on a VM and local mac, and saw no difference in compilation times:
```
time env RUSTC_WRAPPER="" cargo build --all --timings [--release]
-------------

Mac

old: 
    Finished dev [optimized + debuginfo] target(s) in 2m 50s
env RUSTC_WRAPPER="" cargo build --all --timings  1438.32s user 58.24s system 876% cpu 2:50.81 total

old, release:
    Finished release [optimized + debuginfo] target(s) in 2m 04s
env RUSTC_WRAPPER="" cargo build --all --release --timings  1021.31s user 42.76s system 856% cpu 2:04.28 total

new: 
    Finished dev [optimized + debuginfo] target(s) in 2m 48s
env RUSTC_WRAPPER="" cargo build --all --timings  1462.21s user 59.64s system 902% cpu 2:48.57 total

new, release:
    Finished release [optimized + debuginfo] target(s) in 2m 16s
env RUSTC_WRAPPER="" cargo build --all --release --timings  1180.12s user 47.42s system 897% cpu 2:16.79 total

-------------

Linux

old:
    Finished dev [optimized + debuginfo] target(s) in 6m 39s
env RUSTC_WRAPPER="" cargo build --all --timings  2978.42s user 111.40s system 773% cpu 6:39.22 total

old, release:
    Finished release [optimized + debuginfo] target(s) in 4m 40s
env RUSTC_WRAPPER="" cargo build --all --release --timings  2085.85s user 79.98s system 771% cpu 4:40.78 total

new: 
    Finished dev [optimized + debuginfo] target(s) in 6m 41s
env RUSTC_WRAPPER="" cargo build --all --timings  3012.22s user 112.60s system 777% cpu 6:41.83 total

new, release:
    Finished release [optimized + debuginfo] target(s) in 4m 41s
env RUSTC_WRAPPER="" cargo build --all --release --timings  2100.62s user 80.31s system 775% cpu 4:41.35 total
```

SomeoneToIgnore:
I've attached the timing profiles of the release builds:

Interesting, that despite the build time not changing, new S3-related crate takes longer to compile generally.
image

image


[cargo_build_timings.zip](https://github.com/neondatabase/neon/files/10004985/cargo_build_timings.zip)

SomeoneToIgnore (commented):
LGTM, let's see the tests running.

Two notes:

1. I've noticed that you've added `checksum_*` usages in `s3_bucket.rs` but they are unused now.

Let's keep those reverted for the next iteration, where you use the checksums actually?
We switch frameworks and that could impact things such as performance, better to do that morae incremental, if possible, to reduce the number of things that could go wrong.

For instance, I have no idea what are the limitations of the checksums' requests: why do we need to enable them separately? Is it more costly or allows less requests per second? We also have some objects that were not `put` with the checksums enabled — should there be some migration?
IMO, too many questions and I'd rather kept the checksums enabling separately.

2. `cargo hakari generate` is needed to be run on your code to update the `workspace_hack` dependencies, I think the CI Rust style checks will fail now due to that.

SomeoneToIgnore:
I see an increased failure rate in the tests, check out the Allure reports, there are logs (pytest, pageserver, safekeeper, pg, etc.) in the attachments per test.

image


Debug: https://neon-github-public-dev.s3.amazonaws.com/reports/pr-2812/debug/3463396532/index.html#categories/8fb3a91ba5aaf9de24cc8a92edc82b5d

Release: https://neon-github-public-dev.s3.amazonaws.com/reports/pr-2812/release/3463396532/index.html#suites/3de20ced56df482626cff043a3f2c817

There's also a docker-compose set-up and test we have in our repo, and while that does not fail explicitly (since PS could work without the remote storage), it shows some interesting errors:

```
docker-compose-pageserver-1        | 2022-11-14T16:50:09.593840Z ERROR storage_sync_loop:storage_sync_loop_step:process_sync_task_batch{sync_id=f75d15f012765a22bdda57cdeea17eda/9d2ae9dcaea4e1aff81d7ed6be860ec8}:upload_timeline_data: Failed to upload a layer for timeline f75d15f012765a22bdda57cdeea17eda/9d2ae9dcaea4e1aff81d7ed6be860ec8: Failed to upload layer file for f75d15f012765a22bdda57cdeea17eda/9d2ae9dcaea4e1aff81d7ed6be860ec8
[685](https://github.com/neondatabase/neon/actions/runs/3463396532/jobs/5783718923#step:4:686)
docker-compose-pageserver-1        | 
[686](https://github.com/neondatabase/neon/actions/runs/3463396532/jobs/5783718923#step:4:687)
docker-compose-pageserver-1        | Caused by:
[687](https://github.com/neondatabase/neon/actions/runs/3463396532/jobs/5783718923#step:4:688)
docker-compose-pageserver-1        |     0: Failed to upload from '/data/.neon/tenants/f75d15f012765a22bdda57cdeea17eda/timelines/9d2ae9dcaea4e1aff81d7ed6be860ec8/000000000000000000000000000000000000-FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF__0000000001696070-00000000016960E9' to storage path '"pageserver/tenants/f75d15f012765a22bdda57cdeea17eda/timelines/9d2ae9dcaea4e1aff81d7ed6be860ec8/000000000000000000000000000000000000-FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF__0000000001696070-00000000016960E9"'
[688](https://github.com/neondatabase/neon/actions/runs/3463396532/jobs/5783718923#step:4:689)
docker-compose-pageserver-1        |     1: failed to construct request: error during request construction: Only request bodies with a known size can be checksum validated.
[689](https://github.com/neondatabase/neon/actions/runs/3463396532/jobs/5783718923#step:4:690)
docker-compose-pageserver-1        |     2: error during request construction: Only request bodies with a known size can be checksum validated.
[690](https://github.com/neondatabase/neon/actions/runs/3463396532/jobs/5783718923#step:4:691)
docker-compose-pageserver-1        |     3: Only request bodies with a known size can be checksum validated.
[691](https://github.com/neondatabase/neon/actions/runs/3463396532/jobs/5783718923#step:4:692)
```

https://github.com/neondatabase/neon/actions/runs/3463396532/jobs/5783718923#step:4:685

andresrsanchez:
Thanks for the review @SomeoneToIgnore! 
Yep i trimmed the `default-features` from aws-config because initially the results were much worse... 

I will remove all the checksum thing and check the errors, thanks!!!!

andresrsanchez:
refer: https://github.com/awslabs/aws-sdk-rust/issues/668

SomeoneToIgnore (approved):
I think both libraries are bad and good in relatively similar way, when combined, so have not found anything against the transition, since it eventually has to happen due to `rusoto` maintenance status.
The change is isolated enough also, now especially, when no extra checksum logic is mixed inside 🙂 

I will let it stay for a while in case some objections appear, but the tests (even with real S3 bucket) seems to pass, the compilation time seems to be almost unchanged (still, not good in total).

Thank you for the contribution!
Interesting to see how checksums are related to this.

PR #2816, opened 2022-11-15
perf: simple walredo bench

koivunej:
adds a simple walredo bench to allow some comparison of the walredo throughput.

Cc: #1339, #2778


koivunej:
The benchmark shows in this form that when we add more threads, the time to process grows linearly -- more work, more time. With pipelining (#2778) we see that single threaded use is a bit slower (6%), but already with 4 threads pipelining wins by 12%, but then tapers off to 10% with 8 and 16 threads on a 12 physical core system.

Logging the total write buffer sizes reveals:
- `short` is 1132 bytes
- `medium` is 26393 bytes

knizhnik (approved):

koivunej:
While artificially making walredo slower by adding a `nanosleep` of `1ms` the benefits from pipelining get smaller on `medium` input (6% at 16 threads) which is surprising. With `small` input the benefit is only 3% at 16 threads.

Perhaps this would need a mixed scenario, or looking at the percentage is wrong here. The time distributions are distinct for `small` input.

koivunej:
> The performance difference is (approximately) `1 - (WalRedo + communication') / (WalRedo + communication)`, where `WalRedo` is the processing time, `communication` the original time we spend on communicating between threads (writing, reading, waiting), and `communication'` the new (pipelined) time we spend communicating between threads.
> 
> If you increase `WalRedo` by nanosleep-ing, the relative impact of the difference between `communication` and `communication'` will decrease and the fraction will tend to 1 as WalRedo is increased to infinity.

I was expecting the artificial delay to allow running at the limit of the pipe but you make a good point about how the opposite happens, but wasn't really thinking straight. Most of the "trial" time was spent on hunting which syscall does the nanosleep end up using (clock_nanosleep). Thanks for thinking this through!

koivunej:
Rebased to see if ci works this time around, no changes, just squashed the commits together.

PR #2829, opened 2022-11-16
Add neon_broker binary.

arssher:
Which ought to replace neon_broker. This patch only adds the binary and adjusts Dockerfile to include it; subsequent ones will add deploy of helm chart and the actual replacement.

It is a simple and fast pub-sub message bus. In this patch only safekeeper message is supported, but others can be easily added.

Compilation now requires protoc to be installed. Installing protobuf-compiler package is fine for Debian/Ubuntu.

ref
https://github.com/neondatabase/neon/pull/2733
https://github.com/neondatabase/neon/issues/2394


broker/src/bin/neon_broker.rs
```diff
@@ -0,0 +1,563 @@
...
+impl Publisher {
+    // Send msg to relevant subscribers.
+    pub fn send_msg(&mut self, msg: &SafekeeperTimelineInfo) -> Result<(), Status> {
+        // send message to subscribers for everything
+        let mut cleanup_subs_to_all = false;
+        for sub in self.subs_to_all.iter() {
+            match sub.try_send(msg.clone()) {
+                Err(TrySendError::Full(_)) => {
+                    trace!("dropping message, channel is full");
+                }
```
koivunej, line 358: This Drop when full behaviour makes me wonder if per publisher [tokio::sync::broadcast](https://docs.rs/tokio/latest/tokio/sync/broadcast/index.html) channel would had been more applicable. On being full it "drops" oldest messages and resumes from oldest still reachable, compared to this which drops the later messages but continues to send older, skipping some in the middle.

Memory-wise it would allow all subscribers to follow a single queue instead of having their own, removing some cleanup lines at the cost of having more complexity inside subscriber's `try_stream!`, unsure how to handle the `subs_to_all` at least right away... EDIT: I don't think that's an issue since one subscriber only subscribes to all or specific timelines.

Reading this more I think it would help with not needing to be concerned about removal or even book keeping of subscribers.

If you already considered it but it cannot work for some reason I cannot see, could a comment be added somewhere?

broker/src/bin/neon_broker.rs
```diff
@@ -0,0 +1,563 @@
...
+            .subscription_key
+            .ok_or_else(|| Status::new(Code::InvalidArgument, "missing subscription key"))?;
+        let sub_key = SubscriptionKey::from_proto_subscription_key(proto_key)?;
+        let mut subscriber = self.registry.register_subscriber(sub_key).await;
+
+        // transform rx into stream with item = Result, as method result demands
+        let output = async_stream::try_stream! {
+            while let Some(info) = subscriber.sub_rx.recv().await {
+                    yield info
+                }
```
koivunej, line 464: Small nit about the formatting since rustfmt doesn't touch these blocks.

```suggestion
                yield info
            }
```

broker/src/bin/neon_broker.rs
```diff
@@ -0,0 +1,563 @@
...
+
+// Announcements subscriber sends to publisher(s) asking it to stream to the
+// provided channel, or forget about it, releasing memory.
+#[derive(Clone)]
+enum SubAnnounce {
+    AddAll(Sender<SafekeeperTimelineInfo>), // add subscription to all timelines
+    AddTimeline(TenantTimelineId, SubSender), // add subsciption to the specific timeline
+    RemoveTimeline(TenantTimelineId, SubId), // remove subscription to the specific timeline
+                                            // RemoveAll is not needed as publisher will notice closed channel while
+                                            // trying to send the next message.
```
koivunej, line 115: Since this is new code a nit: would prefer to see rustdoc friendly comments before the variants.

```suggestion
    /// Add subscription to all timelines
    AddAll(Sender<SafekeeperTimelineInfo>),
    /// Add subscription to the specific timeline
    AddTimeline(TenantTimelineId, SubSender),
    /// Remove subscription to the specific timeline
    RemoveTimeline(TenantTimelineId, SubId),
    
    // RemoveAll is not needed as publisher will notice closed channel while
    // trying to send the next message.
```

koivunej (commented):
I think this is looking good, with very nice comments. Added some comment style nits since this is new code, and a question about tokio::sync::broadcast usage instead.

Could some sort of smoke test be added, for example that a published update is seen both by an "all" subscriber and a specific tenant-timeline subscriber?

broker/src/bin/neon_broker.rs
```diff
@@ -0,0 +1,563 @@
...
+
+impl Publisher {
+    // Send msg to relevant subscribers.
+    pub fn send_msg(&mut self, msg: &SafekeeperTimelineInfo) -> Result<(), Status> {
+        // send message to subscribers for everything
+        let mut cleanup_subs_to_all = false;
+        for sub in self.subs_to_all.iter() {
+            match sub.try_send(msg.clone()) {
+                Err(TrySendError::Full(_)) => {
+                    trace!("dropping message, channel is full");
```
SomeoneToIgnore, line 357: I think we did not change that and the other `trace!` to warn (or added the metrics, or something else)
See https://github.com/neondatabase/neon/pull/2733#discussion_r1012156020

broker/readme.md
```diff
@@ -0,0 +1,4 @@
+```
```
SomeoneToIgnore, line 1: The readme is still odd, see https://github.com/neondatabase/neon/pull/2733#discussion_r1011764691 
I think, current commit message is better than this readme 🙂 

> Compilation now requires protoc to be installed

This belongs to the main readme, where we list all things needed for compilation: https://github.com/neondatabase/neon/blame/c4b417ecdbc030f245673362dc23fbf4f279180e/README.md#L37

I think we should change all installation instructions, for Mac, the package is named `protobuf-c`

SomeoneToIgnore (commented):
Not sure when the "formal" addition of the tool is supposed to happen, so have transferred important unresolved comments from https://github.com/neondatabase/neon/pull/2733

IMO, we should adjust READMEs now, but ignore me if you plan to do that later.

broker/proto/broker.proto
```diff
@@ -0,0 +1,44 @@
...
+    // Subscribe to safekeeper updates.
+    rpc SubscribeSafekeeperInfo(SubscribeSafekeeperInfoRequest) returns (stream SafekeeperTimelineInfo) {};
+
+    // Publish safekeeper updates.
+    rpc PublishSafekeeperInfo(stream SafekeeperTimelineInfo) returns (google.protobuf.Empty) {};
+}
+
+message SubscribeSafekeeperInfoRequest {
+    oneof subscription_key {
+        google.protobuf.Empty all = 1; // subscribe to everything
```
petuhovskiy, line 17: I'm not sure if `Empty` inside oneof will work. As I understand, oneof requires at least one non-zero value to be passed in one of the fields and `Empty` type is always zero-value, so you probably can't set `all` field and subscribe to everything? I may be wrong here, because I've never used oneof before.

broker/src/metrics.rs
```diff
@@ -0,0 +1,22 @@
+//! Broker metrics.
+
+use metrics::{register_int_gauge, IntGauge};
+use once_cell::sync::Lazy;
+
+pub static NUM_PUBS: Lazy<IntGauge> = Lazy::new(|| {
+    register_int_gauge!("broker_num_pubs", "Number of publications")
```
petuhovskiy, line 7:
```suggestion
    register_int_gauge!("broker_active_publishers", "Number of publications")
```

broker/src/metrics.rs
```diff
@@ -0,0 +1,22 @@
...
+        .expect("Failed to register metric")
+});
+
+pub static NUM_SUBS_TIMELINE: Lazy<IntGauge> = Lazy::new(|| {
+    register_int_gauge!(
+        "broker_num_subs_timeline",
```
petuhovskiy, line 13:
```suggestion
        "broker_per_timeline_active_subscribers",
```

broker/src/metrics.rs
```diff
@@ -0,0 +1,22 @@
...
+        "Number of subsciptions to particular tenant timeline id"
+    )
+    .expect("Failed to register metric")
+});
+
+pub static NUM_SUBS_ALL: Lazy<IntGauge> = Lazy::new(|| {
+    register_int_gauge!("broker_num_subs_all", "Number of subsciptions to all keys")
```
petuhovskiy, line 20:
```suggestion
    register_int_gauge!("broker_all_keys_active_subscribers", "Number of subsciptions to all keys")
```

petuhovskiy:
We should add `protoc` installation instructions to the README.md in the root of the repo

petuhovskiy:
https://github.com/fullstorydev/grpcurl can be used as a client for the broker. For example, this query will subscribe for everything:

```
grpcurl -proto broker.proto -d '{"all":{}}' -plaintext localhost:50051 neon_broker.NeonBroker/SubscribeSafekeeperInfo
```

broker/proto/broker.proto
petuhovskiy: It works, need to reread protobuf docs to understand what's going on.

petuhovskiy (approved):
`neon_broker.rs` is hard to read even on the 2nd try :)
I found nothing wrong in the code, so approve from me.

broker/src/bin/neon_broker.rs
petuhovskiy: Nice, [tokio::sync::broadcast](https://docs.rs/tokio/latest/tokio/sync/broadcast/index.html) looks like a ready implementation of what a broker does, i.e. Pub/Sub. When we will rewrite the broker code, we should definitely consider using tokio::sync::broadcast.

broker/src/bin/neon_broker.rs
```diff
@@ -0,0 +1,563 @@
...
+    // console_subscriber::init();
+
+    let args = Args::parse();
+    ARGS.set(args).expect("ARGS already initialized");
+
+    logging::init(LogFormat::from_config(&get_args().log_format)?)?;
+    info!("version: {GIT_VERSION}");
+
+    let registry = Registry {
+        shared_state: Arc::new(Mutex::new(SharedState {
```
LizardWizzard, line 509: Can be #[derive(Default)] I guess

broker/src/bin/neon_broker.rs
```diff
@@ -0,0 +1,563 @@
...
+
+        let mut stream = request.into_inner();
+
+        loop {
+            select! {
+                msg = stream.next() => {
+                    match msg {
+                        Some(Ok(msg)) => {publisher.send_msg(&msg)?;},
+                        Some(Err(e)) => {return Err(e);}, // grpc error from the stream
+                        None => {break;} // closed stream
```
LizardWizzard, line 434:
```suggestion
                        Some(Ok(msg)) => publisher.send_msg(&msg)?,
                        Some(Err(e)) => return Err(e), // grpc error from the stream
                        None => break // closed stream
```

broker/src/bin/neon_broker.rs
```diff
@@ -0,0 +1,563 @@
...
+        self.registry.unregister_publisher(self);
+    }
+}
+
+struct NeonBrokerImpl {
+    registry: Registry,
+}
+
+#[tonic::async_trait]
+impl NeonBroker for NeonBrokerImpl {
```
LizardWizzard, line 419: NIT: Maybe call it the service `BrokerService`, and then struct becomes just `Broker`, so ends up `impl BrokerService for Broker`

LizardWizzard (dismissed):
Comments are mostly nitpicking/requests for cleanup.

I think before merge this needs to be covered with proper unittests and probably some integrational ones. Should be easy to write them in Rust.

As an idea I think it is worth considering Mutex from parking_lot instead of std one. It doesnt panic and should be faster.

It is sad to see requirement for Vec<u8> for `bytes` protobuf type. Not sure how bad it is, which alternatives should we consider at this point?

broker/src/bin/neon_broker.rs
arssher: Cool, I didn't know it has support for dropping slow receivers messages, will try to switch.

broker/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+edition = "2021"
+
+[features]
+bench = []
+
+[[bin]]
+name = "neon_broker_bench"
+path = "src/bin/bench.rs"
+# build benchmarking binary only if explicitly requested with '--features bench'
+required-features = ["bench"]
```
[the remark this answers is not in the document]
arssher: Checked out, I don't think it fits well the micro benchmarking format, for example it would be nice to print rps, not execution time. Also, #[bench] is unstable (yes, there is Criterion...). But I will add readme to the binary.

broker/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+futures-util = "0.3"
+git-version = "0.3.5"
+humantime = "2.1.0"
+hyper = {version = "0.14.14", features = ["full"]}
+once_cell = "1.13.0"
+prost = "0.11"
+tonic = "0.8"
+tokio = { version = "1.0", features = ["macros", "rt-multi-thread"] }
+# for exploring with tokio-console
+# tokio = { version = "1", features = ["full", "tracing"] }
```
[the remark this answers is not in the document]
arssher: let's just remove altogether, not sure how to require dep features depending on input features %)

broker/src/bin/bench.rs
```diff
@@ -0,0 +1,175 @@
...
+        Some(c) => c,
+        None => NeonBrokerClient::connect(format!("http://{}", DEFAULT_LISTEN_ADDR))
+            .await
+            .unwrap(),
+    };
+
+    // let key = SubscriptionKey::All(());
+    let key = SubscriptionKey::TenantTimelineId(ProtoTenantTimelineId {
+        tenant_id: vec![0xFF; 16],
+        timeline_id: tli_from_u64(i),
```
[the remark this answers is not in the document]
arssher: If you would like to see fixed size array here, then no, protobuf doesn't support them at all. Yes, somewhat disappointing, but here I don't think this matters.

broker/src/bin/neon_broker.rs
```diff
@@ -0,0 +1,563 @@
...
+//!
+//! Subscriptions to 1) single timeline 2) everything are possible. We could add
+//! subscription to set of timelines to save grpc streams, but testing shows
+//! many individual streams is also ok.
+//!
+//! Message is dropped if subscriber can't consume it, not affecting other
+//! subscribers.
+//!
+//! Only safekeeper message is supported, but it is not hard to add something
+//! else with templating.
```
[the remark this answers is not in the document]
arssher: I mean pub-sub code is agnostic to the carried message (`SafekeeperTimelineInfo` here), it is easy to extract it to template arg if we have another (e.g. similar thing for pageservers). Or alternatively we can use protobuf Any, as Arthur suggested.

broker/src/bin/neon_broker.rs
```diff
@@ -0,0 +1,563 @@
...
+    /// HTTP/2 keepalive interval.
+    #[arg(long, value_parser= humantime::parse_duration, default_value = "5000ms")]
+    http2_keepalive_interval: Duration,
+    /// Format for logging, either 'plain' or 'json'.
+    #[arg(long, default_value = "plain")]
+    log_format: String,
+}
+
+// Args is put into OnceCell to make conf available everywhere.
+static ARGS: OnceCell<Args> = OnceCell::new();
```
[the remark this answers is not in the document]
arssher: Well, simplifies access, but here either is fine to me.

broker/src/bin/neon_broker.rs
```diff
@@ -0,0 +1,563 @@
...
+// SharedState wrapper for post-locking operations (sending to pub_tx chans).
+#[derive(Clone)]
+struct Registry {
+    shared_state: Arc<Mutex<SharedState>>,
+}
+
+impl Registry {
+    // Register new publisher in shared state.
+    pub fn register_publisher(&self) -> Publisher {
+        let (announce_tx, announce_rx) = mpsc::channel(128);
```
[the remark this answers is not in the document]
arssher: yes, they are conceptually different; through none should be large

broker/src/bin/neon_broker.rs
arssher: It would make imports more verbose (already fat), so let's leave as is.

broker/src/bin/bench.rs
```diff
@@ -0,0 +1,175 @@
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::Arc;
+use std::time::{Duration, Instant};
+
+use clap::Parser;
+use neon_broker::neon_broker_proto::neon_broker_client::NeonBrokerClient;
```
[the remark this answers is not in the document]
arssher: `crate` won't work here, as this is binary crate, but I reduced that.

broker/src/bin/bench.rs
```diff
@@ -0,0 +1,175 @@
+use std::sync::atomic::{AtomicU64, Ordering};
```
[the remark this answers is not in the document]
arssher: Added as `-h` flag to the binary.

broker/Cargo.toml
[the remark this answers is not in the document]
LizardWizzard: Please revisit what `harness = false` does. It supports exactly this case. You write usual binary with main function that will be invoked with `cargo bench`. See this example from glommio https://github.com/DataDog/glommio/blob/master/glommio/Cargo.toml#L74

broker/src/bin/bench.rs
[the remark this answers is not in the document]
LizardWizzard: Even without fixed size arrays vec shouldnt be necessary. It should be possible to populate the value from &[u8] or an iterator. There is even an issue in the library for that (not mentioning it to avoid direct linking to it). Two allocations for each message look bad for me

broker/src/bin/neon_broker.rs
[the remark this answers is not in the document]
LizardWizzard: Do you mean generics by templating?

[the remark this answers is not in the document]
arssher: yes

broker/src/bin/neon_broker.rs
LizardWizzard: > It would make imports more verbose (already fat),

But why? As I can see there wont be any difference. And no ugly `Impl` suffix as a plus.

arssher: tonic generated code includes two modules, currently called `neon_broker_server` and `neon_broker_client`, these would be longer. But actually I'm fine with either variant.

broker/src/bin/neon_broker.rs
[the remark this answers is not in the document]
LizardWizzard: Lets use proper terminology then. Its not c++ :). 

As a third option we can also just store message as bytes and let the client decide what needs to be passed here. It can be another protobuf blob, can be json.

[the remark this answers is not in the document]
arssher: changed to generics

broker/Cargo.toml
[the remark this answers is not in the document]
arssher: Checked again, it seems impossible to run `cargo bench` without making it pass `--bench` to the binary, and clap doesn't support arguments with no value %)

arssher:
> https://github.com/fullstorydev/grpcurl can be used as a client for the broker. For example, this query will subscribe for everything:

cool, added to `broker.md`

broker/Cargo.toml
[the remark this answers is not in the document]
arssher: Tried that, everything was more or less ok, but hakari doesn't seem to follow optional features, i.e. it means tokio would be compiled with all (including tracing) features always, which is bad, so reverting.

arssher:
- Added unit tests
- Added `broker.md`
- Went through reviews above
- Went ahead and replaced mpsc with `tokio::broadcast`, per suggestion above, and notify channels with rwlock per @petuhovskiy  suggestion above. Checked that performance is approximately the same, memory consumption is slightly higher on a lot of per-timeline subscriptions (but anyway decent) and obviously much less on many publishers. Saved about ~150 lines of code, and it got simpler.

Let's look at CI, but I consider this mostly ready for merge.

arssher:
Helm chart is being added at
https://github.com/neondatabase/helm-charts/pull/15

broker/Cargo.toml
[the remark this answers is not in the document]
arssher: I was surely wrong, clap handles that, moved to [[bench]].

broker/src/bin/bench.rs
```diff
@@ -0,0 +1,175 @@
...
+    let mut c_min_old = counters
+        .iter()
+        .map(|c| c.load(Ordering::Relaxed))
+        .min()
+        .unwrap_or(0);
+    let mut started_at = None;
+    let mut skipped: u64 = 0;
+    loop {
+        interval.tick().await;
+        // print!(
```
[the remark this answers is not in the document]
LizardWizzard: There are still other places with commented code.

broker/src/bin/bench.rs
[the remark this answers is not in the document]
LizardWizzard: Still do not get it why we cant call it just broker. Imports still look like that `use neon_broker::neon_broker_proto::`. Why not just broker::proto?

broker/src/bin/neon_broker.rs
```diff
@@ -0,0 +1,515 @@
...
+
+        // send message to per timeline subscribers
+        let ttid =
+            parse_proto_ttid(msg.tenant_timeline_id.as_ref().ok_or_else(|| {
+                Status::new(Code::InvalidArgument, "missing tenant_timeline_id")
+            })?)?;
+        if let Some(subs) = shared_state.chans_to_timeline_subs.get(&ttid) {
+            // Err can't happen here, as rx is destroyed after removing from the
+            // map the last subscriber.
+            subs.chan.send(msg.clone()).unwrap();
```
LizardWizzard, line 283: Lets use `.expect(...)` here instead of unwrap

broker/src/bin/bench.rs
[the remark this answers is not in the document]
arssher: removed

broker/src/bin/neon_broker.rs
arssher: done

broker/src/bin/bench.rs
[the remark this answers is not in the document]
arssher: It would be possible if lib exported these names directly in global namespace. However, there is
```
// Code generated by protobuf.
pub mod neon_broker_proto {
    include!("../proto/neon_broker.rs");
}
```

I find it useful that all generated code is put into separate module, helps differentiate with manually written one. I also saw the same structure in all tonic apps/examples I encountered.

broker/src/bin/neon_broker.rs
LizardWizzard: Longer than what? There is a change from NeonBroker to BrokerService and struct can be called just Broker after that. So overall it looks shorter to me.

`neon_broker_proto` module name  is written in our code. So can be just proto. It is in lib.rs so can be either `crate::proto` or `broker::proto`. This is part of lengthy imports discussion above

arssher: > neon_broker_proto module name is written in our code. So can be just proto. 

+1

> There is a change from NeonBroker to BrokerService and struct can be called just Broker after that. So overall it looks shorter to me.

Ok, in this case it won't be longer, but I dislike `BrokerService` names as it generates imports like
```
use neon_broker::proto::broker_service_server::{BrokerService, BrokerServiceServer};
```

So perhaps just `Broker`?

arssher: pondering some more, made it `BrokerService`.

LizardWizzard: There have to be two entities. One is a trait that is generated by tonic and the second one is the implementation struct. So cannot be Broker in both places (or we need to alias something in `use` section). So yeah, `BrokerService` trait and `Broker` struct is IMO is a good way to describe this

broker/src/bin/bench.rs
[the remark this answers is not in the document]
LizardWizzard: I'm ok with separate module, was just wondering why we cant call the module just `proto`. I see you made this change under different discussion item. Thanks!

broker/benches/neon_broker_bench.rs
```diff
@@ -0,0 +1,174 @@
+use std::sync::atomic::{AtomicU64, Ordering};
```
LizardWizzard, line 1: NIT: the file can be named something like `timeline_pubsub` or something similar. `_bench` suffix is redundant to me since now fole lives under `/benches` and `neon_broker` too because everything happens under /broker

broker/src/bin/neon_broker.rs
```diff
@@ -0,0 +1,520 @@
...
+        assert!(
+            matches!(subscriber_2.sub_rx.try_recv(), Ok(msg) if matches!(msg.tenant_timeline_id.as_ref(), Some(ttid) if parse_proto_ttid(ttid).unwrap() == ttid_2))
+        );
+        // but nothing more
+        assert!(
+            matches!(subscriber_2.sub_rx.try_recv(), Err(err) if matches!(err, TryRecvError::Empty))
+        );
+
+        // subscriber_all should receive both messages
+        assert!(matches!(subscriber_all.sub_rx.try_recv(), Ok(..)));
```
LizardWizzard, line 514: Lets check exact messages with `assert_eq`

broker/src/bin/neon_broker.rs
```diff
@@ -0,0 +1,520 @@
...
+        assert!(matches!(subscriber_all.sub_rx.try_recv(), Ok(..)));
+        assert!(
+            matches!(subscriber_all.sub_rx.try_recv(), Err(err) if matches!(err, TryRecvError::Empty))
+        );
```
LizardWizzard, line 518: NIT: a bit more readable to me
```suggestion
        assert_eq!(
            subscriber_all.sub_rx.try_recv().unwrap_err(),
            TryRecvError::Empty
        );
```

broker/benches/neon_broker_bench.rs
```diff
@@ -0,0 +1,174 @@
...
+use std::sync::Arc;
+use std::time::{Duration, Instant};
+
+use clap::Parser;
+use neon_broker::proto::subscribe_safekeeper_info_request::SubscriptionKey;
+use neon_broker::proto::TenantTimelineId as ProtoTenantTimelineId;
+use neon_broker::proto::{SafekeeperTimelineInfo, SubscribeSafekeeperInfoRequest};
+use neon_broker::BrokerClientChannel;
+use neon_broker::DEFAULT_LISTEN_ADDR;
+use tokio::time::{self};
```
vadim2404, line 11: `use tokio::time;` — it seems that you had sth more when developed

LizardWizzard (commented):
Thanks for addressing the comments!

Looks mostly good to me. Minor comments remain, except the need for two extra allocations for each message

***

I would avoid neon prefix completely. In all the names of the files, variables etc. It makes things shorter and is consistent with naming in other components

broker/benches/neon_broker_bench.rs
arssher: Renamed to rps.rs, because that's what it really tests.

arssher:
> I would avoid neon prefix completely.

I considered that, but unlike pageserver/safekeeper, `broker`  feels like too generic term to me. So from this stems binary name `neon_broker` and the rest.

broker/src/bin/neon_broker.rs
arssher: yeah, it also make the check much more concise

petuhovskiy:
WDYT about logging every publisher and subscriber connection with `info!` level? I'd like to have logs like that:
```
INFO Started subscription for {sub_key}, ip={ip}
INFO Finished subscription for {sub_key}, ip={ip}
INFO Started publishing, ip={ip}
INFO Finished publishing, ip={ip}
```

arssher:
I'm +1, but I remember at some point we disabled logging loading timelines on safekeepers thinking that would it be too much log volume. Here it is roughly the same number :)

LizardWizzard:
> Perhaps storage_broker?

If we put control plane in this repo then all storage will be under /storage it may repeat `storage` but we have proxy here anyway. So maybe thats fine

arssher:
Renamed and added remote addr logging.

broker/src/bin/bench.rs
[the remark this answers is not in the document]
LizardWizzard: This still looks like a problem to me. Lets ask someone else to get third opinion

storage_broker/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+            EitherBody::Right(b) => Pin::new(b).poll_data(cx).map(map_option_err),
+        }
+    }
+
+    fn poll_trailers(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Result<Option<hyper::HeaderMap>, Self::Error>> {
+        match self.get_mut() {
+            EitherBody::Left(b) => Pin::new(b).poll_trailers(cx).map_err(Into::into),
```
LizardWizzard, line 100: NIT: Using `.map(Into::into)` is rather confusing to me and I would say this is an anti pattern because its quite hard to read and find actual implementation that will be used. So typically `AnyError::from` is the preferred way.

Might be easier to get away with `Box::new(e) as _` this requires slightly changing bounds of A::Error/B::Error above or as I mentioned with `AnyError::from`

LizardWizzard (commented):
Overall looks good to me. Minor comments. And still this problem with extra allocations

arssher:
> And still this problem with extra allocations

Well, this thingy gives several millions rps on my laptop. Performance/scalability here won't be a problem for a long time, and if it is, I suspect this extra allocation won't be the top priority.

storage_broker/src/lib.rs
arssher: Yes, all EitherBody is somewhat complicated. Tried suggested, but `AnyError::from` requires changing bounds to `A::Error:  std::error::Error + Send + Sync + 'static` in three places, a lot of boilerplate; the same for`Box::new` and `Box::new(e) as _` doesn't help.

storage_broker/benches/rps.rs
```diff
@@ -0,0 +1,174 @@
...
+#[derive(Parser, Debug)]
+#[clap(author, version, about = ABOUT)]
+struct Args {
+    /// Number of publishers
+    #[clap(short = 'p', long, value_parser, default_value_t = 1)]
+    num_pubs: u64,
+    /// Number of subscribers
+    #[clap(short = 's', long, value_parser, default_value_t = 1)]
+    num_subs: u64,
+    // Fake value to satisfy `cargo bench` passing it.
```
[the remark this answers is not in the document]
LizardWizzard: Gotcha, TIL

storage_broker/src/lib.rs
LizardWizzard: It is indeed a bit more verbose, but allows to peek into particular From impl without guessing around Into::into

storage_broker/src/metrics.rs
```diff
@@ -0,0 +1,25 @@
+//! Broker metrics.
+
+use metrics::{register_int_gauge, IntGauge};
+use once_cell::sync::Lazy;
+
+pub static NUM_PUBS: Lazy<IntGauge> = Lazy::new(|| {
+    register_int_gauge!("broker_active_publishers", "Number of publications")
```
lassizci, line 7: Could we prefix these with `storage_broker_` instead of just plain ` broker`? We have `pageserver` metrics with `pageserver_` prefix, `safekeeper` metrics with `safekeeper_` prefix and `proxy` metrics with `proxy_` prefix. I won't insist, though, as it's not that big of a deal.

arssher: Yes, renamed.

PR #2832, opened 2022-11-16
Create one Allure report per attempt, without overwriting

MMeent:
Previously, we overwrote the data of the previous attempt, which would result in us losing access to information of previous runs.


SomeoneToIgnore:
> arbitrary failures in the pipeline

https://github.com/neondatabase/aws/pull/66 + some deployment work hopefully helps drastically with that

bayandin:
> Previously, we overwrote the data of the previous attempt, which would result in us losing access to information of previous runs.

We don't exactly overwrite data but extend the latest runs with results with previous runs; if you open any test in https://neon-github-public-dev.s3.amazonaws.com/reports/main/release/3473978290/index.html#suites/3de20ced56df482626cff043a3f2c817 and switch to `Retries` tab, you see results from previous runs.

Also, there's a button that shows only tests which changed their result in the latest retry

Screenshot 2022-11-16 at 10 36 12

Here's an example of the test with a changed result:
https://neon-github-public-dev.s3.amazonaws.com/reports/main/release/3473978290/index.html#suites/ec4311502db344eee91f1354e9dc839b/6c8546f56d1cdbdf/retries

Does this help?

PR #2838, opened 2022-11-16
Nightly Benchmarks: Move from captest to staging

bayandin:
Migrate Nightly Benchmarks from captest to staging.

- Migrate GitHub Workflows
- Replace `zenith-benchmarker` with regular runners
- Remove `environment` parameter from Neon GitHub Actions, add `postgres_version`
- The only job left on captest is `neon-captest-reuse`, which will be moved to staging after its project migration. 

Ref https://github.com/neondatabase/cloud/issues/2836

### DIFF
--- a/.github/actions/neon-branch-create/action.yml
+++ b/.github/actions/neon-branch-create/action.yml
@@ -8,6 +8,9 @@ inputs:
   project_id:
     desctiption: 'ID of the Project to create Branch in'
     required: true
+  api_host:
+    desctiption: 'Neon API host'
+    default: console.stage.neon.tech
 outputs:
   dsn:
     description: 'Created Branch DSN (for main database)'
@@ -32,7 +35,12 @@ runs:
             --data "{
               \"branch\": {
                 \"name\": \"Created by actions/neon-branch-create; GITHUB_RUN_ID=${GITHUB_RUN_ID} at $(date +%s)\"
-              }
+              },
+              \"endpoints\": [
+                {
+                  \"type\": \"read_write\"
+                }
+              ]
             }")
 
           if [ -z "${branch}" ]; then
@@ -60,8 +68,8 @@ runs:
         host=$(echo $branch | jq --raw-output '.endpoints[0].host')
         echo "host=${host}" >> $GITHUB_OUTPUT
       env:
+        API_HOST: ${{ inputs.api_host }}
         API_KEY: ${{ inputs.api_key }}
-        API_HOST: console.stage.neon.tech
         PROJECT_ID: ${{ inputs.project_id }}
 
     - name: Get Role name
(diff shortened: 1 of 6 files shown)


.github/actions/neon-project-create/action.yml
```diff
@@ -5,12 +5,12 @@ inputs:
...
     required: true
-  environment:
-    desctiption: 'dev (aka captest) or staging'
-    required: true
   region_id:
     desctiption: 'Region ID, if not set the project will be created in the default region'
     required: false
+  postgres_version:
+    desctiption: 'Postgres version; default is 15'
+    required: false
```
SomeoneToIgnore, line 13: Could a `default: 15` field be added here, as in our `run-python-test-set` action? Presumably, we would not need `|| 15` below then.

.github/workflows/benchmarking.yml
```diff
@@ -329,9 +299,9 @@ jobs:
       SAVE_PERF_REPORT: ${{ github.event.inputs.save_perf_report || ( github.ref == 'refs/heads/main' ) }}
       PLATFORM: ${{ matrix.platform }}
 
-    runs-on: [ self-hosted, dev, x64 ]
+    runs-on: [ self-hosted, us-east-2, x64 ]
     container:
-      image: 369495373322.dkr.ecr.eu-central-1.amazonaws.com/rustlegacy:pinned
```
SomeoneToIgnore, line 334: 👍 Is that the last usage? We can remove https://github.com/neondatabase/build/tree/main/rustlegacy then.

SomeoneToIgnore (commented):
Looks nice generally, commented a few oddities.
I'm not aware of the infra behind the changes, so cannot provide a good review though.

.github/actions/neon-branch-create/action.yml
```diff
@@ -85,7 +61,7 @@ runs:
         echo "host=${host}" >> $GITHUB_OUTPUT
       env:
         API_KEY: ${{ inputs.api_key }}
-        API_HOST: ${{ steps.parse-input.outputs.api_host }}
+        API_HOST: console.stage.neon.tech
```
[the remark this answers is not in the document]
bayandin: Unfortunately, composite actions don't support env outside of the steps ([a link](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runs-for-composite-actions)), so I've moved API host to actions input. It also allows using production instances if we decide to.

.github/actions/neon-project-create/action.yml
bayandin: Ah, nice catch! I forget that we can do it :) 
I also moved default region_id to inputs default. Much cleaner now, thanks!

.github/actions/neon-project-create/action.yml
```diff
@@ -61,6 +36,7 @@ runs:
           --data "{
             \"project\": {
               \"name\": \"Created by actions/neon-project-create; GITHUB_RUN_ID=${GITHUB_RUN_ID}\",
+              \"pg_version\": ${POSTGRES_VERSION},
```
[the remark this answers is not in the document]
bayandin: We have defined it below the step.run:
```yml
      env:
        API_HOST: ${{ inputs.api_host }}
        API_KEY: ${{ inputs.api_key }}
        REGION_ID: ${{ inputs.region_id }}
        POSTGRES_VERSION: ${{ inputs.postgres_version }}
```

I prefer converting inputs to env vars and using these vars in a script. It makes it easier to debug the script, which can be copy-pasted into your terminal.

.github/workflows/benchmarking.yml
bayandin: Yep, this was the last one. I'll create a PR with the removal of this image from another repo.

PR #2839, opened 2022-11-16
test_runner: make proxy mgmt port mandatory

bayandin:
Make `mgmt` port mandatory argument for `NeonProxy` (and set it for `static_proxy`) to avoid port collision when tests run in parallel.

### DIFF
--- a/test_runner/fixtures/neon_fixtures.py
+++ b/test_runner/fixtures/neon_fixtures.py
@@ -2073,9 +2073,9 @@ class NeonProxy(PgProtocol):
         self,
         proxy_port: int,
         http_port: int,
+        mgmt_port: int,
         neon_binpath: Path,
         auth_endpoint=None,
-        mgmt_port=None,
     ):
         super().__init__(dsn=auth_endpoint, port=proxy_port)
         self.host = "127.0.0.1"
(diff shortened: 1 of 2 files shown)


funbringer (approved):

hlinnaka (approved):
I don't think this explains the port conflicts we've been seeing though. If you don't specify the `--mgmt` option to the proxy, it uses port 7000. The ports chosen by the "port distributor" in neon_fixtures.py starts at 15000, so they should never conflict with the default proxy mgmt port.

I agree this is a good idea nevertheless.

bayandin:
Not all of them, but these 2 failures were definitely caused by using 7000 in parallel running tests:
- `test_password_hack`: https://neon-github-public-dev.s3.amazonaws.com/reports/main/release/3480613958/index.html#/testresult/91b6fb0597d37c49
- `test_proxy_options`: https://neon-github-public-dev.s3.amazonaws.com/reports/main/release/3480613958/index.html#/testresult/a56cd520fd70ad42

The error is a bit misleading at first glance (at least for me), it mentions a port from the port distributor and has `Error: Address already in use (os error 98)` in stderr logs, but these are messages about 2 different ports:
- `Error: Address already in use (os error 98)` is about 7000, it doesn't allow the proxy to start
- `requests.exceptions.ConnectionError: HTTPConnectionPool(host='127.0.0.1', port=17092)` about 17092, because nothing is listening on this port

Probably we should tune `_wait_until_ready` error message.

PR #2840, opened 2022-11-16
Add remote seqscans test

bojanserafimov:
This test is very sensitive to latency between services so it's good to run remotely. We should maybe run all tests remotely, but this one is a priority, I'll think about the general problem later.

### DIFF
--- a/test_runner/performance/test_seqscans.py
+++ b/test_runner/performance/test_seqscans.py
@@ -6,6 +6,7 @@ import pytest
 from fixtures.benchmark_fixture import MetricReport
 from fixtures.compare_fixtures import PgCompare
 from fixtures.log_helper import log
+from pytest_lazyfixture import lazy_fixture  # type: ignore
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


bojanserafimov:
I also thought about adding prefetch/noprefetch parameterization

```
diff --git a/test_runner/performance/test_seqscans.py b/test_runner/performance/test_seqscans.py
index 26258355..849a8131 100644
--- a/test_runner/performance/test_seqscans.py
+++ b/test_runner/performance/test_seqscans.py
@@ -21,6 +21,13 @@ from pytest_lazyfixture import lazy_fixture  # type: ignore
         pytest.param(10000000, 1, 4),
     ],
 )
+@pytest.mark.parametrize(
+    "prefetch",
+    [
+        pytest.param(True, id="prefetch"),
+        pytest.param(False, id="noprefetch"),
+    ],
+)
 @pytest.mark.parametrize(
     "env",
     [
@@ -30,12 +37,16 @@ from pytest_lazyfixture import lazy_fixture  # type: ignore
         pytest.param(lazy_fixture("remote_compare"), id="remote", marks=pytest.mark.remote_cluster),
     ],
 )
-def test_seqscans(env: PgCompare, rows: int, iters: int, workers: int):
+def test_seqscans(env: PgCompare, prefetch: bool, rows: int, iters: int, workers: int):
     with closing(env.pg.connect()) as conn:
         with conn.cursor() as cur:
             cur.execute("create table t (i integer);")
             cur.execute(f"insert into t values (generate_series(1,{rows}));")

+            if True:
+                cur.execute("set enable_seqscan_prefetch = on;")
+                cur.execute("set seqscan_prefetch_buffers = 100;")
+
             # Verify that the table is larger than shared_buffers
             cur.execute(
                 """
```

but IMO there's no reason to test without prefetching. When we're ready we should make prefetching the default, or at the very least make it the default for this test

hlinnaka (approved):

PR #2855, opened 2022-11-18
Nightly Benchmarks: add OLAP-style benchmark (clickbench)

bayandin:
Add ClickBench benchmark, an OLAP-style benchmark, to Nightly Benchmarks.

The full run of 43 queries on the original dataset takes more than 6h (only 34 queries got processed on in 6h) on our default-sized compute. Having this, currently, would mean having some really unstable tests because of our regular deployment to staging/captest environment (see https://github.com/neondatabase/cloud/issues/1872).

I've reduced the dataset size to the first 10^7 rows from the original 10^8 rows. Now it takes ~30-40 minutes to pass.

Ref https://github.com/ClickHouse/ClickBench/tree/main/aurora-postgresql
Ref https://benchmark.clickhouse.com/

### DIFF
--- a/.github/workflows/benchmarking.yml
+++ b/.github/workflows/benchmarking.yml
@@ -110,8 +110,14 @@ jobs:
         rm -rf perf-report-staging
         mkdir -p perf-report-staging
         # Set --sparse-ordering option of pytest-order plugin to ensure tests are running in order of appears in the file,
-        # it's important for test_perf_pgbench.py::test_pgbench_remote_* tests
-        ./scripts/pytest test_runner/performance/ -v -m "remote_cluster" --sparse-ordering --out-dir perf-report-staging --timeout 5400
+        # it's important for test_perf_pgbench.py::test_pgbench_remote_* tests.
+        # Do not run tests from test_runner/performance/test_perf_olap.py because they require a prepared DB. We run them separately in `clickbench-compare` job.
+        ./scripts/pytest test_runner/performance/ -v \
+          -m "remote_cluster" \
+          --sparse-ordering \
+          --out-dir perf-report-staging \
+          --timeout 5400 \
+          --ignore test_runner/performance/test_perf_olap.py
 
     - name: Submit result
       env:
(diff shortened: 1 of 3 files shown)


.github/workflows/benchmarking.yml
```diff
@@ -207,7 +213,7 @@ jobs:
             CONNSTR=${{ steps.create-neon-project.outputs.dsn }}
             ;;
           rds-aurora)
-            CONNSTR=${{ secrets.BENCHMARK_RDS_CONNSTR }}
+            CONNSTR=${{ secrets.BENCHMARK_RDS_AURORA_CONNSTR }}
```
bayandin, line 216: Renamed it for consistency

test_runner/performance/test_perf_olap.py
```diff
@@ -0,0 +1,89 @@
+from typing import Dict, List, Tuple
+
+import pytest
+from fixtures.compare_fixtures import RemoteCompare
+from fixtures.log_helper import log
+
+# A list of queries to benchmark from ClickBench.
+# The first element of each tuple is the query id to make it distinguishable in logs/grafana charts/etc.,
```
vadim2404, line 8: what do you think about having a more structural view of this Tuple? `dataclass`, `namedtuple`, etc. Because I don't think that it will be cool enough to recall what do we store in each index

test_runner/performance/test_perf_olap.py
```diff
@@ -0,0 +1,89 @@
...
+@pytest.mark.remote_cluster
+def test_clickbench(query: Tuple[int, str], remote_compare: RemoteCompare):
+    """
+    An OLAP-style ClickHouse benchmark
+
+    Based on https://github.com/ClickHouse/ClickBench/tree/c00135ca5b6a0d86fedcdbf998fdaa8ed85c1c3b/aurora-postgresql
+    The DB prepared manually in advance
+    """
+
+    q_id, q = query
```
vadim2404, line 88: if you're operating with structural data, it's no sense to deconstruct tuple here. It'll be better to pass it to the `run_psql` function as is

test_runner/performance/test_perf_olap.py
```diff
@@ -0,0 +1,89 @@
...
+
+import pytest
+from fixtures.compare_fixtures import RemoteCompare
+from fixtures.log_helper import log
+
+# A list of queries to benchmark from ClickBench.
+# The first element of each tuple is the query id to make it distinguishable in logs/grafana charts/etc.,
+# without mentioning the query itself (which may be too long); please keep query ids for existing queries.
+# The second element is the query itself
+QUERIES: List[Tuple[int, str]] = [
```
vadim2404, line 11: for const values it's better to use `Tuple`

```python
QUERIES: Tuple[Tuple[int, str], ...] = (
)
```

vadim2404 (approved):

PR #2859, opened 2022-11-18
Nightly Benchmarks: Add RDS Postgres

bayandin:
Add RDS Postgres ~`db.t3.medium`~ `db.m5.large` instance to Nightly Benchmarks.

I also changed the RDS Aurora Postgres instance from 1 ACU to me enable autoscaling from 0.5 ACU to 2 ACU (max 4 GiB of RAM). Before it was static 1 ACU

### DIFF
--- a/.github/workflows/benchmarking.yml
+++ b/.github/workflows/benchmarking.yml
@@ -144,7 +144,9 @@ jobs:
         # neon-captest-new: Run pgbench in a freshly created project
         # neon-captest-reuse: Same, but reusing existing project
         # neon-captest-prefetch: Same, with prefetching enabled (new project)
-        platform: [ neon-captest-new, neon-captest-reuse, neon-captest-prefetch ]
+        # rds-aurora: Aurora Postgres Serverless v2 with autoscaling from 0.5 to 2 ACUs
+        # rds-postgres: RDS Postgres db.m5.large instance (2 vCPU, 8 GiB) with gp3 EBS storage
+        platform: [ neon-captest-new, neon-captest-reuse, neon-captest-prefetch, rds-postgres ]
         db_size: [ 10gb ]
         include:
           - platform: neon-captest-new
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmarking.yml
```diff
@@ -144,7 +144,9 @@ jobs:
         # neon-captest-new: Run pgbench in a freshly created project
         # neon-captest-reuse: Same, but reusing existing project
         # neon-captest-prefetch: Same, with prefetching enabled (new project)
-        platform: [ neon-captest-new, neon-captest-reuse, neon-captest-prefetch ]
+        # rds-aurora: Aurora Postgres Serverless v2 with autoscaling from 0.5 to 2 ACUs
+        # rds-postgres: RDS Postgres db.t3.medium instance (2 vCPU, 4 GiB) with gp3 EBS storage
```
vadim2404, line 148: t3 instance has credits that can drastically decrease performance when we overuse them. Do you control it somehow?

vadim2404 (approved):

.github/workflows/benchmarking.yml
bayandin: Ah, didn't know about it, thanks for pointing it out! 
I've switched the instance to `db.m5.large` (2 vCPU, 8 GiB). The drawback here could be network throughput:
> * These DB instance classes can support maximum performance for 30 minutes at least once every 24 hours. For more information on baseline performance of these instance types, see [Amazon EBS-optimized instances](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSOptimized.html) in the Amazon EC2 User Guide for Linux Instances. 

from https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.DBInstanceClass.html

We potentially could hit it with client-side generated data in `pgbench --initialize`. Let's check it and switch `pgbench` to server-side generated load if required

PR #2863, opened 2022-11-18
Refactoring: replace `Url`-based connection strings with a hand-crafted struct

yeputons:
Previous attempt: #2730

My thoughts are in the first commit's description.


libs/utils/src/postgres_connection.rs
```diff
@@ -0,0 +1,162 @@
+use anyhow::Context;
+use std::fmt;
+use url::Host;
+
+pub fn parse_host_port<S: AsRef<str>>(host_port: S) -> Result<(Host, Option<u16>), anyhow::Error> {
```
yeputons, line 5: There is a similar function [in proxy](https://github.com/neondatabase/neon/blob/f564dff0e320a859ab9eb8b40caa9a70599c829a/proxy/src/auth/backend/console.rs#L252-L255), but I wouldn't touch it in this PR, so there's less to review. Maybe a separate PR, one can argue that IPv6 support is good.

control_plane/src/pageserver.rs
```diff
@@ -77,30 +77,24 @@ pub struct PageServerNode {
...
+                .set_password(password),
             env: env.clone(),
             http_client: Client::new(),
             http_base_url: format!("http://{}/v1", env.pageserver.listen_http_addr),
         }
     }
 
-    /// Construct libpq connection string for connecting to the pageserver.
-    fn pageserver_connection_config(password: &str, listen_addr: &str) -> PgConnectionConfig {
-        format!("postgresql://no_user:{password}@{listen_addr}/no_db")
```
yeputons, line 99: `no_user`/`no_db` were never parsed into `PgConnectionConfig` and are ok to ignore.

libs/utils/src/postgres_connection.rs
```diff
@@ -0,0 +1,199 @@
...
+        impl fmt::Debug for Redacted {
+            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+                f.write_str("REDACTED-STRING")
+            }
+        }
+
+        f.debug_struct("PgConnectionConfig")
+            .field("host", &self.host)
+            .field("port", &self.port)
+            .field("password", &self.password.as_ref().map(|_| Redacted {}))
```
vadim2404, line 129: Won't be easier to put "******" here directly instead of creating a temporary struct for one field debugging?

yeputons: It will be displayed as `"******"`, which looks exactly like a string of asterisks. Not that it's _that_ confusing, but still.

libs/utils/src/postgres_connection.rs
funbringer: You might want to use [`SocketAddr`](https://doc.rust-lang.org/std/net/enum.SocketAddr.html) or [`ToSocketAddrs`](https://doc.rust-lang.org/std/net/trait.ToSocketAddrs.html) here instead.

In fact, I'll likely use one of those in the proxy.

libs/utils/src/lib.rs
```diff
@@ -12,6 +12,8 @@ pub mod simple_rcu;
 /// append only ordered map implemented with a Vec
 pub mod vec_map;
 
+pub mod postgres_connection;
```
funbringer, line 15: This module is big enough (200+ lines including tests), so I think it deserves to be a crate in `libs/`. That way you won't re-add `tokio_postgres` to already bloated `utils`.

libs/utils/src/postgres_connection.rs
```diff
@@ -0,0 +1,199 @@
...
+}
+
+impl fmt::Debug for PgConnectionConfig {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        struct Redacted {}
+        impl fmt::Debug for Redacted {
+            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+                f.write_str("REDACTED-STRING")
+            }
+        }
```
funbringer, line 124: I don't think this is needed. `Option<&str>` implements Debug.

libs/utils/src/postgres_connection.rs
```diff
@@ -0,0 +1,199 @@
...
+#[derive(Clone)]
+pub struct PgConnectionConfig {
+    host: Host,
+    port: u16,
+    password: Option<String>,
+    options: Vec<String>,
+}
+
+impl PgConnectionConfig {
+    pub fn new_host_port(host: Host, port: u16) -> PgConnectionConfig {
```
funbringer, line 53: PgConnectionConfig -> Self

libs/utils/src/postgres_connection.rs
```diff
@@ -0,0 +1,199 @@
...
+
+    /// Return a `<host>:<port>` string.
+    pub fn raw_address(&self) -> String {
+        format!("{}:{}", self.host(), self.port())
+    }
+
+    /// Build a client library-specific connection configuration.
+    /// Used for testing and when we need to add some obscure configuration
+    /// elements at the last moment.
+    pub fn build_tokio_postgres_config(&self) -> tokio_postgres::Config {
```
funbringer, line 90: Most of the time rust's naming conventions use `to_` (ref to owned), `into_` (owned to owned) and `as_` (ref to ref), so this should be `to_tokio_postgres_config`.

libs/utils/src/postgres_connection.rs
```diff
@@ -0,0 +1,199 @@
...
+        }
+    }
+
+    pub fn host(&self) -> &Host {
+        &self.host
+    }
+
+    pub fn port(&self) -> u16 {
+        self.port
+    }
```
funbringer, line 68: Why allow changing everything but this? What's special about host and port?

libs/utils/src/postgres_connection.rs
```diff
@@ -0,0 +1,199 @@
...
+        config.host(&self.host().to_string()).port(self.port);
+        if let Some(password) = &self.password {
+            config.password(password);
+        }
+        if !self.options.is_empty() {
+            config.options(
+                &self
+                    .options
+                    .iter()
+                    .map(|s: &String| s.replace('\\', "\\\\").replace(' ', "\\ "))
```
funbringer, line 103: What's going on here?

libs/utils/src/postgres_connection.rs
```diff
@@ -0,0 +1,199 @@
...
+
+    pub fn set_password(mut self, s: Option<String>) -> Self {
+        self.password = s;
+        self
+    }
+
+    pub fn add_options<I: IntoIterator<Item = S>, S: Into<String>>(mut self, i: I) -> Self {
+        for s in i {
+            self.options.push(s.into());
+        }
```
funbringer, line 78: I guess you could use `self.options.extend()` here.

funbringer (commented):
Thanks for the hard work! I've left a bunch of comments.

libs/utils/src/lib.rs
yeputons: Moved into a separate `postgres_connection` crate

libs/utils/src/postgres_connection.rs
yeputons: Added a comment. Let me know if you still think it's better to display `"REDACTED"` instead of `REDACTED` to save some lines, your call. That's the way `tokio_postgres::Config` does it and it distinguishes between "a redacted string" and "a string which contains the value `REDACTED`".

libs/utils/src/postgres_connection.rs
yeputons: Added a comment.

libs/utils/src/postgres_connection.rs
yeputons: Good idea, thanks. Also renamed to `extend_options`.

libs/utils/src/postgres_connection.rs
yeputons: `Host` also includes unresolved hostnames. I suspect it's more obvious to resolve on connect rather than on configuration parsing. So I'd keep it as it is.

yeputons:
@funbringer I've just addressed all the comments, marked ones I consider unambiguous as resolved, two unresolved are rest. I did not change any other code, rebase was without conflicts.

libs/postgres_connection/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests_pg_connection_config {
+    use crate::PgConnectionConfig;
+    use once_cell::sync::Lazy;
+    use url::Host;
+
+    static STUB_HOST: Lazy<Host> = Lazy::new(|| Host::Domain("stub.host.example".to_owned()));
```
vadim2404, line 163: Do you really need to have the dev-dependency for this line? 

maybe you can replace it with a short method in this module, can't you?

```rust
fn stub_host() -> Host  {
    Host::Domain("stub.host.example".to_owned())
}
```

libs/utils/src/postgres_connection.rs
funbringer: I see! How about this:

```rust
struct Kek {
    value: u32,
}

impl std::fmt::Debug for Kek {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        f.debug_struct("Kek").field("value", &format_args!("REDACTED")).finish()
    }
}

fn main() {
    println!("{:?}", Kek { value: 10 });
}
```

I find `format_args!` immensely useful, because it's essentially `format!` but w/o allocations.

libs/postgres_connection/src/lib.rs
funbringer: I have something different on my mind. [`Host`](https://docs.rs/url/latest/url/enum.Host.html) can be parameterized by `&str` instead of String, so can make `new_host_port` more generic and  call [`to_owned`](https://docs.rs/url/latest/url/enum.Host.html#method.to_owned) inside.

This will let us use

```rust
const STUB_HOST: Host<&str> = Host::Domain("stub.host.example");
```

API users will likely benefit from this too.

vadim2404: More elegant proposal than mine! 

My idea was devoted not to adding new dev-dependency for this one line, but

yeputons: Nice idea! Unfortunately, I bumped into the comparison code: `Host<String>` can only be compared with `Host<String>`, not `Host<&str>`. Therefore I have to convert `STUB_HOST: Host<&str>` to `Host<String>` in `assert_eq!` anyway. Even if I support both `Host<String>` and `Host<&str>` in `new_host_port` (the only way to "overload" in Rust is to create a trait, isn't it..?), I still get `.to_owned()` in the test.

So I prefer to leave the code as is. What I have here is a global const, initialized once, not a value generated dynamically on the fly for each test (e.g. `assert_eq!` do not clone the const), so `Lazy` expresses the intention better. Also, we use `once_cell` heavily in our code a lot already, and its standard counterpart is in progress (https://github.com/rust-lang/rust/issues/74465).

libs/postgres_connection/src/lib.rs
```diff
@@ -0,0 +1,222 @@
+use anyhow::Context;
+use std::fmt;
+use url::Host;
+
+pub fn parse_host_port<S: AsRef<str>>(host_port: S) -> Result<(Host, Option<u16>), anyhow::Error> {
+    let url = url::Url::parse(&format!("postgres://{}", host_port.as_ref()))
+        .context("Invalid host:port pair specified")?;
+    let host = url.host().context("Host is not specified")?.to_owned();
+    Ok((host, url.port()))
```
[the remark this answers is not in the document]
yeputons: Good idea, I forgot about splitting from the right. Splitting from the left does not work with `[::1]`, so I discarded it altogether.

libs/postgres_connection/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+            // and all backslashes should be escaped before that. Although we don't expect options
+            // with spaces at the moment, they're supported by PostgreSQL. Hence we support them
+            // in this typesafe interface.
+            config.options(
+                &self
+                    .options
+                    .iter()
+                    .map(|s: &String| s.replace('\\', "\\\\").replace(' ', "\\ "))
+                    .collect::<Vec<_>>()
+                    .join(" "),
```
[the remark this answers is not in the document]
yeputons: Added `Cow`, turned out to be not much code. However, there are still some allocations in the worst case (each option contains a char to be escaped). Not a problem, I think, this function is called infrequently.

If we strictly aim for fewer allocations, I prefer something imperative which guarantees O(1) allocations (depending on `String`'s growth factor, if it's 2, we have 2 allocations at most):

```rust
            let mut options_str = String::new();
            options_str.reserve(
                self.options.iter().map(|s| s.len()).sum::<usize>()
                    + std::cmp::max(1, self.options.len())
                    - 1,
            );
            let mut first_option = true;
            for opt in self.options.iter() {
                if !first_option {
                    options_str.push(' ');
                }
                first_option = false;
                for c in opt.chars() {
                    match c {
                        '\\' => options_str.push_str("\\\\"),
                        ' ' => options_str.push_str("\\ "),
                        c => options_str.push(c),
                    }
                }
            }
            config.options(&options_str);
```

Rust's `std::str::escape_debug` [does something with custom iterators](https://doc.rust-lang.org/src/core/str/mod.rs.html#2473), so I suspect there is no faster standard way. I'm wondering if there is a functional way to write the escaping logic. I've tried like this, but I need an iterator whose `Item` alternates between `Map` and `Once`, so it doesn't compile:

```rust
                &self
                    .options
                    .iter()
                    // Iter<Iter<&str>>, options, each represented as a iterator of chunks
                    .map(|s| {
                        s.split("").map(|c| match c {
                            "\\" => "\\\\",
                            " " => "\\ ",
                            c => c,
                        })
                    })
                    .intersperse(std::iter::once(" ")) // TODO: use impl from std once it's stabilized
                    .flatten()
                    .collect::<String>(),
```

libs/postgres_connection/src/lib.rs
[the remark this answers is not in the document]
funbringer: Unfolding since I still have something to add here.

```rust
>> "-0".parse::<u16>() // can't reproduce
Err(ParseIntError { kind: InvalidDigit })
>> "+10".parse::<u16>() // this is unfortunate, yes
Ok(10)
```

I think we could add an `is_numeric` check on top if you'd like to.

libs/postgres_connection/src/lib.rs
funbringer: Ok! It's definitely a nice-to-have, not a must. I might revisit this later on my own.

libs/postgres_connection/src/lib.rs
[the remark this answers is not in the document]
yeputons: Good catch, thank you. I've missed that `if` while reading `from_str_radix`.

funbringer (approved):
Cool! This is a good start. We'll see how this changes in time.

pageserver/src/walreceiver/connection_manager.rs
```diff
@@ -586,7 +591,7 @@ impl WalreceiverState {
                     if latest_interaciton > self.wal_connect_timeout {
                         return Some(NewWalConnectionCandidate {
                             safekeeper_id: new_sk_id,
-                            wal_source_connstr: new_wal_source_connstr,
+                            wal_source_connconf: new_wal_source_connstr,
```
LizardWizzard, line 594: A bit inconsistent naming, variable remains to be named `new_wal_source_connstr`

pageserver/src/walreceiver/connection_manager.rs
```diff
@@ -834,28 +841,24 @@ fn wal_stream_connection_string(
...
-                 options='-c timeline_id={timeline_id} tenant_id={tenant_id}'"
-            ))
-        })
-        .with_context(|| format!("Failed to parse pageserver connection URL '{sk_connstr}'"))
+) -> anyhow::Result<PgConnectionConfig> {
+    let (host, port) =
+        parse_host_port(&listen_pg_addr_str).context("Unable to parse listen_pg_addr_str")?;
+    let port = port.unwrap_or(5432);
+    Ok(
+        PgConnectionConfig::new_host_port(host, port).extend_options(vec![
```
LizardWizzard, line 849: There is no need to use a  `vec!` here. Slice or an array (`&[` or `[`) should be ok
```suggestion
        PgConnectionConfig::new_host_port(host, port).extend_options([
```

pageserver/src/walreceiver/connection_manager.rs
```diff
@@ -834,28 +841,24 @@ fn wal_stream_connection_string(
...
-            let port = url.port().unwrap_or(5432); // default PG port
-
-            Ok(format!(
-                "host={host} \
-                 port={port} \
-                 options='-c timeline_id={timeline_id} tenant_id={tenant_id}'"
-            ))
-        })
-        .with_context(|| format!("Failed to parse pageserver connection URL '{sk_connstr}'"))
+) -> anyhow::Result<PgConnectionConfig> {
```
LizardWizzard, line 844: `wal_stream_connection_string` should be probably renamed to `wal_stream_connection_config`

LizardWizzard (approved):
Thanks! Looks good to me. Left minor comments

Not much to add after @funbringer's review :)

yeputons:
Squashed and rebased, minor conflicts during squash only (I had to rearrange some commits), will merge once tests pass.

PR #2864, opened 2022-11-18
sourcetree/docs: mention hakari generate

yeputons:


SomeoneToIgnore (approved):
We can also mention 2nd command here too, or just link to our CI section?
https://github.com/neondatabase/neon/blob/f564dff0e320a859ab9eb8b40caa9a70599c829a/.github/workflows/codestyle.yml#LL134C11-L134C35

PR #2873, opened 2022-11-21
page_service: serve pages in spawn_blocking

koivunej:
this will help with allowing the use of tokio's `block_on` on the walredo implementation. separating this to allow benchmarking separatedly.


koivunej:
This seems to have a negative perf impact, quite large even. The need could be avoided technically by using `flume::bounded(1)` channel in place of the `tokio::sync::oneshot`.

koivunej:
Perhaps sprinkling around many `tokio::task::block_in_place` is a good stop-gap measure before we have all async locks or `spawn_blocking`.

koivunej:
Closing per last comment, this needs to be approached differently.

PR #2875, opened 2022-11-21
Tokio based walredo

koivunej:
Re-implements the walredo process handling as an tokio managed process and task. I think the implementation is at feature parity but not all of these are easy to test:

- lazy start (without it, many tests will fail)
- kill and relaunch the process if any request fails (untested)

Timeout is now handled as "from the write completing to the read completing", which might not be correct with the request pipelining.

#1700 is not handled, as in requests are not retried. That could be implemented.

Builds upon #2873.

While working on this, I noticed that tokio does not yet support vectored writes for child processes: https://github.com/tokio-rs/tokio/pull/5216 -- the implementation will benefit from upgrading to this future tokio version.

Cc: #2778.


libs/pageserver_api/src/models.rs
```diff
@@ -246,21 +246,21 @@ pub enum PagestreamBeMessage {
...
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
 pub struct PagestreamNblocksRequest {
     pub latest: bool,
     pub lsn: Lsn,
     pub rel: RelTag,
 }
 
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
```
koivunej, line 263: These are extra.

pageserver/src/walredo.rs
```diff
@@ -201,25 +200,27 @@ impl PostgresRedoManager {
     ///
     /// Create a new PostgresRedoManager.
     ///
+    // called from:
+    // - attach_local_tenants <= main
+    // - create_tenant <= http
```
koivunej, line 205:
```suggestion
```

Unrelated changes.

pageserver/src/walredo.rs
```diff
@@ -148,7 +145,7 @@ impl WalRedoManager for PostgresRedoManager {
         key: Key,
         lsn: Lsn,
         base_img: Option<Bytes>,
-        records: Vec<(Lsn, NeonWalRecord)>,
+        records: std::sync::Arc<[(Lsn, NeonWalRecord)]>,
```
koivunej, line 148: This needs to be moved to `walredo.rs` completly.

pageserver/src/walredo.rs
```diff
@@ -228,34 +229,30 @@ impl PostgresRedoManager {
...
-
         // Relational WAL records are applied using wal-redo-postgres
         let buf_tag = BufferTag { rel, blknum };
-        let result = process
-            .apply_wal_records(buf_tag, base_img, records, wal_redo_timeout)
-            .map_err(WalRedoError::IoError);
 
-        let end_time = Instant::now();
-        let duration = end_time.duration_since(lock_time);
+        let result = self
```
koivunej, line 244: Add comment that this `block_on` requires that we are:
- entered into tokio context
- outside of async context

Basically within `spawn_blocking` which #2873 provides.

koivunej:
Drafting this while I make the adjustments I've noticed right away.

pageserver/src/walredo.rs
```diff
@@ -587,419 +687,579 @@ impl<C: CommandExt> CloseFileDescriptors for C {
...
-        tenant_id: TenantId,
-        pg_version: u32,
-    ) -> Result<PostgresRedoProcess, Error> {
-        // FIXME: We need a dummy Postgres cluster to run the process in. Currently, we
-        // just create one with constant name. That fails if you try to launch more than
-        // one WAL redo manager concurrently.
-        let datadir = path_with_suffix_extension(
-            conf.tenant_path(&tenant_id).join("wal-redo-datadir"),
-            TEMP_FILE_SUFFIX,
+    pg_version: u32,
```
koivunej, line 693: `pg_version` is still unused.

libs/pageserver_api/src/models.rs
koivunej: Removed.

pageserver/src/walredo.rs
koivunej: Removed.

koivunej:
That is inline with the microbenchmark... Thanks for running that! I don't think this path is viable anymore.

koivunej:
From what I can tell, the biggest slowdown happens with the read returning EAGAIN then epoll then retrying it, then from cross thread communication with the channels.

koivunej:
Ok, I've finally understood what affects this thanks to a bit commute and working on the laptop which has less cores.

The perf difference to `walredo-channel` from fa73b8eb is explained by cross thread communication as by default the use `pageserver::task_mgr::BACKGROUND_RUNTIME` is used. It is created with 24 workers on my workstation. Lowering that to 1 thread makes the `async-walredo` (or a version I'll next clean up and push) faster or the same as `walredo-channel`. This does not of course include the larger scale performance difference, but I'll see to trying ways around that next, maybe starting with flume channels again.

koivunej:
Fixing the tests to have less noise on next benchmark run without #2873, flume but still the default `pageserver::task_mgr::BACKGROUND_RUNTIME`.

koivunej:
Testing out `flume` to avoid the spawn_blocking + block_on:

Getting tests fixed is a bit more work than expected. There was an interesting deadlock alike situation was cleared up as tokio worker's LIFO slot not being stealable right now, so I will need to add `--cfg tokio_unstable` to the build and configure all runtimes not to use the unstealable lifo slot. Will probably separate it to it's own PR if this turns out useful.

koivunej:
Periodic updates:

> so I will need to add --cfg tokio_unstable to the build

This turned out to be difficult, since cargo never appends `RUSTFLAGS`, it just overrides. So very likely, the project setting would get overridden by user setting (use mold) and finally might be overridden by environment RUSTFLAGS. Going ahead sprinkling `tokio::task::block_in_place` around the many functions where we take blocking locks and/or do blocking io. I don't think this is a good idea either, because we could starve the runtime(s) this way too. Tradeoff to disabling the lifo slot would had been performance while leaving the door open to yet-unknown kind of starvation following from "one should not block in async context".

Best workaround for the blocking-in-async so far is to just minimize the locks needed altogether by more single-threaded approaches. For example, if two things can't run concurrently, we could just only ever launch one of them. We could also handle this with very small amount or no locks by having an async task for tenant.

koivunej:
Rebased on top of main, cleaned up some of the commits. Lets see how the benchmarks look now.

pageserver/src/walredo.rs
koivunej: Alternative is to use tokio::task::block_in_place somewhere higher up which should have less of a perf penalty, at the cost runtime of starvation hazard. This is being tested now, resolving and moving these as PR tasks.

knizhnik:
Now performance is much better: 16.5 sec for Q1 with main and  1.5 with Heikki's parallel requests patch.

koivunej:
Thanks @knizhnik for running the bench. I am actually a bit surprised why is it now that much faster. I'll hook up the multiprocess pooling perhaps next, though I am annoyed by the test failures I didn't get before rebase.

koivunej:
> Now performance is much better: 16.5 sec for Q1 with main and 14.5 with Heikki's parallel requests patch. So it provides the same performance as sync version. Nice, but it is a pity that it is not faster;)

I missed responding to the point about *being faster* this last week, I would never expect the async version to ever be faster than the owning thread version, at least the epoll based. As noted, tokio and epoll and the large thread count do bring their own challenges, which I haven't been able to understand. This however affects only the microbenchmarked single-request case, none of the pipelined.

Update on the failures: I still have test failures locally (2 real + 2 backward/forward compat).

workspace_hack/Cargo.toml
```diff
@@ -24,6 +24,7 @@ fail = { version = "0.5", default-features = false, features = ["failpoints"] }
 futures-channel = { version = "0.3", features = ["alloc", "futures-sink", "sink", "std"] }
 futures-task = { version = "0.3", default-features = false, features = ["alloc", "std"] }
 futures-util = { version = "0.3", features = ["alloc", "async-await", "async-await-macro", "channel", "futures-channel", "futures-io", "futures-macro", "futures-sink", "io", "memchr", "sink", "slab", "std"] }
+getrandom = { version = "0.2", default-features = false, features = ["js", "js-sys", "rdrand", "std", "wasm-bindgen"] }
```
vadim2404, line 27: `js`, `js-sys`, `wasm-bindgen` look weird for our use-case

koivunej: Yeah, they come through `flume`. Testing it out in the next push to leave them behind once more.

koivunej:
Need to wait for the allure report (probably after benches) for:
- https://github.com/neondatabase/neon/actions/runs/3563889915/jobs/5987541223#step:4:843
- https://github.com/neondatabase/neon/actions/runs/3563889915/jobs/5987541223#step:4:915

First one seems like a "port not free" transient issue. Luckily the debug regress tests passed.

workspace_hack/Cargo.toml
koivunej: Kept the `flume` for mpmc queue as I got the #2880 reimplemented in the latest commits.

pageserver/src/walredo.rs
```diff
@@ -587,419 +692,889 @@ impl<C: CommandExt> CloseFileDescriptors for C {
...
+            .and_then(|x| x);
+
+        match res {
+            Ok(()) => {
+                let page = Bytes::copy_from_slice(&page_buf[..8192]);
+                // we don't care about the result, because the caller could be gone
+                drop(completion.send(Ok(page)));
+                // if we did read more than one page, drain will move the one byte to the front, so we
+                // can continue with that next round.
+                page_buf.drain(..8192);
```
koivunej, line 1279: I am actually quite unsure on this because I didn't read `impl bytes::BufMut for Vec<u8>` how exactly does it advance, will it override already received byte or not. But this doesn't happen in practice, so not a huge blocker for reading this.

koivunej:
Immediate shutdown does not give enough time to kill/reap the children.

koivunej:
Regression tests will probably fail as I forgot that the spawn_blocking was no longer in place.

pageserver/benches/bench_walredo.rs
```diff
@@ -30,33 +30,51 @@ fn redo_scenarios(c: &mut Criterion) {
...
             &thread_count,
             |b, thread_count| {
-                add_multithreaded_walredo_requesters(b, *thread_count, &manager, medium, 10);
+                add_multithreaded_walredo_requesters(b, *thread_count, &manager, medium);
             },
         );
     }
+    drop(group);
+
+    if false {
```
vadim2404, line 73: do you still need this?

koivunej: It's part of the "scale up, scale down" demo, current commits:
- [6d3ac69](https://github.com/neondatabase/neon/pull/2875/commits/6d3ac69a7cfb8e5dc5c63497a7b848d2df382a85)
- [f856477](https://github.com/neondatabase/neon/pull/2875/commits/f8564775f3492a58d77f2ff9c6e4a5139cd66c47)

These will most likely be thrown out to a different PR, but I've mentioned them a few times to Heikki so didn't do that pre-emptively. Commits are now more readable.

pageserver/src/walredo.rs
```diff
@@ -587,419 +583,1147 @@ impl<C: CommandExt> CloseFileDescriptors for C {
...
-                ));
+    };
+
+    let stdout_task = stdout_task(stdout, result_rx);
+
+    let stderr_task = stderr_task(stderr);
+
+    async move {
+        // changed during splitting this up: run all in current task.
+        let reason = tokio::try_join!(stdin_task, stdout_task, stderr_task);
```
koivunej, line 979: Use of try_join! to take down all tasks forces a weird "Ok error" to be returned from stdout task when everything is done, will change this.

pageserver/src/walredo.rs
koivunej: Almost forgot, adding a test case for this.

koivunej:
The test failures don't seem too applicable to walredo... Well, I'll remove the run-benchmarks for now to get a faster allure report, enhance logging, add missing test case and maybe rebase if I just missed some test fix...

pageserver/src/walredo.rs
koivunej: Fixed, mothballed already into the one large commit.

pageserver/src/walredo.rs
koivunej: Added.

koivunej:
Ok has to be something to do with the shutdown, not the #2880 commit I dropped.

pageserver/src/walredo.rs
```diff
@@ -587,419 +583,1287 @@ impl<C: CommandExt> CloseFileDescriptors for C {
...
+                            None => {
+                                if cancel.is_cancelled() {
+                                    // on shutdown, now the children have all exited, so should we.
+                                    break;
+                                } else if first.is_some() {
+                                    // we are yet to launch our first process
+                                    true
+                                } else {
+                                    debug!("all wal redo processes have been stopped, waiting for the next redo request before launching new process");
+                                    continue;
```
koivunej, line 742:
```suggestion
                                    break;
```

This fixes the hanging. Logic bug which prevented the scale back up from zero.

koivunej:
Oki last one is the forward compatibility which is confused because of the new naming strat for walredo pg directories.

Bench report for the bench_walredo: https://gist.github.com/koivunej/7299083aba3d8c97a22a3e79229767ef

pageserver/src/walredo.rs
koivunej: Made tests for this, tucked them nicely within scale to zero and scale up. That was an interesting exercise.

koivunej:
Trying the benchmarks with the prefetch on default on next rebase.

pageserver/src/walredo.rs
```diff
@@ -587,419 +583,1506 @@ impl<C: CommandExt> CloseFileDescriptors for C {
...
+                &mut buffers,
+            );
+        }
+
+        let write_res = async {
+            while buffers.has_remaining() {
+                poll_fn(|cx| poll_write_buf(Pin::new(&mut stdin), cx, &mut buffers))
+                    .await
+                    .map_err(anyhow::Error::new)?;
+            }
```
problame, line 1216: This loop is just distracting.
Why is there no `write_full` method on stdin?
I guess probably because of some `async` Rust limitation.
Can we define an `util method ourselves, then, to declutter this a little?

pageserver/src/walredo.rs
```diff
@@ -587,419 +583,1506 @@ impl<C: CommandExt> CloseFileDescriptors for C {
...
+            // in general flush is not needed, does nothing on pipes, but since we now accept
+            // AsyncWrite
```
problame, line 1218: ... now what?

pageserver/src/walredo.rs
```diff
@@ -587,419 +583,1506 @@ impl<C: CommandExt> CloseFileDescriptors for C {
...
-                    // reading from the stderr, before checking the stdout.
-                    continue;
+        let write_res = tokio::time::timeout_at(timeout_at, write_res)
+            .map_err(|_| anyhow::anyhow!("write timeout"));
+
+        // wait the write to complete before sending the completion over, because we cannot fail
+        // the request after it has been moved. this could be worked around by making the oneshot
+        // into a normal mpsc queue of size 2 and making the read side race against the channel
+        // closing and timeouted read.
+        match write_res.await.and_then(|x| x) {
```
problame, line 1229: What is the deal with this `.and_then(|x| x)?

I think it wouldn't hurt to have a preceding line like

`let write_res = write_res.await.and_then(|x| x); // explain why`

pageserver/src/walredo.rs
```diff
@@ -587,419 +583,1506 @@ impl<C: CommandExt> CloseFileDescriptors for C {
...
+        1024 * 16
+    } else {
+        // without vectoring, use full buffer
+        1024 * 64
+    });
+
+    let work_timeout = Duration::from_secs(1);
+
+    loop {
+        let ((request, response), reservation) = {
```
problame, line 1131: let's call it `response_tx` ?

pageserver/src/walredo.rs
```diff
@@ -587,419 +583,1506 @@ impl<C: CommandExt> CloseFileDescriptors for C {
...
-                    // To make sure we capture all log from the process if it fails, keep
-                    // reading from the stderr, before checking the stdout.
-                    continue;
+        let write_res = tokio::time::timeout_at(timeout_at, write_res)
+            .map_err(|_| anyhow::anyhow!("write timeout"));
+
+        // wait the write to complete before sending the completion over, because we cannot fail
+        // the request after it has been moved. this could be worked around by making the oneshot
+        // into a normal mpsc queue of size 2 and making the read side race against the channel
+        // closing and timeouted read.
```
problame, line 1228: I don't get this comment, probably because I'm not as deep into it as you are.
To me it's plain obvious that we need to `.await` the write.
If we don't `.await` it here, the walredo process will simply never see it.
This is the first time we poll that future.

pageserver/src/walredo.rs
```diff
@@ -587,419 +583,1506 @@ impl<C: CommandExt> CloseFileDescriptors for C {
...
+        match write_res.await.and_then(|x| x) {
+            Ok(()) => {
+                // by writing and then completing later we achieve request pipelining with the
+                // walredo process. some workloads fit well into the default 64KB buffer, so we
+                // have an opportunity to keep walredo busy by buffering the requests.
+
+                if inter_tx.is_closed() {
+                    drop(response.send(Err(anyhow::anyhow!(
+                        "Failed to read walredo response: stdout task closed already"
+                    ))));
```
problame, line 1238: Shouldn't we retry this request on a new process, then? Same question for the `Err(io_or_timeout)` case below.

problame (changes requested):
Took an initial look at `walredo.rs` only. Didn't look at anything past `stderr_task`.
Generally, this looks pretty great.

Apart from the various comments, I think what's deeply needed here is a block comment that explains how the std{in,out} tasks work together.

With a separate section on cancellation.

And another section of what happens if the walredo process crashes.

pageserver/src/walredo.rs
```diff
@@ -587,419 +583,1506 @@ impl<C: CommandExt> CloseFileDescriptors for C {
...
+    // "hack". this buffer is reused between reads.
+    let mut page_buf = Vec::with_capacity(8192 + 4096);
+
+    loop {
+        let (completion, timeout_at) = match inter_rx.recv().await {
+            Some(t) => t,
+            None => return Ok(()),
+        };
+
+        let res = tokio::time::timeout_at(timeout_at, read_page(&mut stdout, &mut page_buf))
```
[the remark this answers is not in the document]
koivunej: > `BytesMut`

I must've meant `BufMut`.

> let's assert that it's len() is 0.

I don't think that would be a good idea. There's a test case which shows that this read works fine, even if reading the 3*4096 at once then the last 4096.

> line before this that makes it obvious we're using `page_buf` as `BytesMut`?

You mean a comment? I don't think it's too relevant, but yeah why not. The method used only accepts BufMut though, so, unsure. Also, I don't want to make this a `BytesMut` because that would then create a reference counted 3*4096 long `Bytes` out of first two consecutive pages, until degrading to some other buffering which I haven't tested.

I was hoping the the `page_buf.drain(..8192)` would be explicit enough.

pageserver/src/walredo.rs
```diff
@@ -587,419 +583,1506 @@ impl<C: CommandExt> CloseFileDescriptors for C {
...
-            // If we have more data to write, wake up if 'stdin' becomes writeable or
-            // we have data to read. Otherwise only wake up if there's data to read.
-            let nfds = if nwrite < writebuf.len() { 3 } else { 2 };
-            let n = loop {
-                match nix::poll::poll(&mut pollfds[0..nfds], wal_redo_timeout.as_millis() as i32) {
-                    Err(e) if e == nix::errno::Errno::EINTR => continue,
-                    res => break res,
+        // using tokio::try_join! later we get the wanted "channel close or shutdown takes
+        // down all tasks after finishing the in-flight work"
+        tokio::select! {
```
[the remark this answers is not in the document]
koivunej: You are right, and with the other comments about the comments, I should do exactly what you meant and have one big picture instead of these left overs. I went through a quite a few permutations of this :)

pageserver/src/walredo.rs
```diff
@@ -587,419 +583,1506 @@ impl<C: CommandExt> CloseFileDescriptors for C {
...
+                        warn!("stderr task failed: {e:?}");
+                        break;
+                    }
+                },
+            };
+        }
+
+        // only reason why this would fail is that it has exited already, which is
+        // possible by manually killing it outside of pageserver or by it dying while
+        // processing.
```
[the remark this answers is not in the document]
koivunej: Tokio installs on unix SIGCHILD reaper which also wait will update the same process as here, so I don't agree... I guess that should be tested properly.

pageserver/src/walredo.rs
koivunej: Now I included the flush here, so you could test this with anything, like a tcpstream. But with pipes, that flush does nothing.

pageserver/src/walredo.rs
koivunej: I use `.and_then(|x| x)` for [`Result::flatten`](https://doc.rust-lang.org/stable/std/result/enum.Result.html#method.flatten) (still unstable, unfortunately).

So the timeout returns `Result<Inner, tokio::time::Elapsed>` which I `map_err` into `anyhow::Result<Inner>`, but the `Inner` is `anyhow::Result<()>`, so flattening gives me one `anyhow::Result<()>`.

pageserver/src/walredo.rs
koivunej: That's a good point. I already wrote it more than once but I guess I deemed it unlikely, given how the most needed would be the already pipelined requests for which the `request` has already been dropped.

But yeah, thinking this again, I can't see why not save what we can.

pageserver/src/walredo.rs
koivunej: Quickly reading through this, I'm probably confusing `request` and `response[_tx]` here.

pageserver/src/walredo.rs
koivunej: Yes, this would make sense.

pageserver/src/walredo.rs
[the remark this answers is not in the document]
koivunej: The comment is about the `child.start_kill()` failing, which has an explicit check for "has this process been waited already or not" and fail if it has been.

Apologies if I fail to see something which we may have already covered.

pageserver/src/walredo.rs
```diff
@@ -587,419 +583,1506 @@ impl<C: CommandExt> CloseFileDescriptors for C {
...
+{
+    use tokio::io::AsyncReadExt;
+
+    async fn read_page<AR>(mut stdout: AR, buffer: &mut Vec<u8>) -> Result<(), StdoutTaskError>
+    where
+        AR: tokio::io::AsyncRead + Unpin,
+    {
+        loop {
+            let read = stdout
+                .read_buf(buffer)
```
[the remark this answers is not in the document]
koivunej: The docs are a bit vague for `read_buf`, implementation is quite clear "one read". My understanding is that it will go to sleep until there's a readyness event, then try to read the remaining vec. Because of how the response protocol goes, it is that one 8192 (or more, if unlucky/stalled) byte read.

To be honest I haven't tested any interesting configurations for this, for example 7kB pipe size, but I can't see how it could go wrong.

pageserver/src/walredo.rs
```diff
@@ -587,419 +583,1506 @@ impl<C: CommandExt> CloseFileDescriptors for C {
...
-        let child = Command::new(pg_bin_dir_path.join("postgres"))
-            .arg("--wal-redo")
-            .stdin(Stdio::piped())
-            .stderr(Stdio::piped())
-            .stdout(Stdio::piped())
-            .env_clear()
-            .env("LD_LIBRARY_PATH", &pg_lib_dir_path)
-            .env("DYLD_LIBRARY_PATH", &pg_lib_dir_path)
-            .env("PGDATA", &datadir)
-            // The redo process is not trusted, and runs in seccomp mode that
```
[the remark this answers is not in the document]
koivunej: Good catch, I should go over these lost comments.

koivunej:
This PR is not viable at least now while it does look good on the OLAP benches, the OLTP has the 7-12 usec to 24usec penalty (in microbenchmark `short/short/1`) which is too much. I haven't been able to understand root cause for this slowdown, but assume it's "threads". So closing and focusing on #2947.

koivunej:
Deleting branch on `neondatabase/neon`, archiving in `koivunej/neon`: https://github.com/koivunej/neon/tree/async_walredo

koivunej:
Tested this after c1731bc4 landed on main. Even on COMPUTE_REQUEST runtime, getting similar or perhaps a bit better perf than main with the OLTP test, but not that much faster. Performance for OLAP/prefetched would be better. Performance for for example image layer creation would most likely be slower, because there's a cross runtime penalty vs. on COMPUTE_REQUEST the page requests can benefit from the LIFO slot optimization.

PR #2877, opened 2022-11-21
GitHub Actions: Add branch related actions

bayandin:
Add `neon-branch-create` / `neon-branch-delete` to allow using branches in tests.
I have a couple of use cases in mind:
- For destructive tests with a big DB, we can create the DB once in advance and then use branches without the need to recreate the DB itself after tests change it.
- We can run tests in parallel (if there're compute-bound); I'd like to try it out to speed up our OLAP test (in https://github.com/neondatabase/neon/pull/2855)

Also migrate API v2 for `neon-project-create` / `neon-project-delete`

### DIFF
--- a/.github/actions/neon-branch-create/action.yml
+++ b/.github/actions/neon-branch-create/action.yml
@@ -1,0 +1,158 @@
+name: 'Create Branch'
+description: 'Create Branch using API'
+
+inputs:
+  api_key:
+    desctiption: 'Neon API key'
+    required: true
+  environment:
+    desctiption: 'dev (aka captest) or staging'
+    required: true
+  project_id:
+    desctiption: 'ID of the Project to create Branch in'
+    required: true
+outputs:
+  dsn:
+    description: 'Created Branch DSN (for main database)'
+    value: ${{ steps.change-password.outputs.dsn }}
+  branch_id:
+    description: 'Created Branch ID'
+    value: ${{ steps.create-branch.outputs.branch_id }}
+
+runs:
+  using: "composite"
+  steps:
+    - name: Parse Input
+      id: parse-input
+      shell: bash -euxo pipefail {0}
+      run: |
+        case "${ENVIRONMENT}" in
+          dev)
+            API_HOST=console.dev.neon.tech
+            REGION_ID=${REGION_ID:-eu-west-1}
+            ;;
+          staging)
+            API_HOST=console.stage.neon.tech
+            REGION_ID=${REGION_ID:-us-east-1}
+            ;;
+          *)
+            echo 2>&1 "Unknown environment=${ENVIRONMENT}. Allowed 'dev' or 'staging' only"
+            exit 1
+            ;;
+        esac
+
+        echo "api_host=${API_HOST}" >> $GITHUB_OUTPUT
+        echo "region_id=${REGION_ID}" >> $GITHUB_OUTPUT
+      env:
+        ENVIRONMENT: ${{ inputs.environment }}
+        REGION_ID: ${{ inputs.region_id }}
+
+    - name: Create New Branch
+      id: create-branch
+      shell: bash -euxo pipefail {0}
+      run: |
+        for i in $(seq 1 10); do
+         branch=$(curl \
+            "https://${API_HOST}/api/v2/projects/${PROJECT_ID}/branches" \
+            --header "Accept: application/json" \
+            --header "Content-Type: application/json" \
+            --header "Authorization: Bearer ${API_KEY}" \
+            --data "{
+              \"branch\": {
+                \"name\": \"Created by actions/neon-branch-create; GITHUB_RUN_ID=${GITHUB_RUN_ID} at $(date +%s)\"
+              }
+            }")
+
+          if [ -z "${branch}" ]; then
+            sleep 1
+            continue
+          fi
+
+          branch_id=$(echo $branch | jq --raw-output '.branch.id')
+          if [ "${branch_id}" == "null" ]; then
+            sleep 1
+            continue
+          fi
+
+          break
+        done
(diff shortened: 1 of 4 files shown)


.github/actions/neon-branch-create/action.yml
```diff
@@ -0,0 +1,162 @@
...
+      id: parent-branch-id
+      shell: bash -euxo pipefail {0}
+      run: |
+        branches=$(curl "https://${API_HOST}/api/v2/projects/${PROJECT_ID}/branches" \
+          --fail \
+          --header "Accept: application/json" \
+          --header "Content-Type: application/json" \
+          --header "Authorization: Bearer ${API_KEY}"
+          )
+        parent_branch_id=$(echo $branches | jq --raw-output '.branches[0].id')
```
Daniel-ef, line 63: You're getting not the parent branch but random. It's possible now to omit `parent_id` in the create request so that the first main branch will be taken.

Or user should pass desired `parent_branch_id` via inputs

.github/actions/neon-branch-create/action.yml
```diff
@@ -0,0 +1,162 @@
...
+      id: create-branch
+      shell: bash -euxo pipefail {0}
+      run: |
+        branch=$(curl \
+          "https://${API_HOST}/api/v2/projects/${PROJECT_ID}/branches" \
+          --fail \
+          --header "Accept: application/json" \
+          --header "Content-Type: application/json" \
+          --header "Authorization: Bearer ${API_KEY}" \
+          --data "{
```
Daniel-ef, line 80: I'm not sure, does curl use POST request automatically if `--data` is presented?

.github/actions/neon-branch-create/action.yml
```diff
@@ -0,0 +1,162 @@
...
+        API_HOST: ${{ steps.parse-input.outputs.api_host }}
+        PROJECT_ID: ${{ inputs.project_id }}
+        BRANCH_ID: ${{ steps.create-branch.outputs.branch_id }}
+
+    - name: Change Password
+      id: change-password
+      # A shell without `set -x` to not to expose password/dsn in logs
+      shell: bash -euo pipefail {0}
+      run: |
+        for i in $(seq 1 30); do
```
Daniel-ef, line 123: I think 30 requests is too much. 3-5 is enough. 
Because it begins to work in a few requests or doesn't work at all :)

I'm just worried about load to the instances

.github/actions/neon-branch-delete/action.yml
```diff
@@ -0,0 +1,60 @@
...
+            echo 2>&1 "Unknown environment=${ENVIRONMENT}. Allowed 'dev' or 'staging' only"
+            exit 1
+            ;;
+        esac
+
+        echo "api_host=${API_HOST}" >> $GITHUB_OUTPUT
+      env:
+        ENVIRONMENT: ${{ inputs.environment }}
+
+    - name: Delete Neon Project
```
Daniel-ef, line 42:
```suggestion
    - name: Delete Neon Branch
```

.github/actions/neon-branch-create/action.yml
bayandin: Yes, it is!
There's a post from the curl creator that you shouldn't use `-X POST` if possible: https://daniel.haxx.se/blog/2015/09/11/unnecessary-use-of-curl-x/

.github/actions/neon-branch-delete/action.yml
bayandin: Ah, copy-pasting got me!
Thanks!

.github/actions/neon-branch-create/action.yml
bayandin: Thanks for pointing that out! I've switched to using default parent_branch_id (i.e. do not set this parameter)

.github/actions/neon-branch-create/action.yml
bayandin: Reduced it to 10 (just to make it more than our typical compute start time, just in case mostly).

.github/actions/neon-branch-create/action.yml
```diff
@@ -0,0 +1,162 @@
...
+        if [ -z "${branch_id}" ] || [ "${branch_id}" == "null" ]; then
+          echo 2>&1 "Failed to create branch after 10 attempts, the latest response was: ${branch}"
+          exit 1
+        fi
+
+        branch_id=$(echo $branch | jq --raw-output '.branch.id')
+        echo "branch_id=${branch_id}" >> $GITHUB_OUTPUT
+
+        endpoint_id=$(echo $branch | jq --raw-output '.endpoints[0].id')
+        echo "endpoint_id=${endpoint_id}" >> $GITHUB_OUTPUT
```
nikitakalyanov, line 92: It is now possible to get the whole host right from the response with .endpoints[0].host
https://github.com/neondatabase/cloud/pull/3018

.github/actions/neon-branch-create/action.yml
```diff
@@ -0,0 +1,162 @@
...
+      run: |
+        roles=$(curl \
+          "https://${API_HOST}/api/v2/projects/${PROJECT_ID}/branches/${BRANCH_ID}/roles" \
+          --fail \
+          --header "Accept: application/json" \
+          --header "Content-Type: application/json" \
+          --header "Authorization: Bearer ${API_KEY}"
+          )
+
+        role_name=$(echo $roles | jq --raw-output '.roles[] | select(.name != "web_access") | .name')
```
nikitakalyanov, line 110: It's better to filter by `.protected = false`

koivunej (approved):
Approving this on the basis of earlier reviews and that the bash looks good.

PR #2878, opened 2022-11-21
Fix remote seqscan test

bojanserafimov:
I haven't tested yet if the tests pass now, and how long they take

### DIFF
--- a/test_runner/performance/test_seqscans.py
+++ b/test_runner/performance/test_seqscans.py
@@ -22,15 +22,20 @@ from pytest_lazyfixture import lazy_fixture  # type: ignore
     ],
 )
 @pytest.mark.parametrize(
-    "env",
+    "env, scale",
     [
-        # Run on all envs
-        pytest.param(lazy_fixture("neon_compare"), id="neon"),
-        pytest.param(lazy_fixture("vanilla_compare"), id="vanilla"),
-        pytest.param(lazy_fixture("remote_compare"), id="remote", marks=pytest.mark.remote_cluster),
+        # Run on all envs. Use 50x larger table on remote cluster to make sure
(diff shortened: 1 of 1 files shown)


test_runner/performance/test_seqscans.py
```diff
@@ -22,15 +22,20 @@
...
-        pytest.param(lazy_fixture("neon_compare"), id="neon"),
-        pytest.param(lazy_fixture("vanilla_compare"), id="vanilla"),
-        pytest.param(lazy_fixture("remote_compare"), id="remote", marks=pytest.mark.remote_cluster),
+        # Run on all envs. Use 50x larger table on remote cluster to make sure
+        # it doesn't fit in shared buffers, which are larger on remote than local.
+        pytest.param(lazy_fixture("neon_compare"), 1, id="neon"),
+        pytest.param(lazy_fixture("vanilla_compare"), 1, id="vanilla"),
+        pytest.param(
+            lazy_fixture("remote_compare"), 50, id="remote", marks=pytest.mark.remote_cluster
+        ),
```
bayandin, line 33: Are we ok that these test results will be even more incomparable between local Neon, vanilla Postgres and remote Neon?

bayandin (approved):

test_runner/performance/test_seqscans.py
bojanserafimov: Yes. We're comparing local neon with local postgres, and remote neon with aurora, etc. IMO local neon and remote neon are not comparable. Configs are too different already. Priorities are different (fast development vs realistic benchmarks)

I do have a slight issue with the way the test names will look. It will seem like remote is 50x slower than local, because of the scaling factor. But I couldn't find a perfect solution without over engineering.

PR #2884, opened 2022-11-22
Consumption metering RFC

kelvich:
Rendered: https://github.com/neondatabase/neon/blob/metering_rfc/docs/rfcs/021-metering.md


docs/rfcs/021-metering.md
```diff
@@ -0,0 +1,175 @@
...
+    "metric": "traffic_in_bytes",
+    "type": "increment",
+    "endpoint_id": "blazing-warrior-34",
+    "event_start_time": ...,
+    "event_stop_time": ...,
+    "value": 12345454,
+}
+```
+
+Since we report deltas over some period of time, it makes sense to include `event_start_time`/`event_stop_time` where `event_start_time` is the time of the previous report. That will allow us to identify metering gaps better (e.g., failed send/delivery).
```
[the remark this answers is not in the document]
stepashka: not sure if graphite and statsd are the way to go.
they force to fit into the typical format and metadata for their destinations, e.g.
- graphite would have one timestamp and some clients may be optimized for a predefined resolution, tags are a late addition to graphite, typically you would have a dot.separated.metric.name
- does statsd include the timestamp of the measurement?

there's also a question of the support on the recipient. e.g. i haven't seen billing vendors supporting graphite as an input, not sure we want to lock ourselves into the format & schema transformation on the way towards the consumers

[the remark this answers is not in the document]
stepashka: IMHO OTel should be more modern, developing, flexible and more widely supported for the consumption too compared to graphite, stasd and influx.

docs/rfcs/021-metering.md
```diff
@@ -0,0 +1,175 @@
...
+
+* S3 storage size. That is the size of the tenant data on S3. Per-tenant.
+
+That info should be enough to build an internal model that predicts AWS price (hence tracking `written data` and `real storage size`). As for the billing model we probably can get away with mentioning only `CPU time`, `synthetic storage size`, and `traffic` consumption.
+
+# Services participating in metrics collection
+
+## Proxy
+
+Proxy is the only place that knows about traffic flow, so it tracks it and reports it with quite a small interval, let's say 1 minute. A small interval is needed here since the proxy is stateless, and any restart will reset accumulated consumption. Also proxy should report deltas since the last report, not an absolute value of the counter. Such kind of events is easier to integrate over a period of time to get the amount of traffic during some time interval.
```
stepashka, line 39: > Also proxy should report deltas since the last report, not an absolute value of the counter.

delta counters would make it harder to work around the loss of the signal compared to the absolute counters.
e.g. if a message with a delta is lost, we can't recover it from the following messages unless we also report the absolute counters.

could we report both?
depending on the recepient, it may be easier to work with one or another.

docs/rfcs/021-metering.md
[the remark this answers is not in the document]
kelvich: Collection code could be fairly involved, e.g. we'll probably end up with saving some disk state about last sent values. So decent collector could be a good reason to stick to some existing format. I went checked:
* https://github.com/markstory/rust-statsd -- forces UDP, that will complicate the receiver side
* https://github.com/influxdb-rs/influxdb-rust -- assumes influxdb on the other side, so I guess receiver has emulate a bit more then a message format to pretend to be an influx instance
* https://github.com/tureus/graphite-rust -- not maintained for 7 years

Also formats there are not super attractive -- some custom text-based representations with spaces and commas.

With that I'd rather write billing metrics collection manually and own it -- we don't need a super involved logic there as well. I think of that reporting pipeline more like kafka events or RPC append calls, rather then a usual observability metrics.

> I'm not sure start time actually helps that much: if we have constant interval, we already know that the start time is now - interval.

It is a bit different with CPU time in this proposal -- it is num_cores*cpu_time, so it can report 240 seconds when using 4 cores during 60 seconds.

docs/rfcs/021-metering.md
kelvich: Hm, right. Let's do that.

docs/rfcs/021-metering.md
```diff
@@ -0,0 +1,175 @@
...
+
+# Data collection
+
+## Push vs. pull
+
+We already have pull-based Prometheus metrics, so it is tempting to use them here too. However, in our setup, it is hard to tell when some metric changes. For example, garbage collection will constantly free some disk space over a week, even if the project is down for that week. We could also iterate through all existing tenants/branches/endpoints, but that means some amount of code to do that properly and most likely we will end up with some per-metric hacks in the collector to cut out some of the tenants that are surely not changing that metric.
+
+With the push model, it is easier to publish data only about actively changing metrics -- pageserver knows when it performs s3 offloads, garbage collection and starts/stops consuming data from the safekeeper; proxy knows about connected clients; console / autoscaler-agent knows about active cpu time.
+
+Hence, let's go with a push-based model.
```
stepashka, line 114: i suggest to design for both push and pull clients.

we can totally use push for our source of truth for billing, but support the pull readers too.
e.g. for expensive calculations, whenever they are triggered by an internal routine or by an external call, we could cache locally the latest value and the time when it was calculated. we can provided this into to pull-clients for debugging purposes or for ensuring the eventual consistency.

IOW, even though console or billing vendor would receive push, we would have the last measurement cached and available cheap and easy on demand for reporting to prom or to any other client via the API.

towards prometheus we could report something like last_known_{metric} and last_known_{metric}_age

docs/rfcs/021-metering.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Common bus vs. proxying through the console
+
+We can implement such push systems in a few ways:
+
+a. Each component pushes its metrics to the "common bus", namely segment, Kafka, or something similar. That approach scales well, but it would be harder to test it locally, will introduce new dependencies, we will have to distribute secrets for that connection to all of the components, etc. We would also have to loop back some of the events and their aggregates to the console, as we want to show some that metrics to the user in real-time.
+
+b. Each component can call HTTP `POST` with its events to the console, and the console can forward it to the segment for later integration with metronome / orb / onebill / etc. With that approach, only the console has to speak with segment. Also since that data passes through the console, the console can save the latest metrics values, so there is no need for constant feedback of that events back from the segment.
```
stepashka, line 122: >  Each component can call HTTP `POST` with its events to the console, and the console can forward it to the segment for later integration with metronome / orb / onebill / etc. With that approach, only the console has to speak with segment. Also since that data passes through the console, the console can save the latest metrics values, so there is no need for constant feedback of that events back from the segment.

this approach makes the console a single point of failure.
it forces to heavily optimize for the lack of any downtimes of the console and brings up hard questions about the error handling strategies.

what should be the strategies for the other components for handling the failures of the POSTs? retry?
i think there could be the situations when the sender thinks that the message wasn't delivered, when in fact it was.

> the console can save the latest metrics values

with the deltas approach, not only the latest metrics have to be sent, but also the data potentially needs to be aggregated with the messages and measurements received previously. 
horizontal scaling of the processing in the console may be interesting, given the questions about the retries and possibilities of multiple deliveries.

docs/rfcs/021-metering.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Push vs. pull
+
+We already have pull-based Prometheus metrics, so it is tempting to use them here too. However, in our setup, it is hard to tell when some metric changes. For example, garbage collection will constantly free some disk space over a week, even if the project is down for that week. We could also iterate through all existing tenants/branches/endpoints, but that means some amount of code to do that properly and most likely we will end up with some per-metric hacks in the collector to cut out some of the tenants that are surely not changing that metric.
+
+With the push model, it is easier to publish data only about actively changing metrics -- pageserver knows when it performs s3 offloads, garbage collection and starts/stops consuming data from the safekeeper; proxy knows about connected clients; console / autoscaler-agent knows about active cpu time.
+
+Hence, let's go with a push-based model.
+
+## Common bus vs. proxying through the console
```
stepashka, line 116: > console 

where would this live after the control plane split?

docs/rfcs/021-metering.md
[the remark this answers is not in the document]
lassizci: > IMHO OTel should be more modern
Hence ”such as”, didn’t mean exclude any specific protocols.

> there's also a question of the support on the recipient. e.g. i haven't seen billing vendors supporting graphite as an input

So recipient would be the collector agent. What happens after that would just have more options as the selection of protocols would be larger.

docs/rfcs/021-metering.md
```diff
@@ -0,0 +1,175 @@
...
+```
+
+I'd also suggest reporting one value, `cpu time`*`compute_units_count`, instead of two separate fields as it makes event schema simpler (it is possible to treat it the same way as traffic) and preserves additivity.
+
+## Pageserver
+
+Pageserver knows / has access to / can calculate the rest of the metrics:
+
+* Written size -- that is basically `last_received_lsn`,
+* Synthetic storage size -- there is a way to calculate it, albeit a costly one,
```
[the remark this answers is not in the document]
kelvich: I think some adjective like "synthetic" could help to understand that we are talking about @hlinnaka's algorithm value. Running `df -h` in tenants directory also yields a size.

docs/rfcs/021-metering.md
```diff
@@ -0,0 +1,175 @@
...
+* Real storage size -- there is a way to calculate it using a layer map or filesystem,
```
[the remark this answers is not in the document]
kelvich: physical size tenant size, yes. Or, in other words, result of `df -h` in tenant directory.

docs/rfcs/021-metering.md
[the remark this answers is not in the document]
kelvich: Also, IIUC, that synthetic size could be meaningfully reported per branch as well giving something similar to `min(history_size, logical_branch_size)`

docs/rfcs/021-metering.md
```diff
@@ -0,0 +1,175 @@
...
+* S3 storage size -- can calculate it by S3 API calls
+
+Some of those metrics are expensive to calculate, so the reporting period here is driven mainly by implementation details. We can set it to, for example, once per hour. Not a big deal since the pageserver is stateful, and all metrics can be reported as an absolute value, not increments. At the same time, a smaller reporting period improves UX, so it would be good to have something more real-time.
```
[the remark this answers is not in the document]
kelvich: > what does this mean?

I think @hlinnaka mentioned that the current synthetic size calculation is quite expensive.

> what are the metrics that we can afford updating close to realtime for our user?

Don't have too much of the context about why is it slow now, but I think we can make it fast for at least for the main branch.

But generally agree that size reporting delay better be minutes not hours.

docs/rfcs/021-metering.md
kelvich: makes sense, let's keep Prometheus endpoints as well. That will also help to structure collection code on the source side -- that could be an usual metric which is periodically manually pushed somewhere.

Regarding the `last_known_*` I think that applies only to synthetic size. With all other metrics we can update them live.

docs/rfcs/021-metering.md
kelvich: > this approach makes the console a single point of failure.

It is already. As well as the console database. We can later split that collector to a separate service, but for now I think it fine, given that we write data to the database and segment.

Or we can just report to segment right from the service. I thought about that initially, but then it is tricky for the console to subscribe to that events, and not sure moving data through the segment will improve our reliability in practical terms.

> horizontal scaling of the processing in the console may be interesting

that should just work since we touch database (or redis optionally)

> what should be the strategies for the other components for handling the failures of the POSTs? retry?
i think there could be the situations when the sender thinks that the message wasn't delivered, when in fact it was.

sender should consider data as delivered only when it received confirmation from the collector. That will result in at least once semantics. On failure/timeout we can just avoid resetting delta counter / last send at, so instead of retry we will report delta over a bigger time frame.

docs/rfcs/021-metering.md
kelvich: in the control plane

docs/rfcs/021-metering.md
```diff
@@ -0,0 +1,175 @@
...
+]
+```
+
+![data flow](./images/metering.jpg)
+
+Events could be either:
+* `incremental` -- change in consumption since the previous event or service restart. That is `effective_cpu_seconds`, `traffic_in_bytes`, and `traffic_out_bytes`.
+* `absolute` -- that is the current value of a metric. All of the size-related metrics are absolute.
+
+Each service can post events at its own pace and bundle together data from different tenants/endpoints.
```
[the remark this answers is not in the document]
kelvich: >  there should be no consumption events, right?

Nope. For example, garbage collection will constantly free some disk space over a week, even if the project is down for that week.

docs/rfcs/021-metering.md
```diff
@@ -0,0 +1,175 @@
...
+    "event_stop_time": ...,
+    "value": 12345454,
+}
+```
+
+I'd also suggest reporting one value, `cpu time`*`compute_units_count`, instead of two separate fields as it makes event schema simpler (it is possible to treat it the same way as traffic) and preserves additivity.
+
+## Pageserver
+
+Pageserver knows / has access to / can calculate the rest of the metrics:
```
vadim2404, line 83: When should PageServer call control-plane about the tenant size? Can you provide kind of sequence diagram that shows the whole path from initiating the action to the final processing?

docs/rfcs/021-metering.md
```diff
@@ -0,0 +1,175 @@
...
+
+# Goals
+
+This proposal is made with two mostly but not entirely overlapping goals:
+
+* Collect info that is needed for consumption-based billing
+* Cross-check AWS bills
+
+
+# Metrics
```
vadim2404, line 11: I'd like to propose to meter cross-AZ traffic on SK level to, it costs us money. In future, we need to understand what to do with this.

docs/rfcs/021-metering.md
```diff
@@ -0,0 +1,175 @@
...
+
+The console algorithm upon receive of events could be the following:
+
+1. Create and send a segment event with the same content (possibly enriching it with tenant/timeline data for endpoint-based events).
+2. Update the latest state of per-tenant and per-endpoint metrics in the database.
+3. Check whether any of that metrics is above the allowed threshold and stop the project if necessary.
+
+Since all the data comes in batches, we can do the batch update to reduce the number of queries in the database. Proxy traffic is probably the most frequent metric, so with batching, we will have extra `number_of_proxies` requests to the database each minute. This is most likely fine for now but will generate many dead tuples in the console database. If that is the case, we can change step 2 to the following:
+
+2.1. Check if there $tenant_$metric / $endpoint_$metric key in Redis
```
vadim2404, line 159: You don't need to check the presence of metric in Redis in this case. If you call `incr` on absent key, the key will be created

vadim2404, line 159: Do we need to bet on Redis for such solution? If so, we need to turn it in the RDB + AOF mode

docs/rfcs/021-metering.md
vadim2404, line 122: What should we do if while calling control_plane, it doesn't respond? or redis is not available?

docs/rfcs/021-metering.md
```diff
@@ -0,0 +1,175 @@
...
+
+With the push model, it is easier to publish data only about actively changing metrics -- pageserver knows when it performs s3 offloads, garbage collection and starts/stops consuming data from the safekeeper; proxy knows about connected clients; console / autoscaler-agent knows about active cpu time.
+
+Hence, let's go with a push-based model.
+
+## Common bus vs. proxying through the console
+
+We can implement such push systems in a few ways:
+
+a. Each component pushes its metrics to the "common bus", namely segment, Kafka, or something similar. That approach scales well, but it would be harder to test it locally, will introduce new dependencies, we will have to distribute secrets for that connection to all of the components, etc. We would also have to loop back some of the events and their aggregates to the console, as we want to show some that metrics to the user in real-time.
```
vadim2404, line 120: Why didn't you describe the solution with Kafka? It's possible to achieve better reliability with transactional outbox, DLQ. And we can build windows on top of Kafka for customers to show their metrics per day/5 minutes/ or how they want via kSQL and streams/tables (taking into account that 
@arssher
 is working on incorporating protobuf, which will also be helpful here).

docs/rfcs/021-metering.md
```diff
@@ -0,0 +1,175 @@
...
+
+2.2. If no stored value is found and the metric is incremental, then fetch the current value from DWH (which keeps aggregated value for all the events) and publish it.
+
+2.3. Publish a new value (absolute metric) or add an increment to the stored value (incremental metric)
+
+## Consumption watchdog
+
+Since all the data goes through the console, we don't have to run any background thread/coroutines to check whether consumption is within the allowed limits. We only change consumption with `POST /usage_events`, so limit checks could be applied in the same handler.
+
+## Extensibility
```
vadim2404, line 169: I assume that it will be valid requirement to see overall product consumption in order to understand its profitability (cc @stepashka ). Having such metrics in Redis it will be hard to build this view.

docs/rfcs/021-metering.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Common bus vs. proxying through the console
+
+We can implement such push systems in a few ways:
+
+a. Each component pushes its metrics to the "common bus", namely segment, Kafka, or something similar. That approach scales well, but it would be harder to test it locally, will introduce new dependencies, we will have to distribute secrets for that connection to all of the components, etc. We would also have to loop back some of the events and their aggregates to the console, as we want to show some that metrics to the user in real-time.
+
+b. Each component can call HTTP `POST` with its events to the console, and the console can forward it to the segment for later integration with metronome / orb / onebill / etc. With that approach, only the console has to speak with segment. Also since that data passes through the console, the console can save the latest metrics values, so there is no need for constant feedback of that events back from the segment.
+
+# Implementation
```
vadim2404, line 124: Who will be the source of truth of this data? Should it be stored locally in all services who meters sth?

docs/rfcs/021-metering.md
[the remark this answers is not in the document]
lassizci: > https://github.com/markstory/rust-statsd -- forces UDP, that will complicate the receiver side

Hence suggesting to use collection agent in between. They all support udp with statsd as because it's very common to use standard and has low impact even with very high volume of metrics sent.

> https://github.com/influxdb-rs/influxdb-rust -- assumes influxdb on the other side, so I guess receiver has emulate a bit more then a message format to pretend to be an influx instance

vmagent and telegraf will happily take influxdb, although I think vector might be more interesting here, as it for example supports unit testing with configurations. 


> Also formats there are not super attractive -- some custom text-based representations with spaces and commas.

What is more custom than own format? :) But that's honestly besides the point. These are designed for very high volume metrics and therefore opt to simpler format, which takes less cpu cycles to parse. 

But as a clarification, what I'm, thinking is more like:
```


  ┌───────────┐                     ┌─────────┐
  │ pageserver├────────┐            │         │
  └───────────┘        │            │ console │
                     ┌─▼──────┐     │         │
  ┌───────────┐      │        ├────►└───┬─────┘
  │ proxy     ├──────► vector │         │
  └───────────┘      │        │     ┌───▼─────┐
                     └─▲──────┴────►│ external│
  ┌───────────┐        │        ?   │ services│
  │ autosaler ├────────┘            │         │
  └───────────┘                     └─────────┘
```

Then we would be able to send metrics to vector even ridiculously often, have it aggregate and act as a durable queue and have it either push to somewhere (even to external services, while not necessarily the best option, depending on service) or allow console to scrape e.g using prometheus style scraping.

There very likely has to be a queue in between console and metrics sent in any case, and it has to be something pretty simple and small, which won't be updated often.

docs/rfcs/021-metering.md
kelvich: More of less by using a timer. Period ideally is minutes, but some calculations there are expensive so for "synthetic" size it would be a bigger period (like 1 hour). We could later think on how to reduce that period.

Talking about the exact implementation details, most of the metrics described here could defined as Prometheus counters (except "synthetic" size) and updated on the fly. Then background job just collects all that info for all of the tenant and sends it to the console. Straightforward optimization here is not to send values that were not changed since last transmission.

docs/rfcs/021-metering.md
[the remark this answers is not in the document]
hlinnaka: We really need to nail down these terms.. Here's the terminology that we currently use in the pageserver:

Tenant size -- this is what you're calling "synthetic storage size". It's one number per tenant.
Logical size -- What you see if you connect with psql, do `\l+`, and sum up the sizes of all databases. This is calculated separately for each timeline.
Physical size -- Size on disk on pageserver, if you did `du /data/storage/tenant/.../timelines/...`. Calculated separately for each timeline.

Currently, we don't measure "s3 storage size", but it's currently the same as "physical size". When we have on-demand download, they can differ.

I like the "synthetic" name.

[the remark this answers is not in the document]
hlinnaka: > Also, IIUC, that synthetic size could be meaningfully reported per branch as well giving something similar to min(history_size, logical_branch_size)

It's not that simple. The synthetic tenant size is calculated based on all the timelines.

docs/rfcs/021-metering.md
kelvich: Not sure what that means in practical terms. All the WAL traffic that goes in and out of the safekeeper is metered by that `last_lsn` which is also reported to compute and pageserver. So regardless of where we meter it value would be the same. Also neither safekeeper not pageserver have any information about whether counterpart is in the same AZ or not.

But we can deduce how much times WAL traffic is crossing AZ boundaries by tacking into account deploy topology. Basically there are two variants:

Untitled - Frame 2

In the first setup WAL will cross AZ boundaries two times, and in a second setup it will cross it three times. We don't have too much control now over which safekeeper is selected (it is done automatically) but we can report the current source safekeeper in the event so that we can keep track of it.

There is also a question in close placement of proxy and compute and compute and pageserver. Ideally all the pipeline in healthy situation should look like that:

Untitled - Frame 3

Right now we are more or less randomly throw that boxes in AZ's, so on average we cross AZ lots of times.

But let's discuss that separately -- both how to detect cross-az traffic and how to better colocate services.

vadim2404: agreed to discuss it separately

docs/rfcs/021-metering.md
kelvich: I kind of assumed implicit knowledge about our segment pipeline here, but rather should have written that explicitly. So the systems of record here are segment (for recent events) and Snowflake where this event are collected in a big table.

So we can have that windows with existing systems without introducing kafka

docs/rfcs/021-metering.md
kelvich: All events end up in Snowflake

docs/rfcs/021-metering.md
kelvich: Sources of truth are Segment and Snowflake. Redis/postgres are only caching last values to render current consumption in the UI

vadim2404: will you outline it somehow on the chart?

docs/rfcs/021-metering.md
vadim2404: segment and snowflake are not presented in this schema. Nonetheless, how can we be sure that the data was delivered durably to the Snowflake?

docs/rfcs/021-metering.md
vadim2404: yes and no. On one hand, users won't be able to create new projects, and it's ok. But on the other hand, we don't need to stop counting consumptions of up-and-running projects. 

I'm talking not only about emergency, but also about planned maintenance

docs/rfcs/021-metering.md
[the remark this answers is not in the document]
kelvich: @lassizci few questions about the picture:

* what is used on pageserver/proxy side to report metrics in this picture? prometheus endpoint?
* how can console receive that stream from vector? any existing packages?
* can vector downsample the data? e.g. poll proxy each 15 sec but push to the console each five minutes.

Generally I like that idea -- it could remove some amount of complexity about pushing the events / doing retries from the services themselves. OTOH I feel a bit uneasy about introducing quite a big service in the middle of our billing pipeline (however may be it is okay since we mainly can lose traffic events but other thing can be regenerated from the data stored in services)

docs/rfcs/021-metering.md
vadim2404: I reread the passage — understood what you meant. But, in terms of performance, need to think about this proposal.

docs/rfcs/021-metering.md
lassizci: Retries would be concern of collector agent, if we used one. That's part of their very core business.

Sending something with tcp requires dealing with connect timeouts, connection (as in response) timeouts and error responses. This is the reason statsd uses udp by default. It allows outsourcing the complexity.

docs/rfcs/021-metering.md
kelvich: Taking counter value is not an expensive operation and we collect them each 15sec for all tenants by prometheus. Synthetic size should be taken more carefully though.

> PageServer will start the calculation with a timer for all tenants?

I didn't say that. Timer can trigger walk across all tenants calculating sizes one after another. Or do that with some amount of parallelism.

docs/rfcs/021-metering.md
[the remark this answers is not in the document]
kelvich: > It's not that simple. The synthetic tenant size is calculated based on all the timelines.

Hm, can we assign some meaningful size to each timeline using that algorithm? That hopefully sum up to that one number per tenant.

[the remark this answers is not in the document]
hlinnaka: No, it's complicated. Let me give an example:

----

Imagine that you create two branches, A and B, at the same point from main branch, and do a couple of small updates on both branches. Then six months pass, and during those six months the data on the main branch churns over completely multiple times. The retention period is, say 1 month.

```
                       +------> A
                     /
--------------------*--------------------------------------------------------------------> main
                     \
                      +--------> B
``` 

In that situation, the synthetic tenant size would be calculated based on a "logical snapshot" at the branch point, that is, the logical size of the database at that point. Plus the WAL on branches A and B. Let's say that the snapshot size is 10 GB, and the WAL is 1 MB on both branches A and B. So the total synthetic storage size is 10002 MB. (Let's ignore the main branch for now, that would be just added to the sum)

How would you break that down per branch? I can think of three different ways to do it, and all of them have their own problems:

1. Subtraction method

For each branch, calculate how much smaller the total synthetic size would be, if that branch didn't exist. In other words, how much would you save if you dropped the branch. With this method, the size of branches A and B is 1 MB.

With this method, the 10 GB shared logical snapshot is not included for A nor B. So the size of all branches is *not* equal to the total synthetic size of the tenant. If you drop branch A, you save 1 MB as you'd expect, but also the size of B suddenly jumps from 1 MB to 10001 MB, which might feel surprising.

2. Division method

Divide the common parts evenly across all branches that need them. With this method, the size of branches A and B would be 5001 MB.

With this method, the sum of all branches adds up to the total synthetic size. But it's surprising in other ways: if you drop branch A, you might think that you save 5001 MB, but in reality you only save 1 MB, and the size of branch B suddenly grows from 5001 to 10001 MB.

3. Addition method

For each branch, include all the snapshots and WAL that it depends on, even if some of them are shared by other branches. With this method, the size of branches A and B would be 10001 MB.

The surprise with this method is that the sum of all the branches is *larger* than the total synthetic size. And if you drop branch A, the total synthetic size doesn't fall by 10001 MB as you might think.

----

A sort of cop-out method would be to show the whole tree of branches graphically, and for each section of WAL or logical snapshot, display the size of that section. You can then see which branches depend on which sections, which sections are shared etc. That would be good to have in the UI anyway.

Or perhaps calculate per-branch numbers using the subtraction method, and in addition to that, one more number for "shared size" that includes all the data that is needed by more than one branch. 

But the bottom line is that it's not straightforward to attribute the synthetic size to individual branches. There are things we can do, and all of those methods are pretty straightforward to implement, but they all have their own problems. What makes sense depends a lot on what you want to do with the number, what question you are trying to answer.

docs/rfcs/021-metering.md
kelvich: Screen Shot 2022-11-22 at 16 27 46

kelvich: > Nonetheless, how can we be sure that the data was delivered durably to the Snowflake?

Segment is basically managed Kafka sending data to the Snowpipe. I hope they did integration properly, but can't guarantee it as well. At the same time don't have too big concerns either -- segment will store event for a few day if can't connect to Snowflake. We can iterate on that one more time and have some system of record on our premises.

docs/rfcs/021-metering.md
kelvich: It is on the chart. I didn't describe that in text mostly because it is pre-existing infra.

docs/rfcs/021-metering.md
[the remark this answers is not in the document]
lassizci: @kelvich I think I would
* prefer to push with statsd to vector
* vector can for example expose prometheus format /metrics
* vector can also aggregate data. It can actually do a lot of things with [transforms](https://vector.dev/docs/reference/configuration/transforms/), [VRL](https://vector.dev/docs/reference/vrl/)

@antonyc mentioned having used it. Would you have insights about could it suit for this kind of use case?
I've just tried briefly for deduplicating rust stacktraces from plaintext logs, which actually seemed possible with VRL and combining transforms. [Unit testing](https://vector.dev/docs/reference/configuration/unit-tests/) is also possible, which could probably be considered as a requirement, considering we are talking about biling events.

> OTOH I feel a bit uneasy about introducing quite a big service in the middle of our billing pipeline

Understand the concern and it is a service on its' own and can get complicated too. On the other hand it's quite simple to get started with. A very simplified 2 minute demo would be:

```
$ cat vector.yaml 
data_dir: /tmp/vector
sources:
  statsd_source:
    type: statsd
    address: 0.0.0.0:8125
    mode: udp
transforms:
  aggregator:
    type: aggregate
    inputs:
      - statsd_source
    interval_ms: 5000
sinks:
  debugger:
    type: console
    inputs:
      - statsd_source
    target: stdout
    encoding:
      codec: json
  prom:
    type: prometheus_exporter
    inputs:
      - aggregator
    address: 0.0.0.0:9000
    default_namespace: mymetrics
    
$ ./vector -c vector.yaml
```

```
$ while [ 1 ]; do curl http://localhost:9000/metrics; sleep 1; done
```

```
$ cat metrics.py 
import socket, time

sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
sock.sendto(b"my_metric:1|c|#tenant:foo,", ("localhost", 8125))

time.sleep(6)

for i in range(1000):
  sock.sendto(b"my_metric:1|c|#tenant:foo,", ("localhost", 8125))

$ python metrics.py
```

So sending counter 1, waiting for 6 seconds, then sending 1000.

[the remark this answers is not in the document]
lassizci: Could also post over http from vector, but wasn’t as quick to set up

docs/rfcs/021-metering.md
stepashka: > > this approach makes the console a single point of failure.
> 
> It is already.

it is for establishing new connections, but now it's on a critical path for accurate billing too..

> Or we can just report to segment right from the service. I thought about that initially, but then it is tricky for the console to subscribe to that events, and not sure moving data through the segment will improve our reliability in practical terms.

could be segment or kafka or something else.


> > i think there could be the situations when the sender thinks that the message wasn't delivered, when in fact it was.
> 
> On failure/timeout we can just avoid resetting delta counter / last send at, so instead of retry we will report delta over a bigger time frame.

we need to perform careful testing with all of our LBs and reverse proxies in place to make sure the sender does think the data wasn't accepted when in fact it was.
going away from deltas could help with the case that I mentioned too.

stepashka: > and also not able to connect to the existing ones, which is worse

do we plan to stay with this design for good?

docs/rfcs/021-metering.md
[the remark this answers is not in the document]
kelvich: I took a closer look on Vector. Originally I liked it a lot, so explored on how we can use and almost with each of the use cases there is some nit in details.

* Using it to export data to segment or directly to the snowflake. That is not currently supported https://github.com/vectordotdev/vector/issues/1460 and https://github.com/vectordotdev/vector/issues/4365
* With some VRL we could calculate sum of increments and publish result in redis key for console to query it. But currently key/values updates are not supported, only lists/channels.
* Using existing client library could harden our pushes. Statsd client is super simple, just UDP packets without acks. So we would even have to change approach to deltas. That feels like a downgrade compared to http. I went through the list of sources and one possible push client is open telemetry, but vector only supports otel logs, not metrics. There is also generic http source which we could use but then the pipeline is `neon service -> (http) -> vector -> (http) -> console -> segment`. And vector doesn't add a lot of value in such a setup.
* Alternative would be to actually don't do any pushes from services, let `vector` scrape prometheus endpoints and then deliver them to the console. That setup has a nice property of not changing things in our services. However it is still wasteful to scrape everything and there is no good way to filter out unchanged events in the pipeline. So IMO better to go with pushes.

So with that I'd suggest to just POST jsons to the console for now.

Also, having few more chats with @ololobus on possible future ways to avoid segment->snowflake data flow and ways to show traffic / compute usage in console on different intervals I think we will end up with Kafka/Redpanda/Materialize for that pipeline later. But still with console acting as a proxy for all events we can play with few approaches only changing one component. (cc @vadim2404 -- I'm about to give up on no Kafka policy)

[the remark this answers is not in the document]
kelvich: (or we could use existing VictoriaMetrics to show that info)

docs/rfcs/021-metering.md
kelvich: > do we plan to stay with this design for good?

we do have to check auth credentials before letting connections in. And something need to wake up computes when they are down. So i guess answer here is yes.

docs/rfcs/021-metering.md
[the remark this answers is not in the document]
lassizci: Fair enough, doesn't have to be vector. Let me still make a couple points that might be worth thinking about in general:

> Using it to export data to segment or directly to the snowflake

This is the part I would actually be the most hesitant about (also hence the `? `). I think what we collect is rather standard -> use standard collection methods. But what we want to have eventually in Snowflake is much more open ended -> the last mile is probably where the magic needs to happen. Probably would prefer separate microservice from console, but that could happen later too. Scalability requirements are different for different parts of console and that will be an issue at some point probably.

> With some VRL we could calculate sum of increments and publish result in redis key for console to query it. But currently key/values updates are not supported, only lists/channels.

If redis is treated like a queue, list is probably the right data structure. If it was a key where value is updated constantly, it would be essentially the same as scrape.

 > So we would even have to change approach to deltas.
  
I think there are couple mixed things here: my point relates to @stepashka 's [comment](https://github.com/neondatabase/neon/pull/2884#discussion_r1029098828) a little. 

* We don't have reliable absolute counters. Proxy [increments counters](https://github.com/neondatabase/neon/blob/05db6458df5430fbd659962d8b6c5a16ef735da9/proxy/src/proxy.rs#L276-L286) on send/receive. The internal registry is cleared when proxy is restarted.
* But if we could receive every counter increment and would be able to aggregate these as close as possible to the source (like something receiving on every node), and then have a reliable method to pass these forward, we would be able to worry less about loss of signal. It's then also separate concern whether Segment/Snowflake wants deltas or absolute counter.

>  That feels like a downgrade compared to http
* proxy doing cross continent http might have concerns as the latencies are large, which is why this would be nice to be able to offload from proxy to _something_ that's local, either pushing to console or to queue where console can read
* I wouldn't use statsd either, unless the aggregator was something that supports it out of the box. [Datadog's format](https://docs.datadoghq.com/developers/dogstatsd/datagram_shell/?tab=metrics#datagram-format) is rather de-facto nowdays and that's what vector uses too. It's not end to end format, but neither is prometheus.

docs/rfcs/021-metering.md
LizardWizzard, line 120: > we will have to distribute secrets for that connection to all of the components, etc.

Isnt it the same with "send everything to console" approach?

> will introduce new dependencies

For storage we create dependency on console, so anyway it is a new dependency

docs/rfcs/021-metering.md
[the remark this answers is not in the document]
antonyc: I think we should start with something that would allow us to move quickly, to give the needed agility to understand what we bill for, how frequently we count usage, and bill for it. That would allow us to strongly define business logic for billing and understand the limits of our solution to improve the bottlenecks later. 
A pipeline with vector accepting events via HTTP (any protocol it supports really) + Kafka as a bus + Kafka-Snowflake connector to store raw events in Snowflake + Kafka to Postgres connector to store events in our own single RDS could be a reliable solution. And the bills/usage could be calculated in a separate service later. That would deliver the final numbers in terms of minutes to the customers. But it's also a very expensive pipeline (did not do a napkin maths, but it seems to take very much time to build correctly). So we probably need to iterate?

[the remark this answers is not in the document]
kelvich: > Pushing everything to console at some point will create scalability problems

Can you share your though process behind that? My back of the napkin math tells me that hardly ever can burn one cpu core. May be we just have somehow different models of how this thing should operate?

docs/rfcs/021-metering.md
```diff
@@ -0,0 +1,175 @@
...
+* `incremental` -- change in consumption since the previous event or service restart. That is `effective_cpu_seconds`, `traffic_in_bytes`, and `traffic_out_bytes`.
+* `absolute` -- that is the current value of a metric. All of the size-related metrics are absolute.
+
+Each service can post events at its own pace and bundle together data from different tenants/endpoints.
+
+The console algorithm upon receive of events could be the following:
+
+1. Create and send a segment event with the same content (possibly enriching it with tenant/timeline data for endpoint-based events).
+2. Update the latest state of per-tenant and per-endpoint metrics in the database.
+3. Check whether any of that metrics is above the allowed threshold and stop the project if necessary.
```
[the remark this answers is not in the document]
kelvich: For some reason I can't comment to the next two comments, so replying to them here:

> A pipeline with vector accepting events via HTTP (any protocol it supports really) + Kafka as a bus + Kafka-Snowflake connector to store raw events in Snowflake + Kafka to Postgres connector to store events in our own single RDS could be a reliable solution.

Yes, that should work, but IMO we can start with the current approach as it quite dumb and experiment with different event delivery schemes after.

Why do we need vetror in that scheme instead of just pushing events to kafka? To avoid kafka-client dependency?

I started thinking of Kafka mostly due to the Kafka streams / ksql -- we could have cheap running aggregates with them. Otherwise it look that it would be pretty similar to just inserting batch of events in snowflake+postgres right in the `usage_events` handler. But with postgres/snowflake such queries will still be expensive (both in wall time and cpu consumption).

Another alternative to Kafka streams is Materialize, I'd be interested to try it out experimentally.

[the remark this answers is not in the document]
antonyc: > Why do we need vetror in that scheme instead of just pushing events to kafka? To avoid kafka-client dependency?

That's what I meant, yes, to not introduce new dependencies on side of the event sources. Vector has a lot of inputs, HTTP is one of them, as long as we already introduce support for sending events via HTTP. Also, Vector can 
1. enrich events on its side, adding `region` field to the payload, for example.
2. take care of retries in cases of network problems. It has both a simple in-memory buffer and an on-disk buffer for persistence.

docs/rfcs/021-metering.md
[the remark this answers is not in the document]
LizardWizzard: > I think it should say control-plane, not console

Currently it is console. If we plan to implement it in console then it is console and not control plane. 

> Can you share your though process behind that?

I think the main problem with that is that shape of our infra is underspecified in terms of how many tenants we can handle per "cell" (what cell actually is?). I imagine that we can have hundred pageservers in one cell (taking <100 as up to 100 from @ololobus comment above). One pageserver can have thousands (tens of thousands?) of assigned tenants. Do we have statistics on how many of them can be active at the same time? We can have multiple active computes per tenant (several timelines). For one entity metric is reported from 6 places one of them being the console itself. Other ones are 3 safekeepers, proxy and pageserver. Lets assume we have 10k tenants on each pageserver and 20% of them are active (not sure how realistic is it). Then we have 2k of active tenants per pageserver which is 2k*100 = 200k. Then we have safekeepers and proxy, then we have multiple timelines and multiple time series per each one. We should do batching, because doing one request per tenant will be quite inefficient. So we have 2k*100ps tenants, we have 3x on safekeepers and one for proxy. Assuming one time series per tenant it is 600k data points per minute. Obviously it wont directly translate to RPS. How do we do batching? Even if we consolidate all data from each node in one request then requests will be big. With these numbers it looks necessary to do batching on the client side. If it is our own protocol then we need to code and maintain this. If we use stable third party protocol it may be way easier to support. Here we can split transport from schema and discuss it separately

> I think that current API backends can ingest thousands of requests per second.

What kind of processing they do? It is simple to do many RPS if there is not much to do (pipe request to /deb/null) but sending data over the network to another system is different.

Were there any load tests for API backends?

> My back of the napkin math tells me that hardly ever can burn one cpu core. 

What was your path to "hardly ever can burn one cpu core"?

Looks like design of this pipeline will be with us for a long time so to me it makes sense to go beyond napkin math and plan it carefully with load testing involved and various tools tested against expected volume of data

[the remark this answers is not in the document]
antonyc: I think in terms of cell capacity, the upper bound limit is dictated by the number of pods/nodes the k8s cluster could handle, and the most recent figures k8s has are 5000 nodes, 150000 pods per cluster. These are recent figures for bleeding edge deployments and k8s v1.25, I doubt these figures are easy to have out of the box in a managed AWS Kubernetes. In the worst case, that would mean 150000 timelines in a cell, and 150000 tenants in a cell. Which could mean 150k*6=900k events per 1 minute if we send them each minute, which is 15k RPS load. This is rather big volume, and it's too much for the services we intend to use as the consumers of usage data and invoice producers. But we need to get to that volume first.

petuhovskiy (commented):
A lot of good points in the RFC and discussions. I think this RFC is a good starting point, but we can create more smaller RFCs on smaller independent topics, and move our comments with possible approaches there.

Topics for these RFCs can be:

## 1. Consumption events calculation

It seems that everyone mostly agrees to the following schema:
* All services can produce "consumption events" at any time
* We collect all events in a single place
* End-user costs are calculated based on these events

We can further discuss how to calculate these events in different services, how often we should report them and what content should be in these events. For example, there was a discussion about what is "Synthetic storage size" and how to calculate it, it can be discussed in this separate RFC.

## 2. Events collection

This RFC can cover what solution we should use to collect events from all our services, e.g. should we use pull/push model or get a ready-made solution like vector.
We can also cover the format of events, e.g. should we use JSON, protobuf or something else.

Our collection solution should:
- don't lose events, and make retries if one of the services is down
- don't lose events on restart, that means we likely need to have persistent buffer with events stored on disk
- be simple to use and configure, to make it easy to test
- be easy to enable collection from new services and disable collection from old stopped services
- provide exactly-once event delivery, can be done by appending UUID to events or single batch
- have reasonable latency for events delivery

## 3. Events storage

There is an argument that collecting all events in a single place is bad for scalability, we can discuss this in this RFC. From my point of view, we can easily scale by sharding events by tenant_id. I know that large companies use event sourcing and it works for them, so my opinion is that it should work for us too.

We can discuss where we should store events. Requirements for the storage:
- should be able to store a lot of events
- it should should be easy to push new events
- events should be durably stored for some time
- events should be easy to query/consume at any time

Kafka sounds like a good solution for this, but we can discuss other options too.

## 4. Further processing

We can discuss how we should process events to calculate costs and pipelines for pushing events to other services.

---

.

I mostly agree with the proposed solution for the first iteration. I see the overall schema as follows:

```
  ┌──────────────┬────────────────┐
  │  pageserver  │                │
  └──────────────┘      ┌─────────▼──────────┐        ┌───────┐       ┌─────────────┐
                        │                    │        │       │       │             │
  ┌──────────────┐      │ POST /usage-events ├────────► Kafka ├──────►│   console   │
  │    proxy     ├──────►  (control-plane)   │        │       │       │             │
  └──────────────┘      │                    │        └───────┘       └─────────────┘
                        └─────────▲──────────┘
┌──────────────────┐              │
│ autoscaler-agent ├──────────────┘
└──────────────────┘
```

vadim2404:
@petuhovskiy, I saw several implementations like you proposed, except for one moment:
the HTTP endpoint was replaced with a local log file that was pushed to Kafka via filebeat directly (https://www.elastic.co/guide/en/beats/filebeat/current/kafka-output.html).

The one point of failure (HTTP endpoint) is replaced with filebeat, which supports retries/continuation from a previously committed state.

antonyc:
I personally like d) as an intermediate option. Vector could work with pipelines of [HTTP server](https://vector.dev/docs/reference/configuration/sources/http_server/) source -> [log to metric](https://vector.dev/docs/reference/configuration/transforms/log_to_metric/) transform -> [Aggregate](https://vector.dev/docs/reference/configuration/transforms/aggregate/) transform aggregating needed metrics with needed granularity -> [HTTP](https://vector.dev/docs/reference/configuration/sinks/http/) sink -> console to 


Pipeline `sources` -> `multiple Vector instances for a cell` -> `Kafka` -> `single postgres table for all the cloud` -> `some accounting engine` seems to be bright future with have following properties: not losing events, tolerant to repeating events, bill multiple events on per-hour basis, low costs by sending only the needed events to the accounting engine.
**Vector:**
Gives agility in sources, can accept quite a several formats; HTTP, Prometheus is easy to implement on storage nodes. Push does not need to implement storage nodes discovery, so the proposed HTTP or statsd seems easier to maintain.
Gives batching, that increases the throughput when working with Kafka. 
Not losing events (or losing very little): Vector retries the events.
**Kafka**
Given Kafka streams, we can aggregate and bill on per-hour basis, reducing the volume of events (which we are gonna pay for). If the events have idempotency field (timestamp+hostname) we can probably ignore duplicate events on the Kafka side with [log compaction](https://medium.com/swlh/introduction-to-topic-log-compaction-in-apache-kafka-3e4d4afd2262). So at-least-once seems to be OK for us.
Put everything to Postgres with [Kafka Postgresql Sink](https://docs.confluent.io/cloud/current/connectors/cc-postgresql-sink.html).
**postgres table**
Having the events in a Postgres table would allow us to enrich them in the background with client id / other data. Then send it to an external system in the background (Like Metronome, others) or use our home-grown (which does not exists and is not a rocket science to build, but we need to get it first). And add taxes etc. And we could try to show usage from that table to the users and bills from some external system.

But this is a lot to build.

c) the Clickhouse sounds great for showing historical data, to downsample it automatically. But it has weak guarantees on ingestion, so I would not use it as a primary source of data. But a secondary for graphs, etc.
Option d) seems to be fast to build, but there might be a problem with it in scalability. If we leave >1 instance of vector accepting and aggregating the events than we got each GB-hour of storage billed twice. We could add another layer of single instanced vector post-aggregating the events for each hour, but I doubt that would be consistent solution - it still allows billing each hour twice if events come with significant intervals into that post-aggregator. 
And having 1 instance handling all the traffic (15k RPS load) would probably work if we give it enough CPU and mem, it's not that big a deal for vector. I see this scheme as intermediate and thought it will work for sure, we will need to rework ingest pipeline later, we probably do not want to live with single instance for a long time, as long as we don't want to introduce retries to event emitters.
f) needs storage service discovery, I would try to avoid that.
g) not sure we have any guarantees on this metrics storage, probably losing them is not big a deal now, and we probably don't want to heavily rely on that storage.

So I would vote for scheme a) with minor additions (see above) and I would start with from the back: deliver data to some accounting engine (which is pricy orb, it probably has the needed aggregations) with `POST /usage` via HTTP. That would allow us to start with paying customers. Then add Vector just to bring kafka later. Then we could add temporary `Kafka -> Vector -> POST /usage` to just the reduce amount of events to the Orb. And then add Postgres table to store the data ourselves. That would allow us to gradually move step by step to our own solutions, while already having 100% working billing solution. Because building those pipelines from the scratch would put us I would say human-months away from billing customers.

lubennikovaav (approved):

PR #2891, opened 2022-11-22
Disable statement timeout for performance tests

bayandin:
Fix `test_seqscans` by disabling statement timeout.

Also, replace increasing statement timeout with disabling it for performance tests. This should make tests more stable and allow us to observe performance degradation instead of test failures.

Ref https://github.com/neondatabase/neon/actions/runs/3520063845/jobs/5900633702#step:6:277

### DIFF
--- a/test_runner/performance/test_perf_pgbench.py
+++ b/test_runner/performance/test_perf_pgbench.py
@@ -88,7 +88,7 @@ def run_test_pgbench(env: PgCompare, scale: int, duration: int, workload_type: P
     env.zenbenchmark.record("scale", scale, "", MetricReport.TEST_PARAM)
 
     password = env.pg.default_options.get("password", None)
-    options = "-cstatement_timeout=1h " + env.pg.default_options.get("options", "")
+    options = "-cstatement_timeout=0 " + env.pg.default_options.get("options", "")
     # drop password from the connection string by passing password=None and set password separately
     connstr = env.pg.connstr(password=None, options=options)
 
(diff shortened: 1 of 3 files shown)


bayandin:
I've triggered the bench job; it finally should be green: https://github.com/neondatabase/neon/actions/runs/3522641461

LizardWizzard (commented):
Do we exceed 1h timeout? 

Feels better to have higher timeout than not having a timeout at all (there is a default limit for an action job duration of 6h). At some point we'll have some job that will analyze results and fail the check if there is significant performance degradation. Failing to complete a test in reasonable time looks like a good way to signal problems until we have something like that

bayandin:
> Do we exceed 1h timeout?

Currently, we do exceed it for 50GB database init with pgbench. Enabling prefetch fixes this, but for this case, it would be nice to know what speed-up do we get (just nice to have, not really necessary).

> Failing to complete a test in reasonable time looks like a good way to signal problems until we have something like that

That's true, but for now, it fails much more frequently when we increase the dataset size and forget about the timeout rather than finding some real performance issues.

> At some point we'll have some job that will analyze results and fail the check if there is significant performance degradation.

I'm ready to pick this as one of the following tasks.

LizardWizzard (approved):
> it would be nice to know what speed-up do we get (just nice to have, not really necessary).

Ok, lets try it. 

We still have default timeout in gh actions. Another way might be to allow configuring it through workflow params so you can set this value from workflow_dispatch event. But probably too complicated without clear benefits

PR #2894, opened 2022-11-22
List `latest_gc_cutoff_lsn` in the Pageserver OpenAPI spec

ololobus:
It seems that it's present in the API response for quite a while. It's just not listed in the spec, fix it.


ololobus:
Not sure whether it's optional or required, though. It looks like we de-reference the result of `timeline.get_latest_gc_cutoff_lsn()` unconditionally, so it's a required attribute?

vadim2404:
Should be always returned

PR #2896, opened 2022-11-22
[safekeeper] Serialize LSN in the `term_history` according to the spec

ololobus:
Use string format in the timeline status HTTP API reponse.

### DIFF
--- a/safekeeper/src/http/routes.rs
+++ b/safekeeper/src/http/routes.rs
@@ -12,7 +12,6 @@ use tokio::task::JoinError;
 
 use crate::safekeeper::ServerInfo;
 use crate::safekeeper::Term;
-use crate::safekeeper::TermHistory;
 
 use crate::timelines_global_map::TimelineDeleteForceResult;
 use crate::GlobalTimelines;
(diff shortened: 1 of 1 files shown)


arssher:
That would break sk control file, as `TermHistory` is serialized there. We need either bump it, or transfuse it into separate struct like `TimelineStatus` in routes.rs -- as far as I know serde doesn't offer better option. I incline to the latter.

petuhovskiy:
Can we somehow recursively apply `#[serde(serialize_with = "display_serialize")]` only to the field inside `routes.rs`? For example, make something like this work:

```
#[derive(Debug, Serialize)]
struct AcceptorStateStatus {
    term: Term,
    epoch: Term,
    #[serde(serialize_with = "display_serialize")]
    term_history: TermHistory,
}
```

ololobus:
> Can we somehow recursively apply #[serde(serialize_with = "display_serialize")] only to the field inside routes.rs? 

Also thought about something like that, will try

ololobus:
OK, it applies to the whole struct
```
#[serde(serialize_with = "display_serialize")]
   |                              ^^^^^^^^^^^^^^^^^^^ `TermHistory` cannot be formatted with the default formatter
   |
   = help: the trait `std::fmt::Display` is not implemented for `TermHistory`
   = note: in format strings you may be able to use `{:?}` (or {:#?} for pretty-print) instead
```

Probably it's possible to define some custom serializer, or just define some different response type as Arseny suggested

petuhovskiy (approved):

PR #2909, opened 2022-11-23
Implement index vacuum performance tests and comparisons for all included index methods

MMeent:
This assumes we can use pg_prewarm to clear out caches.

### DIFF
--- a/test_runner/performance/test_index_vacuum_perf.py
+++ b/test_runner/performance/test_index_vacuum_perf.py
@@ -1,5 +1,4 @@
 from contextlib import closing
-from typing import Tuple
 
 import pytest
 from fixtures.compare_fixtures import PgCompare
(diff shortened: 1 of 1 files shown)


test_runner/performance/test_index_vacuum_perf.py
```diff
@@ -0,0 +1,71 @@
...
+            cur.execute(
+                f"INSERT INTO {table} "
+                + f"(SELECT ({coldef}) as icol "
+                + f"from generate_series(1, {table_size}) gen(seed))"
+            )
+
+            cur.execute("SELECT pg_prewarm('index_testdata')")
+
+            cur.execute(
+                f"CREATE INDEX {index} " f"   ON {table} " f"       USING {index_method} (icol)"
```
vadim2404, line 59: you don't need to have a bunch of f-strings.

```python
f"CREATE INDEX {index} ON {table} USING {index_method} (icol)"
```

LizardWizzard:
As discussed moving to draft

PR #2922, opened 2022-11-25
[WIP] Initial version of values, not yet working in the new clusters.

lassizci:
helm values for the new `storage-broker`. gRPC, no authentication. The goal would be to have internal ingress and TLS still. Falling back to internal LoadBalancer for now, if we can't get this working otherwise.


arssher (approved):
Checked that broker is reachable on old staging and on us-east-2 staging.

.github/helm-values/prod-eu-central-1-gamma.neon-storage-broker.yaml
```diff
@@ -0,0 +1,57 @@
...
+    cert-manager.io/cluster-issuer: "cert-manager-clusterissuer"
+
+  hosts:
+    - host: storage-broker-gamma.eu-central-1.aws.neon.tech
+      paths:
+        - path: /
+          pathType: Prefix
+  tls:
+    - hosts:
+        - storage-broker-beta.us-east-2.aws.neon.build
```
arssher, line 22: A typo?

.github/workflows/build_and_test.yml
```diff
@@ -888,7 +888,13 @@ jobs:
           helm upgrade ${{ matrix.proxy_job }}       neondatabase/neon-proxy --namespace neon-proxy --install -f .github/helm-values/${{ matrix.proxy_config }}.yaml --set image.tag=${DOCKER_TAG} --wait --timeout 15m0s
           helm upgrade ${{ matrix.proxy_job }}-scram neondatabase/neon-proxy --namespace neon-proxy --install -f .github/helm-values/${{ matrix.proxy_config }}-scram.yaml --set image.tag=${DOCKER_TAG} --wait --timeout 15m0s
 
-  deploy-proxy-new:
+      - name: Re-deploy storage-broker
```
arssher, line 891: We'd better run this only when storage_broker directory changes. Unfortunately, GH actions doesn't support such filtering out of the box on job/step level, but there is 
https://github.com/dorny/paths-filter

Feel free to add it here or leave that for a separate PR.

Also, deploy-prod-new still misses the deploy, but I'm sure this is known.

Also, I don't see much value in customizing namespace name, but that's nitpicking.

lassizci: > We'd better run this only when storage_broker directory changes.

We would need to do this also when ../libs or ../utils change, as the storage_broker seems to depend on those paths, am I right? I have a bit of a worry, it would just get confusing, as we update the same image, but would then run old version of the image. I think we probably need to have a separate discussion about how we handle this, as we don't really do that for anything else that's in the same image.

> Also, deploy-prod-new still misses the deploy, but I'm sure this is known.

Yeah, a bit work in progress.

> Also, I don't see much value in customizing namespace name, but that's nitpicking.

Namespaces are great way to separate apps and their resources. When app deployments (or environment, in case of neon-stress) are separated to their own namespaces, implementing  NetworkPolicies and RBAC becomes possible in a sane way.

arssher: > We would need to do this also when ../libs or ../utils change

Yes, good catch.

PR #2926, opened 2022-11-25
Sentry integration for storage.

KlimentSerafimov:
Added basic instrumentation to integrate sentry with the proxy, pageserver, and safekeeper processes.
Currently in sentry there are three projects, one for each process.


safekeeper.pid
```diff
@@ -0,0 +1 @@
+68975
```
ololobus, line 1: Looks like it was committed by accident?

ololobus (commented):
Also PR contains some seemingly unrelated changes to the vendor/postgres, was it intentional?

safekeeper.pid
KlimentSerafimov: true, just removed it.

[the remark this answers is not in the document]
KlimentSerafimov: I just removed it.

pageserver/src/bin/pageserver.rs
```diff
@@ -52,6 +52,18 @@ fn version() -> String {
 }
 
 fn main() -> anyhow::Result<()> {
+    let _guard = sentry::init((
+        "https://0a872afd470f42bf9f5cdfb851459ad6@o1373725.ingest.sentry.io/4504220031582208",
```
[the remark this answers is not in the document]
KlimentSerafimov: make sense.

hlinnaka:
Seems straightforward, as far as the code changes go.

What are we going to do with Sentry? With this PR, do we already start sending some events? Panics?

proxy/src/main.rs
```diff
@@ -40,6 +40,18 @@ async fn flatten_err(
 
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
+    let _guard = sentry::init((
+        "https://0b4e22f1b09d47a8a0058d6d7eb87dcb@o1373725.ingest.sentry.io/4504220230483968",
```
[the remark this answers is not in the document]
funbringer: I'd add an optional flag `--sentry=...`.

libs/utils/src/sentry_init.rs
```diff
@@ -0,0 +1,17 @@
+pub fn init_sentry(maybe_sentry_url: Option<&String>, process_name: &str) {
+    match maybe_sentry_url {
+        Some(sentry_url) => {
+            let _guard = sentry::init((
```
vadim2404, line 4: the lifetime of this `_guard` is not too long to catch all everything. Should it be used in `main`?

vadim2404: I'll also recommend to mark this method with `#[must_use]` attribute to generate warning for everyone who will forget to use the return value

yeputons: Nit: if it's used, it should be `guard`, not `_guard`.

libs/utils/src/sentry_init.rs
```diff
@@ -0,0 +1,23 @@
+use sentry::ClientInitGuard;
+
+pub fn init_sentry(
+    maybe_sentry_url: Option<&String>,
+    process_name: &str,
+) -> Option<ClientInitGuard> {
+    return match maybe_sentry_url {
+        Some(sentry_url) => {
+            let _guard = sentry::init((
+                sentry_url.as_str().to_string(),
```
yeputons, line 10: I think you should be able to take `Option<&str>` here and pass `&str` without any conversions. If not, I'd recommend taking `Option<String>` right away instead of cloning.

pageserver/src/bin/pageserver.rs
```diff
@@ -77,6 +78,9 @@ fn main() -> anyhow::Result<()> {
         )
     })?;
 
+    let sentry_url = arg_matches.get_one::<String>("sentry-url");
+    let _guard = init_sentry(sentry_url, "pageserver");
```
yeputons, line 82: `_sentry_guard` here and in other places?

pageserver/src/bin/pageserver.rs
```diff
@@ -77,6 +78,9 @@ fn main() -> anyhow::Result<()> {
         )
     })?;
 
+    let sentry_url = arg_matches.get_one::<String>("sentry-url");
```
yeputons, line 81: I'd call `get_one` right inside `init_sentry` arguments. I don't see it being used outside `init_sentry`

yeputons (approved):
The code looks good to me, as long as it works :)

libs/utils/src/sentry_init.rs
yeputons: Also, maybe add a comment to describe what to do with the returned value: keep it alive while something-something-something.

KlimentSerafimov:
> Seems straightforward, as far as the code changes go.
> 
> What are we going to do with Sentry? With this PR, do we already start sending some events? Panics?

yes, I just need to add one more commit for proxy and resolve the reviews.

libs/utils/src/sentry_init.rs
KlimentSerafimov: So it seems to me that _guard doesn't need to be used. I'm keeping it more as a stylistic copy of the sentry tutorial. It also seems to me that it doesn't need to be kept alive either (but I'm less sure about this).

pageserver/src/bin/pageserver.rs
KlimentSerafimov: I agree it's not used outside init_sentry, but I would still keep it as a separate variable for legibility purposes (otherwise the line gets is long and less legible).

petuhovskiy (commented):
From @cicdteam message in Slack:

> sentry client libs usually use SENTRY_DSN environment variable. I think env vars more comfortable then cli args (and it easy to automate)

I also think that using `SENTRY_DSN` environment variable is better than `--sentry-url` cli argument. At least it's more universal, because we don't need to add separate CLI argument to every binary, we can just call `std::env::var` in `init_sentry()`.

libs/utils/src/sentry_init.rs
KlimentSerafimov: Option<&str> didn't work because get_one returns Option<&String> (and you can't do str because get_one requires the type parameter to be clonable). Also Option<String> didn't work because you are cloning anyways just on the caller side.

The change that did work is: sentry_url.as_str().to_string() -> sentry_url.to_string(). There was no need to cast it as a &str before instantiating it as String. Not sure how to avoid cloning.

libs/utils/src/sentry_init.rs
```diff
@@ -0,0 +1,23 @@
+use sentry::ClientInitGuard;
+
+pub fn init_sentry(
+    maybe_sentry_url: Option<&String>,
```
funbringer, line 4: Why take optional sentry url? The name says "init sentry", yet the actual action is "mostly init, but also do nothing sometimes". Let's use `sentry_url: &str` instead and do `let Option(url) = ...` at call sites.

pageserver/src/bin/pageserver.rs
funbringer: Let's not do that. `init_sentry` should only be responsible for one thing: actually initializing sentry regardless of argument names or parsing or whatever.

pageserver/src/bin/pageserver.rs
funbringer: I don't think `_sentry_` prefix adds value here. if `_guard` is returned from `init_sentry`, it obviously has to do with sentry.

libs/utils/src/sentry_init.rs
KlimentSerafimov: I moved the Option<&String> inside init_sentry to have less lines of code at the call sites (due to the if statement). I'll try to integrate your suggestion compactly.

KlimentSerafimov: this is the alternative

```
    if let Some(sentry_url) = maybe_sentry_url {
        let _guard = init_sentry(sentry_url, "pageserver");
    };
```

KlimentSerafimov: another suggestion is to simply rename init_sentry to maybe_init_sentry to keep the one-line

funbringer: > this is the alternative

Unfortunately, that won't work, because the guard object will be dropped. You could still do this:

```rust
let _guard = arg_matches.get_one::<String>("sentry-url").map(|url| {
  init_sentry(url, "process-name");
});
```

libs/utils/src/sentry_init.rs
yeputons: https://docs.rs/sentry/latest/sentry/fn.init.html

> This returns a client init guard that must be kept in scope and that will help the client send events before the application closes. When the guard is dropped, then the transport that was initialized shuts down and no further events can be sent on it.

Seems like we _have to_ keep it alive, otherwise Sentry reporting won't work.

libs/utils/src/sentry_init.rs
yeputons: Just `.as_str()` works for me: [`init`](https://docs.rs/sentry/latest/sentry/fn.init.html) takes `Into<ClientOptions>`, [one case of which](https://docs.rs/sentry/latest/sentry/struct.ClientOptions.html#impl-From%3C(T%2C%20ClientOptions)%3E-for-ClientOptions) is `(T: IntoDsn, ClientOptions)` , and there is [`IntoDsn<&'a str>`](https://docs.rs/sentry/latest/sentry/trait.IntoDsn.html#impl-IntoDsn-for-%26%27a%20str).

pageserver/src/bin/pageserver.rs
yeputons: It is, but I think it's better if a variable's name is descriptive without reading its initialization. Not a big deal, your call.

libs/utils/src/sentry_init.rs
yeputons: Ah, by "used" I meant "returned from the function". We use it inside the function by returning it.

libs/utils/src/sentry_init.rs
KlimentSerafimov: incorporated.

libs/utils/src/sentry_init.rs
KlimentSerafimov: thanks! Since my editor likes _guard more (since it's not explicitly used) I'll keep it as such.

pageserver/src/bin/pageserver.rs
KlimentSerafimov: I opted for _sentry_guard since I agree with yeptons's reasoning.

KlimentSerafimov:
> From @cicdteam message in Slack:
> 
> > sentry client libs usually use SENTRY_DSN environment variable. I think env vars more comfortable then cli args (and it easy to automate)
> 
> I also think that using `SENTRY_DSN` environment variable is better than `--sentry-url` cli argument. At least it's more universal, because we don't need to add separate CLI argument to every binary, we can just call `std::env::var` in `init_sentry()`.

I would prefer to do this in a subsequent PR.

KlimentSerafimov:
@hlinnaka @problame To answer your question, in short, yes: you can start sending panics, and you can add instrumentation for reporting other types of events. In general sentry will **automatically** reports errors, uncaught exceptions, and unhandled rejections as well as other types of errors depending on the platform (from: https://docs.sentry.io/platforms/rust/usage). Additionally, you can add instrumentation for tracing and profiling amongst other things, as well as vary the level of detail that you want from all events.

libs/utils/src/sentry_init.rs
```diff
@@ -0,0 +1,15 @@
+use sentry::ClientInitGuard;
+
+pub fn init_sentry(sentry_url: &str, process_name: &str) -> ClientInitGuard {
+    let _guard = sentry::init((
```
funbringer, line 4: Nit: This has been brought up already, but `_guard` is used here and thus should be renamed to `guard` instead.

hlinnaka:
Pageserver has a config file, `pageserver.toml`, that contains a bunch of options like S3 URLs, etc. Take a look at `PageServerConf` in `config.rs`. This new option probably belongs there.

If you add a new field there, make sure you grep for all the different places that need changing, there are many places that do something with all the fields. If you grep for one of the existing fields you should find them all.

hlinnaka:
> Pageserver has a config file, `pageserver.toml`, that contains a bunch of options like S3 URLs, etc. Take a look at `PageServerConf` in `config.rs`. This new option probably belongs there.

However, if there's a common convention to use the `SENTRY_DSN` env variable, as discussed here, then we should probably do that.

vadim2404:
Updated in order to use the SENTRY_DSN env var

.github/helm-values/dev-eu-west-1-zeta.neon-proxy-scram.yaml
```diff
@@ -8,6 +8,7 @@ settings:
   authBackend: "console"
   authEndpoint: "http://console-staging.local/management/api/v2"
   domain: "*.eu-west-1.aws.neon.build"
+  sentryURL: "{{ SENTRY_URL_PROXY }}"
```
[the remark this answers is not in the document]
vadim2404: updated to proposed solution

libs/utils/src/sentry_init.rs
```diff
@@ -0,0 +1,21 @@
...
+#[must_use]
+pub fn init_sentry(extra_options: &[(&str, &str)]) -> Option<ClientInitGuard> {
+    let dsn = env::var("SENTRY_DSN").ok()?;
+
+    let guard = sentry::init((
+        dsn,
+        sentry::ClientOptions {
+            release: sentry::release_name!(),
+            ..Default::default()
+        },
```
funbringer, line 13: I don't think placing `release_name!` here is a good idea. Judging by the fact that it's a macro and its [implementation](https://docs.rs/sentry-core/0.29.1/src/sentry_core/macros.rs.html#18), it should be called in the target app and then passed as a parameter to `init_sentry`.

That would also eliminate the need for passing app name in `extra_options`.

libs/utils/src/sentry_init.rs
```diff
@@ -0,0 +1,21 @@
...
+    ));
+    sentry::configure_scope(|scope| {
+        for option in extra_options {
```
funbringer, line 16: for (key, value) in extra_options

proxy/src/main.rs
```diff
@@ -47,6 +48,9 @@ async fn main() -> anyhow::Result<()> {
 
     let arg_matches = cli().get_matches();
 
+    // initialize sentry if SENTRY_DSN is provided
+    let _sentry_guard = init_sentry(&[("process", "proxy")]);
```
funbringer, line 52: I'd place this next to tracing init code, before args parsing.

pageserver/src/bin/pageserver.rs
```diff
@@ -87,8 +87,10 @@ fn main() -> anyhow::Result<()> {
     };
 
     // initialize sentry if SENTRY_DSN is provided
-    let _sentry_guard =
-        init_sentry(&[("process", "pageserver"), ("node_id", &conf.id.to_string())]);
+    let _sentry_guard = init_sentry(
+        release_name!(),
+        &[("process", "pageserver"), ("node_id", &conf.id.to_string())],
```
funbringer, line 92: Here and forth: why add `("process", <NAME>)` when `release_name!` already does that?

vadim2404: Initially, I thought it was added to be able to group by captures by tag. But, we've decided to go with a sentry URL per "process". Therefore it's useless anymore. Will delete it.

funbringer (approved):

LizardWizzard (approved):
Looks good to me, thanks!

PR #2928, opened 2022-11-25
wait for layer file uploads to finish before deleting layer files 

problame:
Test fails because of panic due to task::block_on() inside runtime thread.
I tried converting the call stack into async fn's to avoid use of
task::block_on(), but that doesn't work because layer_removal_cs is held
across await points.

Upon closer inspection, usage of layer_removal_cs doesn't protect all
readers, so, we will need to something about that before we get back
here.


problame:
Superseded by https://github.com/neondatabase/neon/pull/2942

PR #2932, opened 2022-11-25
startup: transition tenant into Broken state if config load fails

problame:
instead of skipping over it and not adding it to the tenants map


hlinnaka (approved):

problame:
rebased and merged

PR #2939, opened 2022-11-26
Fix bulk update test

knizhnik:


LizardWizzard:
As discussed moving to draft

PR #2942, opened 2022-11-28
storage_sync2: `wait_completion()` MVP

problame:
Commit 1:

```
 refactor: turn compaction code async

We need this to insert remote_client.wait_completion() calls.
That'll happen in the next commit.

The major wrinkle in this change is Tenant::timeline_delete().
Before this change, Tenant::timeline_delete() does the following:
1 Acquire timelines.lock()
2 get a ref to the timelines map
3 Set it to `TimelineState::Paused`
4 Get Tenant::layer_removal_guard()
  (Internally, this is a Tenant::layer_removal_cs.try_lock().expect())
5 Remove timeline directory and all its files.
6 drop the guard
7 Remove the timeline from the timelines map
8 drop timelines lock (implicit drop)

Couple of odd things about that:

- I think (3) is supposed to lock out any more timeline ops and and
  (4) is supposed to drain any existing timeline ops, namely, GC & Compactions.
  So that (5) can then remove the layer files, knowing that nobody
  references the corresponding Layer structs anymore, except us.
- If that assumption is correct, I think there are several flaws with
  that approach:
  1. The layer_remvoal_guard() is a try_lock() that will panic if we
     can't get it. The caller only shuts down timeline tasks before
     calling us, but GC & Compaction is a per-tenant loop, so, I think
     we could reach the case where the try_lock() fails. In that case,
     timeline_delete() panicks.
  2. The `layer_removal_guard()` only waits for concurrent GC & Compaction.
     Once we have it, we delete the layer files (5).
     We need to be 100% sure that nobody else will be accessing these layer
     files anymore.
     The caller has killed WAL receiver, that's good.
     But the timeline is still reachable via PS HTTP API.
     So, if there's any HTTP op that started before (3), we haven't
     yet waited for it to finish.

This PR doesn't address those problems.
And possible, it makes things worse because layer_removal_cs needs to
become a tokio::sync::Mutex, so that the next commit's
can hold it across the remote_client.wait_completion().await points
that it adds.

So, the solution in delete_timeline() is to drop the timelines.lock()
after setting the `TimelineState::Paused`.
Then do compaction while holding `layer_removal_cs()`.
Then re-aquire timelines.lock() and delete the timeline.
```

Commit 2:

```
 add remote_storage.wait_completion() calls to compaction() and gc() 
```

### DIFF
--- a/pageserver/src/tenant.rs
+++ b/pageserver/src/tenant.rs
@@ -1191,7 +1191,7 @@ impl Tenant {
     /// This function is periodically called by compactor task.
     /// Also it can be explicitly requested per timeline through page server
     /// api's 'compact' command.
-    pub fn compaction_iteration(&self) -> anyhow::Result<()> {
+    pub async fn compaction_iteration(&self) -> anyhow::Result<()> {
         anyhow::ensure!(
             self.is_active(),
             "Cannot run compaction iteration on inactive tenant"
@@ -1201,16 +1201,21 @@ impl Tenant {
         // while holding the lock. Then drop the lock and actually perform the
         // compactions.  We don't want to block everything else while the
         // compaction runs.
-        let timelines = self.timelines.lock().unwrap();
-        let timelines_to_compact = timelines
-            .iter()
-            .map(|(timeline_id, timeline)| (*timeline_id, timeline.clone()))
-            .collect::<Vec<_>>();
-        drop(timelines);
+        let timelines_to_compact = {
+            let timelines = self.timelines.lock().unwrap();
+            let timelines_to_compact = timelines
+                .iter()
+                .map(|(timeline_id, timeline)| (*timeline_id, timeline.clone()))
+                .collect::<Vec<_>>();
+            drop(timelines);
+            timelines_to_compact
+        };
 
         for (timeline_id, timeline) in &timelines_to_compact {
-            let _entered = info_span!("compact_timeline", timeline = %timeline_id).entered();
-            timeline.compact()?;
+            timeline
+                .compact()
+                .instrument(info_span!("compact_timeline", timeline = %timeline_id))
+                .await?;
         }
 
         Ok(())
@@ -1244,42 +1249,87 @@ impl Tenant {
     }
 
     /// Removes timeline-related in-memory data
-    pub fn delete_timeline(&self, timeline_id: TimelineId) -> anyhow::Result<()> {
-        // in order to be retriable detach needs to be idempotent
-        // (or at least to a point that each time the detach is called it can make progress)
-        let mut timelines = self.timelines.lock().unwrap();
+    pub async fn delete_timeline(&self, timeline_id: TimelineId) -> anyhow::Result<()> {
+        // Transition the timeline into TimelineState::Paused.
+        // This should prevent new operations from starting.
+        let timeline = {
+            let mut timelines = self.timelines.lock().unwrap();
 
-        // Ensure that there are no child timelines **attached to that pageserver**,
-        // because detach removes files, which will break child branches
-        let children_exist = timelines
-            .iter()
-            .any(|(_, entry)| entry.get_ancestor_timeline_id() == Some(timeline_id));
+            // Ensure that there are no child timelines **attached to that pageserver**,
+            // because detach removes files, which will break child branches
+            let children_exist = timelines
+                .iter()
+                .any(|(_, entry)| entry.get_ancestor_timeline_id() == Some(timeline_id));
 
-        anyhow::ensure!(
-            !children_exist,
-            "Cannot delete timeline which has child timelines"
-        );
-        let timeline_entry = match timelines.entry(timeline_id) {
-            Entry::Occupied(e) => e,
-            Entry::Vacant(_) => bail!("timeline not found"),
+            anyhow::ensure!(
+                !children_exist,
+                "Cannot delete timeline which has child timelines"
+            );
+            let timeline_entry = match timelines.entry(timeline_id) {
+                Entry::Occupied(e) => e,
+                Entry::Vacant(_) => bail!("timeline not found"),
+            };
+
+            let timeline = Arc::clone(timeline_entry.get());
+            timeline.set_state(TimelineState::Paused);
+
+            drop(timelines);
+            timeline
         };
 
-        let timeline = timeline_entry.get();
-        timeline.set_state(TimelineState::Paused);
+        info!("waiting for layer_removal_cs.lock()");
+        // No timeout here, GC & Compaction should be responsive to the `TimelineState::Paused` change.
+        let layer_removal_guard = timeline.layer_removal_cs.lock().await;
+        info!("got layer_removal_cs.lock(), deleting layer files");
 
-        let layer_removal_guard = timeline.layer_removal_guard()?;
+        // NB: storage_sync upload tasks that reference these layers have been cancelled
+        //     by the caller.
 
         let local_timeline_directory = self.conf.timeline_path(&timeline_id, &self.tenant_id);
+        // XXX make this atomic so that, if we crash-mid-way, the timeline won't be picked up
+        // with some layers missing.
         std::fs::remove_dir_all(&local_timeline_directory).with_context(|| {
             format!(
                 "Failed to remove local timeline directory '{}'",
                 local_timeline_directory.display()
             )
         })?;
-        info!("detach removed files");
+        info!("finished deleting layer files, releasing layer_removal_cs.lock()");
 
         drop(layer_removal_guard);
-        timeline_entry.remove();
+
+        // Remove the timeline from the map.
+        let mut timelines = self.timelines.lock().unwrap();
+        let children_exist = timelines
+            .iter()
+            .any(|(_, entry)| entry.get_ancestor_timeline_id() == Some(timeline_id));
+        // XXX this can happen because `branch_timeline` doesn't check `TimelineState::Paused`.
+        // We already deleted the layer files, so it's probably best to panic.
+        // (Ideally, above remove_dir_all is atomic so we don't see this timeline after a restart)
+        if children_exist {
+            panic!("Timeline grew children while we removed layer files");
+        }
+        let removed_timeline = timelines.remove(&timeline_id);
+        if removed_timeline.is_none() {
+            // This can legitimately happen if there's a concurrent call to this function.
+            //   T1                                             T2
+            //   lock
+            //   unlock
+            //                                                  lock
+            //                                                  unlock
+            //                                                  remove files
+            //                                                  lock
+            //                                                  remove from map
+            //                                                  unlock
+            //                                                  return
+            //   remove files
+            //   lock
+            //   remove from map observes empty map
+            //   unlock
+            //   return
+            debug!("concurrent call to this function won the race");
+        }
+        drop(timelines);
 
         Ok(())
     }
@@ -1767,7 +1817,7 @@ impl Tenant {
                 );
             }
 
-            let result = timeline.gc()?;
+            let result = timeline.gc().await?;
             totals += result;
         }
 
@@ -3057,7 +3107,7 @@ mod tests {
         drop(writer);
 
         tline.checkpoint(CheckpointConfig::Forced).await?;
-        tline.compact()?;
+        tline.compact().await?;
 
         let writer = tline.writer();
         writer.put(*TEST_KEY, Lsn(0x20), &Value::Image(TEST_IMG("foo at 0x20")))?;
@@ -3065,7 +3115,7 @@ mod tests {
         drop(writer);
 
         tline.checkpoint(CheckpointConfig::Forced).await?;
-        tline.compact()?;
+        tline.compact().await?;
 
         let writer = tline.writer();
         writer.put(*TEST_KEY, Lsn(0x30), &Value::Image(TEST_IMG("foo at 0x30")))?;
@@ -3073,7 +3123,7 @@ mod tests {
         drop(writer);
 
         tline.checkpoint(CheckpointConfig::Forced).await?;
-        tline.compact()?;
+        tline.compact().await?;
 
         let writer = tline.writer();
         writer.put(*TEST_KEY, Lsn(0x40), &Value::Image(TEST_IMG("foo at 0x40")))?;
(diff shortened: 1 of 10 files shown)


pageserver/src/tenant.rs
```diff
@@ -1244,32 +1249,50 @@ impl Tenant {
...
+        let timeline = {
+            let mut timelines = self.timelines.lock().unwrap();
 
-        // Ensure that there are no child timelines **attached to that pageserver**,
-        // because detach removes files, which will break child branches
-        let children_exist = timelines
-            .iter()
-            .any(|(_, entry)| entry.get_ancestor_timeline_id() == Some(timeline_id));
+            // Ensure that there are no child timelines **attached to that pageserver**,
+            // because detach removes files, which will break child branches
```
hlinnaka, line 1258: What happens if this is called concurrently from two threads? Should we check here that the timeline is not already in Paused state?

hlinnaka: Oh I see there's some code to deal with that race later, when the Timeline is removed from the map. I think you can get errors from `std::fs::remove_dir_all`, however, if two threads try to remove the files at the same time.

pageserver/src/tenant.rs
```diff
@@ -1244,32 +1249,50 @@ impl Tenant {
...
+        //    and no other thread will act on it every again.
+        let layer_removal_guard = timeline.layer_removal_guard().await?;
 
-        let layer_removal_guard = timeline.layer_removal_guard()?;
+        // NB: storage_sync upload tasks that reference these layers have been cancelled
+        //     by the caller.
 
         let local_timeline_directory = self.conf.timeline_path(&timeline_id, &self.tenant_id);
+        // XXX make this atomic so that, if we crash-mid-way, the timeline won't be picked up
+        // with some layers missing.
```
hlinnaka, line 1295: Could e.g rename the directory it to have `.temp` suffix first. (Doesn't need to be part of this PR as it's a pre-existing problem)

hlinnaka (approved):
Lots of issues remain as you're saying, but this is good enough to get PR #2785 committed.

pageserver/src/tenant.rs
problame: Yeah, one will fail with an error and another will succeed.
Since we don't (and can't!) rollback the state from `Paused` to `Active`, I think that's acceptable for now.

pageserver/src/tenant.rs
problame: Yeah, I though about renaming it to the uninit suffix, so that on next pageserver restart, we clean up.

problame:
The `test_remote_storage.py` tests are failing now. Kinda makes sense.

```
test_remote_storage_upload_queue_retries
test_timeline_deletion_with_files_stuck_in_upload_queue
```

Need to investigate, but most likely, they do some checkpoints/gc/layer deletions while we artificially fail the upload op.
Not sure how straightforward it is to fix the tests.

pageserver/src/tenant.rs
hlinnaka: I'm worried that they would both fail. Imagine that you have a directory with two files, "foo" and "bar". One thread removes "foo" first, then "bar", and the other one the other way round. Both could succeed on removing the first file, and get a "file not found" error on the second one. I haven't looked at how `std::fs::remove_dir_all` is implemented, so I'm not sure what would happen in that case, but I also don't see any guarantees in the docs for that either.

pageserver/src/tenant/timeline.rs
```diff
@@ -2223,6 +2228,38 @@ impl Timeline {
...
+            new_layers,
+            deltas_to_compact,
+        } = self.compact_level0_phase1(target_file_size).await?;
+
+        // Before deleting any layers, we need to wait for their upload ops to finish.
+        // See storage_sync module level comment on consistency.
+        // Do it here because:
+        // 1. We don't want to hold self.layers.write() while waiting
+        // 2. We know that the new layers aren't going to be deleted, so, we can
+        //    avoid waiting for them.
```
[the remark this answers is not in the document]
hlinnaka: Hmm. There are two operations that delete files: compaction and GC. As long as we have something that prevents GC from running at the same time, it's true.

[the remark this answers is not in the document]
hlinnaka: The comment implies that they're not going to be removed *because* they're new though. Not because we're preventing GC from running at the same time. I'm not sure about that.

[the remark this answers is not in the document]
problame: > Hmm. There are two operations that delete files: compaction and GC. As long as we have something that prevents GC from running at the same time, it's true.

This holds due to `layer_removal_cs`.

> The comment implies that they're not going to be removed because they're new though. Not because we're preventing GC from running at the same time. I'm not sure about that.

Me neither. The assert never triggered, though.

**Pro:** It's a clear latency win over doing wait_completion() after scheduling the level 1 layers for upload. Compaction latency, that is.
**Con:** Then again, does compaction latency matter at all? If we're producing compacted layers faster than upload rate, we will fall behind in the next compaction iteration.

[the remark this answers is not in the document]
problame: Looking up the call stack:
1. We measure `Timeline::compact_level0()` latency, so apparently it matters to someone (`compact_time_histo`).
1. Timeline::compact() is invoked sequentially, in a loop.
1.  `compaction_loop` itself does a fixed amount of `sleep`, i.e., it doesn't subtract the time it took to do the compaction.

Not sure what to make of that.

problame:
In the last run, ` test_get_tenant_size_with_multiple_branches` in release build failed due to not-allowed error in log:

```
2022-11-28T16:54:59.969706Z ERROR gc_loop{tenant_id=6062978df26c2e7797747c0e71737d41}: Gc failed, retrying in 2s: timeline is paused: Paused
```

It's a legitimate error because the test races with 'delete_timeline'.

This is a consequence of no longer failing `delete_timeline` if GC is still running.

To fix this, we need to remove
```
cannot lock compaction critical section
```
from `pageserver.allowed_errors` and add
```
Gc failed, retrying in.*timeline is paused: Paused
```

Alternatively, we could switch back to failing delete_timeline() if GC / Compaction is running.
But seems like lack of priority of `delete_timeline` over GC / Compaction has been a problem in the past, e.g., https://github.com/neondatabase/neon/issues/2442

problame:
Added the following three errors to the global allowed_errors list and linked the follow-up issue 

```
 # Tenant::delete_timeline() can cause any of the three following errors.
 # FIXME: we shouldn't be considering it an error: https://github.com/neondatabase/neon/issues/2946
 "could not flush frozen layer.*queue is in state Stopped",  # when schedule layer upload fails because queued got closed before compaction got killed
 ".*wait for layer upload ops to complete.*",  # .*Caused by:.*wait_completion aborted because upload queue was stopped
 "Gc failed, retrying in.*timeline is paused: Paused",  # When gc/compaction check timeline state after acquiring layer_removal_cs
```

problame:
Hmm, this time the debug build has test failure

```
test_tenants_attached_after_download
```

failed with

```
AssertionError: assert 25927680 == 25960448
  +25927680
  -25960448
```

The difference between that is exactly `32k`.

----


And we got a release build test failure due to not-allowed error

```
test_gc_cutoff
```

```
2022-11-28T18:02:26.938227Z  INFO handle_pagerequests{tenant_id=8faae982b2428b4bdf8cd2c1b712772e timeline_id=ad8b3cec55b86cfd8f07977ab5ad2a7f}:handle_get_page_at_lsn_request{rel=1663/12972/2663 blkno=2 req_lsn=0/CEE9938}:launch{tenant_id=8faae982b2428b4bdf8cd2c1b712772e pg_version=14}: old temporary datadir /tmp/test_output/test_gc_cutoff/repo/tenants/8faae982b2428b4bdf8cd2c1b712772e/wal-redo-datadir.___temp exists, removing
2022-11-28T18:02:26.943210Z ERROR handle_pagerequests{tenant_id=8faae982b2428b4bdf8cd2c1b712772e timeline_id=ad8b3cec55b86cfd8f07977ab5ad2a7f}: error reading relation or page version: Failed to reconstruct a page image:

Caused by:
    Directory not empty (os error 39)

Stack backtrace:
   0: pageserver::tenant::timeline::Timeline::get::{{closure}}
             at /__w/neon/neon/pageserver/src/tenant/timeline.rs:394:42
             at /github/home/.cargo/registry/src/github.com-1ecc6299db9ec823/prometheus-0.13.2/src/histogram.rs:716:19
   1: pageserver::tenant::timeline::Timeline::get
             at /__w/neon/neon/pageserver/src/tenant/timeline.rs:392:9
   2: pageserver::pgdatadir_mapping::<impl pageserver::tenant::timeline::Timeline>::get_rel_page_at_lsn
             at /__w/neon/neon/pageserver/src/pgdatadir_mapping.rs:104:9
   3: pageserver::page_service::PageServerHandler::handle_get_page_at_lsn_request::{{closure}}::{{closure}}
             at /__w/neon/neon/pageserver/src/page_service.rs:610:20
             at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/core/src/future/mod.rs:91:19
   4: <tracing::instrument::Instrumented<T> as core::future::future::Future>::poll
             at /github/home/.cargo/registry/src/github.com-1ecc6299db9ec823/tracing-0.1.37/src/instrument.rs:272:9
   5: pageserver::page_service::PageServerHandler::handle_get_page_at_lsn_request::{{closure}}
             at /__w/neon/neon/pageserver/src/page_service.rs:588:5
             at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/core/src/future/mod.rs:91:19
      pageserver::page_service::PageServerHandler::handle_pagerequests::{{closure}}::{{closure}}
             at /__w/neon/neon/pageserver/src/page_service.rs:344:73
   6: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/core/src/future/mod.rs:91:19
      <tracing::instrument::Instrumented<T> as core::future::future::Future>::poll
             at /github/home/.cargo/registry/src/github.com-1ecc6299db9ec823/tracing-0.1.37/src/instrument.rs:272:9
   7: pageserver::page_service::PageServerHandler::handle_pagerequests::{{closure}}
             at /__w/neon/neon/pageserver/src/page_service.rs:271:5
             at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/core/src/future/mod.rs:91:19
      <pageserver::page_service::PageServerHandler as utils::postgres_backend_async::Handler>::process_query::{{closure}}
```

---

I have no clue about the former, investigating the latter now.

problame:
My best hypothesis for the root cause of test_gc_cutoff failure is that we use the `exit` failpoint action, and then restart the pageserver while the old walredo process is still running.
That causes the `fs::remove_dir_all()` call during walredo process `launch()` to fail.
We don't add enough `.context()` on the way up, so, the cause chain in the log leaves a lot to be desired.
Not our problem in this PR, will kick a re-run.
cc @koivunej @knizhnik  is this a know issue with walredo? If not please convert this analysis into a follow-up issue.

hlinnaka:
> My best hypothesis for the root cause of test_gc_cutoff failure is that we use the `exit` failpoint action, and then restart the pageserver while the old walredo process is still running.
> That causes the `fs::remove_dir_all()` call during walredo process `launch()` to fail.
> We don't add enough `.context()` on the way up, so, the cause chain in the log leaves a lot to be desired.
> Not our problem in this PR, will kick a re-run.
> cc @koivunej @knizhnik is this a know issue with walredo? If not please convert this analysis into a follow-up issue.

Yeah. See https://github.com/neondatabase/neon/issues/2400

hlinnaka:
> > My best hypothesis for the root cause of test_gc_cutoff failure is that we use the `exit` failpoint action, and then restart the pageserver while the old walredo process is still running.
> > That causes the `fs::remove_dir_all()` call during walredo process `launch()` to fail.
> > We don't add enough `.context()` on the way up, so, the cause chain in the log leaves a lot to be desired.
> > Not our problem in this PR, will kick a re-run.
> > cc @koivunej @knizhnik is this a know issue with walredo? If not please convert this analysis into a follow-up issue.
> 
> Yeah. See #2400

Funny, we just talked #2400 with @alexstanovoy earlier today. We were not sure if it's still a problem, but apparently it is :-).

problame:
@hlinnaka regarding

```
AssertionError: assert 25927680 == 25960448
  +25927680
  -25960448
```

Your change e51f2be3d0d2172421016ece22332c079328e851 added that assertion.

What fails is

```
    assert detail_before["current_physical_size"] == detail_after["current_physical_size"]
```

The `after` value is larger.

So, something is generating more layer files (but only 32k in size?) after the checkpoint loop but before we shut down the page server:

https://github.com/neondatabase/neon/blob/2dad3a64455185dffe23751e3f9864c3c18f47de/test_runner/regress/test_tenants_with_remote_storage.py#L140-L172

pageserver/src/tenant/timeline.rs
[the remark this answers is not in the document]
hlinnaka: It's not latency-critical as such, but if compaction is too slow, it might not keep up with the speed at which new WAL is ingested.

The code looks good, let's just change the comment to "Do it here because we don't want to hold self.layers.write() while waiting", leaving out the reason (2).

hlinnaka:
> ```
> AssertionError: assert 25927680 == 25960448
>   +25927680
>   -25960448
> ```
> 
> Your change [e51f2be](https://github.com/neondatabase/neon/commit/e51f2be3d0d2172421016ece22332c079328e851) added that assertion.
> 
> What fails is
> 
> ```
>     assert detail_before["current_physical_size"] == detail_after["current_physical_size"]
> ```
> 
> The `after` value is larger.
> 
> So, something is generating more layer files (but only 32k in size?) after the checkpoint loop but before we shut down the page server:
> 
> [](https://github.com/neondatabase/neon/blob/2dad3a64455185dffe23751e3f9864c3c18f47de/test_runner/regress/test_tenants_with_remote_storage.py#L140-L172)

For the record, @problame addressed this in commit https://github.com/neondatabase/neon/pull/2942/commits/718a58484489cc63bfa7026f23e438597a7f00ce.

PR #2944, opened 2022-11-28
Put .proto compilation result to $OUT_DIR/

arssher:
Sometimes CI build fails with

error: couldn't read storage_broker/src/../proto/storage_broker.rs: No such file or directory (os error 2)
  --> storage_broker/src/lib.rs:14:5
   |
14 |     include!("../proto/storage_broker.rs");
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

The root cause is not clear, but it looks like interference with cachepot. Per cargo docs, build scripts shouldn't output to anywhere but OUT_DIR; let's follow this and see if it helps.


petuhovskiy (approved):
Not sure what's going on and how this helps to fix the error in CI, but I see nothing wrong here

PR #2974, opened 2022-11-30
Resend prefetch request in case of pageserver restart

knizhnik:
refer #2819

### DIFF
--- a/pgxn/neon/pagestore_smgr.c
+++ b/pgxn/neon/pagestore_smgr.c
@@ -1782,6 +1782,13 @@ neon_read_at_lsn(RelFileNode rnode, ForkNumber forkNum, BlockNumber blkno,
 		}
 		else
 		{
+			/*
+			 * Empty our reference to the prefetch buffer's hash entry.
+			 * When we wait for prefetches, the entry reference is invalidated by 
+			 * potential updates to the hash, and when we reconnect to the 
+			 * pageserver the prefetch we're waiting for may be dropped,
+			 * in which case we need to retry and take the branch above.
+			 */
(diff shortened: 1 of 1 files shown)


pgxn/neon/pagestore_smgr.c
```diff
@@ -1780,6 +1780,10 @@ neon_read_at_lsn(RelFileNode rnode, ForkNumber forkNum, BlockNumber blkno,
 												  &request_lsn);
 			slot = GetPrfSlot(ring_index);
 		}
+		else
+		{
+			entry = NULL;
```
bojanserafimov, line 1785: This change itself might be fine but the surrounding code is hard to understand so I can't tell if it might break something. For example, can we get stuck in infinite loop now?

IMO the bigger question is "what is this do block trying to accomplish"? Can we start by adding a comment before 1773?

knizhnik: It is actually @MMeent  code. May be he can better explain what he does.
I only have added this loop.
So prefetch is working now in this way: when we get request for some page P, we first lookup if there are registered prefetch request  for P. If so we wait for it completion. If not, then we register new prefetch request for P (+ some more pages)  and wait completion of this request.
But before we can get response to our prefetch request we need to consume responses to all previous prefetch requests.
The source of the problem in this ticket is that after completion of this select there are several pending prefetch requests which are not consumed. If pageserver restarts, connection to page server is broken. We need to skip all this pending prefetch requests, reestablish connection and register request for page P once again.

If you and @MMeent find my explanation above correct and clear, I can add them as a comment to this fragment.

PR #2977, opened 2022-11-30
Add commands to unload and load the tenant in memory

SomeoneToIgnore:
Closes https://github.com/neondatabase/neon/issues/2537

Follow-up of https://github.com/neondatabase/neon/pull/2950
With the new model that prevents attaching without the remote storage, it has started to be even more odd to add attach-with-files functionality (in addition to the issues raised previously).

I have fixed a few existing style nits and added two separate commands:
* `POST {tenant_id}/ignore` that places a mark file to skip such tenant on every start and removes it from memory
* `POST {tenant_id}/schedule_load` that tries to load a tenant from local FS similar to what pageserver does now on startup, but without odd directory removal

The latter command execution does not wait for the tenant to load and I don't think we should overcompilcate this: for manual invocations, it's ok to check for tenant status later after the load.


pageserver/src/tenant_mgr.rs
```diff
@@ -77,27 +78,34 @@ pub fn init_tenant_mgr(
...
+                                tenant_dir_path.display()
+                            )
+                        }
+                        continue;
+                    }
+
+                    let tenant_ignore_mark_file = tenant_dir_path.join(IGNORED_TENANT_FILE_NAME);
+                    if tenant_ignore_mark_file.exists() {
+                        info!("Found an ignore mark file {tenant_ignore_mark_file:?}, skipping the tenant");
+                        continue;
```
problame, line 99: What if we call attach for that tenant?

problame: I suppose you get tenant that is `Attaching` for a brief period of time, then transitions into `Broken` state after error

```
calling Tenant::attach with a tenant directory that doesn't have the attaching marker file:\ntenant_dir: {}\nmarker_file: {}"
```

pageserver/src/http/openapi_spec.yml
```diff
@@ -354,6 +357,92 @@ paths:
...
+    post:
+      description: |
+        Attempts to load a tenant from the local disk and synchronise it with the remote storage (if enabled),
+        repeating pageserver's restart logic for tenant load.
+        If the tenant was ignored before, removes the ignore mark and continues with load scheduling.
+
+        Errors if the tenant is absent on disk, already present in memory or fails to schedule its load.
+        Scheduling a load does not mean that the tenant would load successfully, check tenant status to ensure load correctness.
+      responses:
+        "200":
```
problame, line 419: ACCEPTED is 202

pageserver/src/http/openapi_spec.yml
```diff
@@ -354,6 +357,92 @@ paths:
...
+    parameters:
+      - name: tenant_id
+        in: path
+        required: true
+        schema:
+          type: string
+          format: hex
+    post:
+      description: |
+        Attempts to load a tenant from the local disk and synchronise it with the remote storage (if enabled),
```
problame, line 412: 1. Let's call this just `/load`, not `/schedule_load`. The `202 ACCEPTED` response is indicative enough that this isn't a blocking operation.

2. Make it clear in the first sentence in the description that this only initiates the operation.

pageserver/src/http/openapi_spec.yml
```diff
@@ -354,6 +357,92 @@ paths:
...
+        required: true
+        schema:
+          type: string
+          format: hex
+    post:
+      description: |
+        Remove tenant data (including all corresponding timelines) from pageserver's memory.
+        Files on local disk and remote storage are not affected.
+
+        Future pageserver restarts won't load the data back until `attach` is called on such tenant.
```
problame, line 373: Not `attach`, but `schedule_load`

test_runner/regress/test_tenant_detach.py
```diff
@@ -167,3 +167,196 @@ def test_detach_while_attaching(
...
+    tenants_before_ignore = [tenant["id"] for tenant in pageserver_http.tenant_list()]
+    tenants_before_ignore.sort()
+    timelines_before_ignore = [
+        timeline["timeline_id"]
+        for timeline in pageserver_http.timeline_list(tenant_id=ignored_tenant_id)
+    ]
+    files_before_ignore = [tenant_path for tenant_path in tenant_dir.glob("**/*")]
+
+    # ignore the tenant and veirfy it's not present in pageserver replies, with its files still on disk
+    pageserver_http.tenant_ignore(ignored_tenant_id)
```
problame, line 194: We should check that the tenant is no longer in the pageserver's memory after this returns.

tenant status and/or or timline list HTTP calls would be good candidates.

test_runner/regress/test_tenant_detach.py
```diff
@@ -167,3 +167,196 @@ def test_detach_while_attaching(
...
+
+    env.pageserver.allowed_errors.append(".*could not load tenant .*?: failed to load metadata.*")
+
+    # now, load it from the local files and expect it to be broken due to inability to load tenant files into memory
+    pageserver_http.schedule_tenant_load(tenant_id=tenant_id)
+    wait_until_tenant_status(pageserver_http, tenant_id, "Broken", 5)
+
+
+# Does not use `wait_until` for debugging purposes
+def wait_until_tenant_status(
```
problame, line 345: This seems generally useful. Do we have a shared space for such functions?

test_runner/regress/test_tenant_detach.py
```diff
@@ -167,3 +167,196 @@ def test_detach_while_attaching(
...
+
+    timelines_after_ignore = [
+        timeline["timeline_id"]
+        for timeline in pageserver_http.timeline_list(tenant_id=ignored_tenant_id)
+    ]
+    assert timelines_before_ignore == timelines_after_ignore, "Should have all timelines back"
+
+
+@pytest.mark.parametrize("remote_storage_kind", [RemoteStorageKind.LOCAL_FS])
+def test_ignored_tenant_download_missing_layers(
```
problame, line 237: Test summary comment missing.

```
Ignore tenant.
Delete local layers files.
Load tenant again.
Ensure that load re-downloads layer files.
```

problame: What's the purpose / real-world use case of this test?

It's _not_ the "attach failed because lack of download retries" case, is it?

problame (commented):
Overall mostly minor nit picks.

But we need to get the naming in order.

With this PR, we'll have the following verbs on a tenant:

- load
- attach
- detach
- ignore
- schedule_load

And: why do we need a separate `schedule_load` at all? I think from API user's perspective, `schedule_load` should be the same as `attach`. Would you agree?

test_runner/regress/test_tenant_detach.py
SomeoneToIgnore: Would the existing 

```python
tenants_after_ignore = [tenant["id"] for tenant in pageserver_http.tenant_list()]
    assert ignored_tenant_id not in tenants_after_ignore, "Ignored tenant should be missing"
    assert len(tenants_after_ignore) + 1 == len(
        tenants_before_ignore
    ), "Only ignored tenant should be missing"
```

be enough? I can sure call some method with the ignored tenant_id, but this way I'm also checking other tenants have not disappeared (there should be a default one in this test)

test_runner/regress/test_tenant_detach.py
SomeoneToIgnore: We have a generic https://github.com/neondatabase/neon/blob/main/test_runner/fixtures/utils.py can move it there, but not sure if needed, since this is the only file we need such things so far?

There's `wait_until` from that utils module, but it somehow does not evaluate `log.error` and similar calls inside it and generally is displayed bad in our tests.

test_runner/regress/test_tenant_detach.py
SomeoneToIgnore: I'm trying to model various bad situations where we need to apply `ignore` command fast, then somehow fix the situation and get that tenant back.

Our general "fix the situation" attempts seem include reusing already stored files on S3, since we consider that storage as durable.

So, the test ensures that we can actually use the remote storage properly when some layers were removed locally during the ignored state.

Maybe, those layers got corrupt by the new logic and we've decided to revert?
Hard to find 100% real use case here, but I definitely want to ensure we can work with the remote storage during `load`, similarly to how pageserver reload does.

pageserver/src/tenant_mgr.rs
SomeoneToIgnore: Good point, I should have added a negative test on attaches/loads, since I've even missed a few points here.

Reworked now to ensure both code paths treat first local files: with this approach, `attach` disallows local files, which simplifies a few things.
Still, not fond how both startup and load logic have to treat all these files differently.

SomeoneToIgnore:
> we'll have the following verbs

We won't have `schedule_load` though, but agree that the naming could be improved.
So far, we have pairs of an operation and its undo version: `attach + detach`, `ignore + load`.

I'm also not fond of the amount of markers we keep for tenants, not sure I want to write logic for a few more and their special combinations. Maybe, in the future we should apply @hlinnaka idea and move away certain tenant directories. 

> And: why do we need a separate schedule_load at all? I think from API user's perspective, schedule_load should be the same as attach. Would you agree?

I would not disagree:
* at first (https://github.com/neondatabase/neon/pull/2950) I have proposed exactly this
* I still think, there's some synergy in there

Yet I've thought about your idea in https://github.com/neondatabase/neon/pull/2950#issuecomment-1330803833

> The allow-local-files=true seems pretty unsafe to me.
Why is that?
The pageserver has no means to tell whether the tenant's local filesystem state is the same or ahead of the remote state.
Remote state here means all timelines of the tenant.
We cannot allow the pageserver to operate on a tenant if we aren't sure about this.
For example, if the remote has a timeline that the local filesystem state doesn't have, for whatever reason, then local GC will not respect that timeline's branch point LSN, which will eventually lead to data loss.

And one thing could be noted for sure: we do some extra things in `attach`, in particular, download **every** timeline for the tenant we attach. And detach, as it's reversible operation, also removes all such files.
While the new command set's goal is maintenance: add some way to quickly silence the misbehaving tenant, but preserve its files for future analysis.
Then, after some manipulations, we need a reverse operation too. 
Before, we could shut down entire pageserver for that purpose, and the reverse operation was obvious: start it back.

I don't think pageserver's startup operation on a single tenant scale is anyhow close to `attach` of that tenant, but it's exactly what we need IMO, as it's simple enough and was good enough already.

problame:
> And one thing could be noted for sure: we do some extra things in attach, in particular, download every timeline for the tenant we attach. 

And that's the key point. In `/schedule_load`, we don't check again whether we have all local timelines.
I think it boils down to the question whether we want to allow the following sequence of events:

```
/attach tenant
wait for attach to finish
ls $tenant_dir
  $tenant_dir/timelineA
  $tenant_dir/timelineB
/ignore tenant
rm -rf $tenant_dir/timelineB
/schedule_load tenant
```

With the current implementation, the `/schedule_load` will succeed and the tenant will become `Active`, even though it's missing one timeline.

Once GC starts running, it will not consider `timelineB`'s branch point.
And so, it could decide to delete layers that `timelineB` depends on.
Eventually, these deletes will reach the remote storage.
Then, we have data loss.

**So, IMO, the invariant should be that a tenant cannot become `Active` until we have checked that we have the most recent remote state of _all_ timelines.**
**If you agree with that, here's my proposal on how to get there:**
1. Pageserver-local on-disk state must have indication that a tenant is associated with a remote storage
  - Like `IndexPart`, but for the list of the tenant's timelines. Present both locally and remotely.
    - We could also use this instead of directory listing to discover the set of present timelines, cc https://github.com/neondatabase/neon/issues/2933
2. A tenant that is marked as associated with a remote storage cannot become `Active` until pageserver has fetched the list of all remote timelines & their metadata
  - this is implicitly true for `Attach` but needs to become true for `Load` as well
  - until on-demand download, this also means we download all the timlines' layer files


**How can we make progress on this PR without letting it scope creep to the above?**

In my eyes, by only allowing  `/schedule_load` on tenants that have the attaching marker file.
Because, as pointed out above, the attach code does the right thing already (get the list of remote-timelines).

SomeoneToIgnore:
This is a nice idea, I have thought of such possibility too, yet discarded it due to possibility to induce similar things now with pageserver restart: stop it, rm a branch, start it, trigger a GC.

I agree that it's bad in general, yet, again, do not agree fully with the fact, that the timeline is not Active before knowing the whole remote storage hierarchy: we still can serve get_page@LSN requests for branches loaded locally and might be able to accept WAL on them (soon after the sync, at least). 
We cannot run GC indeed, but I'm not sure if that's good to disable other operations on that timeline.

So, I still would propose to keep this PR as is, simple, and would solve the remote hierarchy sync separately as a follow-up, since it requires changing the PS startup logic too.

pageserver/src/tenant_mgr.rs
```diff
@@ -321,12 +310,88 @@ pub async fn detach_tenant(
...
+    );
+    anyhow::ensure!(
+        !tenant_path.is_empty_dir().with_context(|| {
+            format!("Failed to check whether {tenant_path:?} is an empty dir")
+        })?,
+        "Cannot load tenant from empty directory {tenant_path:?}"
+    );
+    let tenant_attach_mark = conf.tenant_attaching_mark_file_path(&tenant_id);
+    anyhow::ensure!(
+        !tenant_attach_mark.exists(),
```
problame, line 344: This will break the use case `/ignore` & `/spawn_load` a tenant that has encountered an error while it was attaching.

In general, I think most of the checks in this function before `run_if_no_tenant_in_memory` are kind of redundant. `load_local_tenant` and the `spawn_load` / `spawn_attach` functions already do them.

pageserver/src/tenant_mgr.rs
```diff
@@ -321,12 +310,88 @@ pub async fn detach_tenant(
...
+
+    run_if_no_tenant_in_memory(tenant_id, |vacant_entry| {
+        let tenant_ignore_mark = conf.tenant_ignore_mark_file_path(tenant_id);
+        if tenant_ignore_mark.exists() {
+            std::fs::remove_file(&tenant_ignore_mark)
+                .with_context(|| format!("Failed to remove tenant ignore mark {tenant_ignore_mark:?} during tenant loading"))?;
+        }
+
+        let new_tenant = load_local_tenant(conf, &tenant_path, remote_storage)
+            .with_context(|| format!("Failed to load tenant from path {tenant_path:?}"))?;
```
problame, line 356: Error message should make it clear that we only initiated the load/attach operation.
Admittedly, the `load_local_tenant` function name is bad, since it can spawn either a load or an attach operation internally.

Maybe rename to `spawn_tenant`? Or `start_tenant`? That's generic over load vs. attach

problame (changes requested):
At a minimum, let's resolve my remark on the check for presence of the marker file inside `schedule_tenant_load`.

pageserver/src/tenant_mgr.rs
SomeoneToIgnore: Right

> most of the checks in this function before `run_if_no_tenant_in_memory` are kind of redundant

Start to agree more when I've thought that it's better to put them after we take the lock: more chances to prevent a race on concurrent `attach`/`load` requests scheduled.

So, now we check the in-memory state, grab the lock, and then, inside `load_local_tenant` do more checks to ensure that this sort of public method does not load some ignored tenant by accident.
This way, a few FS checks are duplicated still, but that seems not much.

I've removed the attached file check and added a test for such scenario.
Looks like another "fill the DB and upload layers" common function slowly emerges in Python tests, but I've not moved it into the lib code yet.

pageserver/src/tenant_mgr.rs
SomeoneToIgnore: Names are somewhat odd indeed.
I've named it as `schedule_local_tenant_processing` which does not shine, but at least have some verb, just "spawning" or "starting" does not sound ok to me.

SomeoneToIgnore:
Done.

test_runner/regress/test_tenant_detach.py
problame: Yeah, it's enough. Not sure whether it was always there and I missed it, or whether it was added since I added this comment.

Either way, it's good as it is.

problame (changes requested):
Found one more race condition. Discussed on Slack.

Copy-paste follows:

```
T1: recv /ignore
T1: write-lock tenants lock
T1: write ignore file
T1: remove tenant from tenants hashmap
T1: unlock tenants-lock
T2: recv /attach
T2: run_if_no_tenant_in_memory observes no tenant in hashmap
T2: create new tenant object in Attaching state, spawn attach task
T2: start running code inside Tenant::attach
T1: sets its tenant object to Stopped, but no-one cares, there's a new tenant object in the hashmap
T1: returns with success
T2: Tenant::attach finishes.
Result: after attach finishes, the tenant is Active and in memory but has a ignored marker file on disk.
On next pageserver restart, it will be ignored. Before, nobody will notice.
```

SomeoneToIgnore:
Added the code to eliminate the shutdown race.

pageserver/src/tenant_mgr.rs
```diff
@@ -348,25 +386,90 @@ pub fn list_tenants() -> Vec<(TenantId, TenantState)> {
...
+async fn remove_tenant_from_memory<V, F>(
+    tenant_id: TenantId,
+    tenant_cleanup: F,
+) -> anyhow::Result<V>
+where
+    F: std::future::Future<Output = anyhow::Result<V>>,
+{
+    // XXX:
+    // It's important to keep the tenant in memory after the final cleanup, to avoid cleanup races.
+    // The exclusive lock here ensures we don't miss the stats updates before trying another removal.
```
problame, line 433: stats updates? You mean `TenantState` change?

pageserver/src/tenant_mgr.rs
```diff
@@ -348,25 +386,90 @@ pub fn list_tenants() -> Vec<(TenantId, TenantState)> {
...
+                | TenantState::Active => tenant.set_stopping(),
+                TenantState::Stopping => {
+                    anyhow::bail!("Tenant {tenant_id} is stopping already")
                 }
+            },
+            None => anyhow::bail!("Tenant not found for id {tenant_id}"),
+        }
+    }
+
+    // shutdown all tenant and timeline tasks: gc, compaction, page service)
```
problame, line 452: Add the following here:
```
No new tasks will be started for this tenant because it's in `Stopping` state.
Hence, once we're done here, the `tenant_cleanup` callback can mutate tenant on-disk state freely.
```

pageserver/src/tenant_mgr.rs
```diff
@@ -348,25 +386,90 @@ pub fn list_tenants() -> Vec<(TenantId, TenantState)> {
...
+/// If the cleanup fails, tenant will stay in memory in [`TenantState::Broken`] state, and another removal
+/// operation would be needed to remove it.
+async fn remove_tenant_from_memory<V, F>(
+    tenant_id: TenantId,
+    tenant_cleanup: F,
+) -> anyhow::Result<V>
+where
+    F: std::future::Future<Output = anyhow::Result<V>>,
+{
+    // XXX:
```
problame, line 431: Why XXX? This function is correct, I think.

pageserver/src/tenant_mgr.rs
```diff
@@ -348,25 +386,90 @@ pub fn list_tenants() -> Vec<(TenantId, TenantState)> {
...
+    task_mgr::shutdown_tasks(None, Some(tenant_id), None).await;
+
+    match tenant_cleanup
+        .await
+        .with_context(|| format!("Failed to run cleanup for tenant {tenant_id}"))
+    {
+        Ok(hook_value) => {
+            let mut tenants_accessor = tenants_state::write_tenants();
+            if tenants_accessor.remove(&tenant_id).is_none() {
+                warn!("Tenant {tenant_id} got removed from memory before operation finished");
```
problame, line 462: So, can this even happen?
The point of the `match` at the beginning of the function is that we only admit exactly on tokio task into the remainder of this function. All others are locked out because the tenant state is `Stopping` then.

That's necessary to serialize, say, concurrent detach & ignore.

If we arrive here, we failed to do that.


The only other function that removes entries from the tenants map is `shutdown_all_tenants`.
I guess we could still race with that one.
Any harm in making it use this function here as well, then turn this `warn` into an `assert`?

SomeoneToIgnore: I'm not sure it's worth the code changes: technically, now there's a very small race window for `shutdown_all_tenants` with anything, since before it's called, all pageserver endpoints are stopped.

Also, this function is somewhat special since tries to operate on tenants in bulk, infallible.
In particular, it cleans the tenant map, which is not working well with our model of synchronization, but does work due to the fact that we've stopped all other tasks.

I think, warning is ok here since will have a small chance to appear only when shutdown was called during some slow `tenant_cleanup` call which should be rare and visible in logs thanks to such messages.

pageserver/src/tenant_mgr.rs
SomeoneToIgnore: I'm used to interpret it as WARNING, but indeed all our comments around are like that, removed.

pageserver/src/tenant_mgr.rs
problame: Ok, let's ship this.

problame (approved):

vadim2404:
Possibly can fix #2537 (cc @problame)

PR #2979, opened 2022-11-30
Add synthetic layer map bench

bojanserafimov:
I was experimenting with the most trivial synthetic (but realistic) tests for layer map search and found some of them perform really bad, at least 20x worse than I'd expect. For example, it took 17 sec to answer 100k queries on a layer map with 100k layers in it (on my laptop).

I'm sharing this as a draft PR so people can run the code if they wish. I don't intend to merge the code as is. TODO:
- randomly sample search queries so benchmark can run much faster
- experiment more
- settle on 3-4 representative test cases

Relevant https://github.com/neondatabase/neon/pull/2515

### DIFF
--- a/pageserver/benches/bench_layer_map.rs
+++ b/pageserver/benches/bench_layer_map.rs
@@ -5,6 +5,7 @@ use pageserver::tenant::layer_map::LayerMap;
 use pageserver::tenant::storage_layer::Layer;
 use pageserver::tenant::storage_layer::ValueReconstructResult;
 use pageserver::tenant::storage_layer::ValueReconstructState;
+use rand::prelude::{SeedableRng, SliceRandom, StdRng};
 use std::cmp::{max, min};
 use std::fs::File;
 use std::io::{BufRead, BufReader};
@@ -12,6 +13,7 @@ use std::ops::Range;
 use std::path::PathBuf;
 use std::str::FromStr;
 use std::sync::Arc;
+use std::time::Instant;
 use utils::id::{TenantId, TimelineId};
 use utils::lsn::Lsn;
 
@@ -257,11 +259,18 @@ fn bench_from_real_project(c: &mut Criterion) {
 // Benchmark using synthetic data. Arrange image layers on stacked diagonal lines.
 fn bench_sequential(c: &mut Criterion) {
     let mut layer_map = LayerMap::default();
+
+    // Init layer map. Create 100_000 layers arranged in 1000 diagonal lines.
+    //
+    // TODO This code is pretty slow and runs even if we're only running other
+    //      benchmarks. It needs to be somewhere else, but it's not clear where.
+    //      Putting it inside the `bench_function` closure is not a solution
+    //      because then it runs multiple times during warmup.
+    let now = Instant::now();
     for i in 0..100_000 {
-        // % 100 => 17 sec (very bad, should be way under 1 sec. 20x room for improvement)
-        // % 1000 => 1.6 sec
-        // % 10_000 =>
-        // TODO try inserting a super-wide layer in between every 10
+        // TODO try inserting a super-wide layer in between every 10 to reflect
+        //      what often happens with L1 layers that include non-rel changes.
+        //      Maybe do that as a separate test.
         let i32 = (i as u32) % 100;
         let zero = Key::from_hex("000000000000000000000000000000000000").unwrap();
         let layer = DummyImage {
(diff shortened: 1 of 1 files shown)


pageserver/benches/bench_layer_map.rs
```diff
@@ -254,5 +254,34 @@ fn bench_from_real_project(c: &mut Criterion) {
...
+        // % 10_000 =>
+        // TODO try inserting a super-wide layer in between every 10
+        let i32 = (i as u32) % 100;
+        let zero = Key::from_hex("000000000000000000000000000000000000").unwrap();
+        let layer = DummyImage {
+            key_range: zero.add(10 * i32)..zero.add(10 * i32 + 1),
+            lsn: Lsn(10 * i),
+        };
+        layer_map.insert_historic(Arc::new(layer));
+    }
```
bojanserafimov, line 272: Where's a good place to put init code like this? Right now this will run even if I'm not running this test, and it takes a while to finish. I tried putting it in the fixture below, but that runs multiple times during warmup. Maybe put it in the fixture and use `Lazy`? It feels like I'm abusing the framework.

pageserver/benches/bench_layer_map.rs
```diff
@@ -254,5 +254,34 @@ fn bench_from_real_project(c: &mut Criterion) {
...
 
-criterion_group!(benches, bench_from_captest_env, bench_from_real_project);
+// Benchmark using synthetic data. Arrange image layers on stacked diagonal lines.
+fn bench_sequential(c: &mut Criterion) {
+    let mut layer_map = LayerMap::default();
+    for i in 0..100_000 {
+        // % 100 => 17 sec (very bad, should be way under 1 sec. 20x room for improvement)
+        // % 1000 => 1.6 sec
+        // % 10_000 =>
+        // TODO try inserting a super-wide layer in between every 10
```
knizhnik, line 264: I wonder why we are not reporting time of this initialization if it is so critical?

pageserver/benches/bench_layer_map.rs
```diff
@@ -254,5 +254,34 @@ fn bench_from_real_project(c: &mut Criterion) {
...
+            key_range: zero.add(10 * i32)..zero.add(10 * i32 + 1),
+            lsn: Lsn(10 * i),
+        };
+        layer_map.insert_historic(Arc::new(layer));
+    }
+
+    let queries: Vec<(Key, Lsn)> = uniform_query_pattern(&layer_map);
+
+    // Test with uniform query pattern
+    c.bench_function("sequential_uniform_queries", |b| {
```
knizhnik, line 277: In my case estimated time is 2577.3 s 
```
Benchmarking sequential_uniform_queries: Collecting 100 samples in estimated 2577.3 s (100 iterations)^C
```
It seems to be too large and actually not necessary.
May be add .step_by(100) to `uniform_query_pattern` ?

knizhnik (approved):

pageserver/benches/bench_layer_map.rs
knizhnik: Looks like search time greatly depends on key.
Without `step_by(100)` time is increased not 100, but 100 times:
```
Benchmarking sequential_uniform_queries: Collecting 100 samples in estimated 39sequential_uniform_queries
                        time:   [35.525 s 36.184 s 36.887 s]
                        change: [+14507% +14823% +15137%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe


real	64m34,645s
```

How it is possible to reduce number of samples?

pageserver/benches/bench_layer_map.rs
bojanserafimov: We could. But this framework has a hard-coded minumum of 10 samples for any measurement. I don't want to run this 10 times just to measure the query performance.

So far I haven't been very happy with this framework.

pageserver/benches/bench_layer_map.rs
bojanserafimov: I'll randomly sample 100 out of 100_1000 queries with fixed seed

pageserver/benches/bench_layer_map.rs
```diff
@@ -254,5 +256,53 @@ fn bench_from_real_project(c: &mut Criterion) {
...
+// Benchmark using synthetic data. Arrange image layers on stacked diagonal lines.
+fn bench_sequential(c: &mut Criterion) {
+    let mut layer_map = LayerMap::default();
+
+    // Init layer map. Create 100_000 layers arranged in 1000 diagonal lines.
+    //
+    // TODO This code is pretty slow and runs even if we're only running other
+    //      benchmarks. It needs to be somewhere else, but it's not clear where.
+    //      Putting it inside the `bench_function` closure is not a solution
+    //      because then it runs multiple times during warmup.
```
bojanserafimov, line 268: Maybe criterion is just the wrong tool to use for layer map benchmarks. It's very hostile to any code that takes more than a millisecond to run or set up. I'll try `bencher` after this PR.

PR #2985, opened 2022-12-01
Fix #2907: Remove missing_layers  property from IndexPart

### LINKED ISSUE #2907: Remove `IndexPart::missing_layers`
See  https://github.com/neondatabase/neon/pull/2785#discussion_r1030466943

shanyp:
Since missing_layers is not used, removed it from the struct/unit test and integration tests


problame (commented):
I'm worried that removing this field will break backward-compatibility.
I.e., a pageserver with this change uploads an `index_part.json` to S3, it won't have the `missing_layers` field anymore.
But a pageserver before this change will still expect it to be there. So, we'd get a deserialization failure.

At least that's my mental model of how serde works.

I wonder whether I'm wrong or whether the `test_runner/regress/test_compatibility.py` tests don't cover this.
(cc @bayandin )

bayandin:
> I wonder whether I'm wrong or whether the `test_runner/regress/test_compatibility.py` tests don't cover this.
> (cc @bayandin )

This totally might be. I could add it to the test. Can you suggest a case for this to fail (new binary VS old data / old binary VS new data)?

problame:
> . Can you suggest a case for this to fail (new binary VS old data / old binary VS new data)?

Basic test to ensure this is a forward-compatible change
==  absence of `missing_layers` key in new IndexParts works with old software.

Steps:

1. With new software

    1. Create PS with remote storage.
    2. Insert some data into timeline.
    3. Wait for remote upload to finish.
2. Stop new software
3. Remove local filesystem state.
4. With old software

    1. Create PS with same remote storage as in step 1.
    2. Attach the tenant for which we inserted data.
    3. Wait for tenant to become active.
    4. Ensure that we can read the data
    

Maybe you and @shanyp can pair up on adding such a test? Seems you have the most experience with the compatibility tests.

Issue #2904, opened 2022-11-23
`no unpinned buffers available` while running pgbench


LizardWizzard:
## Steps to reproduce

```
pgbench -i --init-steps dtGvp -s150 $CONNSTR
pgbench -c8 -P2 -T600 $CONNSTR
```

## Expected result

No errors

## Actual result

```
pgbench: error: client 0 script 0 aborted in command 5 query 0: ERROR:  no unpinned buffers available
pgbench: error: client 1 script 0 aborted in command 5 query 0: ERROR:  no unpinned buffers available
pgbench: error: client 5 script 0 aborted in command 5 query 0: ERROR:  no unpinned buffers available
```

## Environment

staging

## Logs, links

Comment by @MMeent 

> Then we should fix something; 16384 buffers at 8 threads means 2048 buffers /thread pinned. That's much more than you'd expect normally.

knizhnik:
Yes, it is fixed.
We have a problem that smgrexend -we do not correctly handle extend on more than one page.
ore details are their: https://neondb.slack.com/archives/C04BLQ4LW7K/p1669848816966029?thread_ts=1669646589.188689&cid=C04BLQ4LW7K

PR #2986, opened 2022-12-01
Fix copying relation in walloged create database in PG15

knizhnik:
refer #2904

### DIFF
--- a/pgxn/neon/pagestore_smgr.c
+++ b/pgxn/neon/pagestore_smgr.c
@@ -1102,7 +1102,7 @@ PageIsEmptyHeapPage(char *buffer)
 }
 
 static void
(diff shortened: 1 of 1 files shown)


knizhnik:
> I'm not super happy with doing N (!) WAL log operations to extend a relation by N blocks, but that's how it is, apparently. I hope we can correctly identify that these blocks are empty and thus don't need to be stored as 8kB blocks in the pageserver.

Yes, I am also not so happy with it. Actually it significantly slowdown "create database" (some noticeable delay). I do not think that it is so critical, because "create database" is not so frequent operation (and actually prohibited for Neon users).
But it will be bad if `CreateAndCopyRelation` will be used somewhere else (right now I checked that it is used only in CREATE DATABASE).

So I do not know some better (simpler or more efficient) solution.

PR #2998, opened 2022-12-03
Implement layer map using immutable BST

bojanserafimov:
Resolves https://github.com/neondatabase/neon/issues/1358

Inspired by @yeputons 's persistent/immutable segment tree idea, I implemented the layer map using a persistent/immutable BST. The advantage of using BST is that runtime depends on `log N`, not `log MAX_KEY`. Also it's nice that there exist persistent/immutable BST libraries already (hopefully they're good). We could go back to the segment tree if it turns out to be better. For now I find the BST to be an easier starting point for benchmarking

Compared to our current layer map RTree implementation, this new implementation is roughly 70x faster on the real data benchmark and roughly 2000x faster on the synthetic data benchmark. It's also much faster on secondary layer map methods besides search, like deciding whether or not to make a new image layer.

Even though more perf improvements can be made, it seems like a good time to stop with experimentation, clean the code up and merge it. It's fast enough.


pageserver/benches/bench_layer_map.rs
```diff
@@ -243,63 +245,117 @@ fn bench_from_captest_env(c: &mut Criterion) {
...
         b.iter(|| {
             for q in queries.clone().into_iter() {
                 layer_map.search(q.0, q.1).unwrap();
             }
         });
     });
+    group.bench_function("persistent_bst", |b| {
+        b.iter(|| {
+            for q in queries.clone().into_iter() {
+                bstlm.query(q.0.to_i128(), q.1 .0);
```
koivunej, line 354: Might be a good idea to have an unwrap here as well (compared to "current_code" above), also `criterion::black_box` for both/all. Unsure what happens to this as is, failed to find it from asm even after adding a `#[inline(never)]`, but that's probably me just not seeing it.

Tried to run these to compare, but it seems all of these fail with panic at `thread 'main' panicked at 'unexpected retroactive insert', pageserver/src/tenant/bst_layer_map.rs:55:17`, as is probably expected.

knizhnik:
Trying to launch this version on real Kontur layout:
```
thread 'main' panicked at 'unexpected retroactive insert', pageserver/src/tenant/bst_layer_map.rs:55:17
stack backtrace:
   0: rust_begin_unwind
             at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/core/src/panicking.rs:142:14
   2: pageserver::tenant::bst_layer_map::PersistentLayerMap<Value>::insert
             at /home/knizhnik/zenith/pageserver/src/tenant/bst_layer_map.rs:55:17
   3: pageserver::tenant::layer_map::LayerMap::insert_historic
             at /home/knizhnik/zenith/pageserver/src/tenant/layer_map.rs:365:9
   4: pageserver::tenant::timeline::Timeline::load_layer_map
             at /home/knizhnik/zenith/pageserver/src/tenant/timeline.rs:921:17
   5: pageserver::tenant::UninitializedTimeline::initialize_with_lock
             at /home/knizhnik/zenith/pageserver/src/tenant.rs:190:21
   6: pageserver::tenant::Tenant::init_attach_timelines
             at /home/knizhnik/zenith/pageserver/src/tenant.rs:1501:19
   7: pageserver::tenant_mgr::attach_local_tenants
             at /home/knizhnik/zenith/pageserver/src/tenant_mgr.rs:194:23
   8: pageserver::tenant_mgr::init_tenant_mgr
             at /home/knizhnik/zenith/pageserver/src/tenant_mgr.rs:121:5
   9: pageserver::start_pageserver
             at /home/knizhnik/zenith/pageserver/src/bin/pageserver.rs:302:9
  10: pageserver::main
             at /home/knizhnik/zenith/pageserver/src/bin/pageserver.rs:105:5
  11: core::ops::function::FnOnce::call_once
             at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/core/src/ops/function.rs:248:5
```

bojanserafimov:
The panics are because in my last commit I semi-integrated the new code into the actual layer map. But the new code expects insertions to be sorted by LSN. If you want to run a bench, go one commit back.

knizhnik:
> The panics are because in my last commit I semi-integrated the new code into the actual layer map. But the new code expects insertions to be sorted by LSN. If you want to run a bench, go one commit back.

Sorry, but my intention was actually to test integration and measure speedup o real layermap methods (like `image_coverage`)
Layers are inserted on pageserver startup in the order of traversing directory. Certainly them can be stored in some temporary collection  and then sort by LSN but may be it is better to do it `LayerMap` and extend it API for it.

bojanserafimov:
Ok there is a solution to get the exact same semantics with this data structure. I wrote down the TODOs in the last commit, now I need to do the typing :)

knizhnik:
I have implemented sorting of files by LSN in `Timeline::load_layer_map` (if you want I can commit this patch to your branch),
but still not able to run compaction test:
```
2022-12-09T10:20:43.282391Z ERROR compaction_loop{tenant_id=3e1314c3dd896fe5cdcd9a277f6c0c8e}:compact_timeline{timeline=26f8423c0335368dfa36777528027959}: could not compact, repartitioning keyspace failed: could not find layer with more data for key 000000067F0000400200008A5900FFFFFFFF at LSN 93/1B9AAA88, request LSN 93/1BC0DF10, ancestor 0/0

Caused by:
    0: layer traversal: result Continue, cont_lsn 93/1B9AAA89, layer: 000000067F000032AC000040040000000000-030000000000000000000000000000000002__000000931B9AAA89-000000931B9D7EF9
    1: layer traversal: result Continue, cont_lsn 93/1B9AAA89, layer: 000000067F000032AC000040040000000000-030000000000000000000000000000000002__000000931B9AAA89-000000931B9D7EF9

Stack backtrace:
   0: pageserver::tenant::timeline::layer_traversal_error
             at /home/knizhnik/zenith/pageserver/src/tenant/timeline.rs:2430:15
   1: pageserver::tenant::timeline::Timeline::get_reconstruct_data
             at /home/knizhnik/zenith/pageserver/src/tenant/timeline.rs:1152:32
      pageserver::tenant::timeline::Timeline::get
             at /home/knizhnik/zenith/pageserver/src/tenant/timeline.rs:403:9
   2: pageserver::pgdatadir_mapping::<impl pageserver::tenant::timeline::Timeline>::collect_keyspace
             at /home/knizhnik/zenith/pageserver/src/pgdatadir_mapping.rs:426:31
   3: pageserver::tenant::timeline::Timeline::repartition
             at /home/knizhnik/zenith/pageserver/src/tenant/timeline.rs:1609:28
   4: pageserver::tenant::timeline::Timeline::compact
             at /home/knizhnik/zenith/pageserver/src/tenant/timeline.rs:566:15
```

So looks like search is still performed not correctly.

bojanserafimov:
> So looks like search is still performed not correctly.

That's not surprising. There are at least 2 correctness TODOs:
- I'm sloppy with off-by-ones everywhere and I'm not testing for it yet
- I just implemented the correct logic in the data structure (return latest by lsn.end, not lsn.begin), which wasn't so trivial. But, I still don't have the `lsn_flloor` logic in `layer_map.rs`.

After I implement those things I plan to compare against the current RTree implementation for correctness, and then I'll try end-to-end testing.

bojanserafimov:
Current status: Implemented live comparison during pgbench test. Results are different because I'm not deleting layers from my index yet. Working on that now.

bojanserafimov:
Now both solutions return the exact same outputs on the queries made during the pgbench python test with default params. On this scale, perf is not improved (number of layers is small I guess). I haven't tested on larger tests yet. If anyone wants to test, make sure to first remove the live correctness checking in `LayerMap::search`

@knizhnik did you have any scenarios in mind?

bojanserafimov:
```
thread 'compute request worker' panicked at 'assertion failed: `(left == right)`
  left: `"000000067F0000334E0000400C0000110000-000000067F0000334E000040120100000000__00000002278B4C80"`,
 right: `"000000000000000000000000000000000000-FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF__00000002224DB0E9-00000002278B4C81"`: query: key 000000067F0000334E000040100000000000, end_lsn: 2/278B4C81', pageserver/src/tenant/layer_map.rs:260:17
```

Looks like a tie, both layers end on the same LSN.

knizhnik:
> Now both solutions return the exact same outputs on the queries made during the pgbench python test with default params. On this scale, perf is not improved (number of layers is small I guess). I haven't tested on larger tests yet. If anyone wants to test, make sure to first remove the live correctness checking in `LayerMap::search`
> 
> @knizhnik did you have any scenarios in mind?

There is test_layer_mappy benchmark.
And also I want to run compact() on real Kontur data (I have setup for it).
As far as I understand now all problems are solved and I can try to dot it?

bojanserafimov:
> As far as I understand now all problems are solved and I can try to dot it?

All the tests I've run so far are passing. No more obvious correctness TODOs. There's room for more perf improvements (probably 2-4x times) by optimizing how we call `RetroactiveLayerMap::search`, but let's perf test what we have first.

The current code maintains the index for both the old and the new solution, and runs both of them in `search` and compares the result. You can run it as is on your workload to check for bugs. Then you can comment out code in `search` there to test one solution or the other.

bojanserafimov:
I ran pgbench init 10gb, 3 times over the same project. 230 layer files total (still small number). There's no visible total perf difference, but old layer_map::search is 11% of the flame graph, while the new one is 4.5%.

Screenshot from 2022-12-13 12-05-20
Screenshot from 2022-12-13 12-05-27

More opportunities for improvements (in order of easy to hard):
- Remove Vec allocation (not worth explaining, I'll remove it in the next commit)
- Arc::clone is expensive and can be avoided
- I'm making an extra call to `RetroactiveLayerMap::search` to get the latest image, which I think is avoidable
- Maybe the BTreeMap search in LSN dimension is avoidable most of the time using caching (I bet 99% of the requests ask about the same 10 LSNs)
- Computing the image coverage of the entire key space can be done in O(N), instead of O(N * logN) (just iter the red black tree).

Do you have ideas on how to check if the rebuilds during update are problematic?

bojanserafimov:
Now down to 3% on the same test.

Screenshot from 2022-12-13 13-42-35

Runtime might have improved, but it's noisy and wal redo might be eating any gains because its troughput is capped:
old:
test_pgbench_flamegraph[45-684].init.duration: 274.650 s
test_pgbench_flamegraph[45-684].init.duration: 305.787 s
test_pgbench_flamegraph[45-684].init.duration: 302.106 s
new:
test_pgbench_flamegraph[45-684].init.duration: 238.280 s
test_pgbench_flamegraph[45-684].init.duration: 274.099 s
test_pgbench_flamegraph[45-684].init.duration: 271.065 s

IMO on this test (pgbench) we should think about perf gains in the hypothetical scenario that wal redo is also solved. It's hard to work with split bottlenecks :)

bojanserafimov:
Zoomed in on the `search_new` call:

Screenshot from 2022-12-13 13-49-34

Worth trying the `im` crate at this point, as people have suggested. `BTreeMap` is 6x faster than the red black tree, so using a good persistent BTree instead of RedBlackTree has big potential.

At the end the total cost should end up being around 3 BTree searches:
1. Search for LSN
2. Find latest delta at that LSN
3. Find latest image at that LSN

With that solution the layer map should theoretically take less than 0.5% of the flame graph on this test. So I think we can improve perf 6x and afterwards there's not much we can do besides caching. Maybe we won't even need to.

bojanserafimov:
I tried using the `im` crate. It's 30% faster on microbenchmarks and passes unit tests, but it disagrees on results during pgbench. I'll abandon it for now, maybe investigate later.

Code is in this branch https://github.com/neondatabase/neon/compare/immutable_bst_layer_map...im_layer_map?expand=1

The error was:
```
thread 'background op worker' panicked at 'assertion failed: `(left == right)`
  left: `"000000067F0000334E0000400C000009742A-030000000000000000000000000000000002__00000000016A9231-000000010107F1B1"`,
 right: `"000000067F0000334E000040220000110000-030000000000000000000000000000000002__00000003F6010000"`: query: key 0100000000000000010000000000FFFFFFFF, end_lsn: 3/F6030001', pageserver/src/tenant/layer_map.rs:260:17
```

It's a strange error because the reference implementation looks incorrect.. but it isn't, because if I don't panic here we get other errors.

My suspicion is that it has something to do with "persistent" vs "fully persistent". Fully persistent means you can branch from past versions (which I do during rebuilds). `rpds` makes the promise of full persistency in their docs, `im` doesn't. More on the topic (I'm not 100% familiar with it yet): https://youtu.be/T0yzrZL1py0?t=822

yeputons:
The only difference I can see is that `version.range(0..=key).rev().next()?` never returns negative keys, while `version.get_prev(&key)?` can. If negative keys exist at that level, that would be the beginning of an explanation.

bojanserafimov:
> f negative keys exist at that level, that would be the beginning of an explanation.

Yep, that `0` is definitely a bug because keys can be negative. But I tested `im` with both the `get_prev` and `range` apis, both failed. `rpds` solution with or without the `0` bug passes. So maybe the bug is not being exercised by this test.

> Having a different answer with a more-or-less-straightforward replacement of the underlying data structure is highly suspicious

I am suspicious, no doubt about that :) The layer map doesn't create layers, so the returned image definitely existed at some point. If it still exists, then the reference implementation is wrong here, because this image is more recent than the delta the reference returns. But if the image was deleted (and not properly deleted in my data structure), then I wonder why the delta wasn't deleted first.

knizhnik:
>    Search for LSN
>    Find latest delta at that LSN
>    Find latest image at that LSN

Sorry, did you mean Key here instead of LSN?
Otherwise I do not understand what you mean by Find latest delta at that LSN.

knizhnik:
I once again tried to test this branch with real Kontur layout.
Now it "hangs" in load `Timeline::load_layer_map`
At least it is not completed in several minutes consuming 100% CPU.
Stack trace is the following:
```
#0  rpds::map::red_black_tree_map::iter_utils::IterStack<i128, core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>, archery::shared_pointer::kind::arc::ArcK>::clear_if_finished<i128, core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>, archery::shared_pointer::kind::arc::ArcK, i128> (self=<optimized out>, start_bound=..., end_bound=...)
    at /home/knizhnik/.cargo/registry/src/github.com-1ecc6299db9ec823/rpds-0.12.0/src/map/red_black_tree_map/mod.rs:1188
#1  rpds::map::red_black_tree_map::iter_utils::IterStack<i128, core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>, archery::shared_pointer::kind::arc::ArcK>::advance<i128, core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>, archery::shared_pointer::kind::arc::ArcK, i128> (self=0x7ffe1031bb68, start_bound=..., end_bound=...)
    at /home/knizhnik/.cargo/registry/src/github.com-1ecc6299db9ec823/rpds-0.12.0/src/map/red_black_tree_map/mod.rs:1275
#2  0x000055617f694fcb in rpds::map::red_black_tree_map::RangeIterPtr<i128, core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>, core::ops::range::RangeToInclusive<i128>, i128, archery::shared_pointer::kind::arc::ArcK>::advance_forward<i128, core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>, i128, core::ops::range::RangeToInclusive<i128>, archery::shared_pointer::kind::arc::ArcK> (self=<optimized out>)
    at /home/knizhnik/.cargo/registry/src/github.com-1ecc6299db9ec823/rpds-0.12.0/src/map/red_black_tree_map/mod.rs:1417
#3  rpds::map::red_black_tree_map::{impl#23}::next<i128, core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>, core::ops::range::RangeToInclusive<i128>, i128, archery::shared_pointer::kind::arc::ArcK> (self=<optimized out>)
    at /home/knizhnik/.cargo/registry/src/github.com-1ecc6299db9ec823/rpds-0.12.0/src/map/red_black_tree_map/mod.rs:1452
#4  0x000055617f6295f8 in core::iter::traits::iterator::Iterator::fold<rpds::map::red_black_tree_map::RangeIterPtr<i128, core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>, core::ops::range::RangeToInclusive<i128>, i128, archery::shared_pointer::kind::arc::ArcK>, core::option::Option<(&i128, &core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>)>, core::iter::adapters::map::map_fold::{closure_env#0}<&archery::shared_pointer::SharedPointer<rpds::map::entry::Entry<i128, core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>>, archery::shared_pointer::kind::arc::ArcK>, (&i128, &core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>), core::option::Option<(&i128, &core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>)>, fn(&archery::shared_pointer::SharedPointer<rpds::map::entry::Entry<i128, core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>>, archery::shared_pointer::kind::arc::ArcK>) -> (&i128, &core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>), fn(core::option::Option<(&i128, &core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>)>, (&i128, &core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>)) -> core::option::Option<(&i128, &core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>)>>> (self=..., f=..., 
    init=<error reading variable: Cannot access memory at address 0x0>)
    at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/core/src/iter/traits/iterator.rs:2365
#5  core::iter::adapters::map::{impl#2}::fold<(&i128, &core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>), rpds::map::red_black_tree_map::RangeIterPtr<i128, core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>, core::ops::range::RangeToInclusive<i128>, i128, archery::shared_pointer::kind::arc::ArcK>, fn(&archery::shared_pointer::SharedPointer<rpds::map::entry::Entry<i128, core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>>, archery::shared_pointer::kind::arc::ArcK>) -> (&i128, &core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>), core::option::Option<(&i128, &core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>)>, fn(core::option::Option<(&i128, &core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>)>, (&i128, &core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>)) -> core::option::Option<(&i128, &core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>)>> (self=..., init=..., 
--Type <RET> for more, q to quit, c to continue without paging--
    g=<optimized out>) at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/core/src/iter/adapters/map.rs:124
#6  0x000055617f6b2a9c in core::iter::traits::iterator::Iterator::last<core::iter::adapters::map::Map<rpds::map::red_black_tree_map::RangeIterPtr<i128, core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>, core::ops::range::RangeToInclusive<i128>, i128, archery::shared_pointer::kind::arc::ArcK>, fn(&archery::shared_pointer::SharedPointer<rpds::map::entry::Entry<i128, core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>>, archery::shared_pointer::kind::arc::ArcK>) -> (&i128, &core::option::Option<(u64, alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>)>)>> (self=...)
    at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/core/src/iter/traits/iterator.rs:243
#7  pageserver::tenant::bst_layer_map::PersistentLayerMap<alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>>::add_node<alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>> (self=0x556182005d28, key=33729647248253425157369480890105856)
    at pageserver/src/tenant/bst_layer_map.rs:54
#8  pageserver::tenant::bst_layer_map::PersistentLayerMap<alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>>::insert<alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>> (self=0x556182005d28, key=..., lsn=..., value=Arc(strong=5, weak=0) = {...})
    at pageserver/src/tenant/bst_layer_map.rs:77
#9  pageserver::tenant::bst_layer_map::RetroactiveLayerMap<alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>>::rebuild<alloc::sync::Arc<dyn pageserver::tenant::storage_layer::Layer>> (self=0x556182005d28) at pageserver/src/tenant/bst_layer_map.rs:358
#10 0x000055617f6bd62c in pageserver::tenant::layer_map::LayerMap::rebuild_index (self=<optimized out>) at pageserver/src/tenant/layer_map.rs:439
#11 pageserver::tenant::timeline::Timeline::load_layer_map (self=0x556182005c30, disk_consistent_lsn=...) at pageserver/src/tenant/timeline.rs:949
#12 0x000055617f60cbb4 in pageserver::tenant::UninitializedTimeline::initialize_with_lock (self=..., timelines=0x556181f3c8b0, load_layer_map=true)
    at pageserver/src/tenant.rs:190
#13 0x000055617f619d68 in pageserver::tenant::Tenant::init_attach_timelines (self=0x556181f3c890, timelines=...) at pageserver/src/tenant.rs:1501
#14 0x000055617f751f59 in pageserver::tenant_mgr::attach_local_tenants (conf=0x556181f41f30, remote_index=0x7ffe1031dd20, tenants_to_attach=...)
    at pageserver/src/tenant_mgr.rs:194
#15 0x000055617f750154 in pageserver::tenant_mgr::init_tenant_mgr (conf=0x7ffe1031bb68, remote_storage=...) at pageserver/src/tenant_mgr.rs:121

#16 0x000055617f46a4f7 in pageserver::start_pageserver (conf=0x556181f41f30) at pageserver/src/bin/pageserver.rs:302
#17 0x000055617f465417 in pageserver::main () at pageserver/src/bin/pageserver.rs:105
```

knizhnik:
Looks like the problem is in the following loop:
```
355	        for ((lsn_start, lsn_end, key_start, key_end), layer) in
356	            self.layers.range((rebuild_since, 0, 0, 0)..)
357	        {
358	            self.map
359	                .insert(*key_start..*key_end, *lsn_start..*lsn_end, layer.clone());
360	        }
```
As far as I understand iterator returns entries we have just inserted.
I least I have check in the debugger that insert method is invoked millions times.

bojanserafimov:
I see what the problem is. I call `layers.rebuild_index();` after insert (no batching) in `load_layer_map`, so initialization ends up being an O(N^2) algorithm. I'll need to check all my other `layers.rebuild_index();` calls too, I wasn't very careful with them.

knizhnik:
> I see what the problem is. I call `layers.rebuild_index();` after insert (no batching) in `load_layer_map`, so initialization ends up being an O(N^2) algorithm. I'll need to check all my other `layers.rebuild_index();` calls too, I wasn't very careful with them.

You are right.
Once I moved `rebuild_index` to the end of ` load_layer_map`, Neon is started (but still much slower than with main - 14 seconds). But the time of `compact()` (which actually does no compaction or image generation - just checking if it is tie for new images) is almost the same as with main:

| Branch                               | min | max | avg  |
| ----------------------------------- | ----- | ------ | ----- |
| main                                   | 255 |  636 | 288 |
| lsn_descending_search     | 278 |  432 | 310 |
| immutable_bst_layer_map | 209 | 899  | 244 |

So either this test is not representative, either there is not so large improvement from this PR on real data.

knizhnik:
O, I see that my test was completely useless:(
You have implemented only search method sing BST.
An I am testing  `image_coverage` and `find_latest_image` which are still using old implementation.

knizhnik:
I have change old ` find_latest_image` implementation with this one:
```
    fn find_latest_image(&self, key: Key, lsn: Lsn) -> Option<Arc<dyn Layer>> {
        self.images.query(key.to_i128(), lsn.0)
    }
```
and now comact() time is significantly faster, although I kept old version of  `image_coverage` which selects affected points using R-Tree.

Branch | min | max | avg
-- | -- | -- | --
main | 255 | 636 | 288
lsn_descending_search | 278 | 432 | 310
immutable_bst_layer_map | 100 | 99 | 118


Profile shows that most of the time is still spent in R-Tree:
```
  42,48%  background op w  pageserver         [.] <rstar::algorithm::iterators::SelectionIterator<T,Func> as core::iter::traits
  19,21%  background op w  pageserver         [.] <rstar::algorithm::selection_functions::SelectInEnvelopeFuncIntersecting<T> a
  18,12%  background op w  pageserver         [.] <rstar::algorithm::selection_functions::SelectInEnvelopeFuncIntersecting<T> a
   3,21%  background op w  pageserver         [.] <pageserver::tenant::image_layer::ImageLayer as pageserver::tenant::storage_l
   2,58%  background op w  pageserver         [.] smallvec::SmallVec<A>::reserve
   2,52%  background op w  pageserver         [.] core::ops::range::Range<Idx>::contains
   2,23%  background op w  pageserver         [.] pageserver::tenant::layer_map::LayerMap::image_coverage
   1,70%  background op w  pageserver         [.] <pageserver::tenant::delta_layer::DeltaLayer as pageserver::tenant::storage_l
   0,93%  background op w  pageserver         [.] rpds::map::red_black_tree_map::iter_utils::IterStack<K,V,P>::new
```

bojanserafimov:
Current status: I removed the R-Tree completely and implemented the remaining methods (besides `search`) using the new index. Pgbench init on 10GB runs to completion, but there might be bugs I haven't triggered yet. The new code is not very well tested.

I haven't measured perf yet but I expect it to be better than before on methods other than `search`, with still some room for improvement if necessary.

knizhnik:
I repeated my "compact" test with real Kontur data.
Now results are the following:


branch | min | max | avg
-- | -- | -- | --
main | 255 | 636 | 288
yet_another_layer_map | 38 | 273 | 52
immutable_bst_layer_map | 6 | 332 |18

If everything is correct, then it is new record on this test.

knizhnik:
But results of image_coverage are not the same.
I am not sure whether them are correct or not, but them are different from main and #3175 
I just replaced `debug!` with `info!` in  `time_for_new_image_layer`
Results are attached - compact1.txt is result for main and #3175  (them are the same), compact2.txt - result for this branch.

[compact1.txt](https://github.com/neondatabase/neon/files/10319035/compact1.txt)
[compact2.txt](https://github.com/neondatabase/neon/files/10319036/compact2.txt)

bojanserafimov:
> But results of image_coverage are not the same.

One difference is that I'm calculating the number of stacked deltas, not the number of total deltas. I think it's a slightly better heuristic, but I did it because it was easier to code. But it doesn't seem like that's the only difference.

knizhnik:
Sorry, what do you mean by "stacked deltas"? `count_deltas` calculates number of deltas in the particular LSN range, actually - from last image layer and till the target LSN. It corresponds to maximal number of layers which need to be accessed in the worst case to reconstruct a page. If this number is not matched with yours than something is wrong, either in original either in your way of deltas counting.

Also number of found image layers is bout 3 times larger in your version.

bojanserafimov:
> Sorry, what do you mean by "stacked deltas"?

If there are two deltas, but have disjoint key ranges, I count 1. I'm counting the worst case number of deltas we'll need to access for reconstruction.

bojanserafimov:
> Also number of found image layers is bout 3 times larger in your version.

There are multiple ways to construct an image coverage. I select the coverage that covers each key as far up in LSN as possible. I'm not sure if the solution on main did that.

knizhnik:
> There are multiple ways to construct an image coverage. I select the coverage that covers each key as far up in LSN as possible. I'm not sure if the solution on main did that.

O, I see. Your solution really seems to be more correct (if you choose maximal value of all possible passes to image layers). But how do you handle overlapping delta layers? Count them s 1 or 2?

bojanserafimov:
> overlapping delta layers? Count them s 1 or 2?

Overlapping count as 2. My algorithm is:
1. Compute a delta cover
2. Recurse on each segment of the cover and compute the delta cover there
3. Keep recursing until we reach lsn.start of the request
4. Report the max depth found

knizhnik:
```
| ---------- image layer ---------------|
<----delta 1 ----->
                           <----delta 2 ---->
             < ------ delta 3 ---->
< ------- delta 5 --------->
                     <-------delta 6 ------>
 | ------ image coverage ------------|
```

what will return `count_deltas` in your case?

bojanserafimov:
> what will return count_deltas in your case?

3

bojanserafimov:
Sorry 4, if layers 2 and 5 intersect there for a bit

bojanserafimov:
> Also number of found image layers is bout 3 times larger in your version.

That's also strange. After reading the code on main I expect it to produce more results because it finds all endpoints in the key range and returns the latest image for each. Many consecutive ranges can have the same answer, so it will produce a separate segment for each. Example of what it could return: `[(1..3, "Image 1"), (3..5, "Image 1"), ...]`

bojanserafimov:
After we figure out correctness, there are 2 perf optimizations that will further improve runtime by a factor of log(N):
- [ ] Make only one call to `image_coverage` in `time_for_new_image_layer`, which will return an iterator which we can traverse in parallel with the partitions
- [ ] Do the same for `count_deltas`. Implement a line-sweep algorithm, where we maintain a set of relevant (in key) deltas sorted by `lsn.start` as we traverse left to right, and we count deltas with a single `range(..image.lsn.start)` call

knizhnik:
Something is wrong in your number of deltas calculation - for Kontor layout it never returns values larger than 1. But main found 3292 partitions with 2 delta layers, for example:
```
Delta1 = 000000067F000032AC000040040000000000..030000000000000000000000000000000002, delta2 = 000000067F000032AC000040040000000000..030000000000000000000000000000000002 for key range 00000006800000000000000B9700FFFFFFFF..00000006800000000000000B970100000000
```
```
Delta1 = 000000067F000032AC000040040000000000..030000000000000000000000000000000002, delta2 = 000000067F000032AC000040040000000000..030000000000000000000000000000000002 for key range 00000006800000000000000E0800FFFFFFFF..00000006800000000000000E080100000000
```

bojanserafimov:
@knizhnik If i bench `time_for_new_image_layer` runtime (and correctness) on `odd-brook-layernames.txt` would that be equivalent to the test you've been running? I want to commit the test to the repo so I don't have to bother you to run tests on every change :)

knizhnik:
> @knizhnik If i bench `time_for_new_image_layer` runtime (and correctness) on `odd-brook-layernames.txt` would that be equivalent to the test you've been running? I want to commit the test to the repo so I don't have to bother you to run tests on every change :)

Yes, certainly. But how are you going to do it?
`time_for_new_image_layer` requires partitioning schema which in turn requires deserialization of directory.
And it in turn requires content of some layers: ~30.
You you going to include them in out test framework?

bojanserafimov:
We can include the `collect_keyspace` result or the partitioning that was used in your test. In that case we should also capture the layer file names, in case they're not identical to what we have in tests.

For now I'll do something synthetic and we can improve whenever.

bojanserafimov:
I'm done messing with the core algorithm. It can be pushed a bit more, but let's merge this first. The main algorithm is ready for review, and it should be readable enough.

I did a self-review pass of the rest of the glue code, and updated the PR description with a TODO list. I'll mark the PR as "ready for review" when those are done.

bojanserafimov:
> Something is wrong in your number of deltas calculation

Compared against bruteforce and fixed lots of bugs. Looks good on the tests I have so far. Let me know if you're still finding issues.

pageserver/benches/bench_layer_map.rs
```diff
@@ -183,64 +150,110 @@ fn bench_from_captest_env(c: &mut Criterion) {
...
+
+        let now = Instant::now();
+        let result_fast = layer_map.get_difficulty_map(latest_lsn, &partitioning);
+        assert!(result_fast.len() == partitioning.parts.len());
+        println!("Finished fast in {:?}", now.elapsed());
+
+        // Assert results are equal. Manually iterate for easier debugging.
+        //
+        // TODO The difficulty numbers are in the 50-80 range. That's pretty bad.
+        //      We should be monitoring it in prod.
```
bojanserafimov, line 187: @knizhnik Is this surprising? I'm wondering if my test setup is wrong. I didn't expect to see 50 deltas without an image in between

knizhnik: It is not surprising for me. But I consider it as not expected. With current setting we should generate new image layer after 3 delta layers. Why it doesn't happen? I do not know precisely. It has to be investigated.
I see two possible reasons for it:
1.  There is issue #2984 reported by you which cause pagestore to generate image layers for almost the whole database. In case of Kontor database is really large so generation all page images will take a lot of time. During this time a lot of new delta layers can be created.
2. Compaction and materialization are performed by one threads. So if we are busy with compaction (which can take a lot of time), we are not performing page reconstruction. And once again - it is possible that during this time a lot of delta layers will be generated.

I have not investigated much layer layout of Kontor data. I just noticed one thing - extraction of serialized directory content requires 46 layers. I found it very strange and suspicious. And it corresponds with your difficulty calculation.
That is one of the motivation of my post about storage refactoring: 
https://neondb.slack.com/archives/C033RQ5SPDH/p1673247652199539
to which I got no any reaction (looks like all reported issues are not considered as a problems and we are completely satisfied with our current layer maintenance  policy).

bojanserafimov: Ok opened an issue. I'll reply to your slack post (I've been snoozing a slack reminder to reply to it for a while :)

pageserver/src/tenant/layer_map.rs
```diff
@@ -499,87 +263,226 @@ where
...
-            assert!(l.get_lsn_range().start <= lsn);
-            let range = l.get_key_range();
-            if key_range.contains(&range.start) {
-                points.push(l.get_key_range().start);
-            }
-            if key_range.contains(&range.end) {
-                points.push(l.get_key_range().end);
-            }
+        // TODO I'm 80% sure the lsn bound is inclusive. Double-check that
+        //      and document it. Do the same for image_layer_exists and count_deltas
```
bojanserafimov, line 267: @hlinnaka It's not that simple to infer from the caller code. Is the bound supposed to be inclusive or exclusive? I'd expect it to be exclusive because that's how it is on `search`. But when I implement it as exclusive I get duplicate image layers, so I'm guessing it's not.

Same question for `image_layer_exists` and `count_deltas` if you happen to know the answer.

pageserver/src/tenant/layer_map/historic_layer_coverage.rs
```diff
@@ -0,0 +1,432 @@
...
+        };
+
+        // Apply buffered updates to self.layers
+        self.buffer.retain(|rect, layer| {
+            match layer {
+                Some(l) => {
+                    let existing = self.layers.insert(rect.clone(), l.clone());
+                    if existing.is_some() {
+                        // TODO this happened once. Investigate.
+                        // panic!("can't overwrite layer");
```
bojanserafimov, line 324: with this assertion a lot of the on-demand download tests fail

pageserver/src/tenant/timeline.rs
```diff
@@ -1210,7 +1211,11 @@ impl Timeline {
                             anyhow::bail!("could not rename file {local_layer_path:?}: {err:?}");
                         } else {
                             self.metrics.resident_physical_size_gauge.sub(local_size);
-                            self.layers.write().unwrap().remove_historic(local_layer);
+                            {
+                                let mut layers = self.layers.write().unwrap();
+                                layers.remove_historic(local_layer);
+                                layers.rebuild_index();
```
bojanserafimov, line 1217: This rebuild (and the two rebuilds below) happen on every insertion. With 100k layers that's 0-100ms on every iteration. I'd prefer to rebuild after the for loop, but not sure what's the best way to do that. If I do it without holding a layer map lock, I risk someone seeing a deleted file in the layer map and panicking. But it might not be wise to hold a layer map lock if this method takes a while.

I'm not familiar with this code, would appreciate some help @SomeoneToIgnore

pageserver/src/tenant/layer_map.rs
bojanserafimov: It's inclusive. I'll update the docs and make `count_deltas` treat the bound as inclusive

bojanserafimov:
I tried using `im` instead of `rpds` again. I ported the code and ran benchmarks to see if the new correctness checks there pass.

At first they passed, but the code was slower. Then I changed `.range(..=key).rev().next()` to `.get_prev(&key)`, and the code got faster, but it failed the checks. The two different `get_difficulty` implementations returned different results, even though they both use the same map.

Then I noticed that I don't even have to return `.get_prev(&key)` to trigger the bug. Just running this function and discarding the result mutates something and makes the result wrong. Seems like `get_prev` mutates something?

Here's my code if someone wants to get to the bottom of it (see line 100) https://github.com/neondatabase/neon/compare/immutable_bst_layer_map...lm_im?expand=1 Run `cargo bench --bench bench_layer_map` to trigger the bug

Meanwhile I'm sticking with `rpds`. We don't actually need the 30% speed boost. We need the most reliable library (for now), and I thought that would be `im` because it's more popular.

knizhnik:
> At first they passed, but the code was slower. Then I changed .range(..=key).rev().next() to .get_prev(&key), and the code got faster, but it failed the checks. The two different get_difficulty implementations returned different results, even though they both use the same map.

I wonder why did you ever use `rev()`  if range returns DoubleEndeIterator?
So it is possible to call `.range(..=key).next_back()`

bojanserafimov:
> I wonder why did you ever use rev() if range returns DoubleEndeIterator?

I didn't know about `next_back` when I wrote this code :)

pageserver/src/tenant/layer_map/historic_layer_coverage.rs
bojanserafimov: @problame Is this expected behavior from https://github.com/neondatabase/neon/issues/3369 ? Could the bug cause us to insert a layer that already exists?

Cargo.toml
```diff
@@ -69,6 +69,7 @@ rand = "0.8"
 regex = "1.4"
 reqwest = { version = "0.11", default-features = false, features = ["rustls-tls"] }
 routerify = "3"
+rpds = "0.12.0"
 rstar = "0.9.3"
```
hlinnaka, line 73: We can drop the dependencies on 'rstar' and 'amplify_num' now.

pageserver/src/tenant/timeline.rs
bojanserafimov: On second thought, holding a write lock on the map while rebuilding is a more general problem I should address somehow before merging. Holding the lock for 100ms (on large dbs) doesn't seem fun

hlinnaka (approved):
Very nice! This was a pleasure to read.

A high-level introduction to the data structure would be nice somewhere. Maybe at the top of `layer_map.rs` or `historic_layer_coverage.rs`. Here's what made it "click" for me:

1. A persistent BST (and persistent data structures in general) allows you to modify the tree in such a way that each modification creates a new "version" of the tree. When you modify it, you get a new version, but all previous versions are still accessible too. So if someone is still holding a reference to an older version, they continue to see the tree as it was then. The persistent BST stores all the different versions in an efficient way.

2. Our persistent BST maintains a map of which layer file "covers" each key. It has only one dimension, the key.

3. We use the persistent-property to handle the LSN dimension. To build the layer map, we insert each layer to the persistent BST in LSN order, starting from the oldest one. After each insertion, we grab a reference to that "version" of the tree, and store it in another tree, a BtreeMap keyed by the LSN.

4. To search for a particular key-LSN pair, you first look up the right "version" in the BTreeMap. Then you search that version of the BST with the key.

5. The persistent BST keeps all the versions, but there is no way to change the old versions afterwards. We can add layers as long as they have larger LSNs than any previous layer in the map, but if we need to remove a layer, or insert anything with an older LSN, we need to throw away the persistent BST and build a new one, starting from the oldest LSN. See `rebuild_index`

All these parts are covered in the existing comments already, but it helps to have get an overview.

pageserver/src/tenant/layer_map.rs
```diff
@@ -250,138 +80,117 @@ where
...
         if layer.get_key_range() == (Key::MIN..Key::MAX) {
-            self.l0_delta_layers.push(layer.clone());
+            self.l0_delta_layers.push(layer);
         }
-        self.historic_layers.insert(LayerRTreeObject::new(layer));
+
         NUM_ONDISK_LAYERS.inc();
     }
 
+    /// Must be called after a batch of insert_historic calls, before querying
```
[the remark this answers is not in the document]
bojanserafimov: Right now It will panic if you don't call it, because there are buffered updates you probably want to see.

API-wise, I was thinking of modifying this to make it harder to forget to call it. Either return a `must_use` object on which you'll have to call `.rebuild()` or `.trust_me_i_will_rebuild_later()`, or maybe just rename `insert_historic` to `insert_historic_noflush`.

But I'm delaying this refactor until I'm sure that this rebuilding strategy is feasible. Other strategies exist. Maybe allow querying with buffered updates (query the buffer itself) or something more complicated if needed.

pageserver/src/tenant/layer_map/layer_coverage.rs
```diff
@@ -0,0 +1,151 @@
...
+            None => None,
+        };
+        self.nodes.insert_mut(key, value);
+    }
+
+    /// Insert a layer.
+    ///
+    /// Complexity: worst case O(N), in practice O(log N). See not in implementation.
+    pub fn insert(&mut self, key: Range<i128>, lsn: Range<u64>, value: Value) {
+        // NOTE The order of the following lines is important!!
```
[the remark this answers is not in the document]
bojanserafimov: Adding nodes at the endpoints **before** making any updates ensures that the coverage of keys outside the range we're working with is not affected.

I'll expand this comment

pageserver/src/tenant/timeline.rs
SomeoneToIgnore: Sorry, noticed the comment now only.
The main idea behind the code here is that on every timeline init we check the layers in the remote storage and trust that as a source of truth: all absent files are added as remote historic layers and all files that have a remote copy with different properties (now, simple size properly check only), gets replaced with the remote layer, local version gets removed.

`create_remote_layers` is supposed to happen (and I see it still does) only when pageserver has the remote storage enabled and 
* loads an existing timeline from disk (happens on pageserver start and `load` HTTP call
* loads a remote timeline from the remote storage during timeline attach (doubt, that's your case)

Based on that, I see two odd things: 

1. Why do we have "some" layer map that we reload? We can fill it properly from the first try, if we have a remote storage. If we can not do that (could be a big change, at least worth a separate PR), at least we could collect the changes and apply them once: could that help a bit?

2. Why is this a hot path for the tests? Odd, that something that happens once per timeline existence in pageserver's memory, has such a big impact. Might be not noticing a bigger picture.

pageserver/src/tenant/layer_map/historic_layer_coverage.rs
bojanserafimov: > Does this check get tripped by that, if the removal and insertion are both in the same batch of buffered updates?

Yes that's definitely a bug in my code :) I'll fix it and re-enable the check to see if it's the only bug

pageserver/src/tenant/timeline.rs
bojanserafimov: Thanks:
1. I didn't realize this is an initialization-only method. Should make things easier. We don't always have "some" pre-existing layer map, it's just that `rebuild_index()` is the `flush` method for batched updates. Maybe I'll rename it to `flush` for clarity
2. I don't know that it's a hot path. I didn't actually see any issues. The 100ms number comes from microbenchmarks, and I assumes it would cause issues. But if this is an init method It's not that big of a deal

pageserver/src/tenant/layer_map/historic_layer_coverage.rs
bojanserafimov: Hm actually I'll remove the check instead. I put it there to validate my assumption that layers are uniquely identified by their bounds. But that assumption is true based on the fact that a layer contains **all** updates within its range.

pageserver/src/tenant/layer_map/historic_layer_coverage.rs
```diff
@@ -0,0 +1,413 @@
...
+///
+/// Why is this needed? We most often insert new layers with newer LSNs,
+/// but during compaction we create layers with non-latest LSN, and during
+/// GC we delete historic layers.
+///
+/// Even though rebuilding is an expensive (N log N) solution to the problem,
+/// it's not critical since we do something equally expensive just to decide
+/// whether or not to create new image layers.
+/// TODO It's not expensive but it's not great to hold a layer map write lock
+///      for that long.
```
bojanserafimov, line 237: I should probably address this before merging

bojanserafimov: One solution is to change `HistoricLayerCoverage::historic` from `BTreeMap` into a `rpds::RedBlackTreeMapSync`. Then during rebuild we can clone the entire layer map in O(1), update the clone (without holding lock on the original), and then swap it with the original.

Though I feel like there should be a simpler solution ..

pageserver/benches/bench_layer_map.rs
koivunej: Perhaps the issue could be linked here, since it was created for this TODO?

```suggestion
        // TODO The difficulty numbers are in the 50-80 range. That's pretty bad.
        //      We should be monitoring it in prod. See: https://github.com/neondatabase/neon/issues/3316
```

pageserver/src/tenant/layer_map.rs
```diff
@@ -250,138 +80,117 @@ where
...
-                // this layer's end LSN is smaller than the requested point. If there's
-                // nothing newer, this is what we need to return. Remember this.
-                if let Some(old_candidate) = &latest_delta {
-                    if l.get_lsn_range().end > old_candidate.get_lsn_range().end {
-                        latest_delta.replace(Arc::clone(l));
-                    }
+            (Some(delta), Some(image)) => {
+                let img_lsn = image.get_lsn_range().start;
+                let image_is_newer = image.get_lsn_range().end > delta.get_lsn_range().end;
+                let image_exact_match = Lsn(img_lsn.0 + 1) == end_lsn;
```
knizhnik, line 137: img_lsn.0 + 1 ==  image.get_lsn_range().end

pageserver/src/tenant/layer_map.rs
```diff
@@ -394,102 +203,57 @@ where
...
-                if l.is_incremental() {
-                    continue;
-                }
-                let img_lsn = l.get_lsn_range().start;
-                if l.get_key_range().contains(&range_remain.start) && lsn_range.contains(&img_lsn) {
-                    made_progress = true;
-                    let img_key_end = l.get_key_range().end;
+    pub fn image_layer_exists(&self, key: &Range<Key>, lsn: &Range<Lsn>) -> Result<bool> {
+        if key.is_empty() {
+            return Ok(true);
```
knizhnik, line 218: Why?

pageserver/src/tenant/layer_map.rs
```diff
@@ -499,87 +263,237 @@ where
...
+        let mut current_key = start;
+        let mut current_val = version.delta_coverage.query(start);
+
+        // Loop through the delta coverage and recurse on each part
+        for (change_key, change_val) in version.delta_coverage.range(start..end) {
+            // If there's a relevant delta in this part, add 1 and recurse down
+            if let Some(val) = current_val {
+                if val.get_lsn_range().end > lsn.start {
+                    let kr = Key::from_i128(current_key)..Key::from_i128(change_key);
+                    let lr = lsn.start..val.get_lsn_range().start;
```
knizhnik, line 335: One small notice: `get_lsn_range/get_key_range` are not so cheap functions as may be expected.
Them perform cloning of range and I quite often saw them in profile.
I though how it wil be possible to avoid clone here by returning reference, but InMemoryLayer constructs range dynamically so it is not possible.
At least it willbe good to assign it to local variable to avoid redudant calls.

knizhnik (approved):

pageserver/src/tenant/layer_map.rs
bojanserafimov: > but InMemoryLayer constructs range dynamically so it is not possible

It's still possible. Construct it dynamically and return a reference to that, just to comply with the API.

pageserver/src/tenant/layer_map.rs
bojanserafimov: If the range is empty then for every key in the range (empty set), there's a newer image.

bojanserafimov: Added a comment

pageserver/src/tenant/layer_map.rs
bojanserafimov: I copied this line from previous code to match the logic. If I don't compare to `end_lsn` it won't work.

Maybe changing `>` to `>=` in `image_is_newer` is what you're asking for?

pageserver/src/tenant/layer_map/historic_layer_coverage.rs
```diff
@@ -0,0 +1,418 @@
...
+/// https://www.youtube.com/watch?v=WqCWghETNDc&t=581s
+pub struct BufferedHistoricLayerCoverage<Value> {
+    /// A persistent layer map that we rebuild when we need to retroactively update
+    historic_coverage: HistoricLayerCoverage<Value>,
+
+    /// We buffer insertion into the PersistentLayerMap to decrease the number of rebuilds.
+    ///
+    /// We implicitly assume that layers are identified by their key and lsn range.
+    /// A value of None means we want to delete this item.
+    buffer: BTreeMap<(u64, u64, i128, i128, bool), Option<Value>>,
```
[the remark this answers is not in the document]
bojanserafimov: `Range` doesn't have `Ord` so I added a struct instead.

pageserver/src/tenant/layer_map.rs
knizhnik: I think it will be more natural to write:
```
                let img_lsn_range =  image.get_lsn_range();
                let img_lsn = img_lsn_range.start;
                let image_is_newer = img_lsn_range.end > delta.get_lsn_range().end;
                let image_exact_match =  img_lsn_range.end == end_lsn;``
```

pageserver/src/tenant/layer_map.rs
knizhnik: It seems to me that it is not possible (returning dynamically allocated value as reference) because it is obvious memory leak. We can use boxing, but I do not like it.

bojanserafimov: Hm I thought that the lifetime of the local variables would contain the lifetime of the returned value. But it's not how it works (maybe for a good reason?) It only works if you directly construct the end result in one step.

I guess `Cow` is another option.

Or if we really want 0-overhead, I bet it's possible after https://github.com/neondatabase/neon/pull/3263/files (or similar) refactor. `InMemoryLayer` has no business being in the layer map, so it doesn't need to implement `LayerRange`, as defined in that PR.

Anyway, i'd leave this as a follow-up to this PR since even with all that cloning this PR is not a perf degradation.

pageserver/src/tenant/layer_map/historic_layer_coverage.rs
bojanserafimov: I see 3 options:
1. use rpds
2. keep it a BTreeMap, clone the whole thing on every rebuild
3. switch to sorted Vec, use Vec::append during rebuild, which uses memcpy

Tradeoffs (cons):
1. rpds: Queries are 70% slower
2. BTreeMap: **Every** rebuild clones the entire map at least
3. Vec: Write lock held for 1 vec memcpy (30 microseconds on my pc)

I'd go with option 2 for now because it's the smallest change from current code. We can follow up later and try other options if necessary.

@hlinnaka @knizhnik This would be a big change (I'll need to refactor the RwLocks), so let me know if you see easier solutions or if you'd prefer to review this separately.

knizhnik: > Even though rebuilding is an expensive (N log N) solution to the problem,
it's not critical since we do something equally expensive just to decide
whether or not to create new image layers.

I do not completely agree with it. mage layers can be created relatively rare (especially with partial image layer approach). But rebuilding it after adding each delta layer seems to be expensive.

May be I missed something,  but as you mentioned - we usually adding layers with  newer LSNs,
why then you are rebuilding index after adding each delta layer?

bojanserafimov:
There's some bug.

I added logging to see in practice how good rebuilds are (see last commit). Resulsts:
```
compaction:
Rebuilt layer map. Did 33 insertions to process a batch of 26 updates.
Rebuilt layer map. Did 66 insertions to process a batch of 53 updates.
Rebuilt layer map. Did 70 insertions to process a batch of 70 updates.
Rebuilt layer map. Did 95 insertions to process a batch of 46 updates.
Rebuilt layer map. Did 80 insertions to process a batch of 77 updates.
Rebuilt layer map. Did 85 insertions to process a batch of 82 updates.
Rebuilt layer map. Did 325 insertions to process a batch of 28 updates.

flush:
Rebuilt layer map. Did 71 insertions to process a batch of 1 updates.
Rebuilt layer map. Did 83 insertions to process a batch of 1 updates.
Rebuilt layer map. Did 1 insertions to process a batch of 1 updates.
... (many more of the last one ^^^)
```

Compaction logs seem fine. Gc logs are non-existing (I guess pgbench 10GB is not enough to cause gc). Those two flush logs are a bit concerning, but they make sense only if:
1. We start an imem layer
2. We create 80 images that intersect with it
3. We flush the inmem layer, causing layer map reinsertion of all the images

80 is not a big number, this rebuild is still pretty fast (way less than a millisecond).

But that's too many images to be creating. See diagram:
Screenshot from 2023-01-19 16-09-53

Something about the new implementation causes frequent images. It's most likely that I didn't carry over this line, which excludes L0 layers from count_deltas https://github.com/neondatabase/neon/blob/main/pageserver/src/tenant/layer_map.rs#L598

Both the spec and the implementation need a fix

bojanserafimov:
This also explains https://github.com/neondatabase/neon/issues/3316 since I didn't exclude L0's from the count

bojanserafimov:
Ok this makes more sense now. Fix coming in the next commit.

Screenshot from 2023-01-19 17-33-46

pageserver/src/tenant/layer_map.rs
```diff
@@ -518,94 +303,288 @@ where
...
+    ///      c) the number of images needed to cover it
+    ///      d) the estimated time until we'll have to reimage over it for GC
+    ///
+    /// Case 2: Since L0 layers by definition cover the entire key space, we consider
+    /// them reimage-worthy only when the entire key space can be covered by very few
+    /// images (currently 1).
+    /// TODO The optimal number should probably be slightly higher than 1, but to
+    ///      implement that we need to plumb a lot more context into this function
+    ///      than just the current partition_range.
+    pub fn is_reimage_worthy(layer: &L, partition_range: &Range<Key>) -> bool {
```
bojanserafimov, line 360: Reviewer note: This function is just a rewrite of these lines https://github.com/neondatabase/neon/blob/main/pageserver/src/tenant/layer_map.rs#L598

but instead of hiding it as an implementation hack, I'm attempting to surface the logic up to the API spec of the layer map

pageserver/src/tenant/layer_map/historic_layer_coverage.rs
bojanserafimov: Tldr: We currently only do expensive rebuilds just to remove layers after GC. I think we can avoid that by just deleting everyting in `HistoricLayerCoverage::historic` that's before `latest_gc_cutoff_lsn`. If any layer is older than the cutoff and still needed for another branch, we can keep that layer in a separate layer map for that branch. I'll need to read some code to verify my logic.

pageserver/src/tenant/layer_map.rs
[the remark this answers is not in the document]
bojanserafimov: I changed the API. Now it's checked at compile time that there are no reads before the batch is flushed

pageserver/src/tenant/layer_map/historic_layer_coverage.rs
bojanserafimov: Created follow-up issue

pageserver/benches/bench_layer_map.rs
bojanserafimov: Good idea! Created a follow-up issue

PR #3002, opened 2022-12-03
Explain prefetch

knizhnik:


knizhnik:
In principle it is possible to show prefetch statistic for each node (as in case of buffers).
But it's IMHO overkill.

knizhnik:
> If we don't plan to show statistics at the query node level, then I don't think that modifying PostgreSQL internals is worth the effort. We can probably "just" hook into PostgreSQL's EXPLAIN infrastructure (see `ExplainOneQuery_hook` in `commands/explain.h`).

I certainly considered  `ExplainOneQuery_hook` but it requires reimplementing (cut&paste) a lot of code from explain.c.
I understand you concerns about chaning postgres core and I also want to avoid them as much as possible.
But just copying postgres code is even worse - in first case we we have to handle merge conflicts after each new potgres version release, in the second case we may just not notice them.

I absolutely sure that we should provide user  way to understand whether prefetch is workig and how efficient it is.
Prefetch has critical impact on seqscan performance and so on many queries: OLAP, bulk update, vacuum,...
It is IMHO not acceptable that now to check whether prefetch is used or not we have to attach debugger.

The question how this statistic should be reported. 
We can include in pg_statistic. We can provide special function in neon extension which shows this information.
But IMHO most natural will be to show it in EXPLAIN ANALYZE, although it requires in postgres core.

PR #3008, opened 2022-12-05
Limit chrono features we use

bayandin:
For some reason, [a security warning](https://github.com/advisories/GHSA-wcg3-cvx6-7396) that we muted in the past has decided to pop up again (we've checked that we're not using anything that is affected).

One of the common fixes for this warning in GitHub projects is limiting the scope of features from `chrono`. So I've decided to give it a try. Not sure how legitimate this is for us (and if it causes any issues), but it compiles at least (If It compiles, it works, right 😄 ):

```diff
-chrono = "0.4"
+chrono = { version = "0.4", default-features = false, features = ["clock"] }
```

Also, I've updated `chrono` to  0.4.23 from 0.4.19 for Pageserver.

Please let me know if we should proceed this way or if I should dismiss the security warning again.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -85,7 +85,7 @@ dependencies = [
  "num-traits",
  "rusticata-macros",
  "thiserror",
- "time 0.3.15",
+ "time",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


SomeoneToIgnore (approved):
This is the right way to proceed with this, thank you.

PR #3012, opened 2022-12-05
Implement consumption metrics collection.

lubennikovaav:
Add new background job to collect billing metrics for each tenant and send them to the HTTP endpoint.
Metrics are cached, so we don't send non-changed metrics.

TODO:

- [x]  use proper endpoint instead of localhost (current `http://localhost:80/post` can be used with local httpbin.org)
- [x]  polish the code and add tests
- [ ]  add syntetic size metric
- [ ] add `num_s3_api_calls` metric
- [x]  add proper S3 metric


lubennikovaav:
@LizardWizzard ,
can you please review the remote size calculation in [this commit](https://github.com/neondatabase/neon/pull/3012/commits/d318ea03ae37fbd60088922dc5dfcd158c1ba033)?

As far as I understand, listing objects in s3 just to get the directory size is not very cheap operation, so we should optimize it somehow.  I am thinking about calculating the amount of bytes sent to s3 in pageserver itself somewhere around `perform_upload_task()` and caching them in `RemoteTimelineClient`. WDYT?.

LizardWizzard:
@lubennikovaav AFAIK LayerFileMetadata was introduced specifically for this use case. It contains size for each file on the remote. This data is stored in `RemoteTimelineClient.upload_queue.latest_files`. It was introduced by @koivunej for size tracking model. You may want to sync on that

pageserver/src/billing_metrics.rs
```diff
@@ -0,0 +1,278 @@
...
+            _ => anyhow::bail!("invalid value \"{s}\" for metric type"),
+        }
+    }
+}
+
+impl fmt::Display for BillingMetricKind {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(match self {
+            BillingMetricKind::WrittenSize => "written_size",
+            BillingMetricKind::SyntheticStorageSize => "syntetic_storage_size",
```
antonyc, line 102: `syntetic` -> `synthetic`?

pageserver/src/billing_metrics.rs
```diff
@@ -0,0 +1,278 @@
...
+    S3StorageSize,
+}
+
+impl FromStr for BillingMetricKind {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "written_size" => Ok(Self::WrittenSize),
+            "syntetic_storage_size" => Ok(Self::SyntheticStorageSize),
```
antonyc, line 90: `syntetic` -> `synthetic`?

pageserver/src/billing_metrics.rs
```diff
@@ -0,0 +1,278 @@
...
+        return Ok(());
+    }
+
+    // Send metrics to billing service.
+    // split into chunks of 1000 metrics to avoid exceeding the max request size
+    const CHUNK_SIZE: usize = 1000;
+    let chunks = filtered_metrics.chunks(CHUNK_SIZE);
+    for chunk in chunks {
+        let chunk_json: serde_json::Value = serde_json::json!({
+            "events": serde_json::to_string(&chunk)?});
```
koivunej, line 255: In python test it is expected to parse:

```json
{ "events": [...] }
```

But I think this will produce

```json
{ "events": "[...]" }
```

With this kind of type, I think one could handle outputting an actual json array:

```rust
#[derive(serde::Serialize)]
struct EventChunk<'a> {
  events: &'a [BillingMetric]
}


for chunk in chunks {
  // then use as serde::Serialize as is currently
  let chunk_json = serde_json::value::to_raw_value(&EventChunk {
    events: chunk
  }).expect("BillingMetric should not fail serialization");
  // snip
}
```

I think there could be a `Bytes` based buffer-reusing way to serialize these, but if the payload is currently wrong this should help forward.

koivunej (commented):
Some initial comments on the serialization.

test_runner/regress/test_metric_collection.py
```diff
@@ -0,0 +1,62 @@
...
+
+@pytest.fixture(scope="session")
+def httpserver_listen_address():
+    return ("localhost", 9091)
+
+
+#
+# verify that metrics look minilally sane
+#
+def metrics_handler(request: Request):
```
vadim2404, line 19:
```python
def metrics_handler(request: Request) -> Response:
  if request.json is None:
    return Response(status=400)

  req: Doct[str, Any] = json.loads(request.json)
  events = req["events"]
  
  checks = {
    "written_size": lambda value: value > 0,
    "physical_size": lambda value: value >= 0,
    "s3_storage_size": lambda value: value == 0,
  }
   
  for event in events:
    assert checks.pop(event["metric"])(event["value"]), f"{event['metric']} isn't valid"
  
  assert not checks, f"{' '.join(checks.keys())} wasn't/weren't received"

  return Response(status=200)
 ```

File: pageserver/src/billing_metrics.rs [hunk 5]
```diff
@@ -0,0 +1,278 @@
...
+        debug!(
+            "collected current metrics for tenant: {}: state={:?} tenant_physical_size={} remote_size={}",
+            tenant_id, tenant_state, tenant_physical_size, tenant_remote_size
+        );
+
+        current_metrics.push((
+            BillingMetricsKey {
+                tenant_id,
+                timeline_id: None,
+                metric: BillingMetricKind::PhysicalSize.to_string(),
```
vadim2404, line 196: Why don't you use the enum and all the time you're converting the value to string?

pageserver/src/billing_metrics.rs
koivunej: I guess this got incorporated in [d9e388f](https://github.com/neondatabase/neon/pull/3012/commits/d9e388f9eff772c24d9a06c48eff95ea4e380ffc)

pageserver/src/billing_metrics.rs
```diff
@@ -0,0 +1,285 @@
...
+            filtered_metrics.push({
+                BillingMetric::new_absolute(
+                    curr_key.metric,
+                    curr_key.tenant_id,
+                    curr_key.timeline_id,
+                    *curr_val,
+                )
+            });
+        }
+    }
```
koivunej, line 248: If the sending fails, shouldn't `cached_metrics` get reset to the old values?

pageserver/src/billing_metrics.rs
```diff
@@ -0,0 +1,285 @@
...
+                if res.status().is_success() {
+                    debug!("metrics sent successfully, response: {:?}", res);
+                } else {
+                    error!("failed to send metrics: {:?}", res);
+                }
+            }
+            Err(err) => {
+                error!("failed to send metrics: {:?}", err);
+            }
+        }
```
koivunej, line 281: Should the chunk be attempted to be resent?

pageserver/src/billing_metrics.rs
```diff
@@ -0,0 +1,285 @@
...
+            .json(&chunk_json)
+            .send()
+            .await;
+
+        match res {
+            Ok(res) => {
+                if res.status().is_success() {
+                    debug!("metrics sent successfully, response: {:?}", res);
+                } else {
+                    error!("failed to send metrics: {:?}", res);
```
koivunej, line 275: Consider a different kind of string to distinguish these from the "sending failed" cases (dns, network, ...).

```suggestion
                    error!("metrics endpoint refused the send metrics: {:?}", res);
```

pageserver/src/billing_metrics.rs
```diff
@@ -0,0 +1,285 @@
...
+// "type": "absolute",
+// "tenant_id": "5d07d9ce9237c4cd845ea7918c0afa7d",
+// "timeline_id": "00000000000000000000000000000000",
+// "time": ...,
+// "value": 12345454,
+// }
+#[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
+pub struct BillingMetric {
+    pub metric: BillingMetricKind,
+    pub metric_type: String,
```
koivunej, line 38:
```suggestion
    pub metric_type: &'static str,
```

If this is always `"absolute"`, it does not need to be an allocated string.

pageserver/src/billing_metrics.rs
```diff
@@ -0,0 +1,285 @@
...
+        }
+    }
+}
+
+// One iteration of metrics collection
+// Gather per-tenant and per-timeline metrics and send them to billing service,
+// cache them to avoid sending the same metrics multiple times.
+pub async fn collect_metrics_task(
+    client: &reqwest::Client,
+    cached_metrics: &Arc<Mutex<HashMap<BillingMetricsKey, u64>>>,
```
koivunej, line 152: Why is this an `Arc<Mutex<_>>` vs. `&mut`?

pageserver/src/config.rs
```diff
@@ -142,6 +148,10 @@ pub struct PageServerConf {
 
     /// Number of concurrent [`Tenant::gather_size_inputs`] allowed.
     pub concurrent_tenant_size_logical_size_queries: ConfigurableSemaphore,
+
+    // How often to collect metrics and send them to the metrics endpoint.
+    pub metric_collection_interval: Duration,
+    pub metric_collection_endpoint: String,
```
koivunej, line 154: `reqwest::Url` would work as well, then error would come sooner than the http send.

pageserver/src/billing_metrics.rs
lubennikovaav: I don't think it is necessary, because we will send metrics in other tick anyway.
Otherwise, we'll end up with too many futures spawned.

Maybe we can retry 1-2 times. WDYT?

pageserver/src/billing_metrics.rs
lubennikovaav: thank you. Fixed

pageserver/src/config.rs
lubennikovaav: Fixed. 
I also changed the logic here: disable metric collection, if no `metric_collection_endpoint` is explicitly provided.

pageserver/src/config.rs
```diff
@@ -55,6 +56,8 @@ pub mod defaults {
     pub const DEFAULT_CONCURRENT_TENANT_SIZE_LOGICAL_SIZE_QUERIES: usize =
         super::ConfigurableSemaphore::DEFAULT_INITIAL.get();
 
+    pub const DEFAULT_METRIC_COLLECTION_INTERVAL: &str = "60 s";
```
antonyc, line 59: This might be a nice interval for development. Can we ensure that in real environments it will be minutes? It might be enough for production to send each 10-15 minutes.

lubennikovaav: this is a pageserver's config parameter `metric_collection_interval`, so we can set it to any value for production.

but why 10-15 minutes? If we're going to use this info for billing, we must be more precise.

pageserver/src/billing_metrics.rs
koivunej: I guess this doesn't really matter, since you already implemented the caching for uploaded values only (comment above). I don't have an idea for this, I was just commenting to make sure it has been considered.

pageserver/src/config.rs
vadim2404: Usually, billing systems provide slices on an hourly basis. Therefore, @antonyc suggestion sounds reasonable to me, at least.

pageserver/src/storage_sync2.rs
```diff
@@ -498,6 +498,18 @@ impl RemoteTimelineClient {
         self.metrics.remote_physical_size_gauge().set(size);
     }
 
+    pub fn latest_files_total_size(&self) -> Result<u64, anyhow::Error> {
```
koivunej, line 501: Few times already I've looked at this, and then reminded myself of what the latest_files are. Since this seems to be only called from the `Tenant::get_remote_size` I wonder if this could be `get_timeline_remote_size`, just to have some sort of ... symmetry?

It will probably need to be changed if for example the remote client becomes shared between multiple timelines or something like that.

test_runner/regress/test_metric_collection.py
```diff
@@ -0,0 +1,73 @@
...
+    log.info(events)
+
+    checks = {
+        "written_size": lambda value: value > 0,
+        "physical_size": lambda value: value >= 0,
+        "s3_storage_size": lambda value: value == 0,
+    }
+
+    for event in events:
+        assert checks.pop(event["metric"])(event["value"]), f"{event['metric']} isn't valid"
```
koivunej, line 36: This was some tricky python for me on the first view.

I'd be more happy if the checks were run against all values, not just the first ones. I understand the first found is popped off, ... but that can't be right, then the second "written_size" would fail the `checks.pop` with some `ValueError`? I don't think I understand this test case after all. There should be two tenants on the pageserver in this test, right? I don't understand how could only 3 metrics come out of 2 tenants with 1 timeline each.

pageserver/src/billing_metrics.rs
```diff
@@ -0,0 +1,281 @@
...
+            .iter()
+            .map(|(curr_key, curr_val)| {
+                BillingMetric::new_absolute(
+                    curr_key.metric,
+                    curr_key.tenant_id,
+                    curr_key.timeline_id,
+                    *curr_val,
+                )
+            })
+            .collect();
```
koivunej, line 250: It feels a bit unfortunate to not being able to use the just the chunk and allocate in loop this quite large vec.

The vec could be moved outside of the loop:

```rust
let mut chunk_to_send = Vec::new();

for chunk in chunks {
  chunk_to_send.clear();
  chunk_to_send.extend(chunk.iter().map(/* the existing BillingMetric conv */));
  
  // rest unchanged
}
```

This would allocate once on the first extend to the largest chunk size (1000 or smaller).

koivunej (commented):
Some comments, I'll chat to you about the python test case I couldn't understand.

test_runner/regress/test_metric_collection.py
lubennikovaav: there's just one tenant in this test, so this is not too complicated.
also mock server only serves the first shot, so all metrics should be present here

koivunej: Ah oki, I misread that a new tenant would be created so there would be two (default and created) but that was only two timelines on 1 tenant. Got it, resolving.

pageserver/src/billing_metrics.rs
koivunej: I guess the retry interval actually affects this. Because if it's 1min, then retrying doesn't seem so tempting, because we will try often enough. But if it's 1h, then I think we start to want retrying.

pageserver/src/storage_sync2.rs
lubennikovaav: fixed. and also reused already existing metric instead of calculating it

pageserver/src/billing_metrics.rs
lubennikovaav: fixed

test_runner/regress/test_metric_collection.py
```diff
@@ -0,0 +1,73 @@
...
+    assert not checks, f"{' '.join(checks.keys())} wasn't/weren't received"
+
+    global num_metrics_received
+    num_metrics_received += 1
+    return Response(status=200)
+
+
+def test_metric_collection(
+    httpserver: HTTPServer, neon_env_builder: NeonEnvBuilder, httpserver_listen_address
+):
```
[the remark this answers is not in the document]
lubennikovaav: done
the test is not very elaborate, but I guess it's better than nothing

koivunej:
The test failure seems transient: https://neon-github-public-dev.s3.amazonaws.com/reports/pr-3012/release/3733158944/index.html#suites/837740b64a53e769572c4ed7b7a7eeeb/9ec7882016d82c95/

pageserver/src/config.rs
```diff
@@ -806,6 +845,8 @@ max_file_descriptors = 333
 initial_superuser_name = 'zzzz'
 id = 10
 
+metric_collection_interval = '222 s'
+metric_collection_endpoint = 'http://localhost:80/metrics'
```
koivunej, line 849: I think this will become a matter of needing to be fixed on the next deploy, related to us needing to keep changelogs and all that. However it doesn't seem like nothing will fail horribly if this is not fixed right away, just some errors from the line 275...

Nothing on https://github.com/neondatabase/cloud/issues/3336.

Just thinking if this should be a comment line instead not to actually enable it with bogus endpoint.

lubennikovaav: these are just nonsense values in the config parsing unit test

real defaults set above, ee `pub mod defaults ` and `DEFAULT_CONFIG_FILE`

koivunej: Aah, yes, misread, sorry.

It seems that this is not in the default config file even as commented out?

koivunej (approved):
Apart from the missing default line in the config, I think there has been enough misunderstandings on my part already.

I would prefer the comments to follow rustdoc format esp. as they are so close as it is, but code-wise this seems good.

pageserver/src/config.rs
antonyc: > but why 10-15 minutes? If we're going to use this info for billing, we must be more precise.

The Pageserver metrics are billed as max GB consumed per hour by a timeline/tenant, so having 6 events per hour per timeline/tenant might be enough to provide reasonable reliability (it's hard to loose all the 6 events), precision, and amount of events.

vadim2404 (approved):

pageserver/src/billing_metrics.rs
```diff
@@ -0,0 +1,283 @@
...
+/// "tenant_id": "5d07d9ce9237c4cd845ea7918c0afa7d",
+/// "timeline_id": "00000000000000000000000000000000",
+/// "time": ...,
+/// "value": 12345454,
+/// }
+/// ```
+#[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
+pub struct BillingMetric {
+    pub metric: BillingMetricKind,
+    pub metric_type: &'static str,
```
antonyc, line 41: This [serializes to](https://github.com/neondatabase/neon/actions/runs/3743454811/jobs/6355797651#step:4:7887) 
```
{
  "metric": "remote_storage_size",
  "metric_type": "absolute",
  "tenant_id": [
    0,
    69,
    128,
    207,
    205,
    193,
    78,
    45,
    241,
    206,
    228,
    159,
    51,
    170,
    110,
    230
  ],
  "timeline_id": null,
  "time": "2022-12-20T19:29:43.727764280Z",
  "value": 0
}
```

While [according to the spec](https://github.com/neondatabase/cloud/blob/a374c0e4c6b94116724137b186684453e785fe61/openapispec/billing/billing-v1.yaml#L133-L154) and to the example above (L28-L36) there should be fields:
```
metric
type (not metric_type)
tenant_id (a string, not an array of integers)
time
value
```
present. 

So there are inconsistencies now.

PS: The spec also requires `idempotency_key`, and while this could be softened on the control plane side, if we add changes here, could we please also add `idempotency_key` with current times + hostname? Because having this field is very much desired for the "exactly once" semantics, we would like to have, if we add retries, later on, any node of the event ingestion chain.

antonyc: UPD: I see that timeline_id is also serialized as an array of integers. OR as "null". Can we please send it 
1. either as string if present 
2. or not send it if not present? I understand 2 might be difficult on your side and will help by changing the spec on the consol plane side.

lubennikovaav: Here is the PR
https://github.com/neondatabase/neon/pull/3215

PR #3019, opened 2022-12-06
draw_timeline_dirs: accept paths as input

problame:


bojanserafimov (approved):
Seems fine. I'm curious what the use case is

PR #3020, opened 2022-12-06
Fix tenant config parsing

SomeoneToIgnore:
Adds missing `trace_read_requests` field parsing to the tenant config code, this fixing recent benchmark failures, example of the CI run:
https://github.com/neondatabase/neon/actions/runs/3631064366

I've moved the test to the regression suite, since it's very short and turned to be useful.

Generally, the whole tenant config code would benefit from further improvements, but let's keep the fixes first and small.

### DIFF
--- a/pageserver/src/config.rs
+++ b/pageserver/src/config.rs
@@ -492,7 +492,7 @@ impl PageServerConf {
         let mut builder = PageServerConfigBuilder::default();
         builder.workdir(workdir.to_owned());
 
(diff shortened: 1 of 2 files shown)


bojanserafimov (approved):
> Generally, the whole tenant config code would benefit from further improvements, but let's keep the fixes first and small.

+1. I updated 10 different places to add this config and knew I'd miss 2-3 more

PR #3026, opened 2022-12-07
refactor: use new type LayerFileName when referring to layer file names in PathBuf/RemotePath

problame:
Please use the 'review commit by commit' functionality of GitHub code review to review the three commits individually.
I can also open separate PRs for each of them if that's preferred, but they build on top of each other, so I think it's ultimately better to keep them in one PR.

If it turns out that one commit is ready to merge while others need more time, we can do that.

```
commit bd27d407d2b0f6594a2180c0aa2ad3654d0df99c
Author: Christian Schwarz <christian@neon.tech>
Date:   Wed Dec 7 09:52:48 2022 -0500

    remove Layer::is_in_memory()
    
    It's only used inside gc_timeline() and there, it always returns `false`
    because `layers.iter_historic_layers()` only returns
    ImageLayer / DeltaLayer
```
```
commit 432e8667a2be069e08ccc9a338c775dec76331e2
Author: Christian Schwarz <christian@neon.tech>
Date:   Wed Dec 7 07:40:08 2022 -0500

    refactor: split off LayerMap's needs from trait Layer into super trait PureLayer
    
    While this complicates the typings in `layer_map.rs`, it removes the
    need for most `unimplemented!()` / `todo!()` impls in the layer types
    used for benchmarking the layer map.
    
    That by itself wouldn't justify the time I put into this refactoring.
    
    But, after this change, I will remove `PathBuf`s from the `Layer` trait's interface,
    which in turn will enable a clean-up of the `storage_sync2` APIs.
```

```
commit f169a4c16862c0269e94d58887d75b8de33f1b32
Author: Christian Schwarz <christian@neon.tech>
Date:   Wed Dec 7 04:54:49 2022 -0500

    refactor: use new type LayerFileName when referring to layer file names in PathBuf/RemotePath
    
    The `RemoteTimelineClient` has all the context it needs to construct the
    local path / remote object key of the layer files.
    
    Before this patch, we would sometimes carry around plain file names in
    `Path` types and/or awkwardly "rebase" paths to have a unified
    representation of the layer file name between local and remote.
    
    This patch introduces a new type `LayerFileName` which replaces the use
    of `Path` / `PathBuf` / `RemotePath` in the `storage_sync2` APIs.
    
    Instead of holding a string, it contains the parsed representation of
    the image and delta file name.
    When we need the file name, e.g., to construct a local path or
    remote object key, we construct the name ad-hoc.
    
    `LayerFileName` is also serde {Dese,Se}rializable and hence
    is used directly in `IndexPart`, replacing `RemotePath`.
```foobar

### DIFF
--- a/pageserver/src/storage_sync2/index.rs
+++ b/pageserver/src/storage_sync2/index.rs
@@ -4,11 +4,12 @@
 
 use std::collections::{HashMap, HashSet};
 
-use remote_storage::RemotePath;
+use std::convert::TryFrom;
+
 use serde::{Deserialize, Serialize};
 use serde_with::{serde_as, DisplayFromStr};
 
-use crate::tenant::metadata::TimelineMetadata;
+use crate::tenant::{filename::LayerFileName, metadata::TimelineMetadata};
 
 use utils::lsn::Lsn;
 
@@ -62,7 +63,10 @@ impl LayerFileMetadata {
 /// remember to add a test case for the changed version.
 #[serde_as]
 #[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
-pub struct IndexPart {
+pub struct IndexPartImpl<L>
+where
+    L: std::hash::Hash + PartialEq + Eq,
+{
     /// Debugging aid describing the version of this type.
     #[serde(default)]
     version: usize,
@@ -70,25 +74,115 @@ pub struct IndexPart {
     /// Layer names, which are stored on the remote storage.
     ///
     /// Additional metadata can might exist in `layer_metadata`.
-    pub timeline_layers: HashSet<String>,
+    pub timeline_layers: HashSet<L>,
 
     /// FIXME: unused field. This should be removed, but that changes the on-disk format,
     /// so we need to make sure we're backwards-` (and maybe forwards-) compatible
     /// First pass is to move it to Optional and the next would be its removal
-    missing_layers: Option<HashSet<String>>,
+    missing_layers: Option<HashSet<L>>,
 
     /// Per layer file name metadata, which can be present for a present or missing layer file.
     ///
     /// Older versions of `IndexPart` will not have this property or have only a part of metadata
     /// that latest version stores.
-    #[serde(default)]
-    pub layer_metadata: HashMap<String, IndexLayerMetadata>,
+    #[serde(default = "HashMap::default")]
+    pub layer_metadata: HashMap<L, IndexLayerMetadata>,
 
     // 'disk_consistent_lsn' is a copy of the 'disk_consistent_lsn' in the metadata.
     // It's duplicated here for convenience.
     #[serde_as(as = "DisplayFromStr")]
     pub disk_consistent_lsn: Lsn,
     metadata_bytes: Vec<u8>,
+}
+
+pub type IndexPart = IndexPartImpl<LayerFileName>;
+
+pub type IndexPartUnclean = IndexPartImpl<UncleanLayerFileName>;
+
+#[derive(Debug, PartialEq, Eq, Hash, Clone)]
+pub enum UncleanLayerFileName {
+    Clean(LayerFileName),
+    BackupFile(String),
+}
+
+impl<'de> serde::Deserialize<'de> for UncleanLayerFileName {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        deserializer.deserialize_string(UncleanLayerFileNameVisitor)
+    }
+}
+
+struct UncleanLayerFileNameVisitor;
+
+impl<'de> serde::de::Visitor<'de> for UncleanLayerFileNameVisitor {
+    type Value = UncleanLayerFileName;
+
+    fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(
+            formatter,
+            "a string that is a valid LayerFileName or '.old' backup file name"
+        )
+    }
+
+    fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
+    where
+        E: serde::de::Error,
+    {
+        match LayerFileName::try_from(v) {
+            Ok(clean) => Ok(UncleanLayerFileName::Clean(clean)),
+            Err(e) => {
+                if v.ends_with(".old") {
+                    Ok(UncleanLayerFileName::BackupFile(v.to_owned()))
+                } else {
+                    Err(E::custom(e))
+                }
+            }
+        }
+    }
+}
+
+impl UncleanLayerFileName {
+    fn into_clean(self) -> Option<LayerFileName> {
+        match self {
+            UncleanLayerFileName::Clean(clean) => Some(clean),
+            UncleanLayerFileName::BackupFile(_) => None,
+        }
+    }
+}
+
+impl IndexPartUnclean {
+    pub fn remove_unclean_layer_file_names(self) -> IndexPart {
+        let IndexPartUnclean {
+            version,
+            timeline_layers,
+            missing_layers,
+            layer_metadata,
+            disk_consistent_lsn,
+            metadata_bytes,
+        } = self;
+
+        IndexPart {
+            version,
+            timeline_layers: timeline_layers
+                .into_iter()
+                .filter_map(UncleanLayerFileName::into_clean)
+                .collect(),
+            missing_layers: missing_layers.map(|missing_layers| {
+                missing_layers
+                    .into_iter()
+                    .filter_map(UncleanLayerFileName::into_clean)
+                    .collect()
+            }),
+            layer_metadata: layer_metadata
+                .into_iter()
+                .filter_map(|(l, m)| l.into_clean().map(|l| (l, m)))
+                .collect(),
+            disk_consistent_lsn,
+            metadata_bytes,
+        }
+    }
 }
 
 impl IndexPart {
@@ -100,23 +194,17 @@ impl IndexPart {
     pub const FILE_NAME: &'static str = "index_part.json";
 
     pub fn new(
-        layers_and_metadata: HashMap<RemotePath, LayerFileMetadata>,
+        layers_and_metadata: HashMap<LayerFileName, LayerFileMetadata>,
         disk_consistent_lsn: Lsn,
         metadata_bytes: Vec<u8>,
     ) -> Self {
         let mut timeline_layers = HashSet::with_capacity(layers_and_metadata.len());
         let mut layer_metadata = HashMap::with_capacity(layers_and_metadata.len());
 
-        for (remote_path, metadata) in &layers_and_metadata {
+        for (remote_name, metadata) in &layers_and_metadata {
+            timeline_layers.insert(remote_name.to_owned());
             let metadata = IndexLayerMetadata::from(metadata);
-            match remote_path.object_name() {
-                Some(layer_name) => {
-                    timeline_layers.insert(layer_name.to_owned());
-                    layer_metadata.insert(layer_name.to_owned(), metadata);
-                }
-                // TODO move this on a type level: we know, that every layer entry does have a name
-                None => panic!("Layer {remote_path:?} has no file name, skipping"),
-            }
+            layer_metadata.insert(remote_name.to_owned(), metadata);
         }
 
         Self {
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

File: pageserver/src/tenant/storage_layer.rs [hunk 1]
```diff
@@ -69,26 +70,8 @@ pub enum ValueReconstructResult {
...
-/// An image layer is a snapshot of all the data in a key-range, at a single
-/// LSN
-///
-pub trait Layer: Send + Sync {
-    fn get_tenant_id(&self) -> TenantId;
-
-    /// Identify the timeline this layer belongs to
-    fn get_timeline_id(&self) -> TimelineId;
-
+/// Supertrait of the [`Layer`] trait that conta
```
bojanserafimov, line 73: unfinished doc

pageserver/src/tenant/layer_map.rs
```diff
@@ -223,12 +241,15 @@ impl LayerRTreeObject {
...
-    pub layer: Arc<dyn Layer>,
+pub struct SearchResult<PersistentLayerT: ?Sized> {
+    pub layer: Arc<PersistentLayerT>,
     pub lsn_floor: Lsn,
 }
 
-impl LayerMap {
+impl<PersistentLayerT> LayerMap<PersistentLayerT>
+where
+    PersistentLayerT: ?Sized + Send + Sync + PureLayer,
```
bojanserafimov, line 251: `Send + Sync` is implied by `PureLayer`

pageserver/src/tenant/layer_map.rs
```diff
@@ -28,11 +27,12 @@ use std::sync::Arc;
...
 use utils::lsn::Lsn;
 
+use super::storage_layer::PureLayer;
+
 ///
 /// LayerMap tracks what layers exist on a timeline.
 ///
-#[derive(Default)]
-pub struct LayerMap {
+pub struct LayerMap<PersistentLayerT: ?Sized> {
```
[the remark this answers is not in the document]
problame: No, It's actually `dyn Layer` in the main code. Only in the benchmark  is it `dyn PureLayer`. Rust doesn't support upcast coercions, and all the non-benchmark code needs the layers to be proper `Layer` types anyways.

pageserver/src/tenant/filename.rs
```diff
@@ -172,6 +172,118 @@ impl fmt::Display for ImageFileName {
...
+        LayerFileName::Delta(fname)
+    }
+}
+
+// include a `/` in the name as an additional layer of robustness
+// because `/` chars are not allowed in UNIX paths
+#[cfg(test)]
+pub const LAYER_FILE_NAME_TEST_PREFIX: &str = "LAYER_FILE_NAME::test/";
+
+impl<'a> TryFrom<&'a str> for LayerFileName {
```
SomeoneToIgnore, line 214: Style NIT: Rust stdlib provides `FromStr` as an alternative https://doc.rust-lang.org/std/str/trait.FromStr.html
that enables a `parse` method on every &str, hence no later trait imports would needed to use the conversion.

`parse` name is also looks more applicable due to `parse_str` invocations inside.

pageserver/src/tenant/layer_map.rs
```diff
@@ -53,15 +53,27 @@ pub struct LayerMap {
...
     /// L0 layers have key range Key::MIN..Key::MAX, and locating them using R-Tree search is very inefficient.
     /// So L0 layers are held in l0_delta_layers vector, in addition to the R-tree.
-    l0_delta_layers: Vec<Arc<dyn Layer>>,
+    l0_delta_layers: Vec<Arc<PersistentLayerT>>,
+}
+
+impl<T: ?Sized> Default for LayerMap<T> {
+    fn default() -> Self {
+        Self {
+            open_layer: Default::default(),
```
SomeoneToIgnore, line 66: > Default::default

IMO, it's quite unreadable and also gets the compiler to do some extra type inference work that it can avoid.
We could use actual container types instead of `Default`: `None`, `VecDeque::default/new`, etc.

pageserver/src/tenant/timeline.rs
```diff
@@ -1056,72 +1059,78 @@ impl Timeline {
         //    b) typical case now is that there is nothing to sync, this downloads a lot
         //       1) if there was another pageserver that came and generated new files
         //       2) during attach of a timeline with big history which we currently do not do
-        for path in remote_filenames.difference(&local_filenames) {
-            let fname = path.to_str().unwrap();
-            info!("remote layer file {fname} does not exist locally");
+        for fname in remote_filenames.difference(&local_filenames) {
+            info!(
+                "remote layer file {} does not exist locally",
```
SomeoneToIgnore, line 1064: Good that your PR makes such things more obvious: `download_missing` function is rather generic and does not print tenant/timeline ids, layer names are also rather uninformative, so just printing the layer name, we would not know in which timeline it is missing.

If possible, we might want to print the full path here.

pageserver/src/tenant/filename.rs
```diff
@@ -172,6 +172,118 @@ impl fmt::Display for ImageFileName {
...
+            (Some(delta), None) => LayerFileName::Delta(delta),
+            (None, Some(image)) => LayerFileName::Image(image),
+            (Some(_), Some(_)) => unreachable!(),
+        };
+        Ok(ok)
+    }
+}
+
+impl LayerFileName {
+    pub fn display(&self) -> impl std::fmt::Display {
```
SomeoneToIgnore, line 239: I think that should be removed and only `file_name` should be used, if anybody wants to print it.
I've checked in the code, and it looks that we now log some of such file names, for example:

> info!("scheduled layer file deletion {}", name.display());

which will be turned into something 

`scheduled layer file deletion 000000000000000000000000000000000000-FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF__00000000016B59D8-00000000016B5A51`

It's not clear from the layer name, where does it gets deleted from.
Some places contain spans with IDs, but at least one (mentioned in the other comment below in `timeline.rs`) seems to miss them.
I think we should check the other places after deletion, and use the full file path where applicable.

pageserver/src/tenant/storage_layer.rs
```diff
@@ -127,14 +108,40 @@ pub trait Layer: Send + Sync {
...
+
+    /// Filename used to store this layer on disk. (Even in-memory layers
+    /// implement this, to print a handy unique identifier for the layer for
+    /// log messages, even though they're never not on disk.)
+    fn filename(&self) -> LayerFileName;
 
-    /// Returns true for layers that are represented in memory.
-    fn is_in_memory(&self) -> bool;
+    /// If a layer has a corresponding file on a local filesystem, return its absolute path.
+    fn local_path(&self) -> Option<LayerFileName>;
```
SomeoneToIgnore, line 144: In every current implementation, this is `Some`.

I think, the trait naming is odd: the key feature of the `Layer` now after the refactoring seems to be file persistence.
We have a `HistoricLayer` term for that, we might use that or rename that term alltogether.

Not sure what "Pure" layers represent due to missing rustdoc, but I'd rather call them "stateless"?
Maybe just name _that_ trait `Layer` instead?


---

Let's also adjust `key_iter` and remove the default implementation, if you agree with the concept of the split.

pageserver/src/tenant/timeline.rs
```diff
@@ -1450,7 +1459,7 @@ impl Timeline {
                     reconstruct_state,
                 )?;
                 cont_lsn = lsn_floor;
-                traversal_path.push((result, cont_lsn, layer));
+                traversal_path.push((result, cont_lsn, layer.as_pure_layer()));
```
SomeoneToIgnore, line 1462: This is current the only place we use this method.
Also, `traversal_path` is only passed into `layer_traversal_error` function which then only calls `l.short_id()` for some message creation.

If my assessments are correct, I would remove the entire upcast logic with `fn as_pure_layer<'a>(self: Arc<Self>) -> Arc<dyn PureLayer>` and its impls, and would do a small conversion before.

SomeoneToIgnore (changes requested):
Looks cool, I like the simplifications on the name side indeed.
Have left some comments, looks some code could be redundant, more docs and better names are worth working on.

-----------
As a follow-up idea: 
I think, what's currently called `Layer` should not exist at all: that's what I've proposed to rename into `HistoricLayer`, as it gets only used in `timeline.rs` in context of historic layer.
I think, instead that has to become an enum and I've tried to propose a PR before: 
https://github.com/neondatabase/neon/tree/kb/arcless-historic-layers
but touched some internal code that failed.

pageserver/src/tenant/timeline.rs
LizardWizzard: > and does not print tenant/timeline ids

And it doesnt need to. Because they are included into parent span variables.

SomeoneToIgnore: Found from the 2nd try, the span is also split in two, as I see.
Not sure I want to check all other such places, so will leave the other comment unresolved.

LizardWizzard: I agree that it is a bit detached with spans, but to me thats better than carrying ids and formatting them for every message

pageserver/src/tenant/storage_layer.rs
problame: > In every current implementation, this is `Some`.

All current implementations return `Some`, but with on-demand download,
a new `RemoteLayer` type will be added that returns `None` here.
No sense in churning so much, IMO.
Will add a comment.

problame: > 
> I think, the trait naming is odd: the key feature of the `Layer` now after the refactoring seems to be file persistence.
We have a `HistoricLayer` term for that, we might use that or rename that term alltogether.

Was thinking along the same lines, but didn't want to cause more churn in areas of the code that don't use `dyn Layer`.
Maybe a third commit that completes the renaming?

pageserver/src/tenant/storage_layer.rs
```diff
@@ -69,26 +70,8 @@ pub enum ValueReconstructResult {
...
-/// LSN
-///
-pub trait Layer: Send + Sync {
-    fn get_tenant_id(&self) -> TenantId;
-
-    /// Identify the timeline this layer belongs to
-    fn get_timeline_id(&self) -> TimelineId;
-
+/// Supertrait of the [`Layer`] trait that conta
+pub trait PureLayer: Send + Sync {
```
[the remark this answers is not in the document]
problame: It's not needed by `LayerMap`, so it's not included. Will add a docstring to the `PureLayer` trait.

pageserver/src/tenant/filename.rs
problame: Switched to `FromStr`

pageserver/src/tenant/timeline.rs
problame: What conversion are you referring to? So that `traversal_path` only contains `short_id()`?

SomeoneToIgnore: Yes, just basically call `short_id()` before sending `traversal_path` into `layer_traversal_error`: this way, we can avoid knowing about traits in that function at all.

pageserver/src/tenant/filename.rs
problame: Done.

I think I checked all the places.

Full path is only appropriate to use

1. during Timeline's get page code where we cross the ancestor boundary
2. in error messages that involve reading from the layer file.

In all other places, we have the tenant_id and timeline_id in tracing scope.

Did appropriate changes, will push them soon.

pageserver/src/tenant/timeline.rs
problame: Will introduce the concept of a `TraversalId` in the next push.

problame:
Changes in force-push:

- Rebased.
  - Minor conflicts with @SomeoneToIgnore 's prior refactor regarding `RemotePath`, etc.
  - Major pain with @LizardWizzard 's fix for handling `.old` files in IndexPart. Resolution:
    ```
       LayerFileName` is also serde {Dese,Se}rializable, and in an initial   
       version of this patch, it was supposed to be used directly inside      
       `IndexPart`, replacing `RemotePath`.                                   
       However,                                                               
         commit 3122f3282f1a7f4639141f5a5a451cefae53a43d                      
             Ignore backup files (ones with .n.old suffix) in download_missing
       fixed handling of `*.old` backup file names in IndexPart, and we need  
       to carry that behavior forward.                                        
       The solution is to remove `*.old` backup files names during            
       deserialization. When we re-serialize the IndexPart, the `*.old` file  
       will be gone.                                                          
       This leaks the `.old` file in the remote storage, but makes it safe    
       to clean it up later.                                                  
    ```
- Addressed review comments:

```
commit 954af20c13bf2a93bc92e0b856dfef31d4b41234                                                             
Author: Christian Schwarz <christian@neon.tech>                                                             
Date:   Fri Dec 9 12:09:43 2022 -0500                                                                       
                                                                                                            
    fixup: layer file name: switch to FromStr, as suggested by kirill                                       
                                                                                                            
commit 56d0e15f35cc81a8cfe433500f0863033009d37c                                                             
Author: Christian Schwarz <christian@neon.tech>                                                             
Date:   Mon Dec 12 07:44:29 2022 -0500                                                                      
                                                                                                            
    WIP layer traversal & remove upcasting                                                                  
                                                                                                            
commit b63cf550cae319e1c5e38d2349c5fa09de665fd3                                                             
Author: Christian Schwarz <christian@neon.tech>                                                             
Date:   Mon Dec 12 08:01:26 2022 -0500                                                                      
                                                                                                            
    make local_path() return PathBuf directly, instead of Option<PathBuf>                                   
                                                                                                            
    We'll need to change this back with on-demand download, but let's do                                    
    that when we get there.                                         get to                                         
                                                                                                            
commit 2b50cb7ec8b9d036998ce775de0ac54b12118651                                                             
Author: Christian Schwarz <christian@neon.tech>                                                             
Date:   Mon Dec 12 08:07:00 2022 -0500                                                                      
                                                                                                            
    complete 'WIP layer traversal & remove upcasting': use layer file paths as traversal ID where applicable
                                                                                                            
commit 1f9e5e4d192af8a61a599a571fc3492a7749f2f3                                                             
Author: Christian Schwarz <christian@neon.tech>                                                             
Date:   Mon Dec 12 08:14:27 2022 -0500                                                                      
                                                                                                            
    remove `Display` impl for LayerFileName, use .file_name() instead                                       
                                                                                                            
    as pre Kirill's review comment                                                                          
                                                                                                            
commit 6eaedb79c80b297a607a0541cea05f9c8a271dbe                                                             
Author: Christian Schwarz <christian@neon.tech>                                                             
Date:   Mon Dec 12 08:17:26 2022 -0500                                                                      
                                                                                                            
    fixup trait refactor: layer map: don't use Default::default                                             
```

pageserver/src/storage_sync2/index.rs
```diff
@@ -227,29 +315,24 @@ mod tests {
         let expected = IndexPart {
             // note this is not verified, could be anything, but exists for humans debugging.. could be the git version instead?
             version: 1,
-            timeline_layers: HashSet::from(["000000000000000000000000000000000000-FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF__0000000001696070-00000000016960E9".to_string()]),
+            timeline_layers: ["000000000000000000000000000000000000-FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF__0000000001696070-00000000016960E9".parse().unwrap()].into_iter().collect(),
```
SomeoneToIgnore, line 318: > .into_iter().collect()

Any particular reason to _replace_ with it a better-typed `HashSet::from`?
Somewhat relate do that `Default::default()` discussion.

SomeoneToIgnore (approved):

pageserver/src/storage_sync2/index.rs
```diff
@@ -91,6 +93,97 @@ pub struct IndexPart {
     metadata_bytes: Vec<u8>,
 }
 
+pub type IndexPart = IndexPartImpl<LayerFileName>;
```
LizardWizzard, line 96: This reminds me about our crate that we dont use as much as it deserves :) https://github.com/neondatabase/aversion

It looks like it is sort of another way to evolve format. We have evolved it from v0 to add layer metadata and used an Option to be able to parse both variants.

I'm not sure if it worth investing into it right now, so lets have a comment at least with a reference to the issue so we're able to determine when we can remove it

pageserver/src/storage_sync2/index.rs
```diff
@@ -91,6 +93,97 @@ pub struct IndexPart {
...
+    fn into_clean(self) -> Option<LayerFileName> {
+        match self {
+            UncleanLayerFileName::Clean(clean) => Some(clean),
+            UncleanLayerFileName::BackupFile(_) => None,
+        }
+    }
+}
+
+impl IndexPartUnclean {
+    pub fn remove_unclean_layer_file_names(self) -> IndexPart {
```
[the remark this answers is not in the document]
problame: I mean, we have `def test_tenant_ignores_backup_file`

pageserver/src/tenant/layer_map.rs
LizardWizzard, line 35: NIT: We dont have any sort of convention for naming of generic parameters, I can see what is the reason behind trailing T, I would still rather leave the parameter name without suffixes/prefixes. But that only my personal preference, feel free to ignore

LizardWizzard (approved):
Looks good to me, two minor nitpicky comments

pageserver/src/storage_sync2/index.rs
SomeoneToIgnore: https://github.com/neondatabase/neon/issues/3072

pageserver/src/tenant/layer_map.rs
SomeoneToIgnore: I also disagree with the naming, given that `PersistentLayerT: ?Sized + PureLayer` below and, in our terms, the persistent part is `Layer` since only that has methods like `filename`.
I've renamed `PersistentLayerT` to just `L` and will rename the traits: `PureLayer` -> `Layer`, `Layer` -> `PersistentLayer` in the next commit.

problame:
I guess with all the fixup commits, let's keep 
```
remove Layer::is_in_memory()
```

separate but squash

```
    refactor: split off LayerMap's needs from trait Layer into super trait PureLayer
```
and

```
    refactor: use new type LayerFileName when referring to layer file names in PathBuf/RemotePath
```

(Unless, of course, someones wants to do the work and split & reorder the fixups appropriately)

Combined commit messages for the latter:

```
refactor: use new type LayerFileName when referring to layer file names in PathBuf/RemotePath

Before this patch, we would sometimes carry around plain file names in
`Path` types and/or awkwardly "rebase" paths to have a unified
representation of the layer file name between local and remote.

This patch introduces a new type `LayerFileName` which replaces the use
of `Path` / `PathBuf` / `RemotePath` in the `storage_sync2` APIs.

Instead of holding a string, it contains the parsed representation of
the image and delta file name.
When we need the file name, e.g., to construct a local path or
remote object key, we construct the name ad-hoc.

`LayerFileName` is also serde {Dese,Se}rializable, and in an initial   
version of this patch, it was supposed to be used directly inside      
`IndexPart`, replacing `RemotePath`.                                   
However,                                                               
  commit 3122f3282f1a7f4639141f5a5a451cefae53a43d                      
      Ignore backup files (ones with .n.old suffix) in download_missing
fixed handling of `*.old` backup file names in IndexPart, and we need  
to carry that behavior forward.                                        
The solution is to remove `*.old` backup files names during            
deserialization. When we re-serialize the IndexPart, the `*.old` file  
will be gone.                                                          
This leaks the `.old` file in the remote storage, but makes it safe    
to clean it up later.           

There is additional churn by a preliminary refactoring that got squashed
into this change:

   split off LayerMap's needs from trait Layer into super trait PureLayer

That refactoring splits up the `Layer` trait into a `PureLayer` base trait
and a `Layer` ext trait.
`PureLayer` implements just the functions needed by `LayerMap`, whereas
`Layer` adds the context of the pageserver (timeline, tenant, file paths, etc).
```

problame:
(Edited the proposed squased commit message after posting & seeing that you renamed the traits)

pageserver/src/tenant/timeline.rs
```diff
@@ -1056,83 +1055,74 @@ impl Timeline {
...
+                        continue;
+                    }
 
-                self.layers
-                    .write()
-                    .unwrap()
-                    .insert_historic(Arc::new(delta_layer));
-                self.metrics.current_physical_size_gauge.add(sz);
-            } else if layer_name.ends_with(".old") {
-                // For details see https://github.com/neondatabase/neon/issues/3024
```
problame, line 1129: We lost this comment and the `warn` message. We should re-add it to `remove_unclean_layer_file_names`

SomeoneToIgnore: Nice catch, thank you.

SomeoneToIgnore:
e2e test fails, but looks like it's reproducible in other PRs too:
https://github.com/neondatabase/neon/actions/runs/3680372776/jobs/6226619129

We did not change anything related, so I'm inclined to merge it with this check not run, to avoid blocking on-demand download PR further.

PR #3029, opened 2022-12-07
merge-allure-report: create report even if benchmarks is skipped

bayandin:
The logic `if: success() || failure()` was added in https://github.com/neondatabase/neon/pull/2813, but the job also got skipped if `benchmarks` jobs is skipped.


bayandin:
It works!
Screenshot 2022-12-07 at 21 01 01

problame (approved):

PR #3031, opened 2022-12-08
Make safekeeper exit code 0.

arssher:
We don't have any useful graceful shutdown mode, so immediate one is normal.


petuhovskiy (approved):
Not sure if there is any real benefit in this, but why not

Issue #3024, opened 2022-12-07
unexpected layer filename in remote storage: A-B__C.0.old


LizardWizzard:
## Steps to reproduce

Found for old project created on staging in May 2022

## Expected result

No error during reconcile with remote, and probably no such files in remote storage

## Actual result
```
2022-12-07T11:58:49.445040Z ERROR could not load tenant 46eb16d232d1e747af7b59e64b4b39a8: load local timeline 35765cf4546bea869078c6f1082f8ed9
Caused by:
    0: failed to reconcile with remote
    1: unexpected layer filename in remote storage: 000000067F000040020000C0AE0000014000-030000000000000000000000000000000002__0000001145498F20.0.old
```

And indeed this file (and a bunch of others with similar `.old` suffix are included in the index file and exist as files

## Environment

staging

## Logs, links
- 

First we should check that we dont upload such files anymore. For dealing with existing files I think we have two options. Skip such files during sync (which can be ok if thats really old thing) or completely fix it with some migration that removes such files from index and from the storage. Will inspect how many of those we have in other places  

cc @hlinnaka @problame

problame:
Is this `.old` file referenced from the `IndexPart` ?

SomeoneToIgnore:
I think, that could be a combination of https://github.com/neondatabase/neon/blob/a46a81b5cbb0f4b9022152c609428247e288804d/pageserver/src/tenant/timeline.rs#L2757-L2759, some abrupt restart and a logic in remote storage that considered every file around as layer on start and attempted to upload it.

LizardWizzard:
> For the long-term solution, what if we attempt to fix this with https://github.com/neondatabase/neon/pull/2993#issuecomment-1337126725 ?

Looks good to me.

> Is this .old file referenced from the IndexPart ?

Yes

problame:
Here's code from before `storage_sync2` that caused the `.old` files to be uploaded & become part of `IndexPart`

https://github.com/neondatabase/neon/blob/c7093545794e5a1072fa358d1aa38c87031bce7e/pageserver/src/storage_sync.rs#L1467-L1489

During download, these files would be downloaded as well, but the code that loads the layer map would then ignore it.

:+1: if you agree with this root-cause analysis.

koivunej:
[...]

PR #3034, opened 2022-12-08
Ignore backup files (ones with .n.old suffix) in download_missing

LizardWizzard:
This is rather a hack to resolve immediate issue:
https://github.com/neondatabase/neon/issues/3024

Properly cleaning this file from index part requires changes to initialization of remote queue. Because we need to clean it up earlier than we start warking around files.

With on-demand there will be no walk around layer files becase download_missing is no longer needed, so I believe it will be natural to unify this with load_layer_map

@problame wdyt? Should I clean this up now or it will only introduce unnecessary conflicts with on-demand patch?


LizardWizzard:
Plan to merge this tomorrow morning and then check affected project on staging

PR #277, opened 2021-06-25
Dockerfile changes for console

ololobus:
~~+ Add debug info into zenith release binaries inside Dockerfile~~

PR #3035, opened 2022-12-08
Switch to clap derive API in safekeeper.

arssher:
Less lines and easier to read/modify.

Practically no functional changes.

### DIFF
--- a/safekeeper/src/bin/safekeeper.rs
+++ b/safekeeper/src/bin/safekeeper.rs
@@ -2,16 +2,19 @@
 // Main entry point for the safekeeper executable
 //
 use anyhow::{bail, Context, Result};
-use clap::{value_parser, Arg, ArgAction, Command};
-use const_format::formatcp;
+use clap::Parser;
 use remote_storage::RemoteStorageConfig;
+use toml_edit::Document;
+
 use std::fs::{self, File};
 use std::io::{ErrorKind, Write};
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::thread;
+use std::time::Duration;
+use storage_broker::Uri;
 use tokio::sync::mpsc;
-use toml_edit::Document;
+
 use tracing::*;
 use utils::pid_file;
 
@@ -20,7 +23,7 @@ use safekeeper::broker;
 use safekeeper::control_file;
 use safekeeper::defaults::{
     DEFAULT_HEARTBEAT_TIMEOUT, DEFAULT_HTTP_LISTEN_ADDR, DEFAULT_MAX_OFFLOADER_LAG_BYTES,
-    DEFAULT_PG_LISTEN_ADDR, DEFAULT_WAL_BACKUP_RUNTIME_THREADS,
+    DEFAULT_PG_LISTEN_ADDR,
 };
 use safekeeper::http;
 use safekeeper::remove_wal;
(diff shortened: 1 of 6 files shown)


arssher:
Based on [#277](https://github.com/neondatabase/neon/pull/2733) PR to less rebase it.

safekeeper/src/bin/safekeeper.rs
```diff
@@ -298,110 +307,21 @@ fn set_id(conf: &mut SafeKeeperConf, given_id: Option<NodeId>) -> Result<()> {
...
-            Arg::new("log-format")
-                .long("log-format")
-                .help("Format for logging, either 'plain' or 'json'")
-        )
+    Ok(my_id)
 }
 
-#[test]
-fn verify_cli() {
-    cli().debug_assert();
```
SomeoneToIgnore, line 406: There's a way to debug assert derive declarations too.
See `Migrating` section in https://github.com/clap-rs/clap/releases/tag/v4.0.0-rc.1

```rust 
#[derive(clap::Parser)]
struct Cli {
    ...
}

#[test]
fn verify_cli() {
    use clap::CommandFactory;
    Cli::command().debug_assert()
}
```

It once detected an error where `-h` was specified as a short version of the flag and the help generation of clap shadowed that.
So, could be helpful and seems very easy to create.

SomeoneToIgnore (commented):
Looks more readable.
Have you tried running `build --timings` to see how procmacros influence that?

safekeeper/src/bin/safekeeper.rs
arssher: Cool, didn't know that. Added.

arssher:
> Have you tried running build --timings to see how procmacros influence that?

I don't think it would be noticeable with other heavy deps, but let's see.

with patch
```

**Comment by arssher** [comment]
> Have you tried running build --timings to see how procmacros influence that?

I don't think it would be noticeable with other heavy deps, but let's see.

with patch
```
cargo build -p safekeeper --timings
Total time: | 466.4s (7m 46.4s)
  | safekeeper v0.1.0 bin "safekeeper" | 14.1s
```

without
```
total 483.2s (8m 3.2s)
safekeeper v0.1.0 bin "safekeeper" | 14.7s
```

and safekeeper v0.1.0 bin "safekeeper" doesn't have code generation part at all.

**Review by SomeoneToIgnore — APPROVED** [review]

**Comment by petuhovskiy** [comment]
GitHub shows a large diff with `+1,225 −2,451`, but it seems that the patch is only a single commit. Not sure if this PR was needed at all, we could review it in the same neon-broker PR.

**Comment by arssher** [comment]
Rebased on main.

File: safekeeper/src/bin/safekeeper.rs [hunk 2]
```diff
@@ -44,136 +47,143 @@ const ID_FILE_NAME: &str = "safekeeper.id";
...
+A fleet of safekeepers is responsible for reliably storing WAL received from
+compute, passing it through consensus (mitigating potential computes brain
+split), and serving the hardened part further downstream to pageserver(s).
+"#;
+
+#[derive(Parser)]
+#[command(name = "Neon safekeeper", version = GIT_VERSION, about = ABOUT, long_about = None)]
+struct Args {
+    /// Path to the safekeeper data directory.
+    #[arg(short = 'D', long, default_value = "./")]
```
yeputons, line 60: It was `long = 'dir'` before. This option was never actually parsed by safekeepers. See #3088

PR #3046, opened 2022-12-09
Remove redundant once_cell usages

SomeoneToIgnore:
rustc 1.63 is now [present](https://tracker.debian.org/pkg/rustc) in Debian unstable and I wanted to check on the "const Mutex, RwLock, Condvar initialization"
[feature](https://blog.rust-lang.org/2022/08/11/Rust-1.63.0.html#const-mutex-rwlock-condvar-initialization).
Tokio provides a `parking_lot` feature with similar possibilities for its lock counterparts.

Alas, current rustc, 1.65, does not have `HashMap::new` as `const`, `BTreeMap::new` as `const` is unstable, only `Vec::new` is `const` and available.
`static TEST: Mutex<Option<HashMap<String, u32>>> = Mutex::new(None);` works, but that is not much different from `once_cell` in terms of usage convenience, so seems that the whole idea is not yet easily applicable for our needs.

As a small solace, I've found a few small places where `once_cell` seems to be redundant.

### DIFF
--- a/libs/postgres_ffi/wal_craft/src/lib.rs
+++ b/libs/postgres_ffi/wal_craft/src/lib.rs
@@ -1,7 +1,6 @@
 use anyhow::*;
 use core::time::Duration;
 use log::*;
-use once_cell::sync::Lazy;
 use postgres::types::PgLsn;
 use postgres::Client;
 use postgres_ffi::{WAL_SEGMENT_SIZE, XLOG_BLCKSZ};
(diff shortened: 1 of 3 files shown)


SomeoneToIgnore:
(I don't seem to be able to unassign Compute from this PR before or after creation, but feel free to ignore)

LizardWizzard (approved):
Thanks!

Issue #3004, opened 2022-12-04
Timeline X/Y was cancelled and cannot be used anymore


LizardWizzard:
## Steps to reproduce

Found in sk logs

## Expected result

No error in logs

## Actual result

```
Dec 04 11:58:21 safekeeper-1.us-east-2.aws.neon.build safekeeper[1360899]: 2022-12-04T11:58:21.550310Z ERROR {tid=50}:WAL acceptor{timeline=daa3f503e63abff68a4011f989fd75f0}: failed to unregister compute connection: Timeline ccec8f9b145737b93a4866903b6a582a/daa3f503e63abff68a4011f989fd75f0 was cancelled and cannot be used anymore
Dec 04 11:58:21 safekeeper-1.us-east-2.aws.neon.build safekeeper[1360899]: 2022-12-04T11:58:21.550339Z ERROR {tid=50}: query handler for 'START_WAL_PUSH' failed: Failed to process query for timeline daa3f503e63abff68a4011f989fd75f0
Dec 04 11:58:21 safekeeper-1.us-east-2.aws.neon.build safekeeper[1360899]: Caused by:
Dec 04 11:58:21 safekeeper-1.us-east-2.aws.neon.build safekeeper[1360899]:     Timeline ccec8f9b145737b93a4866903b6a582a/daa3f503e63abff68a4011f989fd75f0 was cancelled and cannot be used anymore
```

## Environment

New staging 

## Logs, links
-

petuhovskiy:
This error means that we stopped sk timeline and all background tasks associated with this timeline should be eventually canceled. Right now, that happens only on timeline deletion because we want to shut down all background tasks after timeline was deleted. This log is quite useful because it shows the real reason why this particular task/connection was stopped and it also cannot happen more than `O(1)` times for a single timeline.

So, I don't know if there is any genuine issue here that should be fixed and what the fix should look like.

LizardWizzard:
During normal operation there shouldnt be any ERROR messages in logs. So if this is normal lets change the log level to info or debug.

petuhovskiy:
postgres_backend logs were reworked in #3759, but I still see similar logs:

```
2023-03-10T13:29:53.533731Z ERROR {cid=3329}:WAL receiver{ttid=X/Y}: terminated: Other(WAL acceptor
Caused by:
    Timeline X/Y was cancelled and cannot be used anymore
```

Need to take a closer look and probably adjust logs a bit more.

[...]

PR #3050, opened 2022-12-09
Use info! log level for query errors in pg_backend

petuhovskiy:
Fixes https://github.com/neondatabase/neon/issues/3004

### DIFF
--- a/libs/utils/src/postgres_backend.rs
+++ b/libs/utils/src/postgres_backend.rs
@@ -437,12 +437,12 @@ impl PostgresBackend {
                     if root_cause.contains("connection closed unexpectedly")
                         || root_cause.contains("Broken pipe (os error 32)")
                     {
-                        error!(
+                        info!(
                             "query handler for '{}' failed: {}",
                             query_string, short_error
                         );
                     } else {
-                        error!("query handler for '{}' failed: {:?}", query_string, e);
+                        info!("query handler for '{}' failed: {:?}", query_string, e);
                     }
                     self.write_message_noflush(&BeMessage::ErrorResponse(&short_error))?;
                     // TODO: untangle convoluted control flow
@@ -476,7 +476,7 @@ impl PostgresBackend {
                 trace!("got execute {:?}", query_string);
                 // xxx distinguish fatal and recoverable errors?
                 if let Err(e) = handler.process_query(self, query_string) {
-                    error!("query handler for '{}' failed: {:?}", query_string, e);
+                    info!("query handler for '{}' failed: {:?}", query_string, e);
                     self.write_message(&BeMessage::ErrorResponse(&e.to_string()))?;
                 }
                 // NOTE there is no ReadyForQuery message. This handler is used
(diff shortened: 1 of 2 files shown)


arssher (approved):
Generally it would be nice to distinguish errors, as some are important.

petuhovskiy:
> Generally it would be nice to distinguish errors, as some are important.

True, but it seems like the effort for this can overweight the benefits. I searched through the latest logs and found these errors:
```
- connection closed unexpectedly
- end streaming to Some("pageserver")
- Connection reset by peer (os error 104)
- Timeline was cancelled and cannot be used anymore
- invalid xlog page header: xlp_magic=0, expected 53517 at LSN
- unexpected message: Terminate during COPY
- Broken pipe (os error 32)
- Tenant X/Y not found in the local state
- state uninitialized, no data to read
- Timeline X/Y was not found
```

We can add them to the ignore list, but it can be a never-ending battle, because new errors will always appear.

arssher:
> invalid xlog page header: xlp_magic=0, expected 53517 at LSN

that one is scary

libs/utils/src/postgres_backend.rs
```diff
@@ -437,12 +437,12 @@ impl PostgresBackend {
                     if root_cause.contains("connection closed unexpectedly")
                         || root_cause.contains("Broken pipe (os error 32)")
                     {
-                        error!(
+                        info!(
```
LizardWizzard, line 440: This is rather ok-ish for me

libs/utils/src/postgres_backend.rs
```diff
@@ -437,12 +437,12 @@ impl PostgresBackend {
...
                             "query handler for '{}' failed: {}",
                             query_string, short_error
                         );
                     } else {
-                        error!("query handler for '{}' failed: {:?}", query_string, e);
+                        info!("query handler for '{}' failed: {:?}", query_string, e);
```
LizardWizzard, line 445: Logging unknown errors at info level doesnt sound right to me

libs/utils/src/postgres_backend.rs
```diff
@@ -476,7 +476,7 @@ impl PostgresBackend {
                 trace!("got execute {:?}", query_string);
                 // xxx distinguish fatal and recoverable errors?
                 if let Err(e) = handler.process_query(self, query_string) {
-                    error!("query handler for '{}' failed: {:?}", query_string, e);
+                    info!("query handler for '{}' failed: {:?}", query_string, e);
```
LizardWizzard, line 479: The same here

LizardWizzard (changes requested):
That doesnt look like a good solution to me. What if we get real error here? We shouldnt report unknown errors at info level. This way we hide problems, not solve them. Errors that are fine to ignore should be explicitly allowed.

LizardWizzard:
AFAIK The problem was solved by @SomeoneToIgnore adjusting error types to contain more details about particular error

PR #3053, opened 2022-12-09
Try to fix docker image tag in broker deploy.

arssher:


lassizci (approved):
So workflow doesn't have issues, because it's the same as
```console
$ DOCKER_TAG=mytag
$ helm ... --set image.tag=${DOCKER_TAG}
```

which works, but this doesn't work:

```
DOCKER_TAG helm ... --set image.tag=${DOCKER_TAG}
```

On the other hand we don't use the env variable for anything else, so perhaps cleaner. :+1:  :shipit:

PR #3054, opened 2022-12-09
test: kill spawned postgres

### LINKED ISSUE #2604: `test_sync_safekeepers_logs` leaks postgres instance
## Steps to reproduce

1. Clone
2. `./scripts/pysync`
3. `./scripts/pytest test_runner/regress/test_compute_ctl.py`
4. Test passes

## Expected result

`ps aux|grep post` before and after should be the same processes.

## Actual result

The usual postgres processes can be found successful test run (snipping out most columns):

```
$REPO/pg_install/v14/bin/postgres -D $REPO/test_output/test_sync_safekeepers_logs/repo/pgdatadirs/tenants/510da7b8c70a9adc72fcc3d7654494fd/test_compute_ctl_pg_node
postgres: checkpointer
postgres: background writer
postgres: walwriter
postgres: autovacuum launcher
postgres: stats collector
postgres: WAL proposer streaming 0/16D28C0
postgres: logical replication launcher
```

## Environment

Ubuntu 22.04 based linux, python-3.10, on a branch based off e516c376.

## Logs, links

- interestingly enough postgres doesn't have a log file, because that is handled in the test case only?
- unsure why does pageserver logs contain SQL queries (below)

```
2022-10-11T10:35:09.210684Z ERROR query handler for 'SELECT state, to_char(state_change, 'YYYY-MM-DD"T"HH24:MI:SS.US"Z"') AS state_change
                         FROM pg_stat_activity
                         WHERE backend_type = 'client backend'
                            AND pid != pg_backend_pid()
                            AND usename != 'cloud_admin';' failed: unknown command
```

Stacktrace from https://github.com/neondatabase/neon/blob/e52029309059ef2197aaff1a725deb4d42976fd2/libs/utils/src/postgres_backend_async.rs#L458

koivunej:
Fixes #2604.

Using a context manager since I think it's cleaner here than to try-except the assertions, but I'm not a python expert.

### DIFF
--- a/test_runner/regress/test_compute_ctl.py
+++ b/test_runner/regress/test_compute_ctl.py
@@ -1,4 +1,5 @@
 import os
+from pathlib import Path
 from subprocess import TimeoutExpired
 
 from fixtures.log_helper import log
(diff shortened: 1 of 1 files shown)


test_runner/regress/test_compute_ctl.py
```diff
@@ -195,18 +196,61 @@ def test_sync_safekeepers_logs(neon_env_builder: NeonEnvBuilder, pg_bin: PgBin):
...
+    Context manager that kills a process with a pid file on exit.
+    """
+
+    def __init__(self, pid_file: Path):
+        self.path = pid_file
+        self.pid_file = open(pid_file, "r")
+        self.pid = int(self.pid_file.readline().strip())
+
+    def __enter__(self):
+        # unsure what should this be?
```
koivunej, line 228: Oops, self is okay for this.

test_runner/regress/test_compute_ctl.py
```diff
@@ -195,18 +196,60 @@ def test_sync_safekeepers_logs(neon_env_builder: NeonEnvBuilder, pg_bin: PgBin):
...
+
+
+class ExternalProcessManager:
+    """
+    Context manager that kills a process with a pid file on exit.
+    """
+
+    def __init__(self, pid_file: Path):
+        self.path = pid_file
+        self.pid_file = open(pid_file, "r")
```
koivunej, line 224: I tried lockf and flock but it seems neither are held by postgres on the file.

koivunej:
Whoa, first try and regress tests passed. No output in happy case: https://github.com/neondatabase/neon/actions/runs/3659971995/jobs/6186717780#step:4:3132

SomeoneToIgnore (approved):
Somewhat concerning that we have similar logic in https://github.com/neondatabase/neon/pull/2988 with locks on such files and even a `pre_exec` hook.

One particular thing that stands out is that postmaster removes it pid file, while the PR linked has lots of comments on why it's racy.

It's not this PR's issue technically, so LGTM.

PR #3058, opened 2022-12-11
test_perf_pgbench: switch to server-side data generation

bayandin:
To offload the network and reduce its impact, I suggest switching to server-side data generation for the pgbench initialize workflow.

From `man pgbench`:

> g or G (Generate data, client-side or server-side)
>     Generate data and load it into the standard tables, replacing any data already present.
> 
>     With g (client-side data generation), data is generated in pgbench client and then sent to the server. This uses the client/server bandwidth extensively
>     through a COPY.  pgbench uses the FREEZE option with version 14 or later of PostgreSQL to speed up subsequent VACUUM, unless partitions are enabled. Using g
>     causes logging to print one message every 100,000 rows while generating data for the pgbench_accounts table.
> 
>     With G (server-side data generation), only small queries are sent from the pgbench client and then data is actually generated in the server. No significant
>     bandwidth is required for this variant, but the server will do more work. Using G causes logging not to print any progress message while generating data.
> 
>     The default initialization behavior uses client-side data generation (equivalent to g).

### DIFF
--- a/test_runner/fixtures/benchmark_fixture.py
+++ b/test_runner/fixtures/benchmark_fixture.py
@@ -11,7 +11,7 @@ from datetime import datetime
 from pathlib import Path
 
 # Type-related stuff
-from typing import Callable, ClassVar, Iterator, Optional
+from typing import Callable, ClassVar, Dict, Iterator, Optional
 
 import pytest
 from _pytest.config import Config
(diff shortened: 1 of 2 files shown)


test_runner/fixtures/benchmark_fixture.py
```diff
@@ -164,25 +167,35 @@ def parse_from_stderr(
...
-            total, drop_tables, create_tables, client_side_generate, vacuum, primary_keys = [
-                float(v) for v in m.groups() if v is not None
-            ]
-        else:
+        timings: Dict[str, Optional[float]] = {k: None for k in cls.EXTRACTORS.keys()}
+        last_line_items = re.split(r"\(|\)|,", last_line)
+        for item in last_line_items:
+            for key, regex in cls.EXTRACTORS.items():
+                if (m := regex.match(item.strip())) is not None:
+                    if timings.get(key) is not None:
```
LizardWizzard, line 180: Then: `if key in timings:` here

test_runner/fixtures/benchmark_fixture.py
```diff
@@ -164,25 +167,35 @@ def parse_from_stderr(
...
+                        raise RuntimeError(
+                            f"can't store pgbench results for repeated action `{key}`"
+                        )
+
+                    timings[key] = float(m.group(1))
+
+        if all(v is None for v in timings.values()) or timings["total"] is None:
```
LizardWizzard, line 187: Then `not timings` or `"total" not in timings`

LizardWizzard (approved):
Minor subjective python nits, feel free to ignore

test_runner/fixtures/benchmark_fixture.py
```diff
@@ -164,25 +167,35 @@ def parse_from_stderr(
...
         return cls(
-            total=total,
-            drop_tables=drop_tables,
-            create_tables=create_tables,
-            client_side_generate=client_side_generate,
-            vacuum=vacuum,
-            primary_keys=primary_keys,
+            total=timings["total"],
+            drop_tables=timings["drop_tables"] or 0.0,
+            create_tables=timings["create_tables"] or 0.0,
```
vadim2404, line 193: for all of these lines you have to use `get` because `KeyError` will be raised

LizardWizzard: We populate everything with None above, so with current approach it should work

vadim2404: but with proposed – it will be broken

PR #3070, opened 2022-12-12
Deprecate old runner for deploy job

zoete:
As we plan to no longer use them


arssher (approved):
I don't possess context here, but don't mind.

PR #3071, opened 2022-12-12
Add new us-west-2 region

SergeyMelnikov:


LizardWizzard (approved):
Looks correct to me. Should someone else from the infra team take a look?

PR #3073, opened 2022-12-12
neon-project-create/delete: print project id to output

bayandin:
Print project_id to stdout from GitHub actions


petuhovskiy (approved):
Funny how we create `${project_id}`, BUT DELETE `${PROJECT_ID}` :)

bayandin:
> Funny how we create `${project_id}`, BUT DELETE `${PROJECT_ID}` :)

They grow so fast! 🤣

PR #3077, opened 2022-12-13
Create initial timeline without remote storage

SomeoneToIgnore:
Sometimes, on my local machine S3 tests failed due to extra log lines present:
```
2022-12-04T11:25:13.935351Z ERROR remote_upload{tenant=b933a6c8685e3898383c94bb220779d1 timeline=ccf01aab20e4d8f34145ccf1122a5f19 upload_task_id=1}: failed to perform remote task UploadLayer(/Users/someonetoignore/work/neon/neon/test_output/test_ignored_tenant_reattach[mock_s3]/repo/tenants/b933a6c8685e3898383c94bb220779d1/timelines/ccf01aab20e4d8f34145ccf1122a5f19/000000000000000000000000000000000000-FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF__00000000016B59D8-00000000016B5A51, size=Some(25821184)), will retry (attempt 0): Failed to upload a layer from local path '/Users/someonetoignore/work/neon/neon/test_output/test_ignored_tenant_reattach[mock_s3]/repo/tenants/b933a6c8685e3898383c94bb220779d1/timelines/ccf01aab20e4d8f34145ccf1122a5f19/000000000000000000000000000000000000-FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF__00000000016B59D8-00000000016B5A51'

Caused by:
    0: failed to construct request: No credentials in the property bag
    1: No credentials in the property bag
```
[full log](https://github.com/neondatabase/neon/files/10215688/eager_s3_init_pageserver.log)

yet the test itself was working, only the log checking assertion failed.

The failure happened on the very first pageserver start during the test, which is an initial timeline creation.
We shut down pageserver immediately after that happens, so it does not make sense to have the remote storage enabled during the procedure in the first place.

### DIFF
--- a/safekeeper/src/bin/safekeeper.rs
+++ b/safekeeper/src/bin/safekeeper.rs
@@ -318,12 +318,15 @@ fn set_id(workdir: &Path, given_id: Option<NodeId>) -> Result<NodeId> {
 }
 
 // Parse RemoteStorage from TOML table.
-fn parse_remote_storage(storage_conf: &str) -> Result<RemoteStorageConfig> {
+fn parse_remote_storage(storage_conf: &str) -> anyhow::Result<RemoteStorageConfig> {
     // funny toml doesn't consider plain inline table as valid document, so wrap in a key to parse
-    let storage_conf_toml = format!("remote_storage = {}", storage_conf);
+    let storage_conf_toml = format!("remote_storage = {storage_conf}");
     let parsed_toml = storage_conf_toml.parse::<Document>()?; // parse
     let (_, storage_conf_parsed_toml) = parsed_toml.iter().next().unwrap(); // and strip key off again
-    RemoteStorageConfig::from_toml(storage_conf_parsed_toml)
+    RemoteStorageConfig::from_toml(storage_conf_parsed_toml).and_then(|parsed_config| {
+        // XXX: Don't print the original toml here, there might be some sensitive data
+        parsed_config.context("Incorrectly parsed remote storage toml as no remote storage config")
+    })
 }
 
 #[test]
--- a/control_plane/src/background_process.rs
+++ b/control_plane/src/background_process.rs
@@ -51,21 +51,21 @@ pub enum InitialPidFile<'t> {
 }
 
 /// Start a background child process using the parameters given.
-pub fn start_process<
-    F,
-    S: AsRef<OsStr>,
-    EI: IntoIterator<Item = (String, String)>, // Not generic AsRef<OsStr>, otherwise empty `envs` prevents type inference
->(
+pub fn start_process<F, AI, A, EI>(
     process_name: &str,
     datadir: &Path,
     command: &Path,
-    args: &[S],
+    args: AI,
     envs: EI,
     initial_pid_file: InitialPidFile,
     process_status_check: F,
 ) -> anyhow::Result<Child>
 where
     F: Fn() -> anyhow::Result<bool>,
+    AI: IntoIterator<Item = A>,
+    A: AsRef<OsStr>,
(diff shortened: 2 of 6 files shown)


LizardWizzard:
Which test was triggering this issue? Was it with real s3 backend? For running tests we rely on secrets in environment variables which should be present during init

koivunej:
> Which test

On the log it has: `test_ignored_tenant_reattach[mock_s3]` so I am inclined to approve this, though I don't yet see why didn't it fail always. It seems the error is a retryable upload failure.

Aah ok, so it must be that this error doesn't always have time to appear in the logs, so the log error searcher cannot witness it. Ok, makes sense.

safekeeper/src/bin/safekeeper.rs
```diff
@@ -318,12 +318,15 @@ fn set_id(workdir: &Path, given_id: Option<NodeId>) -> Result<NodeId> {
...
+fn parse_remote_storage(storage_conf: &str) -> anyhow::Result<RemoteStorageConfig> {
     // funny toml doesn't consider plain inline table as valid document, so wrap in a key to parse
-    let storage_conf_toml = format!("remote_storage = {}", storage_conf);
+    let storage_conf_toml = format!("remote_storage = {storage_conf}");
     let parsed_toml = storage_conf_toml.parse::<Document>()?; // parse
     let (_, storage_conf_parsed_toml) = parsed_toml.iter().next().unwrap(); // and strip key off again
-    RemoteStorageConfig::from_toml(storage_conf_parsed_toml)
+    RemoteStorageConfig::from_toml(storage_conf_parsed_toml).and_then(|parsed_config| {
+        parsed_config
+            .with_context(|| format!("Incorrectly parsed remote storage toml '{storage_conf_toml}' as no remote storage config"))
```
koivunej, line 328: I wonder if there's a chance of logging some secrets here, should we be using a constant context instead?

koivunej (approved):
I think this makes sense, one question on the error message.

LizardWizzard:
> On the log it has: test_ignored_tenant_reattach[mock_s3]

Oh, thanks, can see it now :) To me this is really weird since mock s3 should be fine with mock credentials. And we should pass envs with credentials to init as well. So its not clear what is causing the issue. Probably I'm fine wirth this solution because it will eliminate the race "whether we were able to upload something during init"

koivunej:
> To me this is really weird since mock s3 should be fine with mock credentials.

I understand that the error in the PR description comes from request sending side, not mock s3, as in the aws library now refuses to send the request if it has no authentication information.

SomeoneToIgnore:
This error is quite hard to reproduce for me too, I've only reliably got it on my Linux machine for some reason.
I also think, we could deal with the "initial/root timeline" concept (as console now requires every tenant to have one) and then rework the code to create the tenant in plain CLI mode, but that has to wait for better design first.

SomeoneToIgnore:
> or it was created in some other commit than this branch.

Yes, the log is older than now indeed and points to some phantom commit, I did not pay attention to that part, sorry.
So I guess you can safely ignore backtrace pointers since they are all wrong now.

LizardWizzard:
> I understand that the error in the PR description comes from request sending side, not mock s3, as in the aws library now refuses to send the request if it has no authentication information.

Yes, but my point is that we still pass keys in the environment variables as usual, so credentials should be in place. Thats why it feels weird to me. If it works in the test then it should work in init too

petuhovskiy (approved):

lubennikovaav (approved):

PR #3080, opened 2022-12-13
Timeline::download_missing: fix handling of mismatched layer size

problame:
Before this patch, when we decide to rename a layer file to backup because of layer file size mismatch, we would not remove the layer from the layer map, but remote the on-disk file.

Because we  re-download the file immediately after, we simply end up with two layer objects in memory that reference the same file in the layer map.  So, GetPage() would work fine until one of the layers gets delete()'d. The other layer's delete() would then fail.

### DIFF
--- a/pageserver/src/tenant/timeline.rs
+++ b/pageserver/src/tenant/timeline.rs
@@ -1013,7 +1013,6 @@ impl Timeline {
         //       1) if there was another pageserver that came and generated new files
         //       2) during attach of a timeline with big history which we currently do not do
         let mut local_only_layers = local_layers;
-        let timeline_dir = self.conf.timeline_path(&self.timeline_id, &self.tenant_id);
         for remote_layer_name in &index_part.timeline_layers {
             let local_layer = local_only_layers.remove(remote_layer_name);
 
@@ -1024,13 +1023,16 @@ impl Timeline {
                 .unwrap_or(LayerFileMetadata::MISSING);
 
             // Is the local layer's size different from the size stored in the
-            // remote index file?
-            // If so, rename_to_backup those files & replace their local layer with
-            // a RemoteLayer so that we re-download them on-demand.
+            // remote index file? If so, rename_to_backup those files & remove
+            // local_layer form the layer map.
+            // We'll download a fresh copy of the layer file below.
             if let Some(local_layer) = local_layer {
                 let local_layer_path = local_layer.local_path();
-                assert!(local_layer_path.exists());
-                assert!(local_layer_path == timeline_dir.join(remote_layer_name.file_name()));
+                ensure!(
+                    local_layer_path.exists(),
+                    "every layer from local_layers must exist on disk: {}",
+                    local_layer_path.display()
+                );
 
                 if let Some(remote_size) = remote_layer_metadata.file_size() {
                     let metadata = local_layer_path.metadata().with_context(|| {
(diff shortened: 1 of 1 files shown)


pageserver/src/tenant/timeline.rs
```diff
@@ -1015,49 +1015,69 @@ impl Timeline {
...
 
-            let local_layer_path = timeline_dir.join(remote_layer_name.file_name());
+            // Is the local layer's size different from the size stored in the
+            // remote index file?
+            // If so, rename_to_backup those files & replace their local layer with
+            // a RemoteLayer so that we re-download them on-demand.
+            if let Some(local_layer) = local_layer {
+                let local_layer_path = local_layer.local_path();
+                assert!(local_layer_path.exists());
+                assert!(local_layer_path == timeline_dir.join(remote_layer_name.file_name()));
```
koivunej, line 1033:
```suggestion
                assert_eq!(local_layer_path, timeline_dir.join(remote_layer_name.file_name()));
```

pageserver/src/tenant/timeline.rs
```diff
@@ -1015,49 +1015,69 @@ impl Timeline {
...
-                info!("remote layer {remote_layer_name:?} does not exist locally");
+                let local_layer_path = timeline_dir.join(remote_layer_name.file_name());
+                assert!(
+                    !local_layer_path.exists(),
+                    "caller did not supply exhaustive list of local_layers: {local_layer_path:?}"
+                );
+                // fall-through to adding the remote layer
             }
 
+            info!(
```
koivunej, line 1076: Assuming you will do the demotions from `info` => `debug` above, and we see only these log lines in the log, not the download progress, I think the outcome will be confusing.

I'd propose adding the next action to this message ("downloading"), and maybe an overall completion message with how many were downloaded before return?

koivunej (approved):
Overall this change looks good for me even though I might not remember the full context. Minor logging related suggestions.

SomeoneToIgnore (approved):
On the other end of this bug, there's `layer_map.rs::insert_historic` method that now does zero checks if duplicate layers are added: that may lead to whatever special effects, if we accidentally misuse this method.

We might want to fix it eventually too.

LizardWizzard (approved):
Looks good to me, plus one for @koivunej thoughts regarding info => debug.

problame:
> On the other end of this bug, there's layer_map.rs::insert_historic method that now does zero checks if duplicate layers are added: that may lead to whatever special effects, if we accidentally misuse this method.
> We might want to fix it eventually too.

Are you referring to, before this PR, if `local_layer_path.exists()` we'd `already_downloaded = true; ...; continue`?

SomeoneToIgnore:
I don't think this PR changes `insert_historic` behavior, but generally, anything, including the bug this PR ends up fixing, ends up bringing extra (duplicate) layers into layer map and we never notice that fast enough.

In your PR, you've mentioned

>  So, GetPage() would work fine until one of the layers gets delete()'d.

to me, we should fail fast the moment such layer is added to the layer map (or, at least log an error and discard the duplicate?). 
`insert_historic` is `pub fn` so it should expect whatever trash input may come, but it does not verify that at all.

But sure I don't require your PR to fix this bug, just a general note that we can make such errors more obvious and eager to appear in a relatively simple manner.

problame:
> to me, we should never get in such state, rather failing fast at the moment such layer is added to the layer map (or, at least log an error and discard the duplicate?).

Agree.

PR #3090, opened 2022-12-13
Remove large database files after every test suite

SomeoneToIgnore:
Closes https://github.com/neondatabase/neon/issues/1984 
Closes https://github.com/neondatabase/neon/pull/2830

A follow-up of https://github.com/neondatabase/neon/pull/2830, I've noticed that benchmarks failed again due to out of space issues.

Removes most of the pageserver and safekeeper files from disk after every pytest suite run.


test_runner/fixtures/neon_fixtures.py
```diff
@@ -2756,6 +2779,25 @@ def test_output_dir(request: FixtureRequest, top_output_dir: Path) -> Iterator[P
...
+                test_file = test_entry
+                if not ATTACHMENT_NAME_REGEX.fullmatch(
+                    test_file.name
+                ) and not SMALL_DB_FILE_NAME_REGEX.fullmatch(test_file.name):
+                    log.debug(f"Removing large database {test_file} file")
+                    test_file.unlink()
+            elif test_entry.is_dir():
+                directories_to_clean.append(test_entry)
+
+        while len(directories_to_clean) > 0:
```
vadim2404, line 2795: NIT:

```python
while directories_to_clean:
```

test_runner/fixtures/neon_fixtures.py
```diff
@@ -2756,6 +2779,25 @@ def test_output_dir(request: FixtureRequest, top_output_dir: Path) -> Iterator[P
...
+            directory_to_clean = directories_to_clean.pop()
+            if len(os.listdir(directory_to_clean)) == 0:
```
vadim2404, line 2797: NIT

```python
if not os.listdir(directory_to_clean):
```

vadim2404 (approved):

vadim2404 (approved):

test_runner/fixtures/neon_fixtures.py
```diff
@@ -718,6 +727,26 @@ def enable_real_s3_remote_storage(self, test_name: str, force_enable: bool = Tru
...
+                    if not ATTACHMENT_NAME_REGEX.fullmatch(
+                        test_file.name
+                    ) and not SMALL_DB_FILE_NAME_REGEX.fullmatch(test_file.name):
+                        log.debug(f"Removing large database {test_file} file")
+                        test_file.unlink()
+                elif test_entry.is_dir():
+                    directories_to_clean.append(test_entry)
+
+            while directories_to_clean:
+                directory_to_clean = directories_to_clean.pop()
```
bayandin, line 745:
```suggestion
            for directory_to_clean in directories_to_clean:
```
I think we don't delete the object itself, so it should be ok (but I haven't tried it)

bayandin (approved):
Nice! Thanks for adding this!

Just a couple of python nitpicking; feel free to mix and match whatever you like

test_runner/fixtures/neon_fixtures.py
SomeoneToIgnore: I think I will ignore this one since: we have to remove the directories in the reverse order, to remove the empty deepest ones first, potentially making their parent dirs also empty.

bayandin: Ah, sorry, I missed this part. Maybe then:

```python
for directory_to_clean in directories_to_clean.reverse():
```

SomeoneToIgnore: Sure, thank you for the review!

PR #3094, opened 2022-12-14
always remove RemoteTimelineClient's metrics when dropping it

problame:
I need this to unblock https://github.com/neondatabase/neon/pull/3087

### DIFF
--- a/pageserver/src/metrics.rs
+++ b/pageserver/src/metrics.rs
@@ -201,7 +201,7 @@ pub static NUM_ONDISK_LAYERS: Lazy<IntGauge> = Lazy::new(|| {
 
 // remote storage metrics
 
-pub static REMOTE_UPLOAD_QUEUE_UNFINISHED_TASKS: Lazy<IntGaugeVec> = Lazy::new(|| {
+static REMOTE_UPLOAD_QUEUE_UNFINISHED_TASKS: Lazy<IntGaugeVec> = Lazy::new(|| {
     register_int_gauge_vec!(
         "pageserver_remote_upload_queue_unfinished_tasks",
         "Number of tasks in the upload queue that are not finished yet.",
@@ -210,14 +210,14 @@ pub static REMOTE_UPLOAD_QUEUE_UNFINISHED_TASKS: Lazy<IntGaugeVec> = Lazy::new(|
     .expect("failed to define a metric")
 });
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum RemoteOpKind {
     Upload,
     Download,
     Delete,
 }
 impl RemoteOpKind {
-    pub fn as_str(&self) -> &str {
+    pub fn as_str(&self) -> &'static str {
         match self {
             Self::Upload => "upload",
             Self::Download => "download",
@@ -226,13 +226,13 @@ impl RemoteOpKind {
     }
 }
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq)]
 pub enum RemoteOpFileKind {
     Layer,
     Index,
 }
 impl RemoteOpFileKind {
-    pub fn as_str(&self) -> &str {
+    pub fn as_str(&self) -> &'static str {
         match self {
             Self::Layer => "layer",
             Self::Index => "index",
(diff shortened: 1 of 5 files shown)


test_runner/regress/test_tenants.py
```diff
@@ -157,9 +157,20 @@ def test_metrics_normal_work(neon_env_builder: NeonEnvBuilder):
         )
 
 
-def test_pageserver_metrics_removed_after_detach(neon_env_builder: NeonEnvBuilder):
+@pytest.mark.parametrize(
+    "remote_storage_kind",
+    [RemoteStorageKind.NOOP, RemoteStorageKind.LOCAL_FS, RemoteStorageKind.MOCK_S3],
+)
```
koivunej, line 163: I wonder if this is really required. Maybe with noop and mock_s3 but not all three? I assume that it's currently still behind that Option whether there is a remote or not, unsure if the implementation has advanced.

koivunej (approved):
If the tests pass I think this looks good, question on if all 3 remote storages are needed.

koivunej:
Two familiar sounding failures in one of the regress-tests:

```
FAILED test_runner/regress/test_remote_storage.py::test_remote_storage_upload_queue_retries[local_fs]
FAILED test_runner/regress/test_remote_storage.py::test_timeline_deletion_with_files_stuck_in_upload_queue[local_fs]
```

test_runner/regress/test_tenants.py
problame: We definitely need `NOOP` and some other impl. I guess `LOCAL_FS` vs `MOCK_S3` makes no idfference at the storage_sync2 level.
Which one should I remove, then?

koivunej: I cannot remember what is the difference between mock_s3 and local_fs, maybe @SomeoneToIgnore has some insights?

SomeoneToIgnore: mock_s3 is a Python lib that is started in a background: http://docs.getmoto.org/en/latest/docs/server_mode.html#non-python-sdk-s-server-mode

local_fs is just a FS directory somewhere

pageserver/src/metrics.rs
```diff
@@ -491,10 +491,87 @@ pub fn remove_tenant_metrics(tenant_id: &TenantId) {
 
 use futures::Future;
 use pin_project_lite::pin_project;
+use std::collections::HashMap;
 use std::pin::Pin;
+use std::sync::{Arc, Mutex};
 use std::task::{Context, Poll};
 use std::time::Instant;
 
+#[derive(Default)]
```
problame, line 500: This one is wrong. We need to set `tenant_id` and `timeline_id`. Will post a fix

pageserver/src/metrics.rs
```diff
@@ -491,10 +491,94 @@ pub fn remove_tenant_metrics(tenant_id: &TenantId) {
...
+    }
+    pub fn remote_operation_time(
+        &self,
+        file_kind: &RemoteOpFileKind,
+        op_kind: &RemoteOpKind,
+        status: &'static str,
+    ) -> Histogram {
+        // XXX would be nice to have an upgradable RwLock
+        let mut guard = self.remote_operation_time.lock().unwrap();
+        let key = (file_kind.as_str(), op_kind.as_str(), status);
```
SomeoneToIgnore, line 524: Minor:

Is there any point to form the key out of `str` not the `Copy` enums that are already present (`status` could be also replaced with an enum)?
`(&'static str, &'static str, &'static str)` would become more self-documentable there (ideally, there could be a separate struct with names created)

SomeoneToIgnore (approved):

pageserver/src/metrics.rs
problame: I'm hoping that in the future we can find some zero-cost way to curry `get_metric_with_label_values`. 
That would certainly be using `str`.
All of this code could be deleted, then, and I feel like it's more obvious to a future if the keys are  just `str` tuples.

See  https://neondb.slack.com/archives/C0277TKAJCA/p1671015002722739

SomeoneToIgnore: > zero-cost
> That would certainly be using `str`.

Not quite sure I agree, but ok, let's see how it gets transformed later.

test_runner/regress/test_remote_storage.py
```diff
@@ -436,8 +436,8 @@ def assert_compacted_and_uploads_queued():
 
     assert not timeline_path.exists()
 
-    # timeline deletion should kill ongoing uploads
-    assert get_queued_count(file_kind="index", op_kind="upload") == 0
+    # timeline deletion should kill ongoing uploads, so, the metric will be gone
+    assert get_queued_count(file_kind="index", op_kind="upload") is None
```
lubennikovaav, line 440: `get_queued_count` returns int()
so this needs some fix

I'd expect that it will fail on `assert matches` (line 256), but it isn't. 
Looks like we don't drop the metric properly

lubennikovaav (changes requested):

test_runner/regress/test_remote_storage.py
problame: > get_queued_count returns int()
> so this needs some fix

Right.

> I'd expect that it will fail on assert matches (line 256), but it isn't.
> Looks like we don't drop the metric properly

It's a different nested function definition.

One is in line 249.
Another is in line 380.

Different test cases as well.

For the one in 249, we don't detach the tenant delete the timeline, so, `!= None` is expected.

For the one in 380, we call it once while the timeline is there and once after we've deleted the timeline.
So, it needs to be able to handle both.

Will fix it.

Issue #3104, opened 2022-12-14
misleading `Task was joined more than once` message


problame:
Test run https://neon-github-public-dev.s3.amazonaws.com/reports/pr-3103/debug/3697230857/index.html#suites/82004ab4e3720b47bf78f312dabe7c55/db4fe5159005a401/ produced unexpected flaky error message 

```
/github/home/.cache/pypoetry/virtualenvs/neon-_pxWMzVK-py3.9/lib/python3.9/site-packages/allure_commons/_allure.py:200: in __call__
    return self._fixture_function(*args, **kwargs)
test_runner/fixtures/neon_fixtures.py:1095: in neon_env_builder
    yield builder
test_runner/fixtures/neon_fixtures.py:834: in __exit__
    self.env.pageserver.assert_no_errors()
test_runner/fixtures/neon_fixtures.py:1904: in assert_no_errors
    assert not errors
E   AssertionError: assert not ['2022-12-14T17:53:37.859939Z ERROR wal_connection_manager{tenant=87deb874f2406f433f5b420c072aa8a8 timeline=0009e8b8ba380ab176a47603d9b3319f}: wal receiver task finished with an error: Task was joined more than once\n']
```

Looking at the code:

https://github.com/neondatabase/neon/blob/32662ff1c42a1f36001eddc9fdf60a087fc92e71/pageserver/src/walreceiver.rs#L125-L146

There is only one call site, inside a `tokio::select!()` statement.

If the `jh.await` gets cancelled, then we've `take()`n the join handle but dropped the result on the floor.

I guess this is yet another annoyance of cancel-by-drop.

PR #3105, opened 2022-12-14
make TaskHandle::next_task_event cancellation-safe

problame:
If we get cancelled before jh.await returns we've take()n the join handle but
drop the result on the floor.

Fix it by setting self.join_handle = None after the .await

fixes https://github.com/neondatabase/neon/issues/3104


koivunej (approved):
Looks good to me, great find.

koivunej:
Weird ci failure, I'll push rerun all jobs.

Issue #3109, opened 2022-12-14
`basebackup timeout waiting for LSN X` _after_ `caught up with LSN X`


problame:
In prod, I found this weird case where we log that we caught up with LSN X and several seconds later, we log a basebackup error caused by timeout while waiting for WAL record at LSN X.

Later, basebackup for same tenant & timeline succeeded.

So, wondering what's going on there.

---

https://neonprod.grafana.net/goto/gUCFWcc4k?orgId=1

```
2022-12-14T21:50:57.675129Z  INFO walreceiver_connection{id=XTENANT/YTIMELINE}: caught up at LSN 0/1F98CA0
2022-12-14T21:50:57.680955Z  INFO walreceiver_connection{id=XTENANT/YTIMELINE}:initial_logical_size_calculation{tenant=XTENANT timeline=YTIMELINE}: Successfully calculated initial logical size
2022-12-14T21:50:59.920669Z  INFO remote_upload{tenant=XTENANT timeline=YTIMELINE upload_task_id=1}: remote task UploadMetadata(lsn: 0/1F958D0) completed successfully
2022-12-14T21:51:00.655897Z ERROR query handler for 'basebackup XTENANT YTIMELINE 0/1F98CA0' failed: Timed out while waiting for WAL record at LSN 0/1F98CA0 to arrive, last_record_lsn 0/1F958D0 disk consistent LSN=0/1F958D0
Caused by:
    SeqWaitError
Stack backtrace:
   0: pageserver::tenant::timeline::Timeline::wait_lsn::{{closure}}
             at /home/nonroot/pageserver/src/tenant/timeline.rs:470:9
      <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/core/src/future/mod.rs:91:19
   1: pageserver::page_service::PageServerHandler::handle_basebackup_request::{{closure}}::{{closure}}
             at /home/nonroot/pageserver/src/page_service.rs:634:35
      <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/core/src/future/mod.rs:91:19
   2: <tracing::instrument::Instrumented<T> as core::future::future::Future>::poll
             at /home/nonroot/.cargo/registry/src/github.com-1ecc6299db9ec823/tracing-0.1.37/src/instrument.rs:272:9
[...]

PR #3110, opened 2022-12-14
seqwait: different error messages per variant

problame:
Would have been handy to get slightly more details in
https://github.com/neondatabase/neon/issues/3109

refs https://github.com/neondatabase/neon/issues/3109


koivunej (approved):
Changes look good but what's the rustfmt going on about... Ah, it was missing docs lint by clippy.

PR #3113, opened 2022-12-15
Hotfix: terminate subscription if channel is full.

arssher:


petuhovskiy (approved):
WDYT about merging to main, and then merging main to release? Non-equal release and main branches can cause problems in the future.

storage_broker/src/bin/storage_broker.rs
```diff
@@ -369,7 +369,9 @@ impl BrokerService for Broker {
                     Err(RecvError::Lagged(skipped_msg)) => {
                         missed_msgs += skipped_msg;
                         if let Poll::Ready(_) = futures::poll!(Box::pin(warn_interval.tick())) {
-                            warn!("dropped {} messages, channel is full", missed_msgs);
+                            warn!("subscription id={}, key={:?}, addr={:?} dropped {} messages, channel is full",
+                            subscriber.id, subscriber.key, subscriber.remote_addr, missed_msgs);
+                            Err(Status::new(Code::Internal, "full channel"))?;
                             missed_msgs = 0;
```
SomeoneToIgnore, line 375: Looks a bit odd, let's improve it a bit, before merging?

* `missed_msgs = 0;` after the "return" code looks odd
* `Err(..)?;` looks like a super implicit return, let's write `return Err();` instead
* `subscriber.id, subscriber. ...` like could use an extra indent

SomeoneToIgnore (approved):

storage_broker/src/bin/storage_broker.rs
LizardWizzard: Plus one for 1 and 3

arssher: It's a hotfix, not intended to stay long.

2 doesn't work in try_stream.

arssher:
I rebased on main, but then encountered this
https://github.com/neondatabase/cloud/wiki/Release:-general-flow#hotfixes

Moreover, in this particular case I'd fix prod and left staging for some time broken for easier investigation. WDYT?

storage_broker/src/bin/storage_broker.rs
```diff
@@ -369,8 +369,9 @@ impl BrokerService for Broker {
                     Err(RecvError::Lagged(skipped_msg)) => {
                         missed_msgs += skipped_msg;
                         if let Poll::Ready(_) = futures::poll!(Box::pin(warn_interval.tick())) {
-                            warn!("dropped {} messages, channel is full", missed_msgs);
-                            missed_msgs = 0;
+                            warn!("subscription id={}, key={:?}, addr={:?} dropped {} messages, channel is full",
+                                subscriber.id, subscriber.key, subscriber.remote_addr, missed_msgs);
+                            Err(Status::new(Code::Internal, "full channel"))?;
```
koivunej, line 374: This message seems wrong, since the subscriber is moving too slowly compared to the publisher.

I am also a bit unsure what does the `futures::poll!(Box::pin(warn_interval.tick()))` do, since it feels like it's used as a debounce to lagged events but will it not get reset if the channel goes back to normal..

koivunej: As such I worry that this might be hiding some loop which is causing new messages to be produced way too fast.

Regardless, originally I had the idea when suggesting the channel type that the messages are okay to lose, as in, the later message will always contain the superset of information from the older. As such, the broadcast channel buffer size could be small.

Issue #3118, opened 2022-12-15
Remove IAM at its current form from pageserver


SomeoneToIgnore:
IAM risks becoming a bottleneck soon after the on-demand functionality will land.

* Pageserver background jobs (size calculations, GC, compaction) may leak S3 quotas and require a few S3 connections per timeline per second.
Example: https://github.com/neondatabase/neon/pull/3111

* on-demand downloads will happen spontaneously per timeline, also requiring S3 quotas

IAM is able to serve ~200 RPS: https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/UsingWithRDS.IAMDBAuth.html
While regular S3 allows ~3500 PUT/COPY/POST/DELETE or 5500 GET/HEAD S3 requests **per prefix**: https://aws.amazon.com/premiumsupport/knowledge-center/s3-request-limit-avoid-throttling/

Subtle things like https://github.com/neondatabase/neon/issues/3006 happen due to this difference and will happen more as we start to query S3 more randomly.
We would need to use a better approach to credential management.

zoete:
Can we use an approach to get and use temporary security credentials through AWS STS?

LizardWizzard:
If it takes time to set it up wdyt about switching to static keys now and add rotation to them as a separate step?

koivunej:
Is this still relevant, after caching the credentials?

SomeoneToIgnore:
We still have IAM with 10 time stricter limits that real S3 has, also feels interesting to try using less middlemen services in the storage path.

The priority had decreased indeed though.

kelvich:
I think we can close this for now

PR #3128, opened 2022-12-16
fix(remote_storage): use cached credentials

koivunej:
IMDSv2 has limits, and if we query it on every s3 interaction we are going to go over those limits. Changes the s3_bucket client configuration to use:
- ChainCredentialsProvider to handle env variables or imds usage
- LazyCachingCredentialsProvider to actually cache any credentials

Related: https://github.com/awslabs/aws-sdk-rust/issues/629
Possibly related: https://github.com/neondatabase/neon/issues/3118

### DIFF
--- a/libs/remote_storage/src/s3_bucket.rs
+++ b/libs/remote_storage/src/s3_bucket.rs
@@ -122,30 +122,18 @@ impl S3Bucket {
             "Creating s3 remote storage for S3 bucket {}",
             aws_config.bucket_name
         );
-        let mut config_builder = Config::builder()
-            .region(Region::new(aws_config.bucket_region.clone()))
-            .credentials_provider(provide_credentials_fn(|| async {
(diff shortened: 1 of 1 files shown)


LizardWizzard (approved):

koivunej:
I think we still need the LazyCaching whatever per: https://github.com/awslabs/aws-sdk-rust/discussions/627#discussioncomment-3807017

problame (changes requested):
My understanding is that `ImdsCredentialsProvider` doesn't cache internally.
Need something like `LazyCachingCredentialsProvider` ?

LizardWizzard:
> My understanding is that ImdsCredentialsProvider doesn't cache internally.

You're right, I read it wrong

vadim2404 (approved):

koivunej:
Now without clippy warnigns and the chain + lazy.

problame (approved):

LizardWizzard (approved):

PR #3137, opened 2022-12-16
Rename ZENITH_AUTH_TOKEN to NEON_AUTH_TOKEN

yeputons:
Follow-up of #2455

There are no other relevant instances of `ZENITH_AUTH_TOKEN`/`AUTH_TOKEN` in our neon/cloud/website repositories. 

However, there may be some hard-coded instances in configuration files (esp. generated with `neon_local`) or whatnot, so people may have to update their installations. As to our prod/staging components:

* Pageserver is configured via `neon.pageserver_connstring` — mentioned in the cloud repository, but it does not pass any password/token, so I conclude we do not do auth with pageserver https://github.com/neondatabase/cloud/blob/5b2a3fa19031172ebd2edd100c161e7efcf750d6/goapp/internal/postgressettings/settings.go#L168 and https://github.com/neondatabase/cloud/blob/5b2a3fa19031172ebd2edd100c161e7efcf750d6/goapp/internal/dto/pageserver.go#L30-L32
* Safekeeper auth were not enabled anywhere in staging/prod either as far as I know
* Broker auth is not implemented yet
* Compute auth is completely separate

### DIFF
--- a/control_plane/src/compute.rs
+++ b/control_plane/src/compute.rs
@@ -201,7 +201,7 @@ impl PostgresNode {
             .stderr(Stdio::piped());
 
         if let Some(token) = auth_token {
-            cmd.env("ZENITH_AUTH_TOKEN", token);
+            cmd.env("NEON_AUTH_TOKEN", token);
         }
 
         let sync_handle = cmd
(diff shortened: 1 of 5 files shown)


SomeoneToIgnore (approved):

yeputons:
@MMeent I made Pageserver read from both the old env variable and the new one. See details in the commit message: 51eaf8007a13b08a567f12b4cda463d9c0b6f188 . All other places have been reading the environment variable's name from a configuration already and did not have code changes, so nothing should break as long as the configuration and the running script stay in sync.

The only place with a hard-coded environment variable name is `control_plane`, but it's for `neon_local` only, as far as I understand. I could not find any mentions of `ZENITH_AUTH_TOKEN` in our `cloud` repository, so I assume it's just never passed.

PR #3135, opened 2022-12-16
test_runner/performance: fix flush for NeonCompare

bayandin:
Fix performance tests:
```
AttributeError: 'NeonCompare' object has no attribute 'pageserver_http'
```

https://neon-github-public-dev.s3.amazonaws.com/reports/main/release/3713487584/index.html#categories/dcb49b9953ce76b182e54e96e1454755/b68ee72a188c0b38/

Issue #2866, opened 2022-11-19
test_runner: use `log.info("%s", user_id)` instead of `log.info(f"{user_id}")`; use f-strings instead of `.format(`


yeputons:
Quoting https://github.com/neondatabase/neon/pull/2062/files#r917153823:

> This method works better for integration with various logging middleware. For example if you use Sentry it will be able do dedup events by taking into account only format string, not the parameters. If parameters are pre filled like you proposed then it wont be able to do so and will create separate tickets for every combination of user_id/request_id/whatever that came with the event. It is highly unlikely that we'll need to use something like Sentry in tests, but this is the way that I used to cook python logging just to adhere to least surprise principle

Similarly with `.format(` calls. We have quite a few of both interpolation ways.

While we are here, we may as well replace `.format` with f-strings.

Currently, we use f-strings a lot more frequently.

PR #3142, opened 2022-12-17
mypy: enable check_untyped_defs

bayandin:
Enable `check_untyped_defs` and fix warnings.

This should prevent cases like https://github.com/neondatabase/neon/pull/3135

### DIFF
--- a/scripts/export_import_between_pageservers.py
+++ b/scripts/export_import_between_pageservers.py
@@ -448,15 +448,15 @@ def add_missing_rels(base_tar, output_tar, log_dir, pg_bin, tmp_pg_port: int):
 
 
 def get_rlsn(pageserver_connstr, tenant_id, timeline_id):
-    conn = psycopg2.connect(pageserver_connstr)
-    conn.autocommit = True
-    with closing(conn.cursor()) as cur:
-        cmd = f"get_last_record_rlsn {tenant_id} {timeline_id}"
-        cur.execute(cmd)
-        res = cur.fetchone()
-        assert res is not None
-        prev_lsn = res[0]
-        last_lsn = res[1]
+    with closing(psycopg2.connect(pageserver_connstr)) as conn:
+        conn.autocommit = True
+        with conn.cursor() as cur:
+            cmd = f"get_last_record_rlsn {tenant_id} {timeline_id}"
+            cur.execute(cmd)
+            res = cur.fetchone()
+            assert res is not None
+            prev_lsn = res[0]
+            last_lsn = res[1]
 
     return last_lsn, prev_lsn
 
--- a/scripts/export_import_between_pageservers.py
+++ b/scripts/export_import_between_pageservers.py
@@ -450,13 +450,13 @@ def add_missing_rels(base_tar, output_tar, log_dir, pg_bin, tmp_pg_port: int):
 def get_rlsn(pageserver_connstr, tenant_id, timeline_id):
     conn = psycopg2.connect(pageserver_connstr)
     conn.autocommit = True
-    with conn.cursor() as cur:
+    with closing(conn.cursor()) as cur:
         cmd = f"get_last_record_rlsn {tenant_id} {timeline_id}"
         cur.execute(cmd)
         res = cur.fetchone()
+        assert res is not None
         prev_lsn = res[0]
         last_lsn = res[1]
-    conn.close()
 
     return last_lsn, prev_lsn
 
--- a/test_runner/fixtures/neon_fixtures.py
+++ b/test_runner/fixtures/neon_fixtures.py
@@ -623,6 +623,7 @@ class NeonEnvBuilder:
         return self.env
 
     def start(self):
+        assert self.env is not None, "environment is not already initialized, call init() first"
         self.env.start()
 
     def init_start(self) -> NeonEnv:
@@ -747,7 +748,7 @@ class NeonEnvBuilder:
     def cleanup_remote_storage(self):
         # here wee check for true remote storage, no the local one
         # local cleanup is not needed after test because in ci all env will be destroyed anyway
-        if self.remote_storage_prefix is None:
+        if not isinstance(self.remote_storage, S3Storage) or self.remote_storage_prefix is None:
             log.info("no remote storage was set up, skipping cleanup")
             return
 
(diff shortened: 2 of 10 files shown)


test_runner/fixtures/neon_fixtures.py
```diff
@@ -786,16 +788,17 @@ def cleanup_remote_storage(self):
...
         # flush rest
         if len(objects_to_delete["Objects"]):
             self.remote_storage_client.delete_objects(
-                Bucket=self.remote_storage.bucket_name, Delete=objects_to_delete
+                Bucket=self.remote_storage.bucket_name,
+                Delete=objects_to_delete,
             )
 
-        log.info("deleted %s objects from remote storage", cnt)
+        log.info(f"deleted {cnt} objects from remote storage")
```
yeputons, line 801: It's better to leave it as is: https://github.com/neondatabase/neon/issues/2866

test_runner/regress/test_compute_ctl.py
```diff
@@ -193,8 +193,9 @@ def test_sync_safekeepers_logs(neon_env_builder: NeonEnvBuilder, pg_bin: PgBin):
             timeout=10,
         )
     except TimeoutExpired as exc:
-        ctl_logs = exc.stderr.decode("utf-8")
-        log.info("compute_ctl output:\n" + ctl_logs)
+        if exc.stderr is not None:
```
yeputons, line 196: Shouldn't this be an `assert`? Output is probably always captured by `raw_cli`.

If not, I'd use `(exc.stderr or b"").decode(...)` to make sure log line is always there.

test_runner/fixtures/neon_fixtures.py
```diff
@@ -771,7 +772,8 @@ def cleanup_remote_storage(self):
             Prefix=self.remote_storage_prefix,
         )
 
-        objects_to_delete = {"Objects": []}
+        # Using Any because DeleteTypeDef (from boto3-stubs) doesn't fit our case
+        objects_to_delete: Any = {"Objects": []}
```
yeputons, line 776: `dict[str, Any]` is probably better

test_runner/fixtures/neon_fixtures.py
```diff
@@ -752,7 +753,7 @@ def cleanup_local_storage(self):
     def cleanup_remote_storage(self):
         # here wee check for true remote storage, no the local one
         # local cleanup is not needed after test because in ci all env will be destroyed anyway
-        if self.remote_storage_prefix is None:
+        if not isinstance(self.remote_storage, S3Storage) or self.remote_storage_prefix is None:
```
yeputons, line 756: Shouldn't this `isinstance` be an assertion? If `remote_storage` is not S3, something is probably off.

scripts/export_import_between_pageservers.py
```diff
@@ -450,13 +450,13 @@ def add_missing_rels(base_tar, output_tar, log_dir, pg_bin, tmp_pg_port: int):
 def get_rlsn(pageserver_connstr, tenant_id, timeline_id):
     conn = psycopg2.connect(pageserver_connstr)
     conn.autocommit = True
-    with conn.cursor() as cur:
+    with closing(conn.cursor()) as cur:
```
yeputons, line 453: That should work without `closing`, there are plenty of other places like this. Would you mind double-checking?

yeputons (dismissed):

scripts/export_import_between_pageservers.py
vadim2404:
```python
with psycopg2.connect(pageserver_connstr) as conn:
  conn.autocommit = True
  with conn.cursor() as cur:
    ...
```

I think it should be sth like this

test_runner/fixtures/neon_fixtures.py
koivunej: Was trying to query @LizardWizzard about this, and I hope we can still use f-strings in tests, as we don't have any middleware looking at the log statements.

test_runner/regress/test_compute_ctl.py
```diff
@@ -240,7 +241,7 @@ def __exit__(self, _type, _value, _traceback):
         with self.pid_file:
             try:
                 os.kill(self.pid, signal.SIGTERM)
-            except os.OsError as e:
+            except OSError as e:
```
koivunej, line 244: I ended up adding that `os.` because my python lsp insisted wasn't finding OSError without it, even though I understood from the docs it should be "in the implicit prelude". Is this following some direct change of something else in this PR, like mypy checking more things?

bayandin: Yep `mypy` started to complain about it with `check_untyped_defs`:
```
test_runner/regress/test_compute_ctl.py:244: error: Module has no attribute "OsError"; maybe "OSError" or "_OnError"?
Found 1 error in 1 file (checked 103 source files)
```

test_runner/fixtures/neon_fixtures.py
bayandin: We call `cleanup_remote_storage` unconditionally  `NeonEnvBuilder.__exit__`

bayandin: Actually, since `self.remote_storage_prefix` is coupled with `S3Storage`, I've moved to `isinstance(self.remote_storage, S3Storage)` to and assert

test_runner/regress/test_compute_ctl.py
bayandin: Also, I'm not sure what `os.OsError` is:
```
$ python3.9 -c "import os; print(os.OsError)"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
AttributeError: module 'os' has no attribute 'OsError'
```
```
$ python3.9 -c "import os; print(os.OSError)"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
AttributeError: module 'os' has no attribute 'OSError'
```
```
$ python3.9 -c "import os; print(OSError)"
<class 'OSError'>
```

test_runner/fixtures/neon_fixtures.py
LizardWizzard: Summary:
I would like to keep it the same everywhere. Keeping it as is is one possible way.

I'm not strongly opposed to fstring usage. Would be good to confirm that this solution is preferable for others as well given the limitations.

bayandin:
```bash
git grep -E 'log\.(debug|info|error|warn)\("' -- test_runner/  | grep '%s' | wc -l
      11
```

```bash
git grep -E 'log\.(debug|info|error|warn)\(f"' -- test_runner/ | wc -l
     140
```

With this rough estimation, we're closer to using f-strings everywhere for logging

test_runner/regress/test_compute_ctl.py
vadim2404: in last example you don't need to `import os`. `OSError` is a part of `builtins`

koivunej (approved):
Looking good, thanks for fixing these.

Issue #3115, opened 2022-12-15
Create a timeline subcommand to print its walreceiver state


SomeoneToIgnore:
If walreceiver is stuck due to some reason, it might be simpler to send an HTTP GET query into pageserver management API instead of checking the logs for error messages.

New HTTP endpoint should return:
* all safekeepers whose messages were recently received from the broker
* Safekeepers' metadata, related to timeline: LSNs, last message time, etc.
* current active walreceiver connection metadata: SK id, time since connected, etc. or time since the last connection happened
* latest feedback message metadata? Latest lsn caught up to would be useful

SomeoneToIgnore:
Might be more generic to expose the data via new metric(s).

koivunej:
We've been happy with the summary of different safekeepers in the seqwait timeout error. But this would be easy to add for querying via http, maybe it could even be in the `timeline_detail` because you'll probably want to see `last_record_lsn` and other lsn's at the same time...?

With "summary of different safekeepers" I mean: https://github.com/neondatabase/neon/blob/a1d6b1a4af5252e1215571257d4e79e40d961c11/pageserver/src/tenant/timeline.rs#L641-L658

Added in #4090. So the same contents, not necessarily formatted to a string but as a json would be handy as `GET /v1/tenant/:tenant_id/timeline/:timeline_id/walreceiver` for example, or as part of `GET /v1/tenant/:tenant_id/timeline/:timeline_id`.

Adding a good_first_issue because this is mostly plumbing in what we already have, perhaps adding a bit of `serde::Serialize` and openapi changes. However the openapi changes can be opaque, as in this is something internal to help with the on-call operator, not something to be relied for stability.

PR #3147, opened 2022-12-19
Tidy up walreceiver logs

SomeoneToIgnore:
Closes https://github.com/neondatabase/neon/issues/3114

### DIFF
--- a/pageserver/src/walreceiver/connection_manager.rs
+++ b/pageserver/src/walreceiver/connection_manager.rs
@@ -502,13 +502,18 @@ impl WalreceiverState {
 
     /// Adds another broker timeline into the state, if its more recent than the one already added there for the same key.
     fn register_timeline_update(&mut self, timeline_update: SafekeeperTimelineInfo) {
-        self.wal_stream_candidates.insert(
-            NodeId(timeline_update.safekeeper_id),
+        let new_safekeeper_id = NodeId(timeline_update.safekeeper_id);
+        let old_entry = self.wal_stream_candidates.insert(
+            new_safekeeper_id,
             BrokerSkTimeline {
                 timeline: timeline_update,
                 latest_update: Utc::now().naive_utc(),
             },
         );
+
+        if old_entry.is_none() {
+            info!("New SK node was added: {new_safekeeper_id}");
+        }
     }
 
     /// Cleans up stale broker records and checks the rest for the new connection candidate.
@@ -735,12 +740,13 @@ impl WalreceiverState {
     /// Remove candidates which haven't sent broker updates for a while.
     fn cleanup_old_candidates(&mut self) {
         let mut node_ids_to_remove = Vec::with_capacity(self.wal_stream_candidates.len());
+        let lagging_wal_timeout = self.lagging_wal_timeout;
 
         self.wal_stream_candidates.retain(|node_id, broker_info| {
             if let Ok(time_since_latest_broker_update) =
                 (Utc::now().naive_utc() - broker_info.latest_update).to_std()
             {
-                let should_retain = time_since_latest_broker_update < self.lagging_wal_timeout;
+                let should_retain = time_since_latest_broker_update < lagging_wal_timeout;
                 if !should_retain {
                     node_ids_to_remove.push(*node_id);
                 }
@@ -750,8 +756,11 @@ impl WalreceiverState {
             }
         });
 
-        for node_id in node_ids_to_remove {
-            self.wal_connection_retries.remove(&node_id);
+        if !node_ids_to_remove.is_empty() {
+            info!("Safekeeper nodes {node_ids_to_remove:?} did not send events for over {lagging_wal_timeout:?}, stopping their retries.");
+            for node_id in node_ids_to_remove {
+                self.wal_connection_retries.remove(&node_id);
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


pageserver/src/walreceiver/connection_manager.rs
```diff
@@ -95,6 +95,8 @@ pub fn spawn_connection_manager_task(
     );
 }
 
+const MAX_BROKER_IDLE_WAIT: Duration = Duration::from_secs(90);
```
SomeoneToIgnore, line 98: https://neon-github-public-dev.s3.amazonaws.com/reports/pr-3147/debug/3733221379/index.html#suites/158be07438eb5188d40b466b6acfaeb3/85f48b6cc4d58c58/

There are surprisingly many tests that manage to get idle during tests. 90 seconds produces two tests in debug only, but when I use 60 seconds, I get 7 tests in Debug runs:

https://neon-github-public-dev.s3.amazonaws.com/reports/pr-3147/debug/3731383425/index.html#suites/3de20ced56df482626cff043a3f2c817

What would be the realistic value, if the PR's approach is correct?

Generally, I'm not sure about the `WARN` message level too: IIRC, safekeepers should have the heuristics to stop sending WAL events if no new data arrives. 
So, maybe turn this to `INFO` and go on?

pageserver/src/walreceiver/connection_manager.rs
```diff
@@ -735,8 +756,11 @@ impl WalreceiverState {
             }
         });
 
-        for node_id in node_ids_to_remove {
-            self.wal_connection_retries.remove(&node_id);
+        if !node_ids_to_remove.is_empty() {
+            info!("Safekeeper nodes {node_ids_to_remove:?} did not send events for over {lagging_wal_timeout:?}, stopping their retries.");
+            for node_id in node_ids_to_remove {
+                self.wal_connection_retries.remove(&node_id);
+            }
```
problame, line 763: Would it be better to put the `info!()` log line inside the loop, one line per node id that is removed?
Makes the log more uniform.

pageserver/src/walreceiver/walreceiver_connection.rs
```diff
@@ -184,7 +184,20 @@ pub async fn handle_walreceiver_connection(
...
+        let replication_message = match replication_message {
+            Ok(message) => message,
+            Err(replication_error) => {
+                if replication_error.is_closed() {
+                    info!("Replication stream got closed");
+                    return Ok(());
+                } else {
+                    return Err(anyhow::anyhow!(
+                        "Replication stream error: {replication_error}"
+                    ));
```
problame, line 196: By printing replication_error's `Display` value we lose a lot of detail here.
Is that intentional?
Why not use `.context()?`

problame (approved):
What motivated the increase of the warning timeout?

SomeoneToIgnore:
> What motivated the increase of the warning timeout?

https://github.com/neondatabase/neon/pull/3147#discussion_r1052437318 gathering data for discussion, welcome.

pageserver/src/walreceiver/walreceiver_connection.rs
SomeoneToIgnore: Direct `.context` invocations are not working on that error, because it does not implement `Display`, but I've wrapped it differently, should be better now.

pageserver/src/walreceiver/connection_manager.rs
petuhovskiy: It's actually ok to have no new events (idle) from the broker when compute is down. That means that current approach can spam with `Got no broker or walreceiver events ...` in the logs, because we always have thousands of timelines with no running compute.

PR's approach does not feel right. IMO would be better to put more walreceiver context into `wait_lsn` instead, e.g. we can print latest broker event and full walreceiver status on WAL timeout error.

SomeoneToIgnore: > better to put more walreceiver context into `wait_lsn` instead

Agree that we should do it, planned to do in a separate issue here: https://github.com/neondatabase/neon/issues/2106

What I wanted to have with this log is to catch slow walreceiver reconnects in our tests.
IMO, there's definitely something abnormal: we don't shut down *all* SK nodes for 60 seconds in 7 tests, yet we cannot get broker messages there for some reason.

I'll think of it more, will remove the logging and will try to add some metric instead (and, maybe, debug logging).

petuhovskiy: I tried to run one failed test locally, and it seems we have tests with a root timeline created by `env = neon_env_builder.init_start()`, which is never used after that. We do that in all failed tests, and that timeline has a spawned walreceiver with no activity (no one starts a compute for it). Without a compute safekeepers don't even know about that timeline and don't push anything to the broker.

It looks like these tests fail because they run longer than MAX_BROKER_IDLE_WAIT.

SomeoneToIgnore: Yes, just realized the same when discussed the issue recently, thank you for confirming.
Thinking of it more, I agree now that we should remove the related commit entirely and work on https://github.com/neondatabase/neon/issues/2106 (https://github.com/neondatabase/neon/issues/3115 later, maybe metrics too?) to get the required info, including the "time spent without etcd events" one.

SomeoneToIgnore: Have pushed another commit, that removes this logging entirely, added a few `debug!` logs in case that place would need it later.

Issue #3088, opened 2022-12-13
neon_local start fails to start safekeeper


lubennikovaav:
## Steps to reproduce

```
./target/debug/neon_local init
./target/debug/neon_local start 
```

## Expected result
all components start successfully

## Actual result
`neon_local start` fails with 
```
...
Starting safekeeper at '127.0.0.1:5454' in '.neon/safekeepers/sk1'.....
safekeeper 1 has not started yet, continuing to wait.....
safekeeper '1' start failed: safekeeper 1 did not start in 10 seconds
Stopping storage_broker with pid 99551 immediately.......
storage_broker has not stopped yet, continuing to wait.....
neon broker stop failed: storage_broker with pid 99551 did not stop in 10 seconds
```
head  .neon/safekeepers/sk1/safekeeper\ 1.log
```
2022-12-13T18:17:28.646485Z  INFO version: git:0bc488b7234924c2a4e505374710a79cd67a9725
Error: No such file or directory (os error 2)

Stack backtrace:
   0: <core::result::Result<T,F> as core::ops::try_trait::FromResidual<core::result::Result<core::convert::Infallible,E>>>::from_residual
             at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/core/src/result.rs:2105:27
      safekeeper::set_id
             at /home/anastasia/zenith/safekeeper/src/bin/safekeeper.rs:307:29
      safekeeper::main
             at /home/anastasia/zenith/safekeeper/src/bin/safekeeper.rs:133:14
```

## Environment
my laptop

yeputons:
Offending commit: 32662ff1c42a1f36001eddc9fdf60a087fc92e71 from #2733 . It fails for me even with `init`:

```
yeputons@yeputons06:~/neon/neon$ ./target/debug/neon_local init
Starting pageserver at '127.0.0.1:64000' in '.neon'.
pageserver started, pid: 17141
Failed to create initial tenant and timeline for pageserver: Error: Error: initdb failed: 'could not identify current directory: No such file or directory
could not identify current directory: No such file or directory
could not identify current directory: No such file or directory
initdb: error: The program "postgres" is needed by initdb but was not found in the
same directory as "initdb".
Check your installation.
'
pageserver init failed: pageserver init failed with exit code Some(1)
```

yeputons:
Ah, not really. The message above is an error I get if I try to `init` while another pageserver is already running and occupies a port. Weird by itself, but it's a separate issue.

[...]

PR #3150, opened 2022-12-19
Fix neon_local start and stop commands

SomeoneToIgnore:
Fixes https://github.com/neondatabase/neon/issues/3088

* Adds a test to ensure `neon_local` `start` and `stop` commands work. The nodes are started on the default ports, but all other nodes in the suites are started on different ports thanks to `PortDistributor`.
The test turned out to be useless for reproducing the current case, since created the config for the nodes, which contained absolute path to the test directory. Still, the test seems useful to keep it.

* Fixes an issue when pageserver was not killed after it got started and safekeeper had failed to start.
For me on Mac, the stopping code fails to get the event that shows that the process does not run, but the processes are actually stopped. The issue is not related to this PR hence not fixed.

* Canonicalize workdir in safekeeper to ensure that it exists and avoid issues with startup: pageserver did that, but safekeeper did not and when no workdir was picked, used a relative path when created an id file:
```
Error: Failed to create id file at ".neon/safekeepers/sk1/safekeeper.id"

Caused by:
    No such file or directory (os error 2)
```
It was not visible which file was causing the issue, so some more context was added to errors.

### DIFF
--- a/control_plane/src/bin/neon_local.rs
+++ b/control_plane/src/bin/neon_local.rs
@@ -900,6 +900,7 @@ fn cli() -> Command {
     let stop_mode_arg = Arg::new("stop-mode")
         .short('m')
         .value_parser(["fast", "immediate"])
+        .default_value("fast")
         .help("If 'immediate', don't flush repository data at shutdown")
         .required(false)
         .value_name("stop-mode");
--- a/control_plane/src/bin/neon_local.rs
+++ b/control_plane/src/bin/neon_local.rs
@@ -747,7 +747,7 @@ fn get_safekeeper(env: &local_env::LocalEnv, id: NodeId) -> Result<SafekeeperNod
     if let Some(node) = env.safekeepers.iter().find(|node| node.id == id) {
         Ok(SafekeeperNode::from_env(env, node))
     } else {
-        bail!("could not find safekeeper '{}'", id)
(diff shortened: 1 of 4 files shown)


lubennikovaav (approved):
thank you for fixing!

control_plane/src/bin/neon_local.rs
```diff
@@ -806,22 +806,22 @@ fn handle_safekeeper(sub_match: &ArgMatches, env: &local_env::LocalEnv) -> Resul
...
     // Postgres nodes are not started automatically
 
+    broker::start_broker_process(env)?;
+
+    let pageserver = PageServerNode::from_env(env);
     if let Err(e) = pageserver.start(&pageserver_config_overrides(sub_match)) {
-        eprintln!("pageserver start failed: {e}");
-        try_stop_storage_broker_process(env);
+        eprintln!("pageserver {} start failed: {:#}", env.pageserver.id, e);
+        try_stop_all(env, true);
```
petuhovskiy, line 816: `try_stop_all(..., true)` here and `try_stop_all(..., false)` later (line 824), not sure if it's intended.

petuhovskiy (approved):

LizardWizzard (approved):
Thanks!

PR #3158, opened 2022-12-20
Use local brokers

SomeoneToIgnore:
I have a few tests that fail on my other PR, example: https://neon-github-public-dev.s3.amazonaws.com/reports/pr-3147/debug/3733759250/index.html#categories/ea3d534fbd71268d72e33bcb1850351d/bcbd4e38aa093fc7/

Yet, this test does not have any broker logs: 
image

This is due to broker being shared between tests as a "singleton", having its logs in `test_output/neon/` directory.
Instead, this PR isolates every test broker and puts its log right into the test directory.

### DIFF
--- a/test_runner/fixtures/neon_fixtures.py
+++ b/test_runner/fixtures/neon_fixtures.py
@@ -286,24 +286,19 @@ def port_distributor(worker_base_port: int) -> PortDistributor:
     return PortDistributor(base_port=worker_base_port, port_number=WORKER_PORT_NUM)
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="function")
 def default_broker(
-    request: FixtureRequest,
     port_distributor: PortDistributor,
-    top_output_dir: Path,
+    test_output_dir: Path,
     neon_binpath: Path,
 ) -> Iterator[NeonBroker]:
     # multiple pytest sessions could get launched in parallel, get them different ports/datadirs
     client_port = port_distributor.get_port()
-    broker_logfile = (
-        get_test_output_dir(request, top_output_dir) / f"storage_broker_{client_port}.log"
-    )
-    broker_logfile.parents[0].mkdir(exist_ok=True, parents=True)
+    broker_logfile = test_output_dir / "repo" / "storage_broker.log"
 
     broker = NeonBroker(logfile=broker_logfile, port=client_port, neon_binpath=neon_binpath)
     yield broker
     broker.stop()
-    allure_attach_from_dir(Path(broker_logfile))
 
 
 @pytest.fixture(scope="session")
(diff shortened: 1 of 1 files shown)


test_runner/fixtures/neon_fixtures.py
```diff
@@ -295,10 +295,7 @@ def default_broker(
 ) -> Iterator[NeonBroker]:
     # multiple pytest sessions could get launched in parallel, get them different ports/datadirs
     client_port = port_distributor.get_port()
-    broker_logfile = (
-        get_test_output_dir(request, top_output_dir) / f"storage_broker_{client_port}.log"
-    )
-    broker_logfile.parents[0].mkdir(exist_ok=True, parents=True)
+    broker_logfile = get_test_repo_dir(request, top_output_dir) / "storage_broker.log"
```
bayandin, line 298: We can replace the usage of `top_output_dir` session fixture with `test_output_dir` function fixture:

```suggestion
    broker_logfile = test_output_dir / "storage_broker.log"
```

And the logs we don't need `allure_attach_from_dir(Path(broker_logfile))` — logs will be picked up automatically by `test_output_dir` fixture

arssher (approved):
We could alternatively just fix `allure_attach_from_dir` call and have it shared like previously, but per test startup/teardown shouldn't take noticeable time and might make debugging slightly easier, so I'm fine with either variant.

test_runner/fixtures/neon_fixtures.py
SomeoneToIgnore: This way, the logs will be outside of the repo dir.
Guess it's fine, will leave it as is for now.

image

bayandin: Feel free to change it to 
```suggestion
    broker_logfile = test_output_dir / "repo" / "storage_broker.log"
```
😄

SomeoneToIgnore: I wanted to avoid this particular hardcode, but if recommended, then sure, will do 😄

test_runner/fixtures/neon_fixtures.py
```diff
@@ -286,19 +286,15 @@ def port_distributor(worker_base_port: int) -> PortDistributor:
     return PortDistributor(base_port=worker_base_port, port_number=WORKER_PORT_NUM)
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="function")
 def default_broker(
-    request: FixtureRequest,
     port_distributor: PortDistributor,
-    top_output_dir: Path,
+    test_output_dir: Path,
```
bayandin, line 292: And now we can delete `allure_attach_from_dir(Path(broker_logfile))` at the end of the method

bayandin (approved):

PR #3164, opened 2022-12-20
Add zenith-1-sk-4

petuhovskiy:


LizardWizzard:
Since the PR is missing description. Can you elaborate why we're doing this? Is it a new safekeeper in old region? Or whats the idea? I dont have context

petuhovskiy:
We want to migrate zenith-1-sk-3 to a new instance, see https://neondb.slack.com/archives/C039YKBRZB4/p1671271014292269 for more context.

PR #3167, opened 2022-12-21
Collect core dump backtraces in compute_ctl.

kelvich:
Scan core dumps directory on exit. In case of existing core dumps call gdb/lldb to get a backtrace and log it. By default look for core dumps in postgres data directory as core.<pid>. That is how core collection is configured in our k8s nodes (and a reasonable convention in general).

### DIFF
--- a/compute_tools/src/compute.rs
+++ b/compute_tools/src/compute.rs
@@ -23,7 +23,7 @@ use std::sync::RwLock;
 
 use anyhow::{Context, Result};
 use chrono::{DateTime, Utc};
-use log::info;
+use log::{info, warn};
 use postgres::{Client, NoTls};
 use serde::{Serialize, Serializer};
 
@@ -328,6 +328,9 @@ impl ComputeNode {
             .wait()
             .expect("failed to start waiting on Postgres process");
 
+        self.check_for_core_dumps()
+            .expect("failed to check for core dumps");
+
         Ok(ecode)
     }
 
@@ -342,5 +345,69 @@ impl ComputeNode {
 
         self.prepare_pgdata()?;
         self.run()
+    }
+
+    // Look for core dumps and collect backtraces.
+    //
+    // EKS worker nodes have following core dump settings:
+    //   /proc/sys/kernel/core_pattern -> core
+    //   /proc/sys/kernel/core_uses_pid -> 1
+    //   ulimint -c -> unlimited
+    // which results in core dumps being written to postgres data directory as core.<pid>.
+    //
+    // Use that as a default location and pattern, except macos where core dumps are written
+    // to /cores/ directory by default.
+    fn check_for_core_dumps(&self) -> Result<()> {
+        let core_dump_dir = match std::env::consts::OS {
+            "macos" => Path::new("/cores/"),
+            _ => Path::new(&self.pgdata),
+        };
+
+        // Collect core dump paths if any
+        info!("checking for core dumps in {}", core_dump_dir.display());
+        let files = fs::read_dir(core_dump_dir)?;
+        let cores = files.filter_map(|entry| {
+            let entry = entry.ok()?;
+            let _ = entry.file_name().to_str()?.strip_prefix("core.")?;
+            Some(entry.path())
+        });
+
+        // Print backtrace for each core dump
+        for core_path in cores {
+            warn!(
+                "core dump found: {}, collecting backtrace",
+                core_path.display()
+            );
+
+            // Try first with gdb
+            let backtrace = Command::new("gdb")
+                .args(["--batch", "-q", "-ex", "bt", &self.pgbin])
+                .arg(&core_path)
+                .output();
+
+            // Try lldb if no gdb is found -- that is handy for local testing on macOS
+            let backtrace = match backtrace {
+                Err(ref e) if e.kind() == std::io::ErrorKind::NotFound => {
+                    warn!("cannot find gdb, trying lldb");
+                    Command::new("lldb")
+                        .arg("-c")
+                        .arg(&core_path)
+                        .args(["--batch", "-o", "bt all", "-o", "quit"])
+                        .output()
+                }
+                _ => backtrace,
+            }?;
+
+            warn!(
+                "core dump backtrace: {}",
+                String::from_utf8_lossy(&backtrace.stdout)
+            );
+            warn!(
+                "debugger stderr: {}",
+                String::from_utf8_lossy(&backtrace.stderr)
+            );
+        }
+
+        Ok(())
     }
 }
--- a/Dockerfile.compute-node-v14
+++ b/Dockerfile.compute-node-v14
@@ -207,7 +207,8 @@ RUN apt update &&  \
         libgeos-c1v5 \
         libgdal28 \
         libproj19 \
-        libprotobuf-c1 && \
+        libprotobuf-c1 \
+        gdb && \
     rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 USER postgres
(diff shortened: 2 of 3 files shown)


compute_tools/src/compute.rs
```diff
@@ -343,4 +345,71 @@ impl ComputeNode {
...
+    //   ulimint -c -> unlimited
+    // which results in core dumps being written to postgres data directory as core.<pid>.
+    //
+    // For local testing I also added support for lldb, which writes core dumps to /cores.
+    fn check_for_core_dumps(&self) {
+        let core_dump_dir = if std::env::consts::OS == "macos" {
+            std::path::Path::new("/cores/")
+        } else {
+            std::path::Path::new(&self.pgdata)
+        };
```
funbringer, line 363:
```rust
let dir = match std::env::consts::OS {
  "macos" => Path::new("/cores/"),
  _ => Path::new(&self.pgdata),
}
```

compute_tools/src/compute.rs
```diff
@@ -343,4 +345,71 @@ impl ComputeNode {
...
+                                "-o",
+                                "thread backtrace all",
+                                "-o",
+                                "quit",
+                            ])
+                            .output()
+                            .expect("cannot start lldb process")
+                    } else {
+                        Command::new("gdb")
+                            .args(&["--batch", "-q", &self.pgbin, &full_core_path, "-ex", "bt"])
```
funbringer, line 398: I think file should be passed via `--core`.

funbringer:
Also worth mentioning that it's possible to use [pipe core pattern](https://man7.org/linux/man-pages/man5/core.5.html) (see _Piping core dumps to a program_) to run an arbitrary program for core dump collection. This means that we could write a small tool for capturing backtraces and sending them via http or by any other means to a dedicated endpoint.

There are caveats, though.

For instance, it's impossible to override `core_pattern` in a container-based runtime (I mean, _it may be possible_, but you wouldn't want that as it'd simultaneously change host's pattern as well). If we specified pipe pattern in the host kernel params, the resulting coredumps would most probably be captured by the host, __not containers__ (however, that in of itself might be ok). VM-based runtimes are not affected, though.

compute_tools/src/compute.rs
```diff
@@ -343,4 +345,71 @@ impl ComputeNode {
...
+                .file_name()
+            {
+                let name = name.to_string_lossy();
+                if name.starts_with("core.") {
+                    warn!("core dump found: {}, collecting backtrace", name);
+                    let full_core_path = std::path::Path::new(core_dump_dir).join(name.as_ref());
+                    let full_core_path = full_core_path.to_string_lossy();
+
+                    let backtrace = if std::env::consts::OS == "macos" {
+                        Command::new("lldb")
```
funbringer, line 384: Another option would be to try using both, one after another. There's gdb for mac and lldb for linux.

kelvich: Was thinking of that too, but didn't come up with a nice way to test if debugger is available or not. There is https://crates.io/crates/which but that is a bit too heavyweight.

compute_tools/src/compute.rs
kelvich: Fixed, thanks

compute_tools/src/compute.rs
kelvich:
```
# echo 'int main(){*(volatile int *)0=42;}' > x.c
# gcc -g x.c
# ./a.out 
Segmentation fault (core dumped)
# gdb --batch -q ./a.out ./core.3996 -ex bt
[New LWP 3996]
Core was generated by `./a.out'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x0000000000400480 in main () at x.c:1
1	int main(){*(volatile int *)0=42;}
#0  0x0000000000400480 in main () at x.c:1
```

man also mentions syntax `gdb [OPTIONS] [prog|prog procID|prog core] `

compute_tools/src/compute.rs
funbringer: Or you could ignore error kind completely and just first try lldb and then gdb and then bail with error.

compute_tools/src/compute.rs
```diff
@@ -343,4 +345,70 @@ impl ComputeNode {
...
+            if let Some(name) = path
+                .expect("can't read pgdata directory entry")
+                .path()
+                .file_name()
+            {
+                let name = name.to_string_lossy();
+                if name.starts_with("core.") {
+                    warn!("core dump found: {}, collecting backtrace", name);
+                    let full_core_path = std::path::Path::new(core_dump_dir).join(name.as_ref());
+                    let full_core_path = full_core_path.to_string_lossy();
```
funbringer, line 380: At your discretion:

```rust
let files = fs::read_dir("/tmp")?;
let cores = files.filter_map(|x| {
    let entry = x.ok()?;
    let _ = entry.path().to_str()?.prefix("core.")?;
    Some(entry)
});

for entry in cores {
  let path = entry.path();
  ...
}
```

compute_tools/src/compute.rs
```diff
@@ -343,4 +345,70 @@ impl ComputeNode {
...
+
+        info!(
+            "checking for core dumps in {}",
+            core_dump_dir.to_string_lossy()
+        );
+
+        let paths = fs::read_dir(core_dump_dir).expect("cannot list pgdata for core dumps");
+        for path in paths {
+            if let Some(name) = path
+                .expect("can't read pgdata directory entry")
```
funbringer, line 372: It might be `/cores`, though.

funbringer (approved):
This should do the job. There are comments, but feel free to ignore them if you don't feel like fixing them.

compute_tools/src/compute.rs
kelvich: > Oh, you don't have to! We could just catch execve's error:

Nice, rewrote it this way

compute_tools/src/compute.rs
kelvich: Thanks, I didn't like that part as well. Will rewrite.

compute_tools/src/compute.rs
```diff
@@ -343,4 +346,67 @@ impl ComputeNode {
...
+
+        // Print backtrace for each core dump
+        for core_path in cores {
+            let core_path = core_path.to_string_lossy();
+
+            warn!("core dump found: {}, collecting backtrace", core_path);
+
+            // Try first with gdb
+            let backtrace = Command::new("gdb")
+                .args(["--batch", "-q", &self.pgbin, &core_path, "-ex", "bt"])
```
funbringer, line 386:
```rust
Command::new("gdb")
  .args(["--batch", "--quiet", "-ex", "bt", &self.pgbin])
  .arg(core_path) // this might have a different type!
```

compute_tools/src/compute.rs
```diff
@@ -343,4 +346,67 @@ impl ComputeNode {
...
+            let entry = entry.ok()?;
+            let _ = entry.file_name().to_str()?.strip_prefix("core.")?;
+            Some(entry.path())
+        });
+
+        // Print backtrace for each core dump
+        for core_path in cores {
+            let core_path = core_path.to_string_lossy();
+
+            warn!("core dump found: {}, collecting backtrace", core_path);
```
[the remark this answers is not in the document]
kelvich: 👌

compute_tools/src/compute.rs
kelvich: > // this might have a different type!

hm, how so? Now it is `Cow<str>`, is there any problem with that?

funbringer: This is to say that you don't have to convert Path to anything, because it's possible to construct command using objects of different types.

kelvich: Gotcha, fixed

compute_tools/src/compute.rs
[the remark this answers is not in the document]
kelvich: Ok, used `.display()` instead of `to_string_lossy()`

funbringer (approved):

PR #3178, opened 2022-12-21
Fix 1.66 Clippy warnings

SomeoneToIgnore:
1.66 release speeds up compile times for over 10% according to tests in https://neondb.slack.com/archives/C033RQ5SPDH/p1671653357785879

Also its Clippy finds plenty of old nits in our code:
* useless conversion, `foo as u8` where `foo: u8` and similar, removed `as u8` and similar
* useless references and dereferenced (that were automatically adjusted by the compiler), removed various `&` and `*`
* bool -> u8 conversion via `if/else`, changed to `u8::from`
* Map `.iter()` calls where only values were used, changed to `.values()` instead

Standing out lints:
* `Eq` is missing in our protoc generated structs. Silenced, does not seem crucial for us.
* `fn default` looks like the one from `Default` trait, so I've implemented that instead and replaced the `dummy_*` method in tests with `::default()` invocation
* Clippy detected that
```
if retry_attempt < u32::MAX {
    retry_attempt += 1;
}
```
is a saturating add and proposed to replace it.


pageserver/src/tenant_config.rs
```diff
@@ -220,29 +219,4 @@ impl TenantConf {
...
-        TenantConf {
-            checkpoint_distance: defaults::DEFAULT_CHECKPOINT_DISTANCE,
-            checkpoint_timeout: Duration::from_secs(600),
-            compaction_target_size: 4 * 1024 * 1024,
-            compaction_period: Duration::from_secs(10),
-            compaction_threshold: defaults::DEFAULT_COMPACTION_THRESHOLD,
-            gc_horizon: defaults::DEFAULT_GC_HORIZON,
-            gc_period: Duration::from_secs(10),
-            image_creation_threshold: defaults::DEFAULT_IMAGE_CREATION_THRESHOLD,
-            pitr_interval: Duration::from_secs(60 * 60),
```
funbringer, line 234: изображение

But seriously, `pitr_interval` used to be 60 mins, and now it's `DEFAULT_PITR_INTERVAL`, which is

```rust
pub const DEFAULT_PITR_INTERVAL: &str = "30 days";
```

right?

funbringer (commented):
Thanks a lot!

pageserver/src/tenant_config.rs
SomeoneToIgnore: These values were not considered (by me, at least) for a while.
I've removed the ones used in unit tests and the default ones are now used everywhere, so technically I've not changed anything with this PR, since the tests are green.

Yet, I agree that we should adjust those.

funbringer: Cool, I just wanted to double-check! :D

funbringer (approved):

PR #3187, opened 2022-12-22
Enable billing metric_collection_endpoint on staging

lubennikovaav:


vadim2404 (approved):

cicdteam:
What about other staging regions? What will be If that param absent there?

lubennikovaav:
> What about other staging regions?

Other regions are the `staging.hosts.yaml` and `staging.eu-west-1.hosts.yaml`, right? 
I can add same changes there.

> What will be If that param absent there?

If the endopoint setting is absent, the metric sending is disabled.

Issue #2760, opened 2022-11-04
Epic: Improve LastWrittenLsn mechanism


MMeent:
## Motivation

Currently, we have a limit of N=128 block ranges (or 1028 blocks after #2493) with LRU eviction. This is OK, but we could evict many pages at once, resulting in a relatively high overall evicted LSN.

Because a high LastWrittenLSN means a short WAL propagation time to PageServer, it is better to have the lowest possible (within reason) LastWrittenLSN (or: the lowest reasonably possible LSN in the request). 
Even for pages that haven't been updated in a long time and where the WAL propagation delay is not an issue, a lower LastWrittenLsn will result in higher overall performance because we are able to exclude more layers from the search for WAL in the Pageserver.

### Pairingheap (?) cache eviction

We could change our cache eviction policy to one based on the lowest LSN in the cache, e.g., through a pairing heap on the lowest LSN.
This way, we could have 

```
overallLastWrittenLsn - lowest granularity LSN for all blocks not in the rest of the infra
recentlyEvicted - N entries associated with BufferTag
currentLsn - max LSN we can expect in any block

overallLastWrittenLsn < recentlyEvicted (N entries) < currentLsn
```

This results in more precise LSN lookups for pages where it matters - those with a low LSN.

Note that we use this LSN stuff only to ensure that Pageserver is up-to-date enough, so it makes sense to keep granular data for only the most recently updated evicted pages.

### Probabilistic lastWrittenLsn

We could update our LSN cache to one based on count-min sketches: we maintain N buckets, and for all evicted pages, we hash the page ID to correlate to M (M < N) buckets and update those buckets' lastWrittenLsn value to Max(currentVal, pageLsn). 
When looking up the lastWrittenLsn for a page, we do Min(associated buckets' lastWrittenLsn), which is the actual last written LSN for that page. The benefit is a better estimated last written LSN in less memory than a hash map on all pages (no hashmap-related maintenance infrastructure necessary), and it covers the whole database.
I'll refer to "count-min sketches" and "Counting bloom filters" for more details on why this would work, though here we use "Max(insert_value, current_value)" instead of "current_value++" in the buckets.

This could replace only XLogCtl->maxLastWrittenLsn, or both XLogCtl->maxLastWrittenLsn and the lastWrittenLsnCache. I don't care much about either, but precise data (see Pairingheap cache eviction above) could be useful as a first item, after which this probabilistic structure could provide the next, more granular layer of LSNs.

[...]

PR #3193, opened 2022-12-22
Extract LwLsn cache into extension

MMeent:
Only a few direct references to the existence of the cache remain. Most synchronizations can be replaced by utilizing smgrimmedsync calls to update the LwLsn of the relation to ProcLastRecPtr, which generally is "good enough".

### DIFF
--- a/pgxn/neon/neon_lwlc.c
+++ b/pgxn/neon/neon_lwlc.c
@@ -53,12 +53,10 @@ typedef struct LwLsnCacheEntryData {
 	LwLsnCacheEntryKeyData key;
 	XLogRecPtr	lsn;
 	dlist_node	node;
-	pairingheap_node pnode;
 } LwLsnCacheEntryData;
 
 typedef struct LwLsnCacheData {
 	dlist_head	cache_head;
-	pairingheap	pheap;
 	bool		initialized;
 	int			n_cached_entries;
 	volatile XLogRecPtr	highWaterMark;
@@ -205,13 +203,13 @@ lwlc_setup_shmem(void)
 
 
 	dlist_init(&LwLsnCache->cache_head);
-	/*
-	 * XXX: Manual modification of these fields, because no shared allocation
-	 * method exists for pheap.
-	 */
-	LwLsnCache->pheap.ph_compare = &lwlc_pheap_comparefunc;
-	LwLsnCache->pheap.ph_arg = NULL;
-	LwLsnCache->pheap.ph_root = NULL;
+//	/*
+//	 * XXX: Manual modification of these fields, because no shared allocation
+//	 * method exists for pheap.
+//	 */
+//	LwLsnCache->pheap.ph_compare = &lwlc_pheap_comparefunc;
+//	LwLsnCache->pheap.ph_arg = NULL;
+//	LwLsnCache->pheap.ph_root = NULL;
 	LwLsnCache->n_cached_entries = 0;
 
 	/*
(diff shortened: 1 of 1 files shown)


hlinnaka (commented):
Didn't review in detail, just looked at a high level.

Note: This doesn't merely move code around! This also changes the eviction policy, from LRU to evicting the entry with oldest LSN. In essence, this implements the "pairing heap" idea from https://github.com/neondatabase/neon/issues/2760. I think that's a good idea, and I'm pleasantly surprised how small that change was, but please call it out in the commit message.

Since this changes the behaviour, got performance tests or diagnostics or something to demonstrate the benefit? My gut feeling is that if the cache is large enough, it won't make much difference in practice as the LSN used in getpage requests will be "good enough" anyway. But potentially this let's us get away with a smaller cache.

knizhnik:
> Didn't review in detail, just looked at a high level.
> 
> Note: This doesn't merely move code around! This also changes the eviction policy, from LRU to evicting the entry with oldest LSN. In essence, this implements the "pairing heap" idea from #2760. I think that's a good idea, and I'm pleasantly surprised how small that change was, but please call it out in the commit message.
> 
> Since this changes the behaviour, got performance tests or diagnostics or something to demonstrate the benefit? My gut feeling is that if the cache is large enough, it won't make much difference in practice as the LSN used in getpage requests will be "good enough" anyway. But potentially this let's us get away with a smaller cache.

It was my actually feedback on #2760: I do not think that it is good idea to true to optimize something without proving that it is a bottleneck: without any test demonstrating the problem or benchmark illustrating advantage if new approach.
LRU is IMHO in this case should provide quite good approximation of LSN order (actually I can not imaging situation when it will swap out entries with smaller LSN).  `pairingheap_add` is fast but still much slower than `dlist_push_tail` (just one VAX instruction:). And I am +1 for "pleasantly surprised how small that change was", but still it is ~10 times more lines of code than maintaining LRU list.

So while moving lat written LSN cache to neon extension is definitely right thing, I am not so sure about replacinf LRU with pairing heap. And as far as I understand it will not solve the problem with vacuum+prefetch.

pgxn/neon/pagestore_smgr.c
```diff
@@ -1703,10 +1845,10 @@ neon_extend(SMgrRelation reln, ForkNumber forkNum, BlockNumber blkno,
 	 */
 	if (lsn == InvalidXLogRecPtr)
 	{
-		lsn = GetXLogInsertRecPtr();
-		SetLastWrittenLSNForBlock(lsn, reln->smgr_rnode.node, forkNum, blkno);
+		lsn = ProcLastRecPtr;
```
knizhnik, line 1848: I wonder if using ProcLastRecPtr is correct here?
Assume that some other backend B2 has done something with this relation (for example also extend it).
The this backend B1 is calling smgrextend. But please notice that at this moment it has not yet write something to the WAL, so it's  `ProcLastRecPtr`can be far in the past and smaller than LSN of pages appended to the relation by backend B2.

pgxn/neon/neon_lwlc.c
```diff
@@ -0,0 +1,657 @@
...
+static XLogRecPtr lwlc_lookup_last_lsn(LwLsnCacheRequest lookup_type,
+									   RelFileNode node, ForkNumber fork,
+									   BlockNumber blkno,
+									   XLogRecPtr *effective);
+static bool lwlc_should_insert(XLogRecPtr lsn);
+static void lwlc_insert_last_lsn(XLogRecPtr lsn,
+								 RelFileNode node, ForkNumber fork,
+								 BlockNumber min_blkno, BlockNumber max_blockno);
+
+xlog_set_lwlf_block_hook_type		prev_xlog_set_lwlf_block_hook = NULL;
```
knizhnik, line 111: Will not it be better to have just one hook and pass  LwLsnCacheRequest to it?

pgxn/neon/neon_lwlc.c
```diff
@@ -0,0 +1,657 @@
...
+									HASH_FIND,
+									NULL);
+
+	switch (lookup_type) {
+	case LWLSN_BLOCK:
+		MemSet(&key, 0, sizeof(LwLsnCacheEntryKeyData));
+		key.blockNum = blkno;
+		key.rnode = node;
+		key.forkNum = fork;
+		entry = (LwLsnCacheEntryData *)
```
knizhnik, line 341: Should we perform this lookup after switch to avoid code duplication: all switch branches contain the same call of hash_search_with_hash_value

pgxn/neon/neon_lwlc.c
```diff
@@ -0,0 +1,657 @@
...
+	LwLsnCacheEntryKeyData key;
+	XLogRecPtr	lsn;
+	dlist_node	node;
+} LwLsnCacheEntryData;
+
+typedef struct LwLsnCacheData {
+	dlist_head	cache_head;
+	bool		initialized;
+	int			n_cached_entries;
+	volatile XLogRecPtr	highWaterMark;
```
knizhnik, line 62: Can you please add description of this high/low watermark?

pgxn/neon/pagestore_smgr.c
```diff
@@ -1377,22 +1381,90 @@ nm_adjust_lsn(XLogRecPtr lsn)
  * Return LSN for requesting pages and number of blocks from page server
  */
 static XLogRecPtr
-neon_get_request_lsn(bool *latest, RelFileNode rnode, ForkNumber forknum, BlockNumber blkno)
```
knizhnik, line 1380: I wonder if it is strictly needed to add separate get_XXX_request_lsn functions for all kind of requests?
Them contains mostly the same code, just calling different GetLastWrittenLsnFor...  functions.
But may be once again use wLsnCacheRequest parameter?

LizardWizzard:
As discussed moving to draft

PR #3206, opened 2022-12-27
Run Python tests in 8 threads

SomeoneToIgnore:
I have experimented with the runner threads number, and looks like 8 threads win us a few seconds.
[Before](https://github.com/neondatabase/neon/actions/runs/3801158803) (4 threads): 
image

[After](https://github.com/neondatabase/neon/actions/runs/3801868792) (8 threads):
image

Bumping the thread count more did not improve the situation much:
* 20 threads were not allowed by pytest
* 16 threads were flacking quite notably

My guess would be that all pageservers, safekeepers, and other nodes we start occupy quite much of the CPU and other resources to make this approach more scalable.

### DIFF
--- a/.github/actions/run-python-test-set/action.yml
+++ b/.github/actions/run-python-test-set/action.yml
@@ -123,8 +123,8 @@ runs:
           exit 1
         fi
         if [[ "${{ inputs.run_in_parallel }}" == "true" ]]; then
-          # -n4 uses four processes to run tests via pytest-xdist
-          EXTRA_PARAMS="-n4 $EXTRA_PARAMS"
+          # -n8 uses eight processes to run tests via pytest-xdist
+          EXTRA_PARAMS="-n8 $EXTRA_PARAMS"
 
           # --dist=loadgroup points tests marked with @pytest.mark.xdist_group
           # to the same worker to make @pytest.mark.order work with xdist


.github/actions/run-python-test-set/action.yml
```diff
@@ -123,8 +123,8 @@ runs:
           exit 1
         fi
         if [[ "${{ inputs.run_in_parallel }}" == "true" ]]; then
-          # -n4 uses four processes to run tests via pytest-xdist
-          EXTRA_PARAMS="-n4 $EXTRA_PARAMS"
+          # -n6 uses four processes to run tests via pytest-xdist
```
lubennikovaav, line 126:
```suggestion
          # -n6 uses six processes to run tests via pytest-xdist
```

SomeoneToIgnore: Thank you, still experimenting with the values and gathering numbers.
I will also try running 20 and see the time difference, maybe the bottlenecks will be more visible.

SomeoneToIgnore: So far:
* 20 was not allowed as too big value
* 16 always has some flacky tests
* 8 and 6 have a few seconds increase, I will run them more to see which is less flacky

lubennikovaav (approved):

PR #3211, opened 2022-12-27
[proxy] Add more labels to the pricing metrics

funbringer:


proxy/src/console/messages.rs
```diff
@@ -0,0 +1,190 @@
...
+impl KickSession<'_> {
+    fn parse_db_info<'de, D>(des: D) -> Result<DatabaseInfo, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        #[derive(Deserialize)]
+        enum Wrapper {
+            // Currently, console only reports `Success`.
+            // `Failure(String)` used to be here... RIP.
+            Success(DatabaseInfo),
```
LizardWizzard, line 54: Looks ok to me. FYI @neondatabase/console

LizardWizzard (approved):
LGTM

proxy/src/console/messages.rs
```diff
@@ -0,0 +1,190 @@
+use serde::Deserialize;
+use std::fmt;
+
+/// Generic error response with human-readable description.
+/// Note that we can't always present it to user as is.
+#[derive(Debug, Deserialize)]
+pub struct ConsoleError {
+    pub error: Box<str>,
```
antonyc, line 8: Is this code hand written or autogenerated? Are there code generation tools for OpenAPI spec 3.0 in rust? There is at least [openapi-generator](https://www.twilio.com/docs/openapi/generating-a-rust-client-for-twilios-api) which is not a good solution, being in Java. But at least it allows to not write tests on JSON deserialization.

proxy/src/console/messages.rs
antonyc: Not sure what "Currently, console only reports `Success`." means. The console can respond [both with error and with successful responses](https://github.com/neondatabase/cloud/blob/360de29761e0e00976f5bd449409adbd843fd627/openapispec/v2/management-v2.yaml#L263-L271) in the `  /proxy_wake_compute` endpoint. So what "Failure" means here?

antonyc (commented):
It looks OK from my side but got some questions regarding rust code.

proxy/src/console/messages.rs
funbringer: Good question! This comment refers to the previous definition [over here](https://github.com/neondatabase/neon/pull/3211/files/b7c08055964c7342b23a378a3f013ae1c97cb57e#diff-bc748f50c64625a1b159dbb51dbfb43eaebba5f929298dff9cc03b3ee697a8a5L53-L92
). This is known as `kickResponse` in the console and is used in the link auth. You may notice that previously we had

```rust
#[derive(Debug, Deserialize)]
enum PsqlSessionResult {
    Success(DatabaseInfo),
    Failure(String),
}
```

However, I haven't been able to find any trace of `Failure` in the console code. It occurs to me that this has never been utilized, which is why I've simplified that part and dropped this variant altogether. Please correct me if I'm wrong.

proxy/src/console/messages.rs
funbringer: This is hand written. There are several reasons for that:

* (Historical reasons) That's how it's been done here for quite a while, and it's not that bad, given that:
* The spec is incomplete! It doesn't cover http status codes or headers like `X-Request-ID`.
* The spec lives in a separate private repo, there's no cross-CI checks to validate the generated code.
* It's not like we're doing a lot of different requests here, and handwritten code gives you you more control.

proxy/src/console/messages.rs
funbringer: @antonyc sorry, I accidentally deleted your comment.

proxy/src/console/messages.rs
antonyc: Right, I got you.

proxy/src/console/messages.rs
antonyc: > Please correct me if I'm wrong.

I don't fully understand how the spec would to support both String and DatabaseInfo. So for now only DatabaseInfo can be returned.

antonyc (approved):

Issue #3208, opened 2022-12-27
Refactor code to reflect new separation of concerns & reduce source-level conflicts in the future. 


SomeoneToIgnore:
- [ ] [Fix flaky test, appeared after on-demand PR got merged](https://github.com/neondatabase/neon/issues/3203)
- [ ]  Move storage sync to hypothetical timeline/ module inside tenant.
- [x] Move tenant_* files inside this module.
- [x] Split storage_sync.rs into upload_queue and remote_timeline_client
- [ ] Split big tenant/timeline files.
- [ ] One candidate for separate module is initialization logic. Consider accumulating it in separate structures like TimelineBuilder, TimelineLoader, etc.
 We do this after porting on-demand download as it might add new perspectives to the initialization lifecycle.

LizardWizzard:
Did some steps towards that:
https://github.com/neondatabase/neon/pull/4855
https://github.com/neondatabase/neon/pull/4645
https://github.com/neondatabase/neon/pull/4619

Needs to be continued. Large tenant.rs and timeline.rs frequently cause conflicts. I believe `<Operation>Flow` approach I was using to move out deletion logic can be extended to other operations, like branch creation, tenant initialization, etc. This should move us forward towards better composition inside tenant/timeline modules.

PR #3214, opened 2022-12-27
Move tenant related files to tenant directory

shanyp:
Related to https://github.com/neondatabase/neon/issues/3208


LizardWizzard:
Probably makes sense rename tenant_<name> to just <name> since these modules are under tenant directory anyway

shanyp:
> Probably makes sense rename tenant_ to just since these modules are under tenant directory anyway

Done :-)

SomeoneToIgnore (approved):

LizardWizzard (approved):
Thanks!

PR #3215, opened 2022-12-27
Fix serialization of billing metrics

lubennikovaav:
Fixes:
- serialize TenantId and TimelineId as strings,
- skip TimelineId if none
- serialize `metric_type` field as `type`
- add `idempotency_key` field

### DIFF
--- a/pageserver/src/billing_metrics.rs
+++ b/pageserver/src/billing_metrics.rs
@@ -14,6 +14,7 @@ use pageserver_api::models::TenantState;
 use utils::id::TenantId;
 
 use serde::{Deserialize, Serialize};
+use serde_with::{serde_as, DisplayFromStr};
 use std::collections::HashMap;
 use std::fmt;
 use std::str::FromStr;
@@ -32,16 +33,23 @@ use reqwest::Url;
 /// "tenant_id": "5d07d9ce9237c4cd845ea7918c0afa7d",
 /// "timeline_id": "00000000000000000000000000000000",
 /// "time": ...,
+/// "idempotency_key":
 /// "value": 12345454,
 /// }
 /// ```
+#[serde_as]
 #[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
 pub struct BillingMetric {
     pub metric: BillingMetricKind,
+    #[serde(rename = "type")]
     pub metric_type: &'static str,
+    #[serde_as(as = "DisplayFromStr")]
     pub tenant_id: TenantId,
+    #[serde_as(as = "Option<DisplayFromStr>")]
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub timeline_id: Option<TimelineId>,
     pub time: DateTime<Utc>,
+    pub idempotency_key: String,
     pub value: u64,
 }
 
@@ -58,6 +66,7 @@ impl BillingMetric {
             tenant_id,
             timeline_id,
             time: Utc::now(),
+            idempotency_key: format!("{}", Utc::now()),
             value,
         }
     }
--- a/pageserver/src/billing_metrics.rs
+++ b/pageserver/src/billing_metrics.rs
@@ -56,12 +56,13 @@ pub struct BillingMetric {
 }
 
 impl BillingMetric {
-    pub fn new_absolute(
+    pub fn new_absolute<R: Rng + ?Sized>(
         metric: BillingMetricKind,
         tenant_id: TenantId,
         timeline_id: Option<TimelineId>,
         value: u64,
         node_id: NodeId,
+        rng: &mut R,
     ) -> Self {
         Self {
             metric,
@@ -70,12 +71,7 @@ impl BillingMetric {
             timeline_id,
             time: Utc::now(),
             // key that allows metric collector to distinguish unique events
-            idempotency_key: format!(
-                "{}-{}-{:04}",
-                Utc::now(),
-                node_id,
-                rand::thread_rng().gen_range(0..=9999)
-            ),
+            idempotency_key: format!("{}-{}-{:04}", Utc::now(), node_id, rng.gen_range(0..=9999)),
             value,
         }
     }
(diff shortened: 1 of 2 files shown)


LizardWizzard:
> I haven't find a pageserver's hostname to use idempotency_key here, so now it is just same as time.

I dont have full context here, but if you were looking for hostname then probably node id can fit in this place too. It is available in pageserver config

pageserver/src/billing_metrics.rs
```diff
@@ -58,6 +66,7 @@ impl BillingMetric {
             tenant_id,
             timeline_id,
             time: Utc::now(),
+            idempotency_key: format!("{}", Utc::now()),
```
antonyc, line 69: can we also add a random integer, please, so that the result was "{time}-{hostname_or_nodeid}-{random-4-bytes-number}". Because we need the idempotency_key to be unique across events sent in 38 hours range and if Pageserver sends 2 events at a time, I assume `format!("{}", Utc::now())` called twice for two events can produce the same string.

antonyc (approved):
LGTM

pageserver/src/billing_metrics.rs
lubennikovaav: Fixed. 
New format is
```
{
'metric': 'remote_storage_size', '
 type': 'absolute', 
 'tenant_id': 'a03ebb4f5922a1c56ff7485cc8854143',
 'time': '2022-12-28T11:07:19.317310284Z',
 'idempotency_key': '2022-12-28 11:07:19.317310324 UTC-1-4019', 
'value': 25600000
}
```

pageserver/src/billing_metrics.rs
```diff
@@ -32,16 +35,23 @@ use reqwest::Url;
 /// "tenant_id": "5d07d9ce9237c4cd845ea7918c0afa7d",
 /// "timeline_id": "00000000000000000000000000000000",
 /// "time": ...,
+/// "idempotency_key":
```
LizardWizzard, line 38: NIT: Should there be an example as it is for other fields?

pageserver/src/billing_metrics.rs
```diff
@@ -32,16 +35,23 @@ use reqwest::Url;
...
 /// "value": 12345454,
 /// }
 /// ```
+#[serde_as]
 #[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
 pub struct BillingMetric {
     pub metric: BillingMetricKind,
+    #[serde(rename = "type")]
     pub metric_type: &'static str,
```
LizardWizzard, line 47: NIT: Having struct named Metric with fields prefixed by metric doesnt look natural to me. So to me it makes sense to rename field in the struct itself without serde rename

LizardWizzard (approved):
Left some minor comments

pageserver/src/billing_metrics.rs
```diff
@@ -51,13 +61,21 @@ impl BillingMetric {
...
             metric_type: "absolute",
             tenant_id,
             timeline_id,
             time: Utc::now(),
+            // key that allows metric collector to distinguish unique events
+            idempotency_key: format!(
+                "{}-{}-{:04}",
+                Utc::now(),
+                node_id,
+                rand::thread_rng().gen_range(0..=9999)
```
SomeoneToIgnore, line 77: Note: `rand::thread_rng()` was mentioned to be costly (somewhat similar to Regex::new where it's compiled vs Regex::match on the compiled later).

We might want to reuse this rng, if possible.

SomeoneToIgnore (approved):

pageserver/src/billing_metrics.rs
LizardWizzard: Taking current timestamp is also not that cheap :)

SomeoneToIgnore: Yep, I see nothing critical here and don't ask for changes too :)

Still, `Utc::now` call is inevitable, while `rand::thread_rng` is not and that's what I wanted to notice.

pageserver/src/billing_metrics.rs
lubennikovaav: unfortunately, type is a reserved keyword), so it can't be a field name..
Or did you mean something else?

funbringer: When `type` is a reserved keyword, it's usually written as `type_` or `ty`.

pageserver/src/billing_metrics.rs
lubennikovaav: Please, check the last commit.
It is still pretty limited, though, because it's impossible to reuse thread_rng() with async code (or at least I don't know how to do it).

pageserver/src/billing_metrics.rs
lubennikovaav: fixed

pageserver/src/billing_metrics.rs
```diff
@@ -30,34 +33,45 @@ use reqwest::Url;
...
 /// "value": 12345454,
 /// }
 /// ```
+#[serde_as]
 #[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
 pub struct BillingMetric {
     pub metric: BillingMetricKind,
+    #[serde(rename = "type")]
     pub metric_type: &'static str,
+    #[serde_as(as = "DisplayFromStr")]
```
koivunej, line 48: Apologies for missing this the first time around. I did think of implementing a auto-detecting serializer for these ids but that was as far I got (the seq of u8 is probably used for `!human_readable` bincode and this is used for json/toml always).

pageserver/src/billing_metrics.rs
```diff
@@ -30,34 +33,45 @@ use reqwest::Url;
...
+        rng: &mut R,
     ) -> Self {
         Self {
             metric,
             metric_type: "absolute",
             tenant_id,
             timeline_id,
             time: Utc::now(),
+            // key that allows metric collector to distinguish unique events
+            idempotency_key: format!("{}-{}-{:04}", Utc::now(), node_id, rng.gen_range(0..=9999)),
```
koivunej, line 74: Makes me think of ULID... But perhaps if the other side considers this as opaque string this will be good enough and we can upgrade later.

koivunej (approved):
Looking good, few notes you can just resolve.

pageserver/src/billing_metrics.rs
LizardWizzard: Also reserved keywords can be still used with r#. See https://doc.rust-lang.org/beta/rust-by-example/compatibility/raw_identifiers.html

LizardWizzard: Kind and Type are also quite similar in semantics. Should we replace them with something else? My suggestion:

absolute/delta -> enum Kind
remote_storate_size -> name or id

Wdyt?

PR #3216, opened 2022-12-28
Split upload queue code from storage_sync.rs

shanyp:
https://github.com/neondatabase/neon/issues/3208


SomeoneToIgnore (approved):
Also, timeline client creation on the tenant level feels too high-level, but that's another question to solve later.

LizardWizzard:
> Also, timeline client creation on the tenant level feels too high-level, but that's another question to solve later.

Yep, it may make sense to move remote storage things further down to timeline

PR #3219, opened 2022-12-28
switch to debug from info to produce less noise

LizardWizzard:
Blindly switched most frequently appearing messages to debug. I'll verify the resulting log


lubennikovaav (approved):

LizardWizzard:
Yep, that looks ok to me by looking at logs

SomeoneToIgnore (approved):
`starting op` helped me sometimes, so we might want to info! some operations later, but definitely not all of them.

LizardWizzard:
> so we might want to info! some operations later, but definitely not all of them.

I left in place messages like "scheduling metadata upload", so it may be enough already :)

PR #3221, opened 2022-12-28
Lazy calculation of traversal_id which is needed only for error repoting

knizhnik:
See 
https://neondb.slack.com/archives/C0277TKAJCA/p1672245908989789
and 
https://neondb.slack.com/archives/C033RQ5SPDH/p1671885245981359

### DIFF
--- a/pageserver/src/tenant/timeline.rs
+++ b/pageserver/src/tenant/timeline.rs
@@ -1588,7 +1588,7 @@ trait TraversalLayerExt {
 }
 
 impl TraversalLayerExt for Arc<dyn PersistentLayer> {
-    fn traversal_id(&self) -> String {
+    fn traversal_id(&self) -> TraversalId {
         match self.local_path() {
             Some(local_path) => {
                 debug_assert!(local_path.to_str().unwrap().contains(&format!("{}", self.get_timeline_id())),
@@ -1608,12 +1608,26 @@ impl TraversalLayerExt for Arc<dyn PersistentLayer> {
 }
 
 impl TraversalLayerExt for Arc<InMemoryLayer> {
-    fn traversal_id(&self) -> String {
+    fn traversal_id(&self) -> TraversalId {
         format!(
             "timeline {} in-memory {}",
             self.get_timeline_id(),
             self.short_id()
         )
+    }
+}
+
+enum AnyLayer {
+    Memory(Arc<InMemoryLayer>),
+    Persistent(Arc<dyn PersistentLayer>),
+}
+
+impl AnyLayer {
+    fn traversal_id(&self) -> String {
+        match self {
+            AnyLayer::Memory(layer) => layer.traversal_id(),
+            AnyLayer::Persistent(layer) => layer.traversal_id(),
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


pageserver/src/tenant/timeline.rs
```diff
@@ -1617,6 +1617,20 @@ impl TraversalLayerExt for Arc<InMemoryLayer> {
     }
 }
 
+enum AnyLayer {
```
vadim2404, line 1620: It still looks like a hack, but it's possible to define an additional `trait` and use it in `dyn`. And it's possible to implement traits in custom types (look at example below).

```rust
use std::sync::Arc;

struct InMemoryLayer { x: String }
struct PersistentLayerImpl { y: String }
trait PersistentLayer {
    fn get_id(&self) -> String;
}

impl PersistentLayer for PersistentLayerImpl {
    fn get_id(&self) -> String {
        self.y.clone()
    }
}

type ArcInMemoryLayer = Arc<InMemoryLayer>;
type ArcDynPersistentLayer = Arc<dyn PersistentLayer>;

trait TraversalIdGetter {
    fn traversal_id(&self) -> String;
}

impl TraversalIdGetter for ArcInMemoryLayer {
    fn traversal_id(&self) -> String {
        self.x.clone()
    }
}

impl TraversalIdGetter for ArcDynPersistentLayer {
    fn traversal_id(&self) -> String {
        self.get_id()
    }
}

fn traversal_id(item: &dyn TraversalIdGetter) -> String {
    item.traversal_id()
}

fn main() {
    let in_memory = Arc::new(InMemoryLayer { x: "in_memory".into() });
    let persistent: ArcDynPersistentLayer = Arc::new(PersistentLayerImpl { y: "persistent".into() });
    
    println!("{}", traversal_id(&in_memory));
    println!("{}", traversal_id(&persistent));
}
```

knizhnik: Thank you for suggestion.
Actually there is no need to introduce yet another treat: we already have `TraversalLayerExt`.
So I missed that it is possible to use it instead of `Arc<dyn Layer>`.
Please check new implementation - now AnyLayer enum is removed.

vadim2404 (approved):

pageserver/src/tenant/timeline.rs
vadim2404: much better  👍

SomeoneToIgnore (approved):
I've shared my vision on what I think is the underlying issue:
https://neondb.slack.com/archives/C0277TKAJCA/p1672305471016349?thread_ts=1672245908.989789&cid=C0277TKAJCA

Shortly, that vision sees no use for `trait PersistentLayer` and some `enum PersistentLayer` should exist, to do static dispatch instead of the dynamic one.
This PR is orthogonal to the idea, but still looks nice and small, so why not merge it if it really helps the performance.

If copying over some `Box<dyn TraversalLayerExt>` (where `dyn TraversalLayerExt` is `&Arc<DeltaLayer>` or an image layer) looks good for you, then if/when we implement the static dispatch, that will be replaced with some `Copy` struct copies which should presumably cost us even less.

PR #3222, opened 2022-12-28
neon_cli timeline branch: retrive the branching LSN from Safekeepers

yeputons:
Follow-up of #3140

See https://github.com/neondatabase/neon/issues/2063#issuecomment-1355982449

Previously it was possible for a user to do some SQL queries, branch, and see old data in the branch because pageserver did not get the most recent WAL from safekeepers yet.

Now we wait until Pageserver catches up with the most recent data on Safekeepers. An alternative is to `SELECT pg_current_wal_insert_lsn()`, but that requires running branch's compute node in addition to safekeepers. Which in turn invites concurrent writes and worse semantics for branching.

### DIFF
--- a/control_plane/src/bin/neon_local.rs
+++ b/control_plane/src/bin/neon_local.rs
@@ -22,6 +22,7 @@ use safekeeper_api::{
     DEFAULT_PG_LISTEN_PORT as DEFAULT_SAFEKEEPER_PG_PORT,
 };
 use std::collections::{BTreeSet, HashMap};
+use std::iter;
 use std::path::PathBuf;
 use std::process::exit;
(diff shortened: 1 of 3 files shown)


yeputons:
To do:

* [ ] Rebase on top of `main`. There will be conflicts inside `neon_local.rs`. I believe it's enough to take the version from this PR, the one not using Pageserver API with no LSN.
* [ ] Ensure that the test fails with old `neon_local` binary and passes with the new one.
* [ ] Merge.

LizardWizzard:
As discussed moving to draft

shanyp:
closing as discussed with @arssher

PR #3223, opened 2022-12-28
test_tenant_detach_smoke now does not do compaction before GC

### LINKED ISSUE #3136: test_tenant_detach_smoke: flaky/racy immediate_gc
https://neon-github-public-dev.s3.amazonaws.com/reports/main/release/3713487584/index.html#suites/f9eba3cfdb71aa6e2b54f6466222829b/3f6af6f443ce8d08/

I think it looks like given the activation of failpoint which stalled gc until the detach had been completed:

```
2022-12-16T14:46:12.450766Z  INFO POST /v1/tenant/f5370710742ee6c2a2b76728be13b1ea/detach 200 OK
2022-12-16T14:46:12.462969Z ERROR Error processing HTTP request: InternalServerError(No such file or directory (os error 2)
```

I don't understand why there is no logging about failpoint activation. This might be related to #3124.

yeputons:
The test started failing after 6dec85b19d63fd18b7f64d65d131fae6842f39f0 because a single call to a gc HTTP endpoint was prepended with a compaction endpoint. The former does create a task which detachment waits for, the latter does not.

This test does not seem to care about gc outcome at all, so removing compaction seems safe. The underlying issue is still there: the compaction HTTP endpoint can race with tenant detachment. We only use the endpoint in testing, so it may be worth checking other tests altered in the aforementioned commit.

Fixes #3136

### DIFF
--- a/test_runner/regress/test_tenant_detach.py
+++ b/test_runner/regress/test_tenant_detach.py
@@ -24,7 +24,6 @@ def do_gc_target(
     """Hack to unblock main, see https://github.com/neondatabase/neon/issues/2211"""
     try:
         log.info("sending gc http request")
-        pageserver_http.timeline_checkpoint(tenant_id, timeline_id)
         pageserver_http.timeline_gc(tenant_id, timeline_id, 0)
     except Exception as e:
         log.error("do_gc failed: %s", e)


yeputons:
TODOs:

* [ ] Merge this PR to stabilize the CI a bit
* [ ] Decide whether we should fix the underlying issue: HTTP API endpoint for compaction not adding itself as a `TASK`. If yes, create either an issue or a PR. If no, do nothing.

koivunej:
Now finally looking at this... Wonder if I should push an empty commit to re-test this, the checkpointing still remains in `main`.

This is related to allowing `go_gc` on prod as well, does @LizardWizzard have any thoughts on this review I've failed to look at? We haven't seen this test fail much. Cannot remember a single instance, but that could had just been changed by the changed runners. Previous ones were perhaps more error prone?

koivunej:
Oh well, perhaps this should be closed, now found: https://github.com/neondatabase/neon/pull/3265

Sadly this was not originally linked from the issue so it must've fallen out of cracks. The 3265 is the discussed fix, so perhaps this was even visited earlier.

Doing the empty commit regardless to see how it changes this.

koivunej:
The latest run passed, but I am inclined to close this. Will close in a week unless I forget :)

PR #3224, opened 2022-12-29
Rename billing_metrics to consumption_metrics.

lubennikovaav:
Use more appropriate term, because not all of these metrics are used for billing.

### DIFF
--- a/pageserver/src/consumption_metrics.rs
+++ b/pageserver/src/consumption_metrics.rs
@@ -25,7 +25,7 @@ use chrono::{DateTime, Utc};
 use rand::Rng;
 use reqwest::Url;
 
-/// BillingMetric struct that defines the format for one metric entry
+/// ConsumptionMetric struct that defines the format for one metric entry
 /// i.e.
 ///
 /// ```json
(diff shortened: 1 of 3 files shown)


pageserver/src/consumption_metrics.rs
```diff
@@ -41,8 +41,8 @@ use reqwest::Url;
 /// ```
 #[serde_as]
 #[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
-pub struct BillingMetric {
-    pub metric: BillingMetricKind,
+pub struct ConsumptionMetric {
+    pub metric: ConsumptionMetricKind,
```
LizardWizzard, line 45: Can we rename fields here? Or this is already in some API contract? 

This is out of scope of this PR, but would be nice to do it earlier when we're not committed to backwards compatibility guarantees. Added some suggestions here: https://github.com/neondatabase/neon/pull/3215#discussion_r1058291562

LizardWizzard (approved):

pageserver/src/consumption_metrics.rs
lubennikovaav: there is already existing spec
https://github.com/neondatabase/cloud/blob/a374c0e4c6b94116724137b186684453e785fe61/openapispec/billing/billing-v1.yaml#L133-L154

I don't think it is worth changing as long as there is a spec and everyone is agree on terms.
I'll add a `docs/consumption_metrics.md` that describes this.

PR #3239, opened 2023-01-01
Use neondatabase/ library changes for tokio-tar

SomeoneToIgnore:
Based on https://github.com/neondatabase/tokio-tar/pull/1

If the tokio-tar approach is right, for the library management, we can merge it in master there and use `rev=$SHA` here instead of the branch.


SomeoneToIgnore:
https://github.com/neondatabase/tokio-tar/pull/1

> this is fine for us in the meanwhile

merging

PR #3242, opened 2023-01-02
Allow failure to report metrics in test_metric_collection.

arssher:
Per CI
https://github.com/neondatabase/neon/actions/runs/3822039946/attempts/1 shutdown seems to be racy.


lubennikovaav (approved):
Thank you for the fix!

PR #3247, opened 2023-01-02
Implement wss support in proxy

### LINKED ISSUE #3225: Epic: websockets support in proxy
## Motivation

Now https://github.com/neondatabase/serverless uses https://github.com/petuhovskiy/wsproxy to tunnel traffic to the actual proxy. It was the easiest options to start with, but there are few aspects about why it may be better to embed it in proxy itself:
* maintenance -- proxy already is per region, it has deploy scripts and metrics. So we don't have to re-implement that again for tunnel
* latency -- we can avoid extra pair of send/recv calls by embedding websockets in proxy
* encrypt proxy<->compute part if needed -- proxy parses initial part of the protocol, so we can support setup where js client uses wss and unencrypted postgres connection, but proxy to postgres connection is encrypted.

## DoD

https://github.com/neondatabase/serverless can connect to our staging/prod using wss

## Implementation ideas

I'd suggest to have websockets listener under a separate flag and run it in a separate pod -- that way we'll reduce blast radius and can (in theory) update ws-proxy without restarting psql-proxies.

petuhovskiy:
Resolves #3225

This is a hacky implementation of WebSocket server, embedded into our postgres proxy. The server is used to allow https://github.com/neondatabase/serverless to connect to our postgres from browser and serverless javascript functions.

How it will work (general schema):
- browser opens a websocket connection to `wss://ep-shy-haze-711459.eu-central-1.aws.neon.tech/`
- proxy accepts this connection and terminates TLS (https)
- inside encrypted tunnel (HTTPS), browser initiates plain (non-encrypted) postgres connection
- proxy performs auth as in usual plain pg connection and forwards connection to the compute

How to test it locally (normal flow):
1. Clone this branch and build a proxy binary
2. Start console locally
3. Add ws config to docker-compose.yml
```
...
proxy \
          --proxy 0.0.0.0:4432 \
          --wss 0.0.0.0:8443 \
...
ports: ["4432:4432", "8443:8443"]
```
4. Replace proxy binary inside a docker container
```bash
docker cp proxy zenith_proxy:/usr/local/bin/proxy && docker restart zenith_proxy
```
5. Clone `serverless` repo and test it in the browser
```bash
# don't forget to prepare a database with UNESCO data, as written here https://github.com/neondatabase/serverless-cfworker-demo#how-to-run

git clone -b wss-proxy https://github.com/neondatabase/serverless.git

# specify proxy connstr
cat > .dev.vars
DATABASE_URL=postgres://petuhovskiy:zJyjRY85oDrL@ep-little-pond-851309.localtest.me:4432/neondb

# optional, should work even without build
./build.sh debug

# start server on :8888 and go to /dist/deploy
npx http-server -p 8888
```

Expected:
image

Note: if you have issues with self-signed https certificate, try to (works in my Firefox):
- open https://ep-little-pond-851309.localtest.me:8443/ in new tab and allow "insecure" content
- reload the page with websocket client

I also want to add test for normal flow in the console, I'll do that after this PR will be merged.


proxy/Cargo.toml
```diff
@@ -35,10 +37,12 @@ thiserror = "1.0.30"
 tokio = { version = "1.17", features = ["macros"] }
 tokio-postgres = { git = "https://github.com/neondatabase/rust-postgres.git", rev="43e6db254a97fdecbce33d8bc0890accfd74495e" }
 tokio-rustls = "0.23.0"
+tls-listener = { version = "0.5.1", features = ["rustls", "hyper-h1"] }
```
petuhovskiy, line 40: We can play with `hyper-h2` here probably

proxy/src/http/websocket.rs
```diff
@@ -0,0 +1,255 @@
...
+            ready(false)
+        } else {
+            ready(true)
+        }
+    });
+
+    let make_svc = hyper::service::make_service_fn(|_stream| async move {
+        Ok::<_, Infallible>(hyper::service::service_fn(
+            move |req: Request<Body>| async move {
+                let cancel_map = Arc::new(CancelMap::default());
```
petuhovskiy, line 243: I wanted to make a global `cancel_map`, but Rust beat me. Also I'm not sure what `cancel_map` does actually, so would be nice to have someone familiar with proxy to comment here.

proxy/src/http/websocket.rs
```diff
@@ -0,0 +1,255 @@
...
+        if let Err(err) = conn {
+            error!("failed to accept TLS connection for websockets: {:?}", err);
+            ready(false)
+        } else {
+            ready(true)
+        }
+    });
+
+    let make_svc = hyper::service::make_service_fn(|_stream| async move {
+        Ok::<_, Infallible>(hyper::service::service_fn(
```
petuhovskiy, line 241: NIT: We can try to log `remote_addr` here, not sure how useful it is

proxy/src/http/websocket.rs
kelvich: When the connection is established postgres sends special cancel_key to client. Using this key client can open a new connection and call `select pg_cancel_backend($cancel_key)` to kill currently running query in the main connection. Proxy keeps a map to be able to route that cancel connection.

Usually clients call cancel thing on various timeouts or signals. Not sure is js driver supports it at all.

proxy/src/proxy.rs
```diff
@@ -95,15 +95,31 @@ async fn handle_client(
...
         None => return Ok(()), // it's a cancellation request
     };
 
     // Extract credentials which we're going to use for auth.
     let creds = {
-        let sni = stream.get_ref().sni_hostname();
+        let sni = if config.secure_override_hostname.is_none() {
+            stream.get_ref().sni_hostname()
+        } else {
+            let hostname = config.secure_override_hostname.as_ref().unwrap();
```
SomeoneToIgnore, line 120: > .unwrap()

could be just `match &config.secure_override_hostname` ?

proxy/src/main.rs
```diff
@@ -110,12 +111,23 @@ async fn main() -> anyhow::Result<()> {
...
         tokio::spawn(proxy::task_main(config, proxy_listener)),
         tokio::task::spawn_blocking(move || mgmt::thread_main(mgmt_listener)),
-    ]
-    .map(flatten_err);
+    ];
+
+    if let Some(wss_address) = arg_matches.get_one::<String>("wss") {
+        let wss_address: SocketAddr = wss_address.parse()?;
+        info!("Starting wss on {}", wss_address);
+        let wss_listener = TcpListener::bind(wss_address).await?.into_std()?;
```
SomeoneToIgnore, line 123: > .into_std()?

inside `http::websocket::task_main` we, almost instantly, do 

`let tokio_listener = tokio::net::TcpListener::from_std(ws_listener)?;`

why is it needed?

proxy/src/http/websocket.rs
```diff
@@ -0,0 +1,255 @@
...
+                .advance(amt);
+        }
+    }
+}
+
+async fn serve_websocket(
+    websocket: HyperWebsocket,
+    config: &ProxyConfig,
+    cancel_map: &CancelMap,
+    session_id: uuid::Uuid,
```
SomeoneToIgnore, line 166: Would that make sense to have a span with each session id?

SomeoneToIgnore (commented):
LGTM generally, commented a few odd spots but nothing big.

proxy/src/http/websocket.rs
```diff
@@ -0,0 +1,255 @@
...
+    let host = request.headers().get("host").and_then(|h| h.to_str().ok());
+
+    // Create a new proxy config with the hostname from the request.
+    let config = ProxyConfig {
+        tls_config: config.tls_config.clone(),
+        auth_backend: config.auth_backend.clone(),
+        secure_override_hostname: host
+            .and_then(|h| {
+                // Remove the port from the hostname, if present.
+                h.split(':').next()
```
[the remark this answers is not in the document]
petuhovskiy: > but technically any garbage could come into host HTTP header?

Technically yes, but I'm not sure if there is any good validation we can do to fail fast.

proxy/src/http/websocket.rs
```diff
@@ -0,0 +1,255 @@
...
+pin_project! {
+    /// This is a wrapper around a WebSocketStream that implements AsyncRead and AsyncWrite.
+    pub struct WebSocketRW {
+        #[pin]
+        stream: WebSocketStream<Upgraded>,
+        chunk: Option<bytes::Bytes>,
+    }
+}
+
+unsafe impl Sync for WebSocketRW {}
```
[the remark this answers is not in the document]
petuhovskiy: Sounds good, can you help me with the SAFETY comment? Not sure what should be written in it and haven't looked deep into `WebSocketStream<Upgraded>`.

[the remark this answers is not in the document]
SomeoneToIgnore: I honestly have no idea why it's safe 🙂 
Feels unsafe from the first glance, but there has to be some general reason we use it for, might start with describing that.

jawj:
I have zero Rust experience, so I'm not going to comment on any specific code, but I have some questions about usage/backwards compatibility, SNI and pg auth.

1. Usage and backwards-compatibility

Until now, we've had a single proxy instance, directing all connections via that instance and specifying the actual DB host in an `address` query param — e.g. `wss://ws.neon.build/v1?address=example-db.neon.tech:5432`.

Of course, the proxy is now also the DB host. And having looked at your recent `wss-proxy` branch of the `neondatabase/serverless` repo I think, as things stand, that we now just connect to the host (without an address parameter)?

But: could we instead retain backwards compatibility — and also compatibility for people using the driver with their own pg DB and a separate wsproxy instance — by carrying on doing it the old way?

I guess that would mean connecting to `wss://example-db.neon.tech/v1?address=example-db.neon.tech:5432`, and then enforcing here in the proxy that the address parameter matches the host and that the port is 5432.

2. Authentication and SNI

Until now, the serverless driver deliberately disabled SNI (in the TLS handshake) and then put the project/endpoint name in the password field. It did this to force cleartext password auth (via our hacky workaround for old pg clients) instead of SCRAM-SHA-256.

This is important for two reasons. First, ordinary SCRAM (with 4096 rounds) takes far too long to be run in an edge function: it exceeds the allowed CPU time in a Cloudflare Worker, and produces giant latency numbers. Second, even if we were to set the SCRAM rounds to a much smaller number, it inevitably adds a network round-trip, which again increases latency.

So I think it would be strongly preferable if we could just configure the proxy to always use cleartext password auth for connections via WebSocket. This would simplify the driver, and also ultimately enable us to drop the hacky SNI workaround in general.

I also notice references to SNI workarounds for WebSocket connections in the code, and I wonder if those could then be removed? (There's no reason why we can't enable SNI for the WebSocket connection: we just have to stop deliberately suppressing it).

3. Driver

Finally, I probably ought to have clarified this sooner, but please note that the version of the serverless driver that this needs to work with is the one found on the `no-webassembly` branch: https://github.com/neondatabase/serverless/tree/no-webassembly.

jawj (commented):
Hmm, I'm new to the GitHub review mechanism, so I'm afraid I just left my review comments in the conversation thread (EDIT: which I now see appears directly above this message). Do let me know if I need to copy them here instead.

petuhovskiy:
@jawj your message is in the right place, no need to copy it anywhere else :)

> 1. Usage and backwards-compatibility

We are not going to disable our old (wsproxy) `ws://ws.neon.build/v1?address=example-db.neon.tech:5432` proxy, it can run as long as possible. That means that users of old version of `serverless` library should not experience any issues.

For the new version of `serverless` library we can switch to the new `wss` proxy, and I don't think we should backport `/v1?address=example-db.neon.tech:5432` schema to it. To use driver with the own pg DB you can take wsproxy and old version of `serverless` driver, or we can support legacy mode in the `serverless` driver itself, but I don't think we need to support `/v1` in our proxy.

Regarding future compatibility, maybe it's better to connect to `wss://ep-shy-haze-711459.eu-central-1.aws.neon.tech/v2` with `/v2` in the end, to aid future compatibility.

> 2. Authentication and SNI

When I did experiments with the current version, it used SCRAM in a unencrypted pg connection (there is a TLS, but it comes from WSS/HTTPS, no webassembly required). It should be easy to use plaintext password auth as well, I don't think there will be any issues, but I haven't experimented with it yet.

SNI looks completely irrelevant for unencrypted pg connections and I guess we can already remove all SNI references from our `serverless` driver. I hacked proxy the use `sni` variable to pass the project/endpoint name in it, but in fact we don't have any SNI in here and just take url from the `Host` http header.

> 3. Driver

Thanks, I think we just need to experiment with current `wss` version more and iterate on the way how it works. You can ping or call me in Slack if you have any more questions :)

petuhovskiy:
I looked at the proxy code around the password hack and tried to make pipelining work with it. It worked, but currently there's no good way to do cleartext password auth.

<details>
<summary>
Screenshot of a pipelining with our proxy
</summary>
image
</details>

What I mean by "no good way" is that we don't know the password in the proxy and console, and we need to wake up compute for all auth requests, which is bad. I'm not sure which compromises we want to make to cut latencies, so it's not clear to me what is next steps here.

kelvich:
> What I mean by "no good way" is that we don't know the password in the proxy and console, and we need to wake up compute for all auth requests, which is bad.

Well, I think we are not making situation worse compared to password hack flow, aren't we? Also, we don't have to wake compute without auth -- proxy knows password and console knows SCRAM stored key, and password + stored key is enough to check auth.

petuhovskiy:
This PR is mostly ready and we want to deploy it to staging, so we can go with review after merge.

Issue #3249, opened 2023-01-02
Improve observability for on-demand download


problame:
- [ ] Observability for downloads while attaching tenants. Might punt this if on-demand download arrives quickly. But right now we don't have any observability, not even scoping to `task_mgr`
   - [ ] Download queue size

PR #3250, opened 2023-01-02
remote_timeline_client: more metrics & metrics-related cleanups

problame:
- Clean up redundant metric removal in TimelineMetrics::drop. RemoteTimelineClientMetrics is responsible for cleaning up REMOTE_OPERATION_TIME andREMOTE_UPLOAD_QUEUE_UNFINISHED_TASKS.

- Rename `pageserver_remote_upload_queue_unfinished_tasks` to `pageserver_remote_timeline_client_calls_unfinished`. The new name reflects that the metric is with respect to the entire call to remote timeline client. This includes wait time in the upload queue and hence it's a longer span than what `pageserver_remote_OPERATION_seconds` measures.

- Add the `pageserver_remote_timeline_client_calls_started` histogram. See the metric description for why we need it.

- Add helper functions `call_begin` etc to `RemoteTimelineClientMetrics` to centralize the logic for updating the metrics above (they relate to each other, see comments in code).

- Use these constructs to track ongoing downloads in `pageserver_remote_timeline_client_calls_unfinished`

refs https://github.com/neondatabase/neon/issues/2029 fixes https://github.com/neondatabase/neon/issues/3249

### DIFF
--- a/pageserver/src/tenant/remote_timeline_client.rs
+++ b/pageserver/src/tenant/remote_timeline_client.rs
@@ -298,8 +298,8 @@ impl RemoteTimelineClient {
         conf: &'static PageServerConf,
         tenant_id: TenantId,
         timeline_id: TimelineId,
-    ) -> anyhow::Result<RemoteTimelineClient> {
-        Ok(RemoteTimelineClient {
+    ) -> RemoteTimelineClient {
+        RemoteTimelineClient {
             conf,
             runtime: &BACKGROUND_RUNTIME,
             tenant_id,
@@ -307,7 +307,7 @@ impl RemoteTimelineClient {
             storage_impl: remote_storage,
             upload_queue: Mutex::new(UploadQueue::Uninitialized),
             metrics: Arc::new(RemoteTimelineClientMetrics::new(&tenant_id, &timeline_id)),
-        })
+        }
     }
 
     /// Initialize the upload queue for a remote storage that already received
--- a/pageserver/src/tenant/remote_timeline_client.rs
+++ b/pageserver/src/tenant/remote_timeline_client.rs
@@ -367,7 +367,11 @@ impl RemoteTimelineClient {
 
     /// Download index file
     pub async fn download_index_file(&self) -> Result<IndexPart, DownloadError> {
-        download::download_index_part(
+        let unfinished_gauge_guard = self
+            .metrics
+            .call_begin(&RemoteOpFileKind::Index, &RemoteOpKind::Download);
+
+        let res = download::download_index_part(
             self.conf,
             &self.storage_impl,
             self.tenant_id,
@@ -380,7 +384,11 @@ impl RemoteTimelineClient {
             RemoteOpKind::Download,
             Arc::clone(&self.metrics),
         )
-        .await
+        .await;
+
+        drop(unfinished_gauge_guard);
+
+        res
     }
 
     /// Download a (layer) file from `path`, into local filesystem.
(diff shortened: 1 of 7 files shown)


problame:
Test failures are not due to this patch, re-running.

pageserver/src/tenant/remote_timeline_client.rs
```diff
@@ -410,6 +422,8 @@ impl RemoteTimelineClient {
         )
         .await?;
 
+        drop(unfinished_gauge_guard);
```
shanyp, line 425: instead of dropping, can we use code block, i.e :  {
{
  let unfinished_gauge_guard = self.metrics.call_begin(&RemoteOpFileKind::Layer, &RemoteOpKind::Download);

   let downloaded_size = download::download_layer_file(
            self.conf,
            &self.storage_impl,
	@@ -410,6 +422,8 @@ impl RemoteTimelineClient {
        )
        .await?;

} // not needed anymore as goes out of scope :        drop(unfinished_gauge_guard);

problame:
Addressed @shanyp 's comment & added basic test coverage.
The test unveiled that tenant_attach doesn't use RemoteTimelineClient for downloading the index part.

Fixed that in commit
[`use RemoteTimelineClient for downloading index part during tenant_attach`](https://github.com/neondatabase/neon/pull/3250/commits/00e2242f74f03854c3e0b2fabf8d5a674c4cecd0)
which I extracted into separate PR
https://github.com/neondatabase/neon/pull/3258

Once that separate PR is merged, I'll rebase this PR.

problame:
Force pushed to rebase onto https://github.com/neondatabase/neon/pull/3258

pageserver/src/tenant/remote_timeline_client.rs
```diff
@@ -393,22 +397,27 @@ impl RemoteTimelineClient {
...
-            RemoteOpKind::Download,
-            Arc::clone(&self.metrics),
-        )
-        .await?;
+        let downloaded_size = {
+            let _unfinished_gauge_guard = self
+                .metrics
+                .call_begin(&RemoteOpFileKind::Layer, &RemoteOpKind::Download);
+            download::download_layer_file(
+                self.conf,
```
shanyp, line 405: can't this be a struct within RemoteClientDownload that is shared with download.rs ?

problame: What is `this`? 

Note that this PR didn't change anything about `download::download_layer_file`. I just had to indent it into a new scope so that we drop the (new) `_unfinished_gauge_guard` in a timely manner.

shanyp: What I meant was the download::download_layer_file will receive RemoteClientDownload as a parameter

test_runner/regress/test_remote_storage.py
```diff
@@ -368,6 +369,173 @@ def tenant_active():
...
+    # create some layers & wait for uploads to finish
+    overwrite_data_and_wait_for_it_to_arrive_at_pageserver("a")
+    client.timeline_checkpoint(tenant_id, timeline_id)
+    client.timeline_compact(tenant_id, timeline_id)
+    overwrite_data_and_wait_for_it_to_arrive_at_pageserver("b")
+    client.timeline_checkpoint(tenant_id, timeline_id)
+    client.timeline_compact(tenant_id, timeline_id)
+    gc_result = client.timeline_gc(tenant_id, timeline_id, 0)
+    print_gc_result(gc_result)
+    assert gc_result["layers_removed"] > 0
```
shanyp, line 478: NIT could be a function that is called twice and has iota to make sure it creates different data (a,b,c,d)

shanyp (approved):

pageserver/src/tenant/remote_timeline_client.rs
problame: Sure it could - I guess this comes from the angle of 'long parameter lists are hard to read'.
I personally think 6 parameters is fine iff they all have different types.
That is the case here.

test_runner/regress/test_remote_storage.py
problame: Will be done in next push.

problame:
Rebased to get more stable CI

Issue #3107, opened 2022-12-14
gc_loop starting on 0 timelines is wrong


LizardWizzard:
## Steps to reproduce

the tenant was in broken state and I `ignore` -> `load` it. And this appeared

## Expected result

starting on 1 timeline

## Actual result

```
INFO gc_loop{tenant_id=59b8194a8b0c4021e1e68792dcde69c7}: starting on 0 timelines
INFO compaction_loop{tenant_id=59b8194a8b0c4021e1e68792dcde69c7}:compact_timeline{timeline=ac40625fd35bff607a1c5ad95984fce0}: starting op: Barrier
INFO compaction_loop{tenant_id=59b8194a8b0c4021e1e68792dcde69c7}:compact_timeline{timeline=ac40625fd35bff607a1c5ad95984fce0}: waiting for upload ops to complete
INFO compaction_loop{tenant_id=59b8194a8b0c4021e1e68792dcde69c7}:compact_timeline{timeline=ac40625fd35bff607a1c5ad95984fce0}: starting op: UploadMetadata(lsn: 0/1AAE150) 
INFO wal_connection_manager{tenant=59b8194a8b0c4021e1e68792dcde69c7 timeline=ac40625fd35bff607a1c5ad95984fce0}: Subscribed for broker timeline updates
```

It says `starting on 0 timelines` while there is a timeline

## Environment

Prod

## Logs, links
-

problame:
Why are you expecting `1`? If `refresh_gc_info_internal` returns an empty vec, we log 0.
Would you want a different log message in that case?

Or were you expecting that at least one timeline needs GC?

LizardWizzard:
I expected that it logs number of all timelines, if we log only number of timelines that need gc then lets just change the message. If there are no timelines I would expect "Nothing to gc", or "X timelines need gc, starting" or something similar

Issue #3218, opened 2022-12-28
Epic: Improve page server logging story


shanyp:
## Motivation

Per this discussion started by [@lassizci](https://neondb.slack.com/archives/C033RQ5SPDH/p1672234941320819)  and [@kelvich ](https://neondb.slack.com/archives/C033RQ5SPDH/p1672239384427269?thread_ts=1672234941.320819&cid=C033RQ5SPDH) 
Pageserver creates ~7.5M lines of log at INFO level every hour. Size wise talking about 1.45GB/hour

## DoD
A concise log story for pageserver providing necessary breadcrumbs for actual issues and less verbose in normal operation

## Implementation ideas
https://docs.google.com/document/d/1i4yb102lrZgnKOU7sjKAf4zWXQxdUbU4OrkIDZSHj7M/edit

- Observability of remote storage - metrics
- Request context - missing  (management operations are ok for adding, might be costly for non user operations)
- Many issues arise with string formatting, need to come up with a solution (maybe ETW way)
- Move to collect bread crumbs, and only log them in case of an error
- Add the ability to change log level, per component dynamically, 
- Llvm xray, dynamic instrumentation framework, enable/disable  (https://internals.rust-lang.org/t/pre-rfc-xray-instrumentation-support/17473) 
- support dynamically per Timeline tracing on/off 
- Go over lassi’s noisy messages grafana and flash out the ones that cause most of the noise 
- In the future alert for over quota logs /' alerts over exceeding Errors/Warns  
- Deduplication of trace messages. 
- Use variables for log messages, and not string formatting 

## Tasks

- [x] https://github.com/neondatabase/neon/issues/3114
- [x] https://github.com/neondatabase/neon/pull/3227
- [x] #3253 
- [x] #3314 
- [x] #3337 
- [x] #3338 
- [x] #3339
- [x] #3340 
- [x] #3341 
- [x] #3342 
- [x] #3343 
- [x] #3344 
- [x] #3346 
- [x] #3347 
- [x] #3372 

## Other related tasks and Epics
-

kelvich:
Would also super useful to have:
* per-module tags on logs and ability to change it enforce log level per module in a runtime, e.g. though http api.
* ability to show all logs for specific tenant in runtime

depending on complexity that could be done now or later

shanyp:
the per module logging was moved to epic : #3547

PR #3253, opened 2023-01-02
gc_iteration_internal: better log message

problame:
fixes https://github.com/neondatabase/neon/issues/3107

### DIFF
--- a/pageserver/src/tenant.rs
+++ b/pageserver/src/tenant.rs
@@ -1874,7 +1874,12 @@ impl Tenant {
 
         utils::failpoint_sleep_millis_async!("gc_iteration_internal_after_getting_gc_timelines");
 
-        info!("starting on {} timelines", gc_timelines.len());
+        // If there is nothing to GC, we don't want any messages in the INFO log.
+        if !gc_timelines.is_empty() {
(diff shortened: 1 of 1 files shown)


SomeoneToIgnore (commented):
There'a nothing that prevents 0 timelines from being logged here with INFO level.
If I understand it right, it could be as frequent, as the GC runs?

Let's do an early return maybe, otherwise LGTM.

problame:
> There'a nothing that prevents 0 timelines from being logged here with INFO level.
> If I understand it right, it could be as frequent, as the GC runs?

Right.

> Let's do an early return maybe, otherwise LGTM.

Can't do that, need to account for totals later in the function.

Switched to an `if-else` to log the same message at debug level if `gc_timelines` is 0.
Sadly a little repetitive because `tracing::event!` only accepts `const` levels.

SomeoneToIgnore (approved):

lassizci:
:+1:  Also ref: https://github.com/neondatabase/neon/issues/3218 as these seem to be a large part of the gc_loop logging in practice.

hlinnaka (approved):

shanyp (approved):

PR #3254, opened 2023-01-02
Add more io::Error context when fail to operate on a path

SomeoneToIgnore:
I have a test failure that shows 

```
Caused by:
    0: Failed to reconstruct a page image:
    1: Directory not empty (os error 39)
```

but does not really show where exactly that happens.
https://neon-github-public-dev.s3.amazonaws.com/reports/pr-3227/release/3823785365/index.html#categories/c0057473fc9ec8fb70876fd29a171ce8/7088dab272f2c7b7/?attachment=60fe6ed2add4d82d

The PR aims to add more context in debugging that issue.

### DIFF
--- a/pageserver/src/walredo.rs
+++ b/pageserver/src/walredo.rs
@@ -630,7 +630,7 @@ impl PostgresRedoProcess {
             fs::remove_dir_all(&datadir).map_err(|e| {
                 Error::new(
                     e.kind(),
-                    format!("Failed to remove old temporary dir {datadir:?}: {e}"),
+                    format!("Old temporary dir {datadir:?} removal failure: {e}"),
                 )
             })?;
         }
--- a/pageserver/src/walredo.rs
+++ b/pageserver/src/walredo.rs
@@ -626,24 +626,20 @@ impl PostgresRedoProcess {
 
         // Create empty data directory for wal-redo postgres, deleting old one first.
         if datadir.exists() {
-            info!(
-                "old temporary datadir {} exists, removing",
-                datadir.display()
-            );
-            fs::remove_dir_all(&datadir)?;
+            info!("old temporary datadir {datadir:?} exists, removing");
+            fs::remove_dir_all(&datadir).map_err(|e| {
+                Error::new(
+                    e.kind(),
+                    format!("Failed to remove old temporary dir {datadir:?}: {e}"),
+                )
+            })?;
         }
-        let pg_bin_dir_path = conf.pg_bin_dir(pg_version).map_err(|e| {
-            Error::new(
-                ErrorKind::Other,
-                format!("incorrect pg_bin_dir path: {}", e),
-            )
-        })?;
-        let pg_lib_dir_path = conf.pg_lib_dir(pg_version).map_err(|e| {
-            Error::new(
-                ErrorKind::Other,
-                format!("incorrect pg_lib_dir path: {}", e),
-            )
-        })?;
+        let pg_bin_dir_path = conf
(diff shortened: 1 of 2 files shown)


pageserver/src/walredo.rs
```diff
@@ -626,24 +626,20 @@ impl PostgresRedoProcess {
...
-                ErrorKind::Other,
-                format!("incorrect pg_lib_dir path: {}", e),
-            )
-        })?;
+        let pg_bin_dir_path = conf
+            .pg_bin_dir(pg_version)
+            .map_err(|e| Error::new(ErrorKind::Other, format!("incorrect pg_bin_dir path: {e}")))?;
+        let pg_lib_dir_path = conf
+            .pg_lib_dir(pg_version)
+            .map_err(|e| Error::new(ErrorKind::Other, format!("incorrect pg_lib_dir path: {e}")))?;
```
koivunej, line 642:
```suggestion
            .map_err(|e| Error::new(e.kind(), format!("incorrect pg_bin_dir path: {e}")))?;
        let pg_lib_dir_path = conf
            .pg_lib_dir(pg_version)
            .map_err(|e| Error::new(e.kind(), format!("incorrect pg_lib_dir path: {e}")))?;
```

koivunej:
Also, this original test case failure is rather interesting... Feels like a new failure mode, but I cannot really see why would it be caused here..

pageserver/src/walredo.rs
SomeoneToIgnore: I wish I could apply this (multiline edits are there? nice), but `e` is of type `anyhow::Error` for both of these hence `.kind()` does not exist there.

https://github.com/neondatabase/neon/blob/7920b39a275a7bd13155d8726c92cd417bf7e2f9/pageserver/src/config.rs#L514-L527

SomeoneToIgnore:
Recent CI run with the relevant error: https://neon-github-public-dev.s3.amazonaws.com/reports/pr-3276/debug/3909266252/index.html#categories/b06bc5e2727f596ad6a465f039981fb0/48d11117e142896/

I'd love to see which directory is not empty, to be sure.

test_runner/regress/test_recovery.py
```diff
@@ -11,6 +11,7 @@
 def test_pageserver_recovery(neon_env_builder: NeonEnvBuilder):
     # Override default checkpointer settings to run it more often
     neon_env_builder.pageserver_config_override = "tenant_config={checkpoint_distance = 1048576}"
+    neon_env_builder.preserve_database_files = True
```
problame, line 14: Why is this needed?

pageserver/src/walredo.rs
```diff
@@ -626,24 +626,20 @@ impl PostgresRedoProcess {
...
-            info!(
-                "old temporary datadir {} exists, removing",
-                datadir.display()
-            );
-            fs::remove_dir_all(&datadir)?;
+            info!("old temporary datadir {datadir:?} exists, removing");
+            fs::remove_dir_all(&datadir).map_err(|e| {
+                Error::new(
+                    e.kind(),
+                    format!("Failed to remove old temporary dir {datadir:?}: {e}"),
```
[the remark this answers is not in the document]
SomeoneToIgnore: This PR was created some time before this rule was introduced.
Changed, albeit this particular place is not anyhow's context and probably could be just an error.

test_runner/regress/test_recovery.py
SomeoneToIgnore: Removed, thanks.

problame (approved):

PR #3258, opened 2023-01-03
use RemoteTimelineClient for downloading index part during tenant_attach

problame:
Before this change, we would not .measure_remote_op for index part downloads.

And more generally, it's good to pass not just uploads but also downloads through RemoteTimelineClient, e.g., if we ever want to implement some timeline-scoped policies there.

Found this while working on https://github.com/neondatabase/neon/pull/3250 where I add a metric to measure the degree of concurrent downloads. Layer download was missing in a test that I added there.

### DIFF
--- a/pageserver/src/tenant.rs
+++ b/pageserver/src/tenant.rs
@@ -13,11 +13,13 @@
 
 use anyhow::{bail, Context};
 use bytes::Bytes;
+use futures::FutureExt;
 use futures::Stream;
 use pageserver_api::models::TimelineState;
 use remote_storage::DownloadError;
 use remote_storage::GenericRemoteStorage;
 use tokio::sync::watch;
+use tokio::task::JoinSet;
 use tracing::*;
 use utils::crashsafe::path_with_suffix_extension;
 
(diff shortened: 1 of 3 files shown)


pageserver/src/tenant.rs
```diff
@@ -639,26 +642,59 @@ impl Tenant {
...
 
-        let mut timeline_ancestors: HashMap<TimelineId, TimelineMetadata> = HashMap::new();
-        let mut index_parts: HashMap<TimelineId, IndexPart> = HashMap::new();
-        for (timeline_id, index_part) in remote_timelines {
-            let remote_metadata = index_part.parse_metadata().with_context(|| {
-                format!(
-                    "Failed to parse metadata file from remote storage for tenant {} timeline {}",
-                    self.tenant_id, timeline_id
+        // Download & parse index parts
+        let mut part_downloads = FuturesUnordered::new();
```
SomeoneToIgnore, line 655: NIT

> FuturesUnordered

Before I learned about [`JoinSet`](https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html), I've also used it, but it requires an extra package, `futures`, while `JoinSet` is in `tokio` which we're using heavily anyway.

JoinSet's signatures seem to be slightly simply too, so we might consider using that in the end.

SomeoneToIgnore (approved):

pageserver/src/tenant.rs
problame: Cool, will switch to it.

pageserver/src/tenant.rs
```diff
@@ -639,26 +641,62 @@ impl Tenant {
...
-                    "Failed to parse metadata file from remote storage for tenant {} timeline {}",
-                    self.tenant_id, timeline_id
-                )
-            })?;
+        // Download & parse index parts
+        let mut part_downloads = JoinSet::new();
+        for timeline_id in remote_timeline_ids {
+            let client = RemoteTimelineClient::new(
+                remote_storage.clone(),
+                self.conf,
```
shanyp, line 658: why not pass Tenant (self) and let RemoteTimelineClient take what it needs ?

shanyp (approved):

pageserver/src/tenant.rs
problame: `RemoteTimelineClient` doesn't need anything from `Tenant` except the `TenantId` and `TimelineId`.
So, limit its view to what matters.
I guess some people call this "Interface Segregation Principle".

Also, it makes `RemoteTimelineClient`  easier to test because we don't have to construct a `Tenant` object for it.

Finally, in Rust, if we passed the `Arc<Tenant>`, or better `Arc<Timeline>` down into `RemoteTimelineClient`, we'd have a reference cycle: `Timeline => RemoteTimelineClient => Timeline`

shanyp (approved):

problame:
Rebased to get the more stable CI yml

PR #3227, opened 2022-12-29
Tone down safekeeper and pageserver walreceiver errors

SomeoneToIgnore:
Closes https://github.com/neondatabase/neon/issues/3114

Adds `MaybeIoError` to allow distinguishing between `std::io::Error` and `anyhow::Error`, replace `anyhow::Result<T>` with `Result<T, MaybeIoError>` in all related places.
That allows safekeeper avoid printing lengthy error backtraces.
Many changes related to the error replacement required removing `use anyhow::Result`, hence more changes than intended.

Pageserver logs are now also cleaned of the stacktrace, but through source error downcasting since `tokio_postgres` swallows `std::io::Error` along the way.

<details>
  <summary>Safekeeper logs, before</summary>
  
  ```
  2022-12-29T14:11:23.690976Z  INFO version: git:c0290467fa100c7e2c147e3b804e9a45d108b0b0
2022-12-29T14:11:23.698368Z  INFO initialized safekeeper id 1
2022-12-29T14:11:23.708557Z  INFO claimed pid file at "/Users/someonetoignore/work/neon/neon_main/test_output/test_remote_storage_upload_queue_retries[local_fs]/repo/safekeepers/sk1/safekeeper.pid"
2022-12-29T14:11:23.708697Z  INFO starting safekeeper on 127.0.0.1:15018
2022-12-29T14:11:23.708716Z  INFO auth is disabled
2022-12-29T14:11:23.708876Z  INFO found 0 tenants directories, successfully loaded 0 timelines
2022-12-29T14:11:23.709045Z  INFO broker: started, broker endpoint http://127.0.0.1:15015/
2022-12-29T14:11:23.709163Z  INFO WAL backup launcher started, remote config None
2022-12-29T14:11:23.709174Z  INFO Starting an HTTP endpoint at 127.0.0.1:15019
2022-12-29T14:11:23.788433Z  INFO GET /v1/status 200 OK
2022-12-29T14:11:23.791553Z  INFO GET /v1/status 200 OK
2022-12-29T14:11:24.993355Z  INFO {tid=17}: got query "START_WAL_PUSH" in timeline Some(9921cb8a7c2c1d256587b2274ae9c799)
2022-12-29T14:11:24.993496Z  INFO {tid=17}:WAL acceptor{ttid=b8c2e54bfd8f7ed3d7753959eb28b853/9921cb8a7c2c1d256587b2274ae9c799}: start handshake with walproposer 127.0.0.1:63841 sysid 0 timeline 1
[...]

PR #3262, opened 2023-01-03
Fix one more test log

SomeoneToIgnore:
https://neon-github-public-dev.s3.amazonaws.com/reports/pr-3261/debug/3833043374/index.html#suites/ffbb7f9930a77115316b58ff32b7c719/1f6ebaedc0a113a1/

Spotted a flacky test that appeared after https://github.com/neondatabase/neon/pull/3227 changes


hlinnaka (approved):
Note that there are rough categories in the list of allowed errors, separated by comments. Previously the Broken pipe error was in the "All tests print these, when starting up or shutting down" section. I think that's appropriate for the new one too. I think both of these lines belong in the old location.

problame (approved):
Agree with @hlinnaka 's comment above.

problame (approved):

PR #3263, opened 2023-01-03
Split delta and image layer generics in LayerMap

SomeoneToIgnore:
Follow-up of https://github.com/neondatabase/neon/pull/3261

### DIFF
--- a/pageserver/src/tenant/storage_layer/delta_layer.rs
+++ b/pageserver/src/tenant/storage_layer/delta_layer.rs
@@ -51,7 +51,7 @@ use utils::{
     lsn::Lsn,
 };
 
-use super::{DeltaFileName, Layer, LayerIter, LayerKeyIter, PathOrConf};
+use super::{DeltaFileName, Layer, LayerFileName, LayerIter, LayerKeyIter, LocalLayer, PathOrConf};
 
 ///
 /// Header stored in the beginning of the file
@@ -217,7 +217,7 @@ impl Layer for DeltaLayer {
     }
 
     fn short_id(&self) -> String {
-        self.layer_name().to_string()
+        self.delta_layer_name().to_string()
     }
     /// debugging function to print out the contents of the layer
     fn dump(&self, verbose: bool) -> Result<()> {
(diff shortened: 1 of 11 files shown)


problame:
Note that this change can wait until we've merged RequestContext, see https://github.com/neondatabase/neon/pull/3261#pullrequestreview-1235673708

pageserver/src/tenant/storage_layer/inmemory_layer.rs
```diff
@@ -75,13 +75,7 @@ impl InMemoryLayerInner {
...
 }
 
-impl InMemoryLayer {
-    pub fn get_timeline_id(&self) -> TimelineId {
-        self.timeline_id
-    }
-}
-
-impl Layer for InMemoryLayer {
+impl LayerRange for InMemoryLayer {
```
bojanserafimov, line 78: Is this desirable? The `InMemoryLayer` has a variable lsn range, and someone might expect `get_lsn_range` to be constant, similarly to how `hash()` should be constant for anything that goes in a hashmap. See discussion here https://github.com/neondatabase/neon/pull/2998#discussion_r1080605726 Definitely the new layer map makes such an assumption, and if `LayerRange` is the only requirement for a layer going into the layer map, then it should be required that the range remains fixed.

Maybe just return the full key range (conservative) in `get_lsn_range`, and if necessary add a separate function to narrow the range down. Or don't implement this trait for the InMemoryLayer if it's not needed.

pageserver/src/tenant/storage_layer/delta_layer.rs
```diff
@@ -599,9 +574,24 @@ impl DeltaLayer {
...
-            &self.layer_name(),
+            &self.delta_layer_name(),
         )
     }
+
+    pub fn iter(&self) -> Result<LayerIter<'_>> {
+        let inner = self.load().context("load delta layer")?;
+        Ok(match DeltaValueIter::new(inner) {
+            Ok(iter) => Box::new(iter),
+            Err(err) => Box::new(std::iter::once(Err(err))),
```
LizardWizzard, line 585: Is there a reason why we do not return plain error in this case? But an iterator containing one error value?

pageserver/src/tenant/timeline.rs
```diff
@@ -2271,18 +2267,19 @@ impl Timeline {
...
 
         Ok(layer_paths_to_upload)
     }
 }
+
 #[derive(Default)]
 struct CompactLevel0Phase1Result {
     new_layers: Vec<DeltaLayer>,
-    deltas_to_compact: Vec<Arc<dyn PersistentLayer>>,
+    deltas_to_compact: Vec<LocalOrRemote<DeltaLayer>>,
```
LizardWizzard, line 2282: Can these actually be remote in practice?

LizardWizzard (commented):
I like the direction where its going! Having less dyn is definitely a plus.

PR #3266, opened 2023-01-04
problem: new image layers become visible to GC before they have been …

problame:
…queued for upload

This commit adds comments to illustrate the race condition. I think that it can lead to data loss if the pageserver disk dies at an inopportune time.


problame:
Solved by
* #5086

PR #3275, opened 2023-01-05
Fix README and basic startup example

SomeoneToIgnore:
Follow-up of https://github.com/neondatabase/neon/pull/3270 which made an example from main README.md not working.

Fixes that, by adding a way to specify a default tenant now and modifies the basic neon_local test to start postgres and check branching.
Not all neon_local commands are implemented, so not all README.md contents is tested yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,11 +118,8 @@ Python (3.9 or higher), and install python3 packages using `./scripts/pysync` (r
 # Later that would be responsibility of a package install script
 > ./target/debug/neon_local init
 Starting pageserver at '127.0.0.1:64000' in '.neon'.
-pageserver started, pid: 2545906
-Successfully initialized timeline de200bd42b49cc1814412c7e592dd6e9
-Stopped pageserver 1 process with pid 2545906
 
-# start pageserver and safekeeper
+# start pageserver, safekeeper and broker for their intercommunication
 > ./target/debug/neon_local start
 Starting neon broker at 127.0.0.1:50051
 storage_broker started, pid: 2918372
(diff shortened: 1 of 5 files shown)


test_runner/fixtures/neon_fixtures.py
```diff
@@ -1575,36 +1576,33 @@ def create_tenant(
...
+            "create",
+            "--tenant-id",
+            str(tenant_id),
+            "--timeline-id",
+            str(timeline_id),
+            "--pg-version",
+            self.env.pg_version,
+        ]
+        if conf is not None:
+            args += sum(list(map(lambda kv: (["-c", kv[0] + ":" + kv[1]]), conf.items())), [])
```
vadim2404, line 1591: it's getting hard to read this "one-liner"

```python
from itertools import chain, product

args.extend(chain.from_iterable(product(["-c"], (f"{key}:{value}" for key, value in conf.items()))))
```

that one too, but it will not create new lists. However, I can propose

```python
from itertools import cycle

for arg in zip(cycle(["-c"]), (f"{key}:{value}" for key, value in conf.items())):
    args.extend(arg)
```

README.md
```diff
@@ -118,11 +118,8 @@ Python (3.9 or higher), and install python3 packages using `./scripts/pysync` (r
 # Later that would be responsibility of a package install script
 > ./target/debug/neon_local init
 Starting pageserver at '127.0.0.1:64000' in '.neon'.
-pageserver started, pid: 2545906
-Successfully initialized timeline de200bd42b49cc1814412c7e592dd6e9
-Stopped pageserver 1 process with pid 2545906
 
-# start pageserver and safekeeper
+# start pageserver, safekeeper and broker for their intercommunication
```
vadim2404, line 122: `start pageserver, safekeeper, and broker for their intercommunication`
comma before `and`

vadim2404 (approved):

test_runner/fixtures/neon_fixtures.py
SomeoneToIgnore: I've used the former option, thanks for the suggestion.

vadim2404 (approved):

PR #3276, opened 2023-01-06
Break wal_craft -> postgres_ffi -> wal_craft dependency cycle

SomeoneToIgnore:
rust-analyzer complained about a dependency cycle for a while

> [ERROR project_model::workspace] cyclic deps: wal_craft(CrateId(373)) -> postgres_ffi(CrateId(225)), alternative path: postgres_ffi(CrateId(225)) -> wal_craft(CrateId(373))

I'm not familiar with `wal_craft` tool, but seems that there are two main usages of it:
* the cli, that's now moved under `pageserver/src/bin` list of binaries, but could as well go into `safekeeper`. That did not have any `src/bin/` subdirectory though, ergo the choice.
* the test code in `libs/postgres_ffi/src/xlog_utils.rs` module

The PR aims to preserve both use cases and clean the dependency tree.

Additionally, it avoids `tempfile` dependency and creates unix socket tmp dir manually, see first commit.


SomeoneToIgnore:
My biggest reason for pushing this change finally is to unite `initdb` invocation and its parameters.
3rd `initdb` call is in `wal_craft`, which is simpler to fix when it is in `postgres_ffi` from where the logic of creating the common `initdb` command could be exported to pageserver as `pub fn` or similar.

SomeoneToIgnore:
Includes https://github.com/neondatabase/neon/pull/3281 after the review.

libs/postgres_ffi/src/wal_craft.rs
```diff
@@ -152,17 +148,48 @@ impl Conf {
     }
 }
 
+fn create_unix_socket_directory() -> anyhow::Result<PathBuf> {
+    let current_millis = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .context("Faled to generate random tmp dir name")?
```
vadim2404, line 154: Typo: `failed`

SomeoneToIgnore: And also did not comply the [recent discussion](https://neondb.slack.com/archives/C0277TKAJCA/p1673453329770429) results to keep the context out of parasite "Failed", "could not", etc. words.
Fixed.

pageserver/src/walredo.rs
```diff
@@ -626,35 +592,37 @@ impl PostgresRedoProcess {
...
         let pg_lib_dir_path = conf.pg_lib_dir(pg_version).map_err(|e| {
             Error::new(
                 ErrorKind::Other,
-                format!("incorrect pg_lib_dir path: {}", e),
+                format!("incorrect pg_lib_dir path: {e:#}"),
             )
         })?;
 
-        info!("running initdb in {}", datadir.display());
-        let initdb = Command::new(pg_bin_dir_path.join("initdb"))
```
problame, line 649: This one didn't call `current_dir()`, but `prepare_initdb_command` does. Intentional?

libs/postgres_ffi/src/lib.rs
```diff
@@ -184,6 +190,54 @@ pub fn fsm_logical_to_physical(addr: BlockNumber) -> BlockNumber {
...
+/// keep the same for the same branch.
+///
+/// Various env parameters influence that, same as various postgres versions may still change the
+/// values in the long run, but if this function was used, it's simpler to track such changes.
+///
+/// Ensures that there's a correct directory structure for initdb to run in.
+pub fn prepare_initdb_command(
+    pg_bin_dir: &Path,
+    pg_lib_dir: &Path,
+    target_dir: &Path,
```
problame, line 204: nit about API design: this takes three params, all with the same type. That's easy to mix up at the call site. Consider using builder pattern or introduce an `InitdbArgs` struct

pageserver/src/walredo.rs
```diff
@@ -626,35 +592,37 @@ impl PostgresRedoProcess {
...
-            .args(["-D", &datadir.to_string_lossy()])
-            .arg("-N")
-            .env_clear()
-            .env("LD_LIBRARY_PATH", &pg_lib_dir_path)
-            .env("DYLD_LIBRARY_PATH", &pg_lib_dir_path) // macOS
-            .close_fds()
-            .output()
-            .map_err(|e| Error::new(e.kind(), format!("failed to execute initdb: {e}")))?;
+        info!("running initdb in {datadir:?}");
```
problame, line 611: All callers of prepare_initdb_command to similar kind of logging.
Can we centralize this inside `prepare_initdb_command`?
Ofcourse, that woudl require `RUST_LOG` to include a rule for `postgres_ffi`.

Alternatively, introduce an `InitdbArgs` struct, make it `derive(Debug)`, and log `info!("running initdb: {args:?}")` at the caller.

pageserver/src/tenant.rs
```diff
@@ -2547,35 +2546,21 @@ fn run_initdb(
...
-        .args(["-U", &conf.superuser])
-        .args(["-E", "utf8"])
-        .arg("--no-instructions")
-        // This is only used for a temporary installation that is deleted shortly after,
-        // so no need to fsync it
-        .arg("--no-sync")
-        .env_clear()
-        .env("LD_LIBRARY_PATH", &initdb_lib_dir)
-        .env("DYLD_LIBRARY_PATH", &initdb_lib_dir)
-        .stdout(Stdio::null())
```
problame, line 2570: Why are we not using this for all callers?

problame (changes requested):
Can't say I fully understand all the implications of the various `initdb` invocations.

Anyways, found some small things to improve on.

pageserver/src/tenant.rs
SomeoneToIgnore: Technically, there's some useful stdout that might lead us to thinking:

```
~/Downloads ❯ initdb ./zzz/ 1>stdout.log 2>stderr.log

~/Downloads ❯ cat stdout.log
The files belonging to this database system will be owned by user "someonetoignore".
This user must also own the server process.

The database cluster will be initialized with locale "en_US.UTF-8".
The default database encoding has accordingly been set to "UTF8".
The default text search configuration will be set to "english".

Data page checksums are disabled.

creating directory ./zzz ... ok
creating subdirectories ... ok
selecting dynamic shared memory implementation ... posix
selecting default max_connections ... 100
selecting default shared_buffers ... 128MB
selecting default time zone ... Europe/Helsinki
creating configuration files ... ok
running bootstrap script ... ok
performing post-bootstrap initialization ... ok
syncing data to disk ... ok


Success. You can now start the database server using:

    pg_ctl -D ./zzz/ -l logfile start


~/Downloads ❯ cat stderr.log
initdb: warning: enabling "trust" authentication for local connections
You can change this by editing pg_hba.conf or using the option -A, or
--auth-local and --auth-host, the next time you run initdb.
```

For instance, I find that Helsinki TimeZone even in the programmatic initdb call logs, that could be also pinned instead.
So, I'd leave stdout enabled in the library method: if some tools are not interested in that, `Command` api could be used for disabling.

pageserver/src/walredo.rs
SomeoneToIgnore: Yes, I think this way it's safer for the library code to operate: in pageserver's case, we set `workdir` as current dir, and any kinds of creations are more or less ok, but `wal_craft` binary does not do that and might create the directory in the wrong place.

pageserver/src/walredo.rs
SomeoneToIgnore: Sure, `postgres_ffi` has `log` crate and we use adapters for that, added.

libs/postgres_ffi/src/lib.rs
SomeoneToIgnore: I have considered this, but have thought that it would overcomplicate the PR and left it out.
I still have some hope to merge it, so let's keep it as is for now, unless somebody else objects.

problame: Ok

libs/postgres_ffi/src/lib.rs
```diff
@@ -204,6 +206,8 @@ pub fn prepare_initdb_command(
     target_dir: &Path,
     user: &str,
 ) -> anyhow::Result<Command> {
+    let initdb_bin = pg_bin_dir.join("initdb");
+    info!("running {initdb_bin:?} in {target_dir:?}, libdir: {pg_lib_dir:?}");
```
problame, line 210: Nice!

So, does env var `RUST_LOG` contain  `postgres_ffi=info` in all relevant environments? (unit tests, python tests, staging, prod)?

SomeoneToIgnore: AFAIK, it's not needed since we have `info` there that should handle this enabling either.
> unit tests

Nothing there, but neighbour https://github.com/neondatabase/neon/pull/3284 should enable `info` there too.

PR #3281, opened 2023-01-06
Run initdb with the same commands across the whole storage

SomeoneToIgnore:
Based on https://github.com/neondatabase/neon/pull/3276

Pageserver runs `initdb` in two places during its work: during new timeline creation and during walredo start.
It seems that we can rework it and make pageserver to run `initdb` only once on startup/tenant creation.

For that, some preparations are better to be made, this is one of them.
Slightly improves https://github.com/neondatabase/neon/issues/2592 by using the same `initdb` command in all Rust code that calls it.


libs/utils/src/fs_ext.rs
```diff
@@ -15,6 +15,40 @@ where
...
+                //
+                // NOTE: It's easy to indirectly cause a malloc or lock a mutex,
+                // which is not async-signal-safe. Be careful.
+                close_fds::set_fds_cloexec_threadsafe(3, &[]);
+                Ok(())
+            })
+        }
+    }
+}
+
```
koivunej, line 51: Placing this in `fs_ext.rs` instead of `process_ext.rs` feels a bit off, as this doesn't have anything to do with filesystem, but you probably considered this already and I assume we don't have any other `CommandExt`.

libs/postgres_ffi/src/lib.rs
```diff
@@ -185,6 +190,53 @@ pub fn fsm_logical_to_physical(addr: BlockNumber) -> BlockNumber {
...
+        .context("Cannot run initdb on a root directory")?;
+    if !target_parent.exists() {
+        std::fs::create_dir_all(target_parent).with_context(|| {
+            format!("Failed to create parent directory {target_parent:?} for initdb target dir")
+        })?;
+    }
+
+    let mut command = Command::new(pg_bin_dir.join("initdb"));
+    command
+        .args(["-D", &target_dir.to_string_lossy()])
```
koivunej, line 223: This string conversion is unnecesssary, could just be:

```
.arg("-D")
.arg(&target_dir)
```

koivunej (commented):
Few notes.. Also a bit surprised that the initial timeline creation does still happen blockingly. I wonder if that's the case with the ongoing refactorings still?

libs/utils/src/fs_ext.rs
SomeoneToIgnore: So far, decided to avoid creating new modules and use a small trait that does an fs-related thing in the fs module.

koivunej: Yeah, this can be done later. I will probably make a version of this for tokio, and a version which handles an array of file descriptors, I can split it up then.

koivunej (approved):
Looking good.

PR #3284, opened 2023-01-06
Use actual temporary dir for pageserver unit tests

SomeoneToIgnore:
Before this PR, a `tmp_check` directory was created in the project's root for pageserver's tenant unit tests: every unit test created a subdirectory inside `tmp_check` was created.

The only reason to do so appears to be test file explorations, but in reality those files are relatively standard (hence, could be reproduced with Python tests, if needed) or mock (since there's no proper walreceiver in unit tests).
Does not make much sense to create these files as "permanent", keep them in the temporary dir that's cleaned after test.

### DIFF
--- a/pageserver/src/tenant.rs
+++ b/pageserver/src/tenant.rs
@@ -2626,9 +2626,11 @@ where
 #[cfg(test)]
 pub mod harness {
     use bytes::{Bytes, BytesMut};
+    use once_cell::sync::OnceCell;
     use std::sync::Arc;
     use std::{fs, path::PathBuf};
     use tempfile::TempDir;
+    use utils::logging;
     use utils::lsn::Lsn;
 
     use crate::{
@@ -2694,6 +2696,10 @@ pub mod harness {
 
     impl TenantHarness {
         pub fn new() -> anyhow::Result<Self> {
+            LOG_HANDLE.get_or_init(|| {
+                logging::init(logging::LogFormat::Test).expect("Failed to init test logging")
+            });
+
             let temp_repo_dir = tempfile::tempdir()?;
             // `TempDir` uses a randomly generated subdirectory of a system tmp dir,
             // so far it's enough to take care of concurrently running tests.
--- a/pageserver/src/tenant.rs
+++ b/pageserver/src/tenant.rs
@@ -2626,9 +2626,9 @@ where
 #[cfg(test)]
 pub mod harness {
     use bytes::{Bytes, BytesMut};
-    use once_cell::sync::Lazy;
-    use std::sync::{Arc, RwLock, RwLockReadGuard, RwLockWriteGuard};
+    use std::sync::Arc;
     use std::{fs, path::PathBuf};
+    use tempfile::TempDir;
     use utils::lsn::Lsn;
 
     use crate::{
@@ -2659,8 +2659,6 @@ pub mod harness {
         buf.freeze()
     }
 
-    static LOCK: Lazy<RwLock<()>> = Lazy::new(|| RwLock::new(()));
-
     impl From<TenantConf> for TenantConfOpt {
         fn from(tenant_conf: TenantConf) -> Self {
             Self {
@@ -2681,36 +2679,27 @@ pub mod harness {
         }
     }
 
-    pub struct TenantHarness<'a> {
+    /// The harness saves some boilerplate and provides a way to create functional tenant
+    /// without running pageserver binary. It uses temporary directory to store data in it.
+    /// Tempdir gets removed on harness drop.
+    pub struct TenantHarness {
+        // keep the struct to not to remove tmp dir during the test
+        _temp_repo_dir: TempDir,
         pub conf: &'static PageServerConf,
         pub tenant_conf: TenantConf,
         pub tenant_id: TenantId,
-
-        pub lock_guard: (
-            Option<RwLockReadGuard<'a, ()>>,
-            Option<RwLockWriteGuard<'a, ()>>,
-        ),
     }
 
-    impl<'a> TenantHarness<'a> {
-        pub fn create(test_name: &'static str) -> anyhow::Result<Self> {
-            Self::create_internal(test_name, false)
-        }
-        pub fn create_exclusive(test_name: &'static str) -> anyhow::Result<Self> {
-            Self::create_internal(test_name, true)
-        }
-        fn create_internal(test_name: &'static str, exclusive: bool) -> anyhow::Result<Self> {
-            let lock_guard = if exclusive {
-                (None, Some(LOCK.write().unwrap()))
-            } else {
-                (Some(LOCK.read().unwrap()), None)
-            };
+    static LOG_HANDLE: OnceCell<()> = OnceCell::new();
 
-            let repo_dir = PageServerConf::test_repo_dir(test_name);
-            let _ = fs::remove_dir_all(&repo_dir);
-            fs::create_dir_all(&repo_dir)?;
+    impl TenantHarness {
+        pub fn new() -> anyhow::Result<Self> {
+            let temp_repo_dir = tempfile::tempdir()?;
+            // `TempDir` uses a randomly generated subdirectory of a system tmp dir,
+            // so far it's enough to take care of concurrently running tests.
+            let repo_dir = temp_repo_dir.path();
 
-            let conf = PageServerConf::dummy_conf(repo_dir);
+            let conf = PageServerConf::dummy_conf(repo_dir.to_path_buf());
             // Make a static copy of the config. This can never be free'd, but that's
             // OK in a test.
             let conf: &'static PageServerConf = Box::leak(Box::new(conf));
@@ -2728,10 +2717,10 @@ pub mod harness {
             fs::create_dir_all(conf.timelines_path(&tenant_id))?;
 
             Ok(Self {
+                _temp_repo_dir: temp_repo_dir,
                 conf,
                 tenant_conf,
                 tenant_id,
-                lock_guard,
             })
         }
 
@@ -2825,7 +2814,8 @@ mod tests {
 
     #[tokio::test]
     async fn test_basic() -> anyhow::Result<()> {
-        let tenant = TenantHarness::create("test_basic")?.load().await;
+        let harness = TenantHarness::new()?;
+        let tenant = harness.load().await;
         let tline = tenant
             .create_empty_timeline(TIMELINE_ID, Lsn(0), DEFAULT_PG_VERSION)?
             .initialize()?;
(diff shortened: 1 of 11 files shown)


pageserver/src/tenant.rs
```diff
@@ -2317,6 +2317,7 @@ impl Tenant {
         let uninit_mark_path = self
             .conf
             .timeline_uninit_mark_file_path(tenant_id, timeline_id);
+        dbg!(&uninit_mark_path);
```
problame, line 2320: Remove before merge

pageserver/src/tenant.rs
```diff
@@ -2647,7 +2649,11 @@ pub mod harness {
         }
     }
 
+    /// The harness provides a way to create mock certain tenant processes for testing.
+    /// It also creates a temporary dir that serves as a test repository, that gets removed on the harness drop.
     pub struct TenantHarness<'a> {
+        // keep the struct to not to remove tmp dir during the test
+        _temp_repo_dir: TempDir,
```
problame, line 2656: Let's have an explicit `impl Drop for TenantHarness` that does `if !panicking { TempDir::close().expect("") }`. 

I want the test to make noise if it can't clean up

pageserver/src/tenant.rs
```diff
@@ -2672,8 +2678,8 @@ pub mod harness {
                 (Some(LOCK.read().unwrap()), None)
             };
 
-            let repo_dir = PageServerConf::test_repo_dir(test_name);
-            let _ = fs::remove_dir_all(&repo_dir);
+            let temp_repo_dir = tempfile::tempdir()?;
```
problame, line 2681: I think this will create the dir in `/tmp` on most platforms, some of which use tmpfs to back it.
Are we creating large repos in unit tests?
That would cause tests to consume more DRAM than before.

pageserver/src/tenant.rs
```diff
@@ -2824,9 +2832,8 @@ mod tests {
 
     #[tokio::test]
     async fn no_duplicate_timelines() -> anyhow::Result<()> {
-        let tenant = TenantHarness::create("no_duplicate_timelines")?
-            .load()
-            .await;
+        let harness = TenantHarness::create("no_duplicate_timelines")?;
```
problame, line 2835: Picking out this test (could be any of them) to demonstrate a problem with this patch:

The repository objects must not outlive the TenantHarness.
Otherwise, the harness's repository directory is used after TenantHarness::drop deletes it.
Critically, this includes the background loops of the tenant.

Before this patch, it wasn't a big problem because the temp directory name wouldn't be reused for another test.

But after this patch, tempfile::TempDir might reuse the temp directory of a previous test.

Race condition:

```
T1: start running
T1: create harness with tempdir /tmp/A
T1: start tenant background loop
T1: run test
T1: drop harness at end of test scope, deletes /tmp/A
T2: start running
T2: observes /tmp/A does not exist
T2: create harness with tempdir /tmp/A
T1: background loop does something in repository dir
T1: tokio::test stops polling runtime on which background loop runs
T1: returns
T2: observes changes that T1 made to /tmp/A, fails spuriously
```

problame (changes requested):

pageserver/src/tenant.rs
SomeoneToIgnore: I don't understand this race, in particular, how 

> But after this patch, tempfile::TempDir might reuse the temp directory of a previous test.

is possible?
We do 

```
let temp_repo_dir = tempfile::tempdir()?;
let repo_dir = temp_repo_dir.path().join(test_name);
```

which should give us different paths, given that the test names are different.

pageserver/src/tenant.rs
SomeoneToIgnore: >  Are we creating large repos in unit tests?

No, we barely have unit tests at all, to start with.

>  That would cause tests to consume more DRAM than before.

I'm ok with that behavior as a default.

pageserver/src/tenant.rs
SomeoneToIgnore: Not sure where `panicking` comes from, so just dropping.

pageserver/src/tenant.rs
problame: Fair, I didn't consider that we append the test name to it.

It's the developer's responsibility to ensure that these are unique, though. Easy mistake when copy-pasting a test. Can't we get rid of that as well?

Idea: Include a UUIDv4 in the tempdir name, a la `tempfile::TempDir::new(format!("pageserver.tenant_harness.repo.{}", gen_uuid_v4()))`.

That way, 
1. devs don't need to keep `test_name` unique. We can get rid of it.
2. with the UUIDv4 in the prefix, we don't need to rely on `tempdir` to not re-use directory names

SomeoneToIgnore:
```rust
fn main() {
    let temp_repo_dir = tempfile::tempdir().unwrap();
    dbg!(temp_repo_dir.into_path());
}
```

boils down to https://github.com/Stebalien/tempfile/blob/1c6149d7ce33b2ad8b20a38a8ddce0fc29f68f0e/src/util.rs#L18

and prints on mac:
```
[src/main.rs:3] temp_repo_dir.into_path() = "/var/folders/80/7bn887rd1sj2gt1z64tmrbx80000gn/T/.tmpdIDO7Q"
```
on linux:
```
[src/main.rs:3] temp_repo_dir.into_path() = "/tmp/.tmpm0OMA4"
```

So in terms of races and uniqueness, we should be fine.
Even `TempDir` drops are fine too, since they delete the randomly generated, exclusive, subdirs only.

Specifying the name in general seems redundant indeed, I've removed that.
As a bonus, I've also enabled logs in such tests.

problame: > So in terms of races and uniqueness, we should be fine.
> Even TempDir drops are fine too, since they delete the randomly generated, exclusive, subdirs only.

Yeah, and after that, the name can be re-used.
At least on Linux, that's definitely possible.

So, if the path is remembered in some leaked `async task`, we could theoretically hit the race condition outlined earlier.

problame:
Re commit [Enable logs in unit tests](https://github.com/neondatabase/neon/pull/3284/commits/03292db6bd2fe27c70431f93458aa2891fb6c0f5)

Would be nice if that could be a separate PR

pageserver/src/tenant.rs
SomeoneToIgnore: How probable is this to you?
To me, `.tmpm0OMA4` looks like a good enthropy for the amount of tests we have.
I also think, tempfile is the place to solve this, not here.

SomeoneToIgnore:
> Would be nice if that could be a separate PR

Would it be ok to leave it as a separate commit here? 
I have plenty of open PRs already and this is small.

pageserver/src/tenant.rs
problame: Allright.
Docs say it's 6 bytes of entropy for the name. That's less than a 128bit UUID but probably enough :) 

https://docs.rs/tempfile/latest/tempfile/struct.Builder.html#method.rand_bytes

pageserver/src/tenant/ephemeral_file.rs
```diff
@@ -367,8 +368,10 @@ mod tests {
     }
 
     #[test]
-    fn test_ephemeral_files() -> Result<(), io::Error> {
-        let (conf, tenant_id, timeline_id) = harness("ephemeral_files")?;
+    fn test_ephemeral_files() -> io::Result<()> {
+        let tmp_dir = tempfile::tempdir()?;
+        let repo_dir = tmp_dir.path().join("ephemeral_files");
```
problame, line 373: Can we just used `tempfile::tempdir()` inside `harness()`, thereby avoiding repetition of the test name here?

Same for `test_ephemeral_blobs`

problame (approved):
Nice cleanup, thanks for addressing all the comments :)

pageserver/src/tenant/ephemeral_file.rs
LizardWizzard: This also should eliminate `!repo_dir.exists()` check

pageserver/src/tenant.rs
```diff
@@ -2647,36 +2647,46 @@ pub mod harness {
         }
     }
 
-    pub struct TenantHarness<'a> {
+    /// The harness provides a way to create mock certain tenant processes for testing.
```
LizardWizzard, line 2650: Wording issues. I suggest something like that:

```
The harness saves some boilerplate and provides a way to create functional tenant
without running pageserver binary. It uses temporary directory to store data in it.
Tempdir gets removed on harness drop. 
```

pageserver/src/tenant.rs
```diff
@@ -2647,36 +2647,46 @@ pub mod harness {
...
+    /// It also creates a temporary dir that serves as a test repository, that gets removed on the harness drop.
+    pub struct TenantHarness {
+        // keep the struct to not to remove tmp dir during the test
+        temp_repo_dir: Option<TempDir>,
```
LizardWizzard, line 2654: I'm not sure I get it why it is an option (seems to compile fine without it) and do we really need the drop code? Dir will be cleaned on drop automatically

pageserver/src/tenant.rs
```diff
@@ -2647,36 +2647,46 @@ pub mod harness {
...
+            });
 
-            let repo_dir = PageServerConf::test_repo_dir(test_name);
-            let _ = fs::remove_dir_all(&repo_dir);
-            fs::create_dir_all(&repo_dir)?;
+            let temp_repo_dir = tempfile::tempdir()?;
+            // `TempDir` uses a randomly generated subdirectory of a system tmp dir,
+            // so far it's enough to take care of concurrently running tests.
+            let repo_dir = temp_repo_dir.path();
+            fs::create_dir_all(repo_dir)?;
```
LizardWizzard, line 2687: `create_dir_all` looks redundant since tempdir actually creates it, not just generates a path

LizardWizzard (commented):
I like the idea overall. For my pet project I was thinking about using `target` dir for that purpose. Using tempdir looks like a better option. Might be worth adding some --keep flag or env var for better debugging ux

pageserver/src/tenant/ephemeral_file.rs
SomeoneToIgnore: Reused the harness, thanks.

pageserver/src/tenant.rs
SomeoneToIgnore: Applied the patch, thanks.

SomeoneToIgnore:
> some --keep flag

That would complicate the logic and recent PRs are somewhat slow, so I'd keep this simple for now and if later we'd need this, add separately.

LizardWizzard (approved):
Looks good to me, thanks!

Issue #3160, opened 2022-12-20
Log transitions to broken state, so search by "Broken" can give first error message with the reason.


shanyp:

shanyp:
Fixed in #3289

PR #3289, opened 2023-01-08
Adding transition reason to the log when a tenant is moved to Broken state 

shanyp:
#3160

### DIFF
--- a/pageserver/src/tenant.rs
+++ b/pageserver/src/tenant.rs
@@ -1505,19 +1505,22 @@ impl Tenant {
                     // activated should never be marked as broken. We cope with it the best
                     // we can, but it shouldn't happen.
                     *current_state = TenantState::Broken;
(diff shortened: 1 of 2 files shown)


pageserver/src/tenant.rs
```diff
@@ -1505,7 +1505,7 @@ impl Tenant {
                     // activated should never be marked as broken. We cope with it the best
                     // we can, but it shouldn't happen.
                     *current_state = TenantState::Broken;
-                    warn!("Changing Active tenant to Broken state");
+                    warn!("Changing Active tenant to Broken state, reason:{}", reason);
```
knizhnik, line 1508: Will it be better to add space after column? "Such:message" IMHO is harder to read "than: such".

knizhnik (approved):

hlinnaka (approved):

PR #3290, opened 2023-01-09
Proxy metrics

lubennikovaav:
Implement the worker that periodically collects and sends proxy consumption metrics to the HTTP endpoint.
See spec in https://github.com/neondatabase/cloud/issues/3475

TODO (Need help)
- [x] Write a test. 
Locally running proxy used in python tests doesn't fill `endpoint_id` and other `traffic_labels`, so I don't know how to properly test it.. 
- [x] Set correct `node_id` for proxy. What should I use as a proxy's node_id?
- [x] fix metric collection thread cancellation
- [x] clarify API. Only send metrics for outbound traffic (tx). As discussed with @antonyc 
 ~~What should proxy report: separate metrics for each direction ( "rx"/"tx") or just one of them?~~

### DIFF
--- a/libs/consumption_metrics/src/lib.rs
+++ b/libs/consumption_metrics/src/lib.rs
@@ -1,0 +1,50 @@
+//!
+//! Shared code for consumption metics collection
+//!
+use chrono::{DateTime, Utc};
+use rand::Rng;
+use serde::Serialize;
+
+#[derive(Serialize, Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
+#[serde(tag = "type")]
+pub enum EventType {
+    #[serde(rename = "absolute")]
+    Absolute { time: DateTime<Utc> },
+    #[serde(rename = "incremental")]
+    Incremental {
+        start_time: DateTime<Utc>,
+        stop_time: DateTime<Utc>,
+    },
+}
+
+#[derive(Serialize, Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
+pub struct Event<Extra> {
+    #[serde(flatten)]
+    #[serde(rename = "type")]
+    pub kind: EventType,
+
+    pub metric: &'static str,
+    pub idempotency_key: String,
+    pub value: u64,
+
+    #[serde(flatten)]
+    pub extra: Extra,
+}
+
+pub fn idempotency_key(node_id: String) -> String {
+    format!(
+        "{}-{}-{:04}",
+        Utc::now(),
+        node_id,
+        rand::thread_rng().gen_range(0..=9999)
+    )
+}
+
+pub const CHUNK_SIZE: usize = 1000;
+
+// Just a wrapper around a slice of events
+// to serialize it as `{"events" : [ ] }
+#[derive(serde::Serialize)]
+pub struct EventChunk<'a, T> {
+    pub events: &'a [T],
+}
--- a/pageserver/src/consumption_metrics.rs
+++ b/pageserver/src/consumption_metrics.rs
@@ -3,142 +3,42 @@
 //! and push them to a HTTP endpoint.
 //! Cache metrics to send only the updated ones.
 //!
-
-use anyhow;
-use tracing::*;
-use utils::id::NodeId;
-use utils::id::TimelineId;
-
-use crate::task_mgr;
-use crate::task_mgr::TaskKind;
-use crate::task_mgr::BACKGROUND_RUNTIME;
-
+use crate::task_mgr::{self, TaskKind, BACKGROUND_RUNTIME};
 use crate::tenant::mgr;
+use anyhow;
+use chrono::Utc;
+use consumption_metrics::{idempotency_key, Event, EventChunk, EventType, CHUNK_SIZE};
 use pageserver_api::models::TenantState;
-use utils::id::TenantId;
-
-use serde::{Deserialize, Serialize};
+use reqwest::Url;
+use serde::Serialize;
 use serde_with::{serde_as, DisplayFromStr};
 use std::collections::HashMap;
-use std::fmt;
-use std::str::FromStr;
 use std::time::Duration;
+use tracing::*;
+use utils::id::{NodeId, TenantId, TimelineId};
 
-use chrono::{DateTime, Utc};
-use rand::Rng;
-use reqwest::Url;
+const WRITTEN_SIZE: &str = "written_size";
+const SYNTHETIC_STORAGE_SIZE: &str = "synthetic_storage_size";
+const RESIDENT_SIZE: &str = "resident_size";
+const REMOTE_STORAGE_SIZE: &str = "remote_storage_size";
+const TIMELINE_LOGICAL_SIZE: &str = "timeline_logical_size";
 
-/// ConsumptionMetric struct that defines the format for one metric entry
-/// i.e.
-///
-/// ```json
-/// {
-/// "metric": "remote_storage_size",
-/// "type": "absolute",
-/// "tenant_id": "5d07d9ce9237c4cd845ea7918c0afa7d",
-/// "timeline_id": "a03ebb4f5922a1c56ff7485cc8854143",
-/// "time": "2022-12-28T11:07:19.317310284Z",
-/// "idempotency_key": "2022-12-28 11:07:19.317310324 UTC-1-4019",
-/// "value": 12345454,
-/// }
-/// ```
 #[serde_as]
-#[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
-pub struct ConsumptionMetric {
-    pub metric: ConsumptionMetricKind,
-    #[serde(rename = "type")]
-    pub metric_type: &'static str,
+#[derive(Serialize)]
+struct Ids {
     #[serde_as(as = "DisplayFromStr")]
-    pub tenant_id: TenantId,
+    tenant_id: TenantId,
     #[serde_as(as = "Option<DisplayFromStr>")]
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub timeline_id: Option<TimelineId>,
-    pub time: DateTime<Utc>,
-    pub idempotency_key: String,
-    pub value: u64,
-}
-
-impl ConsumptionMetric {
-    pub fn new_absolute<R: Rng + ?Sized>(
-        metric: ConsumptionMetricKind,
-        tenant_id: TenantId,
-        timeline_id: Option<TimelineId>,
-        value: u64,
-        node_id: NodeId,
-        rng: &mut R,
-    ) -> Self {
-        Self {
-            metric,
-            metric_type: "absolute",
-            tenant_id,
-            timeline_id,
-            time: Utc::now(),
-            // key that allows metric collector to distinguish unique events
-            idempotency_key: format!("{}-{}-{:04}", Utc::now(), node_id, rng.gen_range(0..=9999)),
-            value,
-        }
-    }
-}
-
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Ord, PartialOrd, Serialize, Deserialize)]
-#[serde(rename_all = "snake_case")]
-pub enum ConsumptionMetricKind {
-    /// Amount of WAL produced , by a timeline, i.e. last_record_lsn
-    /// This is an absolute, per-timeline metric.
-    WrittenSize,
-    /// Size of all tenant branches including WAL
-    /// This is an absolute, per-tenant metric.
-    /// This is the same metric that tenant/tenant_id/size endpoint returns.
-    SyntheticStorageSize,
-    /// Size of all the layer files in the tenant's directory on disk on the pageserver.
-    /// This is an absolute, per-tenant metric.
-    /// See also prometheus metric RESIDENT_PHYSICAL_SIZE.
-    ResidentSize,
-    /// Size of the remote storage (S3) directory.
-    /// This is an absolute, per-tenant metric.
-    RemoteStorageSize,
-    /// Logical size of the data in the timeline
-    /// This is an absolute, per-timeline metric
-    TimelineLogicalSize,
-}
-
-impl FromStr for ConsumptionMetricKind {
-    type Err = anyhow::Error;
-
-    fn from_str(s: &str) -> Result<Self, Self::Err> {
-        match s {
-            "written_size" => Ok(Self::WrittenSize),
-            "synthetic_storage_size" => Ok(Self::SyntheticStorageSize),
-            "resident_size" => Ok(Self::ResidentSize),
-            "remote_storage_size" => Ok(Self::RemoteStorageSize),
-            "timeline_logical_size" => Ok(Self::TimelineLogicalSize),
-            _ => anyhow::bail!("invalid value \"{s}\" for metric type"),
-        }
-    }
-}
-
-impl fmt::Display for ConsumptionMetricKind {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.write_str(match self {
-            ConsumptionMetricKind::WrittenSize => "written_size",
-            ConsumptionMetricKind::SyntheticStorageSize => "synthetic_storage_size",
-            ConsumptionMetricKind::ResidentSize => "resident_size",
-            ConsumptionMetricKind::RemoteStorageSize => "remote_storage_size",
-            ConsumptionMetricKind::TimelineLogicalSize => "timeline_logical_size",
-        })
-    }
-}
-
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
-pub struct ConsumptionMetricsKey {
-    tenant_id: TenantId,
     timeline_id: Option<TimelineId>,
-    metric: ConsumptionMetricKind,
 }
 
-#[derive(serde::Serialize)]
-struct EventChunk<'a> {
-    events: &'a [ConsumptionMetric],
+/// Key that uniquely identifies the object, this metric describes.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct PageserverConsumptionMetricsKey {
+    pub tenant_id: TenantId,
+    pub timeline_id: Option<TimelineId>,
+    pub metric: &'static str,
 }
 
 /// Main thread that serves metrics collection
@@ -170,7 +70,7 @@ pub async fn collect_metrics(
 
     // define client here to reuse it for all requests
     let client = reqwest::Client::new();
-    let mut cached_metrics: HashMap<ConsumptionMetricsKey, u64> = HashMap::new();
+    let mut cached_metrics: HashMap<PageserverConsumptionMetricsKey, u64> = HashMap::new();
 
     loop {
         tokio::select! {
@@ -179,7 +79,7 @@ pub async fn collect_metrics(
                 return Ok(());
             },
             _ = ticker.tick() => {
-                if let Err(err) = collect_metrics_task(&client, &mut cached_metrics, metric_collection_endpoint, node_id).await
+                if let Err(err) = collect_metrics_iteration(&client, &mut cached_metrics, metric_collection_endpoint, node_id).await
                 {
                     error!("metrics collection failed: {err:?}");
                 }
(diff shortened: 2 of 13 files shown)


libs/consumption_metrics/src/lib.rs
```diff
@@ -0,0 +1,210 @@
...
+///
+/// ```json
+/// {
+/// "metric": "remote_storage_size",
+/// "type": "absolute",
+/// "tenant_id": "5d07d9ce9237c4cd845ea7918c0afa7d",
+/// "timeline_id": "a03ebb4f5922a1c56ff7485cc8854143",
+/// "time": "2022-12-28T11:07:19.317310284Z",
+/// "idempotency_key": "2022-12-28 11:07:19.317310324 UTC-1-4019",
+/// "value": 12345454,
```
vadim2404, line 30: `NIT`: trailing comma makes JSON invalid

lubennikovaav: fixed

proxy/src/proxy_metrics.rs
```diff
@@ -0,0 +1,188 @@
...
+        metric_collection_endpoint
+    );
+
+    // define client here to reuse it for all requests
+    let client = reqwest::Client::new();
+    let mut cached_metrics: HashMap<ProxyConsumptionMetricsKey, (u64, DateTime<Utc>)> =
+        HashMap::new();
+
+    loop {
+        tokio::select! {
```
lubennikovaav, line 39: @funbringer 
could you please double check this code?

I'm not sure that I fully understand how `try_join_all` works.
Do I need to add some extra check for cancellation in this loop to ensure that it exits or it will happen automagically?

proxy/src/main.rs
```diff
@@ -64,6 +65,18 @@ async fn main() -> anyhow::Result<()> {
...
+            Some(s) => Some(s.parse()?),
+            None => None,
+        };
+
+    let metric_collection_interval: std::time::Duration = humantime::parse_duration(
+        arg_matches
+            .get_one::<String>("metric-collection-interval")
+            .unwrap(),
+    )
+    .expect("cannot parse metric collection interval");
```
funbringer, line 79: Please don't use `expect` like this. I think `?` is more appropriate here.

proxy/src/main.rs
```diff
@@ -95,6 +108,8 @@ async fn main() -> anyhow::Result<()> {
     let config: &ProxyConfig = Box::leak(Box::new(ProxyConfig {
         tls_config,
         auth_backend,
+        metric_collection_endpoint,
+        metric_collection_interval,
```
funbringer, line 112: There's a tls config and an auth backend config, which leads me to think that there should be metric collection config containing those two fields.

I'm not into prefixes.

proxy/src/main.rs
```diff
@@ -126,6 +141,19 @@ async fn main() -> anyhow::Result<()> {
         )));
     }
 
+    if let Some(metric_collection_endpoint) = &config.metric_collection_endpoint {
+        let hostname = hostname::get().unwrap().into_string().unwrap();
```
funbringer, line 145: Please don't use unwrap. This function returns anyhow::Result. Either use `.context(...)?` or just `?`.

proxy/src/main.rs
```diff
@@ -126,6 +141,19 @@ async fn main() -> anyhow::Result<()> {
...
+
+        tasks.push(tokio::spawn(
+            proxy_metrics::collect_metrics(
+                metric_collection_endpoint,
+                config.metric_collection_interval,
```
funbringer, line 150: I think we should extract this into `Option<MetricConfig>`. interval is useless without endpoint and vice versa.

test_runner/fixtures/neon_fixtures.py
```diff
@@ -2301,6 +2301,20 @@ def extra_args(self) -> list[str]:
...
+            self.metric_collection_interval = metric_collection_interval
+
+        def extra_args(self) -> list[str]:
+            return [
+                # Link auth backend params
+                *["--auth-backend", "link"],
+                *["--uri", NeonProxy.link_auth_uri],
+                *["--metric-collection-endpoint", self.metric_collection_endpoint],
+                *["--metric-collection-interval", self.metric_collection_interval],
+            ]
```
funbringer, line 2316: This is misguided. There's no such thing as `link auth with metrics collector`. `AuthBackend` should've been a union of Postgres/Link/whatever, but there's no such notion in python, so we have to use inheritance. Those two params are universal and should be provided in init (default is None).

workspace_hack/Cargo.toml
```diff
@@ -15,7 +15,7 @@ publish = false
 [dependencies]
 anyhow = { version = "1", features = ["backtrace", "std"] }
 bytes = { version = "1", features = ["serde", "std"] }
-chrono = { version = "0.4", default-features = false, features = ["clock", "iana-time-zone", "serde", "std", "winapi"] }
+chrono = { version = "0.4", features = ["clock", "iana-time-zone", "js-sys", "oldtime", "serde", "std", "time", "wasm-bindgen", "wasmbind", "winapi"] }
```
funbringer, line 18: Why add new wasm-related features? This doesn't look right.

funbringer (commented):
Thanks for the PR. I'll add more comments later, this is just a beginning.

workspace_hack/Cargo.toml
lubennikovaav: fixed
it was just unneeded chrono dependency

libs/consumption_metrics/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+    /// This is an absolute, per-tenant metric.
+    RemoteStorageSize,
+    /// Logical size of the data in the timeline
+    /// This is an absolute, per-timeline metric
+    TimelineLogicalSize,
+    /// Number of bytes proxied
+    /// This is an incremental per endpoint proxy metric
+    /// see also NUM_BYTES_PROXIED_COUNTER
+    ProxyIOBytesPerClient,
+}
```
funbringer, line 157: I still don't quite get why we need to define this in `consumption_metrics`. This library is a dependency and as such can't be responsible for enumerating all possible metrics of its users. Besides, that'd mean that e.g. `ProxyIOBytesPerClient` is available in pageserver, which doesn't make sense.

Instead, I propose do define metric names in a following manner:

```rust
// proxy/src/proxy/metrics.rs
const PROXY_IO_BYTES_PER_CLIENT: &str = "...";
...
```

Then it'd be possible to reuse those names everywhere in the proxy (the same logic applies elsewhere).

In `PageserverConsumptionMetric` and `ProxyConsumptionMetric` we'd have

```rust
// because the name should be a compile-time constant
pub metric: &'static str,
```

Furthermore, `PageserverConsumptionMetric` & `ProxyConsumptionMetric` look like twins. It should be possible to unite them and even extract the common denominator. Here's how I would do that:

```rust
use chrono::{DateTime, Utc};
use serde::Serialize;

fn main() {
    // This could be literally anything as long as it implements Serialize.
    // I'd put contextually-dependedent things here (e.g. things which make
    // sense for the proxy but not pageserver).
    //
    // I've noticed that in the cloud OpenAPI spec absolute events use
    // tenant_id while incremental events use endpoint_id, which to me
    // doesn't make sense. However, if that's how it should truly work,
    // we could move them to EventType.
    //
    // If there are any fields that I forgot to describe, we should
    // embed them into Event.
    #[derive(Serialize)]
    struct Ids {
        tenant_id: String,
        timeline_id: String,
    }

    let event = Event {
        kind: EventType::Incremental {
            start_time: Utc::now(),
            stop_time: Utc::now(),
        },
        metric: "metric name goes here",
        idempotency_key: "...".into(),
        value: 100500,
        extra: Ids {
            tenant_id: "foo".into(),
            timeline_id: "bar".into(),
        },
    };

    println!("{}", serde_json::to_string(&event).unwrap());
}

#[derive(Serialize, Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
#[serde(tag = "type")]
pub enum EventType {
    #[serde(rename = "absolute")]
    Absolute {
        time: DateTime<Utc>,
    },
    #[serde(rename = "incremental")]
    Incremental {
        start_time: DateTime<Utc>,
        stop_time: DateTime<Utc>,
    },
}

#[derive(Serialize, Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
pub struct Event<Extra> {
    #[serde(flatten)]
    pub kind: EventType,

    pub metric: &'static str,
    pub idempotency_key: String,
    pub value: u64,

    #[serde(flatten)]
    pub extra: Extra,
}
```

```
$ cargo run
   Compiling jsons v0.1.0 (/tmp/jsons)
    Finished dev [unoptimized + debuginfo] target(s) in 0.48s
     Running `target/debug/jsons`
{"type":"incremental","start_time":"2023-01-12T12:39:06.596256607Z","stop_time":"2023-01-12T12:39:06.596262568Z","metric":"metric name goes here","idempotency_key":"...","value":100500,"tenant_id":"foo","timeline_id":"bar"}
```

One of the benefits is that it closely mirrors the OpenAPI spec and is substantially easier to comprehend.

proxy/src/proxy_metrics.rs
funbringer: Not sure about `try_join_all`, it's not present here. Could you elaborate?

Anyways, currently there's no special cancellation logic in the proxy, so that shouldn't be a problem.

Speaking of `select!`, you shouldn't need here because you're polling just one future. So it should be:

```rust
    loop {
        let _ = ticker.tick().await;

        let res = collect_metrics_task(
            &client,
            &mut cached_metrics,
            metric_collection_endpoint,
            hostname.clone(),
        )
        .await;

        match res {
            Err(e) => {
                error!("Failed to send consumption metrics: {} ", e);
            }
            Ok(_) => {
                trace!("collect_metrics_task completed successfully")
            }
        }
    }
```

However, let's take a closer look at the error handling part. Do you think we should terminate proxy if we are unable to collect or send metrics? I don't think that's a good idea, because clients are unlikely to be happy about that. We should consume the error and log it.

proxy/src/proxy_metrics.rs
```diff
@@ -0,0 +1,188 @@
...
+                    (value, Utc::now()),
+                ));
+            }
+        }
+    }
+
+    current_metrics
+}
+
+pub async fn collect_metrics_task(
```
funbringer, line 96: naming nit: It's not a task, because "task" means the top-level future (aka green thread).

proxy/src/proxy_metrics.rs
```diff
@@ -0,0 +1,188 @@
...
+    client: &reqwest::Client,
+    cached_metrics: &mut HashMap<ProxyConsumptionMetricsKey, (u64, DateTime<Utc>)>,
+    metric_collection_endpoint: &reqwest::Url,
+    hostname: String,
```
funbringer, line 100: Why String instead of &str though?

libs/consumption_metrics/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct ProxyConsumptionMetricsKey {
+    pub endpoint_id: String,
+    pub direction: String,
+}
+
+#[derive(serde::Serialize)]
+pub struct ProxyEventChunk<'a> {
+    pub events: &'a [ProxyConsumptionMetric],
+}
```
funbringer, line 221: I don't think we need this. We wouldn't be able to derive Serialize if there was no way to serialize the slice itself.

proxy/src/proxy_metrics.rs
```diff
@@ -0,0 +1,188 @@
...
+                    .find(|l| l.get_name() == "direction")
+                    .unwrap()
+                    .get_value();
+
+                let endpoint_id = ms
+                    .get_label()
+                    .iter()
+                    .find(|l| l.get_name() == "endpoint_id")
+                    .unwrap()
+                    .get_value();
```
funbringer, line 74: It's possible to do this in one traversal using `for` and two options.

libs/consumption_metrics/src/lib.rs
lubennikovaav: This is a way to serialize chunks into jsons of the following format
`{"events" : [ {***}, {***} ] }`
instead of manual json constuction.

funbringer: Ah, right, I forgot about this. Still, this doesn't have to be tightly coupled with proxy's metrics:

```rust
#[derive(serde::Serialize)]
struct Events<'a, T> {
    pub events: &'a [T],
}
```

proxy/src/proxy_metrics.rs
lubennikovaav: sorry for the confusing question.
I was referring to [this code](https://github.com/neondatabase/neon/blob/9d6f88360eb4703ac42e293276307b083fd73e99/proxy/src/main.rs#L167) in in proxy main.rs
```
   // This will block until all tasks have completed.
```

but `collect_metrics` future just loops forever and doesn't check for any shutdown signal.
Will it be somehow cancelled by tokio when other futures from the set fail, or I need to implement shutdown logic in `collect_metrics` so that main proxy thread would not hang forever?

proxy/src/proxy_metrics.rs
lubennikovaav: renamed to `collect_metrics_iteration`

libs/consumption_metrics/src/lib.rs
lubennikovaav: Thank you for the suggestion!
I pushed the refactoring

proxy/src/proxy_metrics.rs
```diff
@@ -0,0 +1,188 @@
...
+                            // cache new metric
+                            .or_insert((send_metric.value, send_metric.stop_time));
+                    }
+                } else {
+                    error!("metrics endpoint refused the sent metrics: {:?}", res);
+                }
+            }
+            Err(err) => {
+                error!("failed to send metrics: {:?}", err);
+            }
```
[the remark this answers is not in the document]
lubennikovaav: Sure. We've discussed it before and agreed that retries can be implemented later.
We send metrics often enough and cache only what was sent successfully, so small glitches shouldn't be a big deal.

proxy/src/proxy_metrics.rs
lubennikovaav: there are not too many labels, so I'd keep it as is.
to me that looks more readable than alternative

libs/consumption_metrics/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+//!
+
+use utils::id::TimelineId;
+
+use utils::id::TenantId;
+
+use serde::Serialize;
+
+use chrono::{DateTime, Utc};
+use rand::Rng;
```
funbringer, line 12: nit: needless spaces; worth checking other files too.

libs/consumption_metrics/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+///
+/// Both the proxy and the ingestion endpoint will live in the same region (or cell)
+/// so while the project-id is unique across regions the whole pipeline will work correctly
+/// because we enrich the event with project_id in the control-plane endpoint.
+///
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct ProxyConsumptionMetricsKey {
+    pub endpoint_id: String,
+    pub direction: String,
+}
```
funbringer, line 110: We could do this:

```rust
pub enum ProxyConsumptionMetricsKey {
  Tx { endpoint_id: String },
  Rx { endpoint_id: String },
}
```

libs/consumption_metrics/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+    )
+}
+
+/// Key that uniquely identifies the object, this metric describes.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct PageserverConsumptionMetricsKey {
+    pub tenant_id: TenantId,
+    pub timeline_id: Option<TimelineId>,
+    pub metric: &'static str,
+}
```
funbringer, line 88: What do you think about moving this to the pageserver? Same for `ProxyConsumptionMetricsKey`. That'd make this crate even smaller, but I'm reasonably confident that there is some logic that we could put here (maybe later).

proxy/src/metrics.rs
```diff
@@ -0,0 +1,200 @@
...
+                            // cache new metric
+                            .or_insert((send_metric.value, stop_time));
+                    }
+                } else {
+                    error!("metrics endpoint refused the sent metrics: {:?}", res);
+                }
+            }
+            Err(err) => {
+                error!("failed to send metrics: {:?}", err);
+            }
```
funbringer, line 195: It's possible to decrease nesting level here:

```rust
for ... {
    let Ok(res) = res else {
        error!("failed to send metrics: {:?}", err);
        continue;
    };
    
    // yay, res is available here!
}
```

funbringer (commented):
Thanks! We're almost there :+1:

proxy/src/metrics.rs
lubennikovaav: nice!
though, how do I get the `err` using this syntax?

libs/consumption_metrics/src/lib.rs
lubennikovaav: I removed this struct at all, as it is replaced by `Ids`.

proxy/src/metrics.rs
funbringer: Ah, right... I'm getting tired.

```rust
let res = match res {
  Ok(x) => x,
  Err(e) => { error!(...); continue; }
}
```

funbringer (approved):
LGTM, thank you! Now we just need to rebase.

pageserver/src/consumption_metrics.rs
```diff
@@ -300,35 +200,29 @@ pub async fn collect_metrics_task(
 
     // Send metrics.
     // Split into chunks of 1000 metrics to avoid exceeding the max request size
-    const CHUNK_SIZE: usize = 1000;
     let chunks = current_metrics.chunks(CHUNK_SIZE);
 
-    let mut chunk_to_send: Vec<ConsumptionMetric> = Vec::with_capacity(1000);
+    let mut chunk_to_send: Vec<Event<Ids>> = Vec::with_capacity(1000);
```
problame, line 205: `1000` magic constant. Is this `CHUNK_SIZE`?

Also, should update the `1000` in the comment above.

pageserver/src/consumption_metrics.rs
```diff
@@ -300,35 +200,29 @@ pub async fn collect_metrics_task(
 
     // Send metrics.
```
problame, line 201: Why can't we share everything from here on with `proxy`? Looks quite similar to me.

problame (commented):
Only looked at pageserver changes.

pageserver/src/consumption_metrics.rs
lubennikovaav: umm, we collect metrics in a different way and they have slightly different format.

all shared code is already moved in lib/consumption_metrics

pageserver/src/consumption_metrics.rs
lubennikovaav: thank you! fixed

pageserver/src/consumption_metrics.rs
```diff
@@ -300,35 +200,29 @@ pub async fn collect_metrics_task(
...
     // Split into chunks of 1000 metrics to avoid exceeding the max request size
-    const CHUNK_SIZE: usize = 1000;
     let chunks = current_metrics.chunks(CHUNK_SIZE);
```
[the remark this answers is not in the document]
lubennikovaav: i'd leave it as is. 
qualified name looks a bit too wordy to me

pageserver/src/consumption_metrics.rs
problame: I'm speaking of just the chunking + sending part. Those look the same to my eyes.

I won't die on that hill, though.

pageserver/src/consumption_metrics.rs
problame: Comment  is still missing:

```
// Split into chunks of 1000 metrics to avoid exceeding the max request size
```

problame (approved):

pageserver/src/consumption_metrics.rs
funbringer: I like @problame's proposal though. Can we leave a TODO comment? I'm sure somebody will pick this up later.

funbringer (approved):
You probably need to update helm charts to cover new cmdline args. However, this can be done a separate PR.

PR #3291, opened 2023-01-09
Fix log message matching

SomeoneToIgnore:
Spotted https://neon-github-public-dev.s3.amazonaws.com/reports/main/debug/3871991071/index.html#suites/158be07438eb5188d40b466b6acfaeb3/22966d740e33b677/ failing on `main`, fixes that by using a proper regex match string.

Also removes one clippy lint suppression.

### DIFF
--- a/test_runner/fixtures/neon_fixtures.py
+++ b/test_runner/fixtures/neon_fixtures.py
@@ -1923,7 +1923,7 @@ class NeonPageserver(PgProtocol):
             ".*kill_and_wait_impl.*: wait successful.*",
             ".*Replication stream finished: db error: ERROR: Socket IO error: end streaming to Some.*",
             ".*query handler for 'pagestream.*failed: Broken pipe.*",  # pageserver notices compute shut down
-            ".*query handler for 'pagestream.*failed: Connection reset by peer (os error 104).*",  # pageserver notices compute shut down
+            ".*query handler for 'pagestream.*failed: Connection reset by peer.*",  # pageserver notices compute shut down
             # safekeeper connection can fail with this, in the window between timeline creation
             # and streaming start
             ".*Failed to process query for timeline .*: state uninitialized, no data to read.*",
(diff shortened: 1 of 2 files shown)


koivunej (approved):

test_runner/fixtures/neon_fixtures.py
```diff
@@ -1923,7 +1923,7 @@ def __init__(self, env: NeonEnv, port: PageserverPort, config_override: Optional
             ".*kill_and_wait_impl.*: wait successful.*",
             ".*Replication stream finished: db error: ERROR: Socket IO error: end streaming to Some.*",
             ".*query handler for 'pagestream.*failed: Broken pipe.*",  # pageserver notices compute shut down
-            ".*query handler for 'pagestream.*failed: Connection reset by peer (os error 104).*",  # pageserver notices compute shut down
+            ".*query handler for 'pagestream.*failed: Connection reset by peer.*",  # pageserver notices compute shut down
```
koivunej, line 1926: Could you actually clarify how will this help with the original issue?

I glanced this quickly, was thinking this was because of the error logging PR I read through earlier, but now looking I cannot understand why wouldn't the older one match the message just as well as the new one.

Orig message:

```
AssertionError: assert not ["2023-01-09T08:51:47.253480Z ERROR query handler for 'pagestream b777f612133b871f934fd8d23c31053f 911acef9cea27d9b23b6fbd853f7e73b' failed: Connection reset by peer (os error 104)\n"]
```

SomeoneToIgnore: This would also work:

> ".*query handler for 'pagestream.*failed: Connection reset by peer \(os error 104\).*",  # pageserver notices compute shut down

so the issue is with unescaped `()` treated not as text parentheses.
I've decided to remove the code at all, since all I/O errors (well, connection resets by peers for sure) are expected.

koivunej: > issue is with unescaped `()`

I failed to see this, blaming the time off from working :)

koivunej (approved):

PR #3293, opened 2023-01-09
Use GHA environment for per-region deploy approvals on staging

SergeyMelnikov:
Each main deploy will wait for manual approval for each region


shanyp (approved):

shanyp:
Wondering if we want to leave it for staging as well or disable once we have it for production environment

PR #3294, opened 2023-01-09
add script to download all remote layers

problame:
For use in production in case on-demand download turns out to be problematic during tenant_attach, or when we eventually introduce layer eviction.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,6 +33,7 @@ psutil = "^5.9.4"
 types-psutil = "^5.9.5.4"
 types-toml = "^0.10.8"
 pytest-httpserver = "^1.0.6"
+aiohttp = "^3.8.3"
 
 [tool.poetry.dev-dependencies]
 flake8 = "^5.0.4"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,7 @@ psutil = "^5.9.4"
 types-psutil = "^5.9.5.4"
 types-toml = "^0.10.8"
 pytest-httpserver = "^1.0.6"
-aiohttp = "^3.8.3"
+aiohttp = "3.7"
 
 [tool.poetry.dev-dependencies]
 flake8 = "^5.0.4"
--- a/scripts/force_layer_download.py
+++ b/scripts/force_layer_download.py
@@ -14,13 +14,14 @@ class ClientException(Exception):
 
 class Client:
     def __init__(self, pageserver_api_endpoint: str):
-        self.sess = aiohttp.ClientSession(pageserver_api_endpoint)
+        self.endpoint = pageserver_api_endpoint
+        self.sess = aiohttp.ClientSession()
 
     async def close(self):
         await self.sess.close()
 
     async def get_tenant_ids(self):
-        resp = await self.sess.get("/v1/tenant")
+        resp = await self.sess.get(f"{self.endpoint}/v1/tenant")
         body = await resp.json()
         if not resp.ok:
             raise ClientException(f"{resp}")
@@ -29,7 +30,7 @@ class Client:
         return [t["id"] for t in body]
 
     async def get_timeline_ids(self, tenant_id):
-        resp = await self.sess.get(f"/v1/tenant/{tenant_id}/timeline")
+        resp = await self.sess.get(f"{self.endpoint}/v1/tenant/{tenant_id}/timeline")
         body = await resp.json()
         if not resp.ok:
             raise ClientException(f"{resp}")
@@ -40,7 +41,7 @@ class Client:
     async def timeline_spawn_download_remote_layers(self, tenant_id, timeline_id, ongoing_ok=False):
 
         resp = await self.sess.post(
-            f"/v1/tenant/{tenant_id}/timeline/{timeline_id}/download_remote_layers",
+            f"{self.endpoint}/v1/tenant/{tenant_id}/timeline/{timeline_id}/download_remote_layers",
         )
         body = await resp.json()
         if resp.status == 409:
@@ -61,7 +62,7 @@ class Client:
         timeline_id,
     ):
         resp = await self.sess.get(
-            f"/v1/tenant/{tenant_id}/timeline/{timeline_id}/download_remote_layers",
+            f"{self.endpoint}/v1/tenant/{tenant_id}/timeline/{timeline_id}/download_remote_layers",
         )
         body = await resp.json()
 
--- a/scripts/force_layer_download.py
+++ b/scripts/force_layer_download.py
@@ -21,6 +21,12 @@ class Client:
     async def close(self):
         await self.sess.close()
 
+    async def __aenter__(self):
+        return self
+
+    async def __aexit__(self, exc_t, exc_v, exc_tb):
+        await self.close()
+
     async def get_tenant_ids(self):
         resp = await self.sess.get(f"{self.endpoint}/v1/tenant")
         body = await resp.json()
@@ -136,12 +142,10 @@ def handle_sigint():
 
 
 async def main(args):
-    client = Client(args.pageserver_http_endpoint)
 
-    with open(args.report_output, "w") as report_out:
-        exit_code = await main_impl(args, report_out, client)
-
-    await client.close()
+    async with Client(args.pageserver_http_endpoint) as client:
+        with open(args.report_output, "w") as report_out:
+            exit_code = await main_impl(args, report_out, client)
 
     return exit_code
 
--- a/scripts/force_layer_download.py
+++ b/scripts/force_layer_download.py
@@ -2,6 +2,7 @@ import argparse
 import asyncio
 import json
 import logging
+import signal
 import sys
 from typing import Any, List, Tuple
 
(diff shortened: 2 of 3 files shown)


shanyp (approved):

SomeoneToIgnore (commented):
I'm bad with Python, but does not look bad for me.
Feel free to ignore any comments, if this is more ad-hoc script rather than a foundation for a collection of a kind.

scripts/force_layer_download.py
```diff
@@ -0,0 +1,235 @@
...
+    completed = await client.timeline_download_remote_layers(tenant_id, timeline_id)
+    assert completed["state"] == "Completed"
+    return Completed(completed)
+
+
+async def main(args):
+    client = Client(args.pageserver_http_endpoint)
+
+    with open(args.report_output, "w") as report_out:
+        exit_code = await main_impl(args, report_out, client)
```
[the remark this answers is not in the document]
vadim2404: it will be skipped. You're right

scripts/force_layer_download.py
```diff
@@ -0,0 +1,235 @@
...
+
+    tenant_and_timline_ids: List[Tuple[str, str]] = []
+    # fill  tenant_and_timline_ids based on spec
+    for spec in args.what:
+        comps = spec.split(":")
+        if comps == ["ALL"]:
+            logging.info("get tenant list")
+            tenant_ids = await client.get_tenant_ids()
+            tasks = [
+                asyncio.create_task(client.get_timeline_ids(tenant_id)) for tenant_id in tenant_ids
```
vadim2404, line 129: NIT: you don't need to `create_task`. They will be created inside the `asyncio.gather` in any case.

scripts/force_layer_download.py
```diff
@@ -0,0 +1,235 @@
...
+    tenant_and_timline_ids = tmp
+
+    logging.info("spawn tasks")
+    tasks = [
+        asyncio.create_task(do_timeline(client, tid, tlid), name=f"{tid}:{tlid}")
+        for tid, tlid in tenant_and_timline_ids
+    ]
+    logging.info("gather results")
+    results = await asyncio.gather(*tasks, return_exceptions=True)
+    report: dict[str, List[str]] = {
```
vadim2404, line 165:
```python
from collections import defaultdict

report = defaultdict(list)
```

and you're not obliged to define the keys.

scripts/force_layer_download.py
```diff
@@ -0,0 +1,235 @@
...
+    if args.verbose:
+        logging.basicConfig(level=logging.DEBUG)
+    logging.basicConfig(
+        format="%(asctime)s,%(msecs)03d %(levelname)-8s [%(filename)s:%(lineno)d] %(message)s",
+        datefmt="%Y-%m-%d:%H:%M:%S",
+        level=level,
+    )
+
+    loop = asyncio.get_event_loop()
+    exit_code = loop.run_until_complete(main(args))
```
vadim2404, line 233:
```python
loop = asyncio.get_event_loop()
exit_code = loop.run_until_complete(main(args))
```

can be replaced with

```python
exit_code = asyncio.run(main(args))
```

scripts/force_layer_download.py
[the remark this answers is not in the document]
vadim2404: to fix it, in `Client` we need 

```python
async def __aenter__(self):
    self

async def __aexit__(self, exc_t, exc_v, exc_tb):
    self.close()
```

and rewrite this method:

```python
async def main(args):
    async with Client(args.pageserver_http_endpoint) as client:
        with open(args.report_output, "w") as report_out:
            return await main_impl(args, report_out, client)
```

scripts/force_layer_download.py
```diff
@@ -0,0 +1,235 @@
...
+                raise ClientException("download task ids changed while polling")
+
+            if st["state"] != "Completed":
+                await asyncio.sleep(10)
+                continue
+
+            return st
+
+
+class Completed(dict[str, Any]):
```
vadim2404, line 95: it's not usual to inherit from `dict`. It's better to do sth like this.

```python
@dataclass
class Completed:
    result: Dit[str, Any]
```

and use 

```python
x = Completed(task_result)
print(x.result["state"])
```

pyproject.toml
```diff
@@ -33,6 +33,7 @@ psutil = "^5.9.4"
 types-psutil = "^5.9.5.4"
 types-toml = "^0.10.8"
 pytest-httpserver = "^1.0.6"
+aiohttp = "^3.8.3"
```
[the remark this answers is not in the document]
problame: `requests` isn't `async`, so I went with `aiohttp` here.

Also, the script is meant to be free-standing, so that we can scp it into prod pageservers and use it right away.
We don't have the `PageserverHttpClient`  there.

problame:
Changes in the last few hours:

- addressed review comments
- added progress reporting (every 10 seconds)
- added an option to configure the amount of concurrent download tasks created by the script
  - default: `1`, since the download task _inside_ the pageserver downloads _all_ layers of a timeline concurrently
    - Thinking about that, do want to limit the concurrency inside the pageserver as well? If so, we'll need a PR for that.
      - @SomeoneToIgnore ?

Last thing that I'd like to add (not sure if worth the time investment):
the ability to feed a report JSON of a previous run as input into the script, so that it retries all tasks that previously failed. It would enable a fixpoint iteration like so:
    - Suppose 20 out of 1000 tenants fail to download, we'd just want to run the script for those 20 tenants again.
    - In the second run, 10 out of the 20 tenants succeed, 10 continue to have failures.
    - Do a third run
    - ...
    - All tenants succeeded.

Why go through report json instead of a loop inside the script:
1. If user wants to change a knob, e.g. concurrency factor, they'd need to abort the current run, change the CLI argument, and resume. By going through the report.json, we can get that resume state.
2. If the script crashes mid-way, we can use the last report.json (=> guarantees forward progress)

So, do we want that feature in the script?

scripts/force_layer_download.py
```diff
@@ -0,0 +1,315 @@
...
+
+        return body
+
+
+@dataclass
+class Completed:
+    """The status dict returned by the API"""
+
+    status: dict[str, Any]
+    pass
```
vadim2404, line 91: just remove the `pass`. No need in it anymore

scripts/force_layer_download.py
```diff
@@ -0,0 +1,315 @@
...
+                f"download task reached terminal state != Completed: {st['state']}"
+            )
+
+        return Completed(st)
+
+
+def handle_sigint():
+    logging.info("SIGINT received, asyncio event set. Will not start new downloads.")
+    global sigint_received
+    sigint_received.set()
```
vadim2404, line 146: NIT: it's possible to stop all planned tasks if you need it

scripts/force_layer_download.py
```diff
@@ -0,0 +1,315 @@
...
+        f"{tid}:{tlid}": do_timeline(client, tid, tlid) for tid, tlid in tenant_and_timline_ids
+    }
+    for task in tasks.items():
+        task_q.put_nowait(task)
+
+    result_q: asyncio.Queue[Tuple[str, Any]] = asyncio.Queue()
+    taskq_handlers = []
+    for _ in range(0, args.concurrent_tasks):
+
+        async def taskq_handler(task_q):
```
vadim2404, line 210: I don't think that you need to define the function in the loop

scripts/force_layer_download.py
```diff
@@ -0,0 +1,315 @@
...
+        return self
+
+    async def __aexit__(self, exc_t, exc_v, exc_tb):
+        await self.close()
+
+    async def get_tenant_ids(self):
+        resp = await self.sess.get(f"{self.endpoint}/v1/tenant")
+        body = await resp.json()
+        if not resp.ok:
+            raise ClientException(f"{resp}")
```
vadim2404, line 36: NIT: `raise ClientException(str(resp))`

scripts/force_layer_download.py
```diff
@@ -0,0 +1,315 @@
...
+    async def timeline_spawn_download_remote_layers(self, tenant_id, timeline_id, ongoing_ok=False):
+
+        resp = await self.sess.post(
+            f"{self.endpoint}/v1/tenant/{tenant_id}/timeline/{timeline_id}/download_remote_layers",
+        )
+        body = await resp.json()
+        if resp.status == 409:
+            if not ongoing_ok:
+                raise ClientException("download already ongoing")
+            pass  # response body has same shape for ongoing and newly created
```
vadim2404, line 59: just remove the `pass`

LizardWizzard (commented):
Works for me. I dont think we need to spend much time on polishing this script to absolute perfection. People did a great job reviewing this before me :)

> It would enable a fixpoint iteration like so:

I dont think this is needed. As long as download operation is idempotent it will move on to layers that were not downloaded previously

Regarding dependencies I'm leaning towards strategy to use binary clients and spawn processes to invoke them instead of using python client libraries. E g using curl over http client and psql over psycopg. E g in migration script I use curl instead of python pageserver client. It is quite convenient to load stdout with `json.loads`. If you want async, there are subprocess wrappers https://docs.python.org/3/library/asyncio-subprocess.html it is usually not desired to use subprocesses, because of the overhead, but this is an ok approach for such maintenance scripts. For sure I'm not asking to radically change approach here, just to share what I came up with

problame:
As discussed in Slack DMs, I'm handing this PR over to @LizardWizzard as he'll be the one who needs to use it in case on-demand download doesn't perform well enough after migrating the last pageservers to the new AWS account.
Work items left:
- [ ] Address remaining review comments 
- [ ] Send concurrency limit value as request body, see https://github.com/neondatabase/neon/pull/3302#issuecomment-1384194674
  - The script won't work otherwise.
- [ ] Merge this script into the repo after the migration project is done. We'll continue to have a maybe-need for this, at the very least until layer eviction is developed & stable.

LizardWizzard:
I went through the list. Lets merge it even before migration completes. Otherwise it will be lost in the PR's list. It will be useful for some time as a runbook item in case problems with ondemand/eviction

scripts/force_layer_download.py
LizardWizzard: Changed the message

pyproject.toml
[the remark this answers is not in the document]
LizardWizzard: aiohttp was installed as a package on servers. For the future I think using binary clients like curl and subprocesses is a simpler option

problame (commented):
LGTM, don't forget to squash!

PR #3296, opened 2023-01-09
Enable bloom extension

vadim2404:


vadim2404:
I didn't see that we've built contrib extensions in such a way. Should we?

PR #3298, opened 2023-01-09
Update tokio version (RUSTSEC-2023-0001)

vadim2404:
`cargo deny check advisories` didn't pass because of vulnerability at tokio

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3731,9 +3731,9 @@ dependencies = [
 
 [[package]]
 name = "tokio"
-version = "1.21.1"
+version = "1.24.1"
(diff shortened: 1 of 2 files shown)


hlinnaka (approved):

SomeoneToIgnore:
What's odd, is that this vulnerability targets Windows only: https://rustsec.org/advisories/RUSTSEC-2023-0001.html
I wonder, if `deny` can ignore such cases at all, since not relevant to us?

vadim2404:
@SomeoneToIgnore , not so many options `cargo deny` provides to escape such vulnerabilities

koivunej:
Very nice, this PR updates tokio to 1.24.1 which has welcome optimizations.

PR #3301, opened 2023-01-10
Release 2023-01-10

problame:
```                                                                                                                   
Conflicts:                                                                                                         
        .github/helm-values/neon-stress.proxy-scram.yaml                                                           
        .github/helm-values/neon-stress.proxy.yaml                                                                 
        .github/helm-values/staging.proxy-scram.yaml                                                               
        .github/helm-values/staging.proxy.yaml                                                                     
        All of the above were deleted in `main` after we hotfixed them                                             
        in `release. Deleting them here                                                                            
        storage_broker/src/bin/storage_broker.rs                                                                   
        Hotfix toned down logging, but `main` has sinced implemented                                               
        a proper fix. Taken `main`'s side, see                                                                     
        https://neondb.slack.com/archives/C033RQ5SPDH/p1673354385387479?thread_ts=1673354306.474729&cid=C033RQ5SPDH
```

closes https://github.com/neondatabase/neon/issues/3287

PR #3302, opened 2023-01-10
download_all_remote_layers API: require client to specify max_concurr…

problame:
…ent_downloads

Before this patch, we would start all layer downloads simultaneously.

There is at most one download_all_remote_layers task per timeline. Hence, the specified limit is per timeline.

There is still no global concurrency limit for layer downloads. We'll have to revisit that at some point and also prioritize on-demand initiated downloads over download_all_remote_layers downloads. But that's for another day.

### DIFF
--- a/pageserver/src/http/routes.rs
+++ b/pageserver/src/http/routes.rs
@@ -3,6 +3,7 @@ use std::sync::Arc;
 use anyhow::{anyhow, Context, Result};
 use hyper::StatusCode;
 use hyper::{Body, Request, Response, Uri};
+use pageserver_api::models::DownloadRemoteLayersTaskSpawnRequest;
 use remote_storage::GenericRemoteStorage;
 use tokio_util::sync::CancellationToken;
 use tracing::*;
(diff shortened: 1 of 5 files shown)


problame:
(@shanyp and I determined that this is not required for #3301 . I just had some spare cycles)

problame:
Note to self: after merging this, update https://github.com/neondatabase/neon/pull/3294

hlinnaka (approved):
What is the default if you don't specify 'max_concurrent_downloads' ?

Remember to update `openapi_spec.yml`

problame:
> What is the default if you don't specify 'max_concurrent_downloads' ?

It's a required argument, both at python level and API level.

> Remember to update openapi_spec.yml

It's an undocumented API, so, no need to update it.
Only consumer is my script in https://github.com/neondatabase/neon/pull/3294 and the unit tests.

pageserver/src/http/routes.rs
```diff
@@ -778,10 +779,11 @@ async fn timeline_checkpoint_handler(request: Request<Body>) -> Result<Response<
 }
 
 async fn timeline_download_remote_layers_handler_post(
-    request: Request<Body>,
+    mut request: Request<Body>,
 ) -> Result<Response<Body>, ApiError> {
     let tenant_id: TenantId = parse_request_param(&request, "tenant_id")?;
     let timeline_id: TimelineId = parse_request_param(&request, "timeline_id")?;
+    let body: DownloadRemoteLayersTaskSpawnRequest= json_request(&mut request).await?;
```
LizardWizzard, line 786:
```suggestion
    let body: DownloadRemoteLayersTaskSpawnRequest = json_request(&mut request).await?;
```

LizardWizzard (approved):
LGTM

Issue #2963, opened 2022-11-30
`test_forward_compatibility@compatibility` is flaky


yeputons:
A recent failure in `main`: https://github.com/neondatabase/neon/actions/runs/3563254362/jobs/5986129088

`main` seems failing:passing about 50:50 at the moment, each failure is unique in its flakiness.

```

__________________________ test_forward_compatibility __________________________
[gw0] linux -- Python 3.9.2 /github/home/.cache/pypoetry/virtualenvs/neon-_pxWMzVK-py3.9/bin/python
subprocess.CalledProcessError: Command '['/tmp/neon-previous/bin/neon_local', 'pg', 'start', '--tenant-id', '325bbf12962a84e96cf57bcde7b95c22', '--pg-version', '14', '--port=15068', 'main']' returned non-zero exit status 1.

The above exception was the direct cause of the following exception:
test_runner/regress/test_compatibility.py:170: in test_forward_compatibility
    check_neon_works(
test_runner/regress/test_compatibility.py:311: in check_neon_works
    cli.pg_start("main", port=pg_port)
test_runner/fixtures/neon_fixtures.py:1661: in pg_start
    res = self.raw_cli(args)
test_runner/fixtures/neon_fixtures.py:1356: in raw_cli
    raise Exception(msg) from subprocess.CalledProcessError(
E   Exception:             Run ['/tmp/neon-previous/bin/neon_local', 'pg', 'start', '--tenant-id', '325bbf12962a84e96cf57bcde7b95c22', '--pg-version', '14', '--port=15068', 'main'] failed:
E                 stdout: Starting new postgres (v14) main on timeline 88760b3450f4dfb8bdca11cafed67536 ...
E   Extracting base backup to create postgres instance: path=/tmp/test_output/test_forward_compatibility/compatibility_snapshot/repo/pgdatadirs/tenants/325bbf12962a84e96cf57bcde7b95c22/main port=15068
E   Starting postgres node at 'host=127.0.0.1 port=15068 user=cloud_admin dbname=postgres'
E   
E                 stderr: command failed: pg_ctl failed, exit code: exit status: 1, stdout: waiting for server to start.... stopped waiting
E   , stderr: pg_ctl: could not start server
E   Examine the log output.
E   
E   
E   Stack backtrace:
E      0: control_plane::compute::PostgresNode::pg_ctl
E                at ./control_plane/src/compute.rs:453:13
E      1: control_plane::compute::PostgresNode::start
[...]

Issue #1351, opened 2022-03-11
Compute cannot start using recevied basebackup


ololobus:
... or couldn't get a basebackup with LSN received by the sync safekeepers run.

We've recently observed two failures in the staging cloud env, which are likely related to storage.

1) Cluster https://console.stage.zenith.tech/admin/clusters/old-rain-092695 (**UPD: this has been deleted already by Nikita**) failed to start with:
```log
2022-03-10 12:53:16.480 UTC [main] INFO: starting cluster #old-rain-092695, operation #f294a8e8-034e-4cef-a131-403fcf13db5d
2022-03-10 12:53:16.480 UTC [http-endpoint] INFO: starting HTTP server on 0.0.0.0:3080
2022-03-10 12:53:16.480 UTC [main] INFO: starting safekeepers syncing
...
2022-03-10 12:53:16.502 UTC [main] INFO: safekeepers synced at LSN 0/1000000
2022-03-10 12:53:16.502 UTC [main] INFO: getting basebackup@0/1000000 from pageserver host=172.32.24.61 port=6400
Error: failed to get a basebackup@0/1000000 from pageserver host=172.32.24.61 port=6400

Caused by:
    0: db error: ERROR: invalid basebackup lsn
    1: ERROR: invalid basebackup lsn
```

LSN 0/1000000 looks really suspicious, my guess is that something went wrong during safekeepers sync.

2) Cluster https://console.stage.zenith.tech/admin/clusters/withered-fire-280332 failed to start with:
```log
2022-03-10 12:56:53.445 UTC [main] INFO: safekeepers synced at LSN 11/25000000
2022-03-10 12:56:53.445 UTC [main] INFO: getting basebackup@11/25000000 from pageserver host=172.32.24.61 port=6400
...
2022-03-10 12:56:53.586 GMT [14] LOG:  starting with zenith basebackup at LSN 11/25000028, prev 0/0
2022-03-10 12:56:53.586 GMT [14] FATAL:  cannot start in read-write mode from this base backup
```
LSN 11/25000000 looks a bit too even, IMO, but we at least managed to produce basebackup.

For 2) pg comment says:
```clang
/*
 * We cannot start generating new WAL if we don't have a valid prev-LSN
 * to use for the first new WAL record. (Shouldn't happen.)
 */
```
so it looks like that produced basebackup doesn't contain a properly formatted WAL segment?

Any thoughts about what could possible go wrong here?

PR #3315, opened 2023-01-12
Use ready! rustc 1.64 macro

SomeoneToIgnore:
rustc [1.64](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1640-2022-09-22) had brought `ready!` macro: https://doc.rust-lang.org/stable/std/task/macro.ready.html

Use it to shorten the code slightly.

### DIFF
--- a/libs/utils/src/postgres_backend_async.rs
+++ b/libs/utils/src/postgres_backend_async.rs
@@ -7,12 +7,12 @@ use crate::postgres_backend::AuthType;
 use anyhow::Context;
 use bytes::{Buf, Bytes, BytesMut};
 use pq_proto::{BeMessage, ConnectionError, FeMessage, FeStartupPacket, SQLSTATE_INTERNAL_ERROR};
-use std::future::Future;
 use std::io;
 use std::net::SocketAddr;
 use std::pin::Pin;
 use std::sync::Arc;
 use std::task::Poll;
+use std::{future::Future, task::ready};
 use tracing::{debug, error, info, trace};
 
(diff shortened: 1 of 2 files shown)


funbringer (approved):

ololobus (approved):

koivunej (approved):
You probably had this ready to go right after :)

koivunej:
I was thinking at first the failure looked like a port already in use, but this seems new for me: https://neon-github-public-dev.s3.amazonaws.com/reports/pr-3315/release/3903582277/index.html#suites/0fbaeb107ef328d03993d44a1fb15690/d7e5d24ebc8f1c75/

pg says:

```
2023-01-12 15:34:55.341 GMT [13873] LOG:  starting with zenith basebackup at LSN 0/AA31D80, prev 0/0
2023-01-12 15:34:55.341 GMT [13873] FATAL:  cannot start in read-write mode from this base backup
```

Haven't seen this one yet.
```

ololobus:
> FATAL:  cannot start in read-write mode from this base backup

We had this issue (or better say the same symptom) before https://github.com/neondatabase/neon/issues/1351. And that time it was a problem with safekeepers

koivunej:
Hmm but the issue is already closed. Looking through the other log files I cannot find anything notable... Maybe I'll just create a new issue and the workflow can be retried?

koivunej:
Ah, should had searched, better one was found quickly: https://github.com/neondatabase/neon/issues/2963

PR #3322, opened 2023-01-12
Rework Cargo.toml dependencies

SomeoneToIgnore:
* Use workspace variables from cargo, coming with rustc [1.64](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1640-2022-09-22)

See https://doc.rust-lang.org/nightly/cargo/reference/workspaces.html#the-package-table and https://doc.rust-lang.org/nightly/cargo/reference/workspaces.html#the-dependencies-table sections.

Now, all dependencies in all non-root `Cargo.toml` files are defined as 
```
clap.workspace = true
```

sometimes, when extra features are needed, as 
```
bytes = {workspace = true, features = ['serde'] }
```

With the actual declarations (with shared features and version numbers/file paths/etc.) in the root Cargo.toml.
Features are additive:
https://doc.rust-lang.org/nightly/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace

* Uses the mechanism above to set common, 2021, edition across the workspace

* Mechanically bumps a few dependencies

* Updates hakari format, as it suggested:
```
work/neon/neon kb/cargo-templated ❯ cargo hakari generate
info: no changes detected
info: new hakari format version available: 3 (current: 2)
(add or update `dep-format-version = "3"` in hakari.toml, then run `cargo hakari generate && cargo hakari manage-deps`)
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ members = [
 
 [workspace.package]
 edition = "2021"
+license = "Apache-2.0"
 
 ## All dependency versions, used in the project, grouped in some arbitrary order.
 [workspace.dependencies]
@@ -120,8 +121,8 @@ walkdir = "2.3.2"
 log = "0.4"
 env_logger = "0.10"
 
-# TODO fork this and use similar `rev` hash 
-amplify_num = { git = "https://github.com/hlinnaka/rust-amplify.git", branch = "unsigned-int-perf" }
+## TODO switch when the new release is made
+amplify_num = { git = "https://github.com/rust-amplify/rust-amplify.git", tag = "v4.0.0-beta.1" }
 
 ## Libraries from neondatabase/ git forks, ideally with changes to be upstreamed
 postgres = { git = "https://github.com/neondatabase/rust-postgres.git", rev="43e6db254a97fdecbce33d8bc0890accfd74495e" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,9 @@ members = [
     "libs/*",
 ]
 
+[workspace.package]
+edition = "2021"
+
 [profile.release]
 # This is useful for profiling and, to some extent, debug.
 # Besides, debug info should not affect the performance.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,122 @@ members = [
 [workspace.package]
 edition = "2021"
 
+[workspace.dependencies]
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1"
+serde_with = "2.0"
+humantime-serde = "1.1.1"
+anyhow = { version = "1.0", features = ["backtrace"] }
+thiserror = "1.0"
+once_cell = "1.13"
+nix = "0.25"
+bytes = "1.0"
+byteorder = "1.4"
+rand = "0.8"
+futures = "0.3"
+async-stream = "0.3"
+async-trait = "0.1"
+routerify = "3"
+signal-hook = "0.3"
+rustls = "0.20"
+pin-project-lite = "0.2"
+tracing = "0.1"
+tracing-subscriber = { version = "0.3", features = ["env-filter"] }
+regex = "1.4"
+clap = "4.0"
+git-version = "0.3"
+tar = "0.4"
+url = "2.2"
+const_format = "0.2"
+parking_lot = "0.12"
+chrono = { version = "0.4", default-features = false, features = ["clock"] }
+humantime = "2.1"
+reqwest = { version = "0.11", default-features = false, features = ["rustls-tls"] }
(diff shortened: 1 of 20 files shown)


libs/metrics/Cargo.toml
```diff
@@ -1,11 +1,12 @@
 [package]
 name = "metrics"
 version = "0.1.0"
-edition = "2021"
+edition.workspace = true
 license = "Apache-2.0"
```
vadim2404, line 5: The same is possible for license. Will you update more fields?

SomeoneToIgnore: > The same is possible for licens

I don't see it possible for `license` field, only for `license-file`, that's different.

Cargo.toml
```diff
@@ -10,6 +10,156 @@ members = [
...
+webpki-roots = "0.22.5"
+x509-parser = "0.14"
+walkdir = "2.3.2"
+
+## TODO replace this with tracing
+log = "0.4"
+env_logger = "0.10"
+
+# TODO fork this and use similar `rev` hash 
+amplify_num = { git = "https://github.com/hlinnaka/rust-amplify.git", branch = "unsigned-int-perf" }
```
vadim2404, line 124: I guess we can switch back, because @hlinnaka’s PR was merged

libs/metrics/Cargo.toml
vadim2404: https://github.com/EmbarkStudios/rust-gpu/blob/main/crates/rustc_codegen_spirv-types/Cargo.toml

You showed me this example

Cargo.toml
SomeoneToIgnore: I see that this PR seems to be included into `4.0.0-*` branch: https://github.com/rust-amplify/rust-amplify/commit/bd49b737c2e6e623ab8e9ba5ceaed5712d3a3940

but that's still not released.
Switched to the upstream git repo at least.

libs/metrics/Cargo.toml
SomeoneToIgnore: Indeed that works, switched.

vadim2404 (approved):

Cargo.toml
```diff
@@ -10,6 +10,157 @@ members = [
     "libs/*",
 ]
 
+[workspace.package]
+edition = "2021"
+license = "Apache-2.0"
+
+## All dependency versions, used in the project, grouped in some arbitrary order.
```
funbringer, line 17: Can we not have them grouped in some arbitrary order? My personal order of preference:

alphabetical >> chaos >> obscure partitions

funbringer (commented):
This change has an unpleasant side effect: whoever decides to update a dependency in their domain will be forced to deal with migration churn all over the project. I don't think this will scale well.

SomeoneToIgnore:
In that case, they can use a specific version in their domain and use it? Does not look like an issue, unless I miss something.

I plan to do similar thing with tracing-subscriber: leave proxy to use the `workspace=true` version, and make pageserver to use a branch from https://github.com/neondatabase/tracing

Cargo.toml
SomeoneToIgnore: Sorted.
Had left a few small groups below, such as `## TODO replace this with tracing` one.

funbringer:
I think you're right. By the way, does cargo-udeps work with workspace deps?

libs/postgres_ffi/Cargo.toml
```diff
@@ -1,30 +1,32 @@
...
+anyhow.workspace = true
+crc32c.workspace = true
+hex.workspace = true
+once_cell.workspace = true
+log.workspace = true
+memoffset.workspace = true
+thiserror.workspace = true
+serde.workspace = true
+
+utils.workspace = true
```
funbringer, line 21: It's funny how this is now almost indistinguishable from regular crates :) Should we separate it from others? In that case, would it make sense to add a comment on why?

SomeoneToIgnore: Does not look very sensible indeed, the main separation now lies in the root Cargo.toml.
United sections in the other dependency files also.

SomeoneToIgnore:
> does cargo-udeps work with workspace deps?

It's odd, as if there's something broken with my installation.
Even on the regular Cargo.toml, it gives me:

```
work/neon/neon_main main:origin/main* ❯ cargo +nightly udeps
error: the `-Z` flag is only accepted on the nightly channel of Cargo, but this is the `` channel
See https://doc.rust-lang.org/book/appendix-07-nightly-rust.html for more information about Rust release channels.
```

but when I run it with RUSTC_BOOTSTRAP=1, it works, even for the new toml format.

funbringer (approved):

PR #3324, opened 2023-01-12
Add VM informant to vm-compute-node

sharnoff:
No clue if this is the right way to do this, haven't tested it, etc. All changes were basically following the path of least resistance ­— please nitpick.

The general idea is that the VM informant binary is added to the vm-compute-node images only. `compute_tools` then will run whatever's at `/bin/vm-informant`, if the path exists.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -595,6 +595,8 @@ jobs:
     defaults:
       run:
         shell: sh -eu {0}
+    env:
+      VM_INFORMANT_VERSION: 0.1.1
 
     steps:
       - name: Downloading latest vm-builder
@@ -606,9 +608,22 @@ jobs:
         run: |
           docker pull 369495373322.dkr.ecr.eu-central-1.amazonaws.com/compute-node-${{ matrix.version }}:${{needs.tag.outputs.build-tag}}
 
+      - name: Downloading VM informant version ${{ env.VM_INFORMANT_VERSION }}
+        run: |
+          curl -L https://github.com/neondatabase/autoscaling/releases/download/${{ env.VM_INFORMANT_VERSION }}/vm-informant -o vm-informant
+          chmod +x vm-informant
+
+      - name: Adding VM informant to compute-node image
+        run: |
+          ID=$(docker create 369495373322.dkr.ecr.eu-central-1.amazonaws.com/compute-node-${{ matrix.version }}:${{needs.tag.outputs.build-tag}})
+          docker cp vm-informant $ID:/bin/vm-informant
+          docker commit $ID temp-vm-compute-node
+          docker rm -f $ID
+
       - name: Build vm image
         run: |
-          ./vm-builder -src=369495373322.dkr.ecr.eu-central-1.amazonaws.com/compute-node-${{ matrix.version }}:${{needs.tag.outputs.build-tag}} -dst=369495373322.dkr.ecr.eu-central-1.amazonaws.com/vm-compute-node-${{ matrix.version }}:${{needs.tag.outputs.build-tag}}
+          # note: as of 2023-01-12, vm-builder requires a trailing ":latest" for local images
+          ./vm-builder -src=temp-vm-compute-node:latest -dst=369495373322.dkr.ecr.eu-central-1.amazonaws.com/vm-compute-node-${{ matrix.version }}:${{needs.tag.outputs.build-tag}}
 
       - name: Pushing vm-compute-node image
         run: |
--- a/compute_tools/README.md
+++ b/compute_tools/README.md
@@ -19,7 +19,9 @@ Also `compute_ctl` spawns two separate service threads:
 - `http-endpoint` runs a Hyper HTTP API server, which serves readiness and the
   last activity requests.
 
-If the `vm-informant` binary is present, it will also be started.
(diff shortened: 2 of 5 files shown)


compute_tools/src/informant.rs
```diff
@@ -0,0 +1,50 @@
+use log::{info, warn};
+use std::path::Path;
+use std::process;
+use std::thread;
+use std::time::Duration;
+
+use anyhow::{Context, Result};
+
+static VM_INFORMANT_PATH: &str = "/bin/vm-informant";
```
vadim2404, line 9: why is it not a `const`?

```rust
const VM_INFORMANT_PATH: &'static str = "/bin/vm-informant";
```

.github/workflows/build_and_test.yml
```diff
@@ -606,9 +608,22 @@ jobs:
...
           docker pull 369495373322.dkr.ecr.eu-central-1.amazonaws.com/compute-node-${{ matrix.version }}:${{needs.tag.outputs.build-tag}}
 
+      - name: Downloading VM informant version ${{ env.VM_INFORMANT_VERSION }}
+        run: |
+          curl -L https://github.com/neondatabase/autoscaling/releases/download/${{ env.VM_INFORMANT_VERSION }}/vm-informant -o vm-informant
+          chmod +x vm-informant
+
+      - name: Adding VM informant to compute-node image
+        run: |
+          ID=$(docker create 369495373322.dkr.ecr.eu-central-1.amazonaws.com/compute-node-${{ matrix.version }}:${{needs.tag.outputs.build-tag}})
```
vadim2404, line 618: what do you think about adding this into `Dockerfile.compute-node-v1[45]`?

compute_tools/README.md
```diff
@@ -19,6 +19,8 @@ Also `compute_ctl` spawns two separate service threads:
 - `http-endpoint` runs a Hyper HTTP API server, which serves readiness and the
   last activity requests.
 
+If the `vm-informant` binary is present, it will also be started.
```
vadim2404, line 22: please, add more info about `vm-informant` necessity

cicdteam (approved):

.github/workflows/build_and_test.yml
vadim2404: finally, we will have only VMs, and this binary is lightweight.

cicdteam: We are opensource and I think people mostly run neon locally in docker or in their k8s clusters. NeonVM without kvm (bare-metal hosts) too slow.

ololobus (approved):
Went through the code, LGTM. As for the building, I think it's fine to put binary only into VM images in CI

compute_tools/src/informant.rs
sharnoff: Nothing substantial — just habit. With `&'static str` the only difference is that unoptimized builds with `static`s load the length from another static memory location. Unoptimized consts and optimized builds of either inline the length wherever it's used.

vadim2404: as I saw, in the codebase, we use `const` more often for that case.
But agree, that's not a big deal

sharnoff: Ah ok! I'll change it

.github/workflows/build_and_test.yml
sharnoff: I also think the `vm-informant` shouldn't be included in the non-VM compute node images, for two reasons:
1. In general, we shouldn't run programs in environments we don't use them
2. The `vm-informant` binary is ~40MB. That's not _huge_, but it's not insignificant either.

vadim2404: sounds good

compute_tools/README.md
sharnoff: Done

vadim2404 (approved):

sharnoff:
I'm struggling to get `test_compatibility.py::test_forward_compatibility` to pass. The sequence of actions I took were:
* Push 496a535dbbe2f241dac20ec7938e8f0113560528 ("Add VM informant to vm-compute-node") (parent was ~8 commits behind) — CI passed.
* Push 073bea745f181fb8a594f452bab76930e00bd102 ("Review feedback") ­— CI failed (didn't retry)
* Rebase 073bea745f181fb8a594f452bab76930e00bd102 -> 28f57881421d6e2264b0a6e941cebadc9b3d22ad — CI failed (retried, total of 5 attempts)

`pg_dumpall` failed on all of them, but I don't see why the "Review feedback" commit would have triggered that.

I know from #2963 that this test is flaky, but this seems like it's actually failing, even though AFAICT this PR is unrelated _and_ CI on main is green.

sharnoff:
Had a brief look at other CI runs. I think there's an existing issue with the snapshot being used for `test_forward_compatibility` that started sometime between `2023-01-13 16:41 UTC` and `2023-01-13 17:16 UTC`.

Recent PRs with similar failures: #3220, #3331, #3348

bayandin:
> Had a brief look at other CI runs. I think there's an existing issue with the snapshot being used for `test_forward_compatibility` that started sometime between `2023-01-13 16:41 UTC` and `2023-01-13 17:16 UTC`.
> 
> Recent PRs with similar failures: #3220, #3331, #3348, #3349

This should be fixed now. I've restarted failed jobs

PR #3331, opened 2023-01-13
[proxy] Implement cloud API caches

funbringer:
- [x] Find a decent LRU implementation.
- [x] Implement timed LRU on top of that.
- [x] Cache results of `proxy_wake_compute` API call.
- [x] Don't invalidate newer cache entries for the same key.
- [x] Add cmdline configuration knobs (requires some refactoring).
- [x] Add failed connection estab metric.
- [x] Refactor auth backends to make things simpler (retries, cache placement, etc).
- [x] Address review comments (add code comments + cleanup).
- [x] Retry `/proxy_wake_compute` if we couldn't connect to a compute (e.g. stalled cache entry).
- [x] Add high-level description for `TimedLru`.
- [ ] Add cache metrics (hit, spurious hit, miss).
- [ ] Synchronize http requests across concurrent per-client tasks (https://github.com/neondatabase/neon/pull/3331#issuecomment-1399216069).
- [ ] Cache results of `proxy_get_role_secret` API call.
- [ ] Benchmark and deploy to staging.


vadim2404:
Closes #3332

proxy/src/proxy.rs
```diff
@@ -272,18 +277,24 @@ impl<S: AsyncRead + AsyncWrite + Unpin + Send> Client<'_, S> {
...
+        let mut node = auth_result.value;
         let (db, cancel_closure) = node
             .config
             .connect(params)
             .or_else(|e| stream.throw_error(e))
-            .await?;
+            .await
+            .map_err(|e| {
+                // Invalidate the cache entry if we failed to connect.
+                warn!("invalidating stalled compute node info cache entry");
```
koivunej, line 294: Is there enough context for this message. I find for non-ws connections `client=$session_id` but nothing for ws connections?

koivunej (approved):
Patch is looking good, also did some digging into surrounding proxy parts, which are looking great. Some comment specific ideas ~, and question about logging~.

I doubt benchmarking would provide too interesting results, because the time to wakeup currently fluctuates and will dominate the time before this patch and with happen every once per ttl with this patch.

proxy/src/proxy.rs
koivunej: I missed that there is the similar `ws-client=$session_id`, resolving.

proxy/src/proxy.rs
```diff
@@ -272,18 +277,24 @@ impl<S: AsyncRead + AsyncWrite + Unpin + Send> Client<'_, S> {
...
-        let node = auth_result.value;
+        let mut node = auth_result.value;
         let (db, cancel_closure) = node
             .config
             .connect(params)
             .or_else(|e| stream.throw_error(e))
-            .await?;
+            .await
+            .map_err(|e| {
+                // Invalidate the cache entry if we failed to connect.
```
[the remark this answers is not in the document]
funbringer: You're right, this isn't clear enough.

proxy/src/auth/backend/console.rs
```diff
@@ -251,15 +255,33 @@ impl Api<'_> {
...
         .map_err(crate::error::log_error)
         .instrument(info_span!("wake_compute", id = request_id))
         .await
     }
+
+    /// Wake up the compute node and return the corresponding connection info.
+    pub async fn wake_compute(&self) -> Result<CachedNodeInfo, WakeComputeError> {
+        let key = self.creds.project().expect("impossible");
+
+        // Wake-up might not be needed if the node is still alive.
```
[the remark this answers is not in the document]
funbringer: Updated.

LizardWizzard (approved):
Looks very nice!

AM I right that invalidation happens based on either time or error during compute connection establishment?

proxy/src/auth/backend/console.rs
```diff
@@ -251,15 +255,36 @@ impl Api<'_> {
...
     }
+
+    /// Wake up the compute node and return the corresponding connection info.
+    pub async fn wake_compute(&self) -> Result<CachedNodeInfo, WakeComputeError> {
+        let key = self.creds.project().expect("impossible");
+
+        // Every time we do a wakeup http request, the compute node will stay up
+        // for some time (~5 mins; depends on the console's scale-to-zero policy);
+        // The connection info remains the same during that period of time,
+        // which means that we might cache it to reduce the load and latency.
```
ololobus, line 277: > for some time (~5 mins; depends on the console's scale-to-zero policy)

I realized that now it's not strictly correct, or not correct anymore.

Some context, we use NodePort service (it exposes some fixed port on **every** k8s node) and ip address of some random not tainted k8s node as an entry point. But, we started enabling k8s autoscaler, so k8s nodes are not something really persistent, they can appear to handle increased load and disappear, when resources are not needed anymore.

That said, previously picked address of some k8s node may become invalid. To handle that, I added some ad-hoc service discovery into activity monitor in the console. Now on prod it's configured to poke every presumably running compute every ~10 secs, and if it doesn't respond, it tries to pick a different k8s node address and poke again.

Currently, k8s autoscaler is rolled out to `us-east-2` only, but soon to be in every region, as it seems to be decently working.
- The rate of such address updates could be find on this metric https://neonprod.grafana.net/d/S3MPUUonz/neon-console?orgId=1&viewPanel=57&from=now-24h&to=now
- And here is a rate of failed checks (even after address update) https://neonprod.grafana.net/d/S3MPUUonz/neon-console?orgId=1&viewPanel=55&from=now-24h&to=now

Since proxy has some force invalidation after failed connection, that's probably fine, I just stumbled on this comment and decided to clarify some things. Everything explained above can change one day (e.g. we can switch to using stable ClusterIP addresses), but that's how it works right now.

So the best flow could be
- accept client connection
- try connecting using cached address
- if not, try to do wake up without dropping the client connection
Probably that's how it's done already, then cool, otherwise it may increase error rate in client connection.

funbringer: Ok, thanks for the information.

> Since proxy has some force invalidation after failed connection, that's probably fine, I just stumbled on this comment and decided to clarify some things. Everything explained above can change one day (e.g. we can switch to using stable ClusterIP addresses), but that's how it works right now.

I'm starting to feel very uneasy here.

* This info arrived very late in the dev cycle. In your opinion, how do we fix that?
* What's the proposed default TTL of a cache entry, then? 10s?
* Say, we pick ttl=10s, how do we keep it up-to-date? Should we assign a maintainer from @neondatabase/control-plane?

funbringer: I have a proposal (which IMO has been posted elsewhere, but nonetheless):

What do you think about returning the proposed TTL from `proxy_wake_compute`? This would give us the flexibility we need. Each entry might have its own TTL.

ololobus: Unfortunately, I don't think that any TTL will guarantee that address will last this time for real. I once answered / briefly explained this in the Slack (somewhat around this thread https://neondb.slack.com/archives/C04JRCGV8F2/p1673357467738899?thread_ts=1673274880.738879&cid=C04JRCGV8F2)

There are certain user actions + undergoing background activities in the k8s, which can invalidate it almost immediately after proxy cached it. So 5 mins default TTL for a happy path sounds good to me, but having an address refresh backup path seems to be a must have

funbringer: Yeah, this makes sense. I'm going to give this another spin using the new intel.

proxy/src/cache.rs
```diff
@@ -0,0 +1,221 @@
...
+            let value = extract(raw_entry.key(), entry);
+
+            // Update the deadline and the entry's position in the LRU list.
+            raw_entry.get_mut().expires_at = deadline;
+            raw_entry.to_back();
+
+            Some(value)
+        }
+
+        pub fn insert_raw(&self, key: K, value: V) -> (Instant, Option<V>) {
```
hlinnaka, line 119:
```suggestion
        fn insert_raw(&self, key: K, value: V) -> (Instant, Option<V>) {
```

proxy/src/cache.rs
```diff
@@ -0,0 +1,221 @@
...
+                .lock()
+                .insert(key, entry)
+                .map(|entry| entry.value);
+
+            (created_at, old)
+        }
+    }
+
+    impl<K: Hash + Eq + Clone, V: Clone> TimedLru<K, V> {
+        pub fn insert(&self, key: K, value: V) -> (Option<V>, Cached<&Self>) {
```
hlinnaka, line 141:
```suggestion
        /// Insert an entry to the cache. If an entry with the same key already existed, returns
        /// the old value. XXX: What is `Cached<&Self>`?
        pub fn insert(&self, key: K, value: V) -> (Option<V>, Cached<&Self>) {
```

proxy/src/cache.rs
```diff
@@ -0,0 +1,221 @@
...
+    /// Entry's key.
+    type Key;
+    /// Entry's value.
+    type Value;
+    /// Used for entry invalidation.
+    type LookupInfo<Key>;
+
+    /// Invalidate an entry using a lookup info.
+    fn invalidate(&self, _: &Self::LookupInfo<Self::Key>) {}
+}
```
hlinnaka, line 29: I think this trait is an unnecessary abstraction. There is only one implementation, TimedLru, and it makes it harder to understand.

proxy/src/proxy.rs
```diff
@@ -236,6 +245,8 @@ struct Client<'a, S> {
     params: &'a StartupMessageParams,
     /// Unique connection ID.
     session_id: uuid::Uuid,
+    /// Varous caches for the cloud API responses.
+    caches: &'static auth::caches::ApiCaches,
```
hlinnaka, line 249: Seems weird for caches to be a property of a connection. It's shared by all connections.

proxy/src/auth/backend.rs
```diff
@@ -128,22 +129,51 @@ pub struct AuthSuccess<T> {
...
+}
+
+pub type NodeInfoCache = TimedLru<Arc<str>, NodeInfo>;
+pub type CachedNodeInfo = timed_lru::Cached<&'static NodeInfoCache>;
+
+/// Various caches for [`console`].
+pub struct ApiCaches {
+    /// Cache for the `wake_compute` API method.
+    pub node_info: NodeInfoCache,
 }
```
hlinnaka, line 164: Do you have plans for caching other API calls? I would suggest removing ApiCaches, and just use NodeInfoCache directly. Until we actually have other kinds of caches, this is an unnecessary complication, and just makes the code harder to understand.

If we introduce another kind of cache in the future, I'm not sure it would be an "API cache" anyway, and it might not be needed in the same places as NodeInfoCache.

proxy/src/main.rs
```diff
@@ -134,41 +80,108 @@ async fn main() -> anyhow::Result<()> {
...
-    set_build_info_metric(GIT_VERSION);
     // This will block until all tasks have completed.
     // Furthermore, the first one to fail will cancel the rest.
+    let tasks = tasks.into_iter().map(flatten_err);
     let _: Vec<()> = futures::future::try_join_all(tasks).await?;
 
     Ok(())
 }
 
+// Config is expected to live forever.
```
hlinnaka, line 118:
```suggestion
// ProxyConfig is created at proxy startup, and lives forever.
```

proxy/src/config.rs
```diff
@@ -1,16 +1,17 @@
...
-use std::sync::Arc;
+use anyhow::{bail, ensure, Context};
+use std::{str::FromStr, sync::Arc, time::Duration};
 
 pub struct ProxyConfig {
     pub tls_config: Option<TlsConfig>,
     pub auth_backend: auth::BackendType<'static, ()>,
-    pub metric_collection_config: Option<MetricCollectionConfig>,
+    pub metric_collection: Option<MetricCollectionConfig>,
+    pub api_caches: auth::caches::ApiCaches,
```
hlinnaka, line 9: 'api_caches' seems misplaced here. It's not configuration, it's the actual cache. Maybe rename this to ProxyGlobal or something:

hlinnaka:
If you have a burst of 100 connections to the same database simultaneously, does all the connections race and perform 100 lookups in the console, and then all insert the same value in the cache? It would be better to keep track of which lookups are in progress, so that only one connection performs the lookup, and the others wait for it to finish. That is more efficient.

I don't think that's must-have, just something to consider and maybe add on later.

proxy/src/auth/backend.rs
funbringer: > Until we actually have other kinds of caches, this is an unnecessary complication, and just makes the code harder to understand.

That's a strange claim, because the name and the purpose are self-explanatory. Besides, there'll be one more cache (creds cache) per above, so I'm going to leave it as is.

hlinnaka:
This PR includes a lot of seemingly unrelated refactorings. It makes it hard to review. Any chance you could split this into 2-3 smaller PRs? Or are the commits in this PR reviewable separately?

proxy/src/auth/backend/console.rs
funbringer: Done.

proxy/src/cache.rs
```diff
@@ -0,0 +1,221 @@
+use std::{
+    borrow::Borrow,
+    hash::Hash,
+    ops::{Deref, DerefMut},
+    time::{Duration, Instant},
+};
+
+// This seems to make more sense than `lru` or `cached`:
```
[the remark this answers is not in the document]
funbringer: Incorporated with some editing, thanks.

funbringer:
I don't think this PR is too big. Most of the changes are straightforward and actually facilitate the review process. For instance, there used to be an ugly `auth::backend::console::handle_client` which grew completely out of control once I added a few more parameters:

изображение

Now everything related to the console API lives in `src/console` and the rest of the auth backends has been freed from the lower-level boilerplate (http requests and such). We now have a decent `console::Api` abstraction which has helped me tremendously because it battles the overall complexity.

LizardWizzard (approved):
Looks good to me. I've had a call with @funbringer discussing the implementation which resulted in a few extra comments in code. The control flow indeed became simpler

I think we can move forward with testing this on staging

Issue #2031, opened 2022-07-05
Epic: create per-service unit tests for pageserver services


kelvich:
We have a bunch of `lazy_static` variables in the pageserver that make it quite hard to spin up one or more of the pageserver services (like page service, wal service, etc). We did a good job in the safekeeper to allow such tests: https://github.com/neondatabase/neon/blob/main/safekeeper/src/safekeeper.rs#L968-L975 Such test are most quick to run, so we can programmatically generate a lot of edge cases, use more fuzzing, etc.

There are a few blockers in the pageserver codebase for that tests like:
* avoid `lazy_static` global state, as it hard to spin them up in unit tests
* talk with disk and S3 via an intermediary object that could be mocked in tests

Good candidates for such tests:
* branching and GC interaction
* on-demand S3 load
* rebalancing

While this epic is really wide in its scope I suggest writing such tests for one of the suggested topics (e.g. only for rebalancing) to consider this epic as done.

jcsp:
Testing is good, but I don't think this epic is tracking anything specific enough to make the ticket useful

PR #3348, opened 2023-01-15
Add Layer::overlaps method and use it in count_deltas to avoid unnecessary image layer generation

knizhnik:
refer #2948


knizhnik:
For pgbench -i 1000 with subsequent 1000 seconds of updates  the difference in storage size is not so large:

* main: 67GB	
* PR: 56GB

knizhnik:
It is not so easy to reproduce behavior reported in #2948 (so that number of image layers is greater than delta layers).
I was able to get about 2 times difference with main using the following test:
```
pgbench -i -s 1000
pgbench -s 1000 -c 10 -M prepared -T 300 -P 10 -f update.sql
```
update.sql is the following:
```
update pgbench_accounts set abalance = abalance + 1 where aid=1;
update pgbench_accounts set abalance = abalance + 1 where aid=:scale * 100000;
```

main: 22GB
this PR: 14GB

bojanserafimov:
Now we can pile up an unlimited number of deltas. The last lsn cache will help avoid false positives and locate the first relevant layer, but won't help finding the second relevant layer. It's a solveable layer map problem with some effort, but IMO the bigger problem is if we need to download any of these layers on demand. Doesn't feel good to not have any bound on how many layers need to be downloaded to serve a get_page if the layers are not on the pageserver. This was the whole point of compaction from the start: to make on demand download feasible. If we don't care about this we shouldn't be creating L1 layers at all, just use L0.

knizhnik:
> Now we can pile up an unlimited number of deltas. The last lsn cache will help avoid false positives and locate the first relevant layer, but won't help finding the second relevant layer. It's a solveable layer map problem with some effort, but IMO the bigger problem is if we need to download any of these layers on demand. Doesn't feel good to not have any bound on how many layers need to be downloaded to serve a get_page if the layers are not on the pageserver. This was the whole point of compaction from the start: to make on demand download feasible. If we don't care about this we shouldn't be creating L1 layers at all, just use L0.

Yes, I also thought about it.
But do not know some good solution at this moment.

bojanserafimov:
> But do not know some good solution at this moment.

I tried my method of changing the definition of L1 to mean "sufficiently dense layer", such that those 2 out of 10 sparse layers would get re-compacted. I somewhat prefer this approach to adding L2 layers (which was probably the original plan) because at least the L1 layers that come out dense from the start don't need to be compacted.

It was a 2 line change in `LayerMap::get_level0_deltas` but it caused some panics that I haven't investigated yet.

knizhnik:
I do not think that this approach will really work.
First of all because it may lead to infinite compaction loop: layers produced by compactor will never meet "dense" criteria and will e once and once again added to L0 layers list until compacting them will take unacceptable amount of time.

Also, as I found myself with PR trying to limit layer's key range, it is hard if ever possible to find some reasonable key rage threshold. There can be just one big relation (i.e. 1Tb). But it key range will e just 256Mb (so not satisfying your criteria of sparse layer). But it can contain just one page of this relation. Fro the other side, there can be thousands of relatively small relation and splitting them in separate layers will lead to large number of very small layers. Also not good.

We have yesterday discussed this problems with @hlinnaka  and @shanyp 
Followup is here:
https://docs.google.com/document/d/1GcmSW9_DXHou3tuezhjKyL2-ZNuxdRTAaKoi1BC08ns/edit
We decided that first step can be to add to layer map information about ranges, so it will be possible to perform fast lookup wihtout reading layer content from the disk. For our purposes storing ranges seems to be omre efficient than bloom filter.
But still unclear how much ranges can we keep in memory to avoid too much memory footprint of layer map.

So, summarizing all above: right now we can propose four different approaches for solving problems with image layers:
1. Reduce period when image layers are created. For example create them not after 3 but 30 delta layers. Improved layer map can be used to efficiently skip not relevant delta layer. To reduce reconstruction cost with can either store partia images, either rely on reconstructed page cache.
2. Force compaction of sparse layers. I think your approach can be improved by more pricese and sophisticated criteria of layer density (may rely on number of range stored in layer map for this layer)
3. When calculated number of delta layers as criteria of image layer generation, ignore layers not intersecting with target image layer key range (this PR). Once again, instead of accessing B-Tree for checking overlaps we can use ranges stored n memory.
4.  Avoid creation of sparse deltas at all. Looks like key range is not good criteria.Detecting large "holes" seems to be better criteria. But I think we should  take in account not only content of delta layer, but also image coverage. I.e. if we have delta layers with few pages belonging to two relations, we should consider how large are this relations. If them are small, then there is no sense to try to split this delta layer.

bojanserafimov:
Here's the effect on pgbench 10gb:

Screenshot from 2023-01-20 12-48-01

It's not the best workload to test this on, because I don't think this change has much effect on the random updates phase (top of the diagram). But at least we can see that layers are being created later than usual, and all at the same time. So the two long L1s at the bottom probably have large holes and are not affecting the image heuristic.

I don't know if there's a better way to test this than to deploy and see what happens. The cost is not that big (only log_2(10) increased pressure on the new layer map, not sure about current layer map).

Off topic: I don't know how to interpret all the layers on the right end of the key space. Something changed on main branch since fe8cef3427dcd5a2f2d8e96655e898d332859035. See for example: https://github.com/neondatabase/neon/pull/2995#issuecomment-1382293170 which is a PR branched from an older version on main and doesn't have those layers. I also see some 24KB L1 deltas now. There might be some regression on main

hlinnaka:
What would be the worst case workload to demonstrate the problem?

bojanserafimov:
> What would be the worst case workload to demonstrate the problem?

Worst case is probably synthetic uniform random updates. The top 10 holes wouldn't take up a big percentage of the range. But this seems adversarial, and in real workloads I'd expect that the top 10 holes would take up a significant percentage of the range.

hlinnaka:
> > What would be the worst case workload to demonstrate the problem?
> 
> Worst case is probably synthetic uniform random updates. The top 10 holes wouldn't take up a big percentage of the range. But this seems adversarial, and in real workloads I'd expect that the top 10 holes would take up a significant percentage of the range.

I meant, worst case to demonstrate the original problem that this fixes.

bojanserafimov:
> I meant, worst case to demonstrate the original problem that this fixes.

The bigger the database the worse it is. If we have 1TB database we will spontaneously create 1TB images once in a while even if compute is doing nothing but vacuum. We routinely see databases with 5-10x their size in useless images.

It's hard to reproduce this because of nondeterminism. But generally because we do a full reimage of the database once we get 3 batches of L1s that cover the entire key space, and most L1 batches cover the entire keyspace, 30-50GB of wal is (my guess) enough to trigger reimaging. But it can also trigger just from passage of time. We can periodically flush tiny L0s, then compact them into tiny L1s and reimage. We've also seen this in prod.

bojanserafimov:
We could write a binary that we execute against a timelie on prod to download layer metadata: layer range + 100 largest holes.

This gives us enough information to run the `count_deltas` function (locally!!) for each image, and compute how many of those would not have existed with this PR. We can then fine tune the number of splits to create (IMO 10 is the right number, and 100 is also affordable after we pay the tech debt on layer map rebuilds)

Yes, all this testing is extra effort compared to just merging this, waiting a week and then looking at prod metrics, but IMO that local reimaging simulation would make a nice unit test anyway. We should be testing compaction/reimaging/gc/branching/upload/download in unit tests, just using metadata. There's an old issue for this assigned to me that I somehow got distracted from :) https://github.com/neondatabase/neon/issues/2031

pageserver/src/tenant/layer_map.rs
```diff
@@ -200,45 +200,61 @@ where
...
-                    lsn_floor,
-                })
-            }
-            (Some(delta), Some(image)) => {
-                let img_lsn = image.get_lsn_range().start;
-                let image_is_newer = image.get_lsn_range().end >= delta.get_lsn_range().end;
-                let image_exact_match = img_lsn + 1 == end_lsn;
-                if image_is_newer || image_exact_match {
+    pub fn search(&self, key: Key, mut end_lsn: Lsn) -> Option<SearchResult<L>> {
+        loop {
```
bojanserafimov, line 204: It's more efficient to change the insert method instead of search (insert the occupied ranges separately). Maybe leave a TODO if that's not planned for this PR.

pageserver/src/tenant/layer_map.rs
```diff
@@ -417,18 +433,21 @@ where
     /// TODO The optimal number should probably be slightly higher than 1, but to
     ///      implement that we need to plumb a lot more context into this function
     ///      than just the current partition_range.
-    pub fn is_reimage_worthy(layer: &L, partition_range: &Range<Key>) -> bool {
+    pub fn is_reimage_worthy(layer: &L, partition_range: &Range<Key>) -> Result<bool> {
```
bojanserafimov, line 436: nit: It never returns `Err`. Is there a reason for this change?

bojanserafimov (commented):
I left a few nits.

I'm curious about results from at least one of these databases: https://neondb.slack.com/archives/C03UVBDM40J/p1674749418292749

pageserver/src/tenant/layer_map.rs
knizhnik: Will it be ok to have multiple references to the same layers?
Can  it cause returning the same layer twice in iteration through layers?

bojanserafimov: When iterating the coverage it's good to return the same layer multiple times. But `iter_historic` should return each layer only once.

pageserver/src/tenant/layer_map.rs
knizhnik: `overlaps` may return error

knizhnik:
Ugh...
Propagating everywhere `RequestContext` everywhere is really terrible.

Will be pleased if @problame  or @hlinnaka  can review if I did everything as expected.
Also this PR fixes two remote tests added by @hlinnaka - can you look at them please.

bojanserafimov:
Just so the comment doesn't get lost in history:

> I'm curious about results from at least one of these databases: https://neondb.slack.com/archives/C03UVBDM40J/p1674749418292749

these projects have potential for more than 50% images disappearing, and I'm wondering if that will happen

knizhnik:
> Just so the comment doesn't get lost in history:
> 
> > I'm curious about results from at least one of these databases: https://neondb.slack.com/archives/C03UVBDM40J/p1674749418292749
> 
> these projects have potential for more than 50% images disappearing, and I'm wondering if that will happen

`layer_map_analyzer` is now merged and once it is installed at staging/prod it is possible to apply it to all projects=tenants

bojanserafimov (changes requested):
The whole point of the new layer map is to avoid the linear search. This is a regression in that regard. Linear search is avoidable by storing the occupied ranges separately.

knizhnik:
Sorry, which linear search do you mean?
Holes are stored in ascending order and binary search is used.
Or you mean that layer map will have to skip non-overlapping layers by  checking them one-by-one?
But number of delta layers between image layers should not be large. Ideally 3 but even in the worst case - should not be larger than 100. I do not think that it an somehow slowdown search.

bojanserafimov:
> Or you mean that layer map will have to skip non-overlapping layers by checking them one-by-one?

Yes

> But number of delta layers between image layers should not be large. Ideally 3 but even in the worst case - should not be larger than 100.

Is this true? What if we have N layers, where the i-th layer only occupies ranges i..i+1 and i+N..i+N+1? Diagram:
```

           _      |                 _
          _       |                _
         _        |               _
        _         |              _
       _          |             _
      _           |            _
     _            |           _
    _             |          _
   _              |         _
  _               |        _
 _                |       _
_                 |      _
                  v

```


We won't create any image layers because no occupied ranges overlap. But the layer map will have to search N layers, where N could be as large as you want

knizhnik:
I see... Then we should follow your suggestion and handle holes not during search, but in insert: allow tree to contain multiple references to the same layer. And for iteration through historical layer we should either maintain separate collection or exclude duplicates. First seems to e easier.

pageserver/src/tenant/layer_map.rs
```diff
@@ -89,9 +106,11 @@ pub struct LayerMap<L: ?Sized> {
...
     historic: BufferedHistoricLayerCoverage<Arc<L>>,
 
-    /// L0 layers have key range Key::MIN..Key::MAX, and locating them using R-Tree search is very inefficient.
-    /// So L0 layers are held in l0_delta_layers vector, in addition to the R-tree.
-    l0_delta_layers: Vec<Arc<L>>,
+    /// L0 layers are kept in separate set because we need fast way to iterate through L0 layers in compaction
+    l0_delta_layers: HashSet<LayerRef<L>>,
+
+    /// `historic` tree may contain maultuple reference to one layer, so lets maintain separate set instead of eliminating duplicates
+    historic_layers: HashSet<LayerRef<L>>,
```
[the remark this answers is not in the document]
koivunej: No plans by me, I just want to get #3544 merged and continue onwards to adding test cases in #3533. But I should probably review this as well.

pageserver/src/tenant/layer_map.rs
```diff
@@ -63,6 +64,22 @@ pub use historic_layer_coverage::Replacement;
 
 use super::storage_layer::range_eq;
 
+struct LayerRef<L: ?Sized>(Arc<L>);
+
+impl<L: ?Sized> PartialEq for LayerRef<L> {
+    fn eq(&self, other: &LayerRef<L>) -> bool {
+        Arc::ptr_eq(&self.0, &other.0)
```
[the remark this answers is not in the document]
koivunej: Clippy probably doesn't anything because now there's no `dyn Layer` info ever in this context -- note that there's no bound on `L` except `?Sized`. Lack of bound means that when looking at this function, clippy will never have information on the actual usage with `dyn Layer` because it's already erased.

Though I am unsure why is this LayerRef needed all. It's just an arc.

pageserver/src/tenant/layer_map.rs
```diff
@@ -63,6 +64,22 @@ pub use historic_layer_coverage::Replacement;
...
+    }
+}
+
+impl<L: ?Sized> Eq for LayerRef<L> {}
+
+impl<L: ?Sized> Hash for LayerRef<L> {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        Arc::into_raw(Arc::clone(&self.0)).hash(state)
```
koivunej, line 79:
```suggestion
        Arc::as_ptr(&self.0).hash(state)
```

This will avoid incrementing and decrementing the strong counter needlessly around hashing. No comment on if this makes sense ... With the current `Arc::ptr_eq` it might be in violation of the trait if used together with `Eq`. I think I added more comments on that in `LayerMap::compared_arced_layers`.

pageserver/src/tenant/remote_timeline_client/index.rs
```diff
@@ -21,31 +22,41 @@ use utils::lsn::Lsn;
 #[cfg_attr(test, derive(Default))]
 pub struct LayerFileMetadata {
     file_size: Option<u64>,
+    holes: Option<Vec<Hole>>,
```
koivunej, line 25: As noted earlier, would suggest going with `Vec<Hole>`.

pageserver/src/tenant/remote_timeline_client/index.rs
```diff
@@ -233,12 +249,14 @@ impl IndexPart {
 #[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize, Default)]
 pub struct IndexLayerMetadata {
     pub(super) file_size: Option<u64>,
+    pub(super) holes: Option<Vec<Hole>>,
```
koivunej, line 252: The `IndexPart::LATEST_VERSION` should be bumped from 1 to 2, even though it was missed when removing the `missing_layers`.

Could you add an example of this new format with holes populated, similar to `v1_indexpart_is_parsed`?

pageserver/src/tenant/storage_layer/delta_layer.rs
```diff
@@ -290,6 +314,51 @@ impl Layer for DeltaLayer {
         Ok(())
     }
 
+    fn get_occupied_ranges(&self, ctx: &RequestContext) -> Result<Vec<Range<Key>>> {
+        if let Ok(inner) = self.load(ctx) {
```
koivunej, line 318: Shouldn't there be a `?` here, or at least log the error ... feels strange it's just ignored. At least add some justification comment for ignoring it?

test_runner/regress/test_ondemand_download.py
```diff
@@ -212,6 +213,9 @@ def get_resident_physical_size():
     log.info(filled_size)
     assert filled_current_physical == filled_size, "we don't yet do layer eviction"
 
+    # Wait until generated image layers are uploaded to S3
+    time.sleep(3)
```
koivunej, line 217: There's now a `wait_for_upload` helper as well instead of sleeps.

koivunej (commented):
Some preliminary comments. A lot going on. I wonder if we could separate at least the ctx changes from the rest...

pageserver/benches/bench_layer_map.rs
```diff
@@ -33,7 +36,7 @@ fn build_layer_map(filename_dump: PathBuf) -> LayerMap<LayerDescriptor> {
         min_lsn = min(min_lsn, lsn_range.start);
         max_lsn = max(max_lsn, Lsn(lsn_range.end.0 - 1));
 
-        updates.insert_historic(Arc::new(layer));
+        updates.insert_historic(Arc::new(layer), &ctx).unwrap();
```
bojanserafimov, line 39: We can plumb `ctx` everywhere if it's needed. But I wonder what particular change in this PR required the need to plumb `ctx`. Why is it unavoidable?

pageserver/src/tenant/layer_map.rs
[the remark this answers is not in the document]
bojanserafimov: Ok. I'll try doing this change on main branch and I'll see how viable it is

pageserver/src/tenant/layer_map.rs
```diff
@@ -473,18 +543,25 @@ where
...
     ///      than just the current partition_range.
-    pub fn is_reimage_worthy(layer: &L, partition_range: &Range<Key>) -> bool {
+    pub fn is_reimage_worthy(
+        layer: &L,
+        partition_range: &Range<Key>,
+        ctx: &RequestContext,
+    ) -> Result<bool> {
+        if !layer.overlaps(partition_range, ctx)? {
+            return Ok(false);
+        }
```
hlinnaka, line 553: Please update the comment above, for the addition of this case

pageserver/benches/bench_layer_map.rs
hlinnaka: if we really need to add `ctx` to all these places, please extract that to a separate commit, for easier review

pageserver/src/tenant/layer_map.rs
hlinnaka: I'm a little surprised the layer.overlaps function is needed at all. We store the holes in the layer map, and use them during searches to avoid visiting layers unnecessarily, right? Could we use the information from the layer map here, too?

hlinnaka (commented):
The basic idea looks good to me. Needs some more high-level comments and cleanup.

Instead of tracking "holes", how about tracking the ranges that are *present* in a layer, i.e. the opposite of holes? It's usually easier to think of things in the positive sense. For example, it's usually easier toh have "enabled" and "!enabled", than "disabled" and "!disabled".

pageserver/src/tenant/layer_map.rs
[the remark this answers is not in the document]
bojanserafimov: I tried it https://github.com/neondatabase/neon/compare/layer_map_indirection?expand=1

... and microbenchmarks slowed down 2x. I didn't expect a hashmap lookup to have such an effect. Let's postpone this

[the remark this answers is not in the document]
knizhnik: Hmmm, not sure that I completely understand all changes done in this microbenchmark, but are you sure that slowdown is actually caused by hashmap? There are also many extra clones ...

pageserver/src/tenant/remote_timeline_client/index.rs
knizhnik: > I wonder if we could separate at least the ctx changes from the rest...
How?
I still think that TLS is better solution.
Or at least ctx should be added to all methods. Adding it on demand IMHO is most awful approach.

knizhnik: > As noted earlier, would suggest going with Vec<Hole>.

Sorry, I remembered this your suggestion. But failed to find a reference now. And do not remember arguments. Yes, Option can be eliminated and replaced with empty vector. But why it should be done? To save one byte and one check? It seems to be that Option better express situation when information about holes is not present or is not relevant (for example for image layers). And some minor overhead can not be measurable.

pageserver/src/tenant/layer_map.rs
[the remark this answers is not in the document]
bojanserafimov: I'm not sure it's the hashmap. But I just pushed a change to the branch where I remove the extra clones and it's still bad. I don't have a profiling setup right now to see what's the issue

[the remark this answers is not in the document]
bojanserafimov: Could be that hashing a large key takes time..

pageserver/src/tenant/remote_timeline_client/index.rs
knizhnik: Done

pageserver/src/tenant/storage_layer/delta_layer.rs
knizhnik: This is ugly workaround I have to add to make 3 tests passed. Otherwise errors while invalid (specially corrupted) layer loading cause failure in the different place then test expected.

test_runner/regress/test_ondemand_download.py
knizhnik: `wait_for_upload` waits until specified LSN reaches S3, doesn't it?
But how it can enforce uploading of  generated by compaction image layer?
Right now this test performs:
```
    # wait until pageserver has successfully uploaded all the data to remote storage
    wait_for_sk_commit_lsn_to_reach_remote_storage(
        tenant_id, timeline_id, env.safekeepers, env.pageserver
    )
```

but it doesn't help.

pageserver/src/tenant/layer_map.rs
```diff
@@ -89,9 +106,11 @@ pub struct LayerMap<L: ?Sized> {
     /// Index of the historic layers optimized for search
     historic: BufferedHistoricLayerCoverage<Arc<L>>,
 
-    /// L0 layers have key range Key::MIN..Key::MAX, and locating them using R-Tree search is very inefficient.
-    /// So L0 layers are held in l0_delta_layers vector, in addition to the R-tree.
-    l0_delta_layers: Vec<Arc<L>>,
+    /// L0 layers are kept in separate set because we need fast way to iterate through L0 layers in compaction
```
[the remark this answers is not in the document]
knizhnik: Layer may not contain holes (as image layer), but always contain at least one range.
So keeping holes is a little bit more compact and what is more important - easier to maintain (to provide backward compatibility of formats).
Frankly speaking this two arguments for for outweigh negative connotation of holes:)

pageserver/src/tenant/layer_map.rs
knizhnik: `overlaps` was originally introduced in this PR and used in `LayerMap::search`. 
Right now it seems to be really not needed, because counting stacked deltas is based on stored key ranges, not on key range of layer itself. @bojanserafimov  cam you confirm it?

knizhnik:
> if we really need to add ctx to all these places, please extract that to a separate commit, for easier review

Sorry, how you can imagine it?
Some of new methods, i.e. `get_occupied_ranges()` require context. 
So I need to propagate it through all chain of calls.
I can not add `get_occupied_ranges()`  without passing context.
And I can not just pass context - where? without having this method. Do separate commit and pass `_ctx` to all potential callers of  `get_occupied_ranges()` ? How I can find all this places if compiler  will not do this work for me? And even if I managed to do it, then still after adding `get_occupied_ranges()` I will have to change '_ctx' to 'ctx'. So it slightly reduce amount of induced changed in the commit, but not so much.

IMHO we should either force propagation of context in all pageserver methods. Either use TLS.
Adding them on demand is most awful solution (small change can cause avalanche  of other changes).

test_runner/regress/test_ondemand_download.py
koivunej: Oki I was just looking at the comment above the sleep and thought of `wait_for_upload`.

pageserver/src/tenant/layer_map.rs
koivunej: This would had been a memory leak with `into_raw`.

pageserver/src/tenant/layer_map.rs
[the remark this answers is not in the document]
koivunej: > ... and microbenchmarks slowed down 2x

Quite surprising. Cannot see any clear cause right away.

bojanserafimov:
> Some of new methods, i.e. get_occupied_ranges() require context.

Why? We need `ctx` to pass it to `get_occupied_ranges`, which needs it to call `self.load`, because the holes are stored in the delta's `inner`. But the holes are already precomputed. We don't need to specify download behavior. Why not make `holes` accessible without `load`?

bojanserafimov:
It's not that I don't like the eyesore of seeing `ctx` in the layer map. It's that having `ctx` in the layer map is a red flag, that implies we might potentially do expensive operations like layer download, and that should never be the case, especially not inside the `insert_historic` method, under a write lock.

The layer map should use already available metadata.

But I'm not sure if the right fix is to change your code, or change `RequestContext`. Why does `RequestContext` require `TastKind` in the case when `download_behavior` is `Error`? Why would we ever use `Error`, instead of statically preventing the error.. Not sure I haven't read that code yet.

pageserver/src/tenant/remote_timeline_client/index.rs
koivunej: I think you replaced the `v1_indexpart_is_parsed`? Please just copy a new test for it. The idea is to have "images" of all these indexparts we *must* support in the tests. We still need to support v1 indexparts.

pageserver/src/tenant/remote_timeline_client/index.rs
koivunej: First mentions here: https://neondb.slack.com/archives/C0277TKAJCA/p1674829578261489?thread_ts=1674828847.998759&cid=C0277TKAJCA

I did try to ask about it here: https://neondb.slack.com/archives/C033RQ5SPDH/p1674810428051729?thread_ts=1674809569.138119&cid=C033RQ5SPDH based on your answer on the next message I settled with "Option<Vec<_>> is not needed".

This is not a space optimization, because there most likely is no space optimization, or if there is, it's one usize (bringing the size of a Option<Vec<_>> to 4 usizes).

> But why it should be done?

It comes down to what do you want to communicate with this. Do we need to distinguish the "value not set" and "empty vec". Usually one does not, like you answered in slack. 

This would be different if we were able to express `NonEmptyVec<_>` ergonomically but no one usually has time or energy, so we settle with `Option<Vec<_>>` made into `Vec<_>`, because all collections can play the part of "empty" and the reader has less to think about.

In Java circles, I remember a lot of blog posts on the topic "never return null for collection", which is a bit different angle because of the implicit nullability but still the same story.

pageserver/benches/bench_layer_map.rs
koivunej: Or a separate PR, but I think this depends on the "should LayerMap receive layers with full hole information or can it calculate them on demand". I think this would be possible to side-step by requiring that the layers have the metadata before being put into the LayerMap.

pageserver/src/tenant/storage_layer/delta_layer.rs
koivunej: I think this in the center of needing to add `ctx` around in LayerMap. Could there be an another way of always making sure we have the correct information available for layers in the LayerMap..

This comment could also be part of: https://github.com/neondatabase/neon/pull/3348/files#r1098257312

pageserver/src/tenant/layer_map.rs
```diff
@@ -332,28 +363,67 @@ where
...
+                        // expected, it means they differ only on vtables.
+                        //
+                        // otherwise there's no known reason for this to happen as
+                        // compacted layers should have different covering rectangle
+                        // leading to produce Replacement::NotFound.
+
+                        anyhow::bail!(
+                            "replacing downloaded layer into layermap failed because another layer was found instead of expected"
+                        );
+                    }
```
koivunej, line 402: I am quite unsure of the outcome here (making these all errors) at least for on-demand download. I will investigate this more in later PR, so this will have to adapt to those changes.

koivunej:
Related to the ctx requirement, we already have 1-2 threads on it, perhaps keep that discussion there? https://github.com/neondatabase/neon/pull/3348#discussion_r1098257312

Is it obvious for else how will this work in practice? I think the ideal situation would be that we could just pull some existing tenant, and have this new code regenerate all of the files for that, then compare and bench. We probably cannot do that right now? Should we have such testing facilities? I think having such would make testing in future easier, for example compression etc, other layer layout ideas coming in future.

pageserver/src/tenant/remote_timeline_client/index.rs
knizhnik: Ok

pageserver/src/tenant/storage_layer/delta_layer.rs
knizhnik: > Could there be an another way of always making sure we have the correct information available for layers in the LayerMap

I will be very happy if there is such way. But I do not know one except TLS. And looks like nobody likes it.

knizhnik:
> It's not that I don't like the eyesore of seeing `ctx` in the layer map. It's that **having `ctx` in the layer map is a red flag**, that implies we might potentially do expensive operations like layer download, and that should never be the case, especially not inside the `insert_historic` method, under a write lock.
> 
> The layer map should use already available metadata.
> 
> But I'm not sure if the right fix is to change your code, or change `RequestContext`. I haven't read that code yet.

Information about holes is stored in index_part.json and is used when we download file from S3.
But it is not stored locally at pageserver: it is reconstructed when layer is loaded. And it is now done on demand.

I do not understand why "ctx" in LayerMap is red flag. RequestContext is actually just a way to propagate task specific context.
From my point of view it should be everywhere (in each our method) or never (use LS when it is actually needed).

pageserver/src/tenant/layer_map.rs
bojanserafimov: > counting stacked deltas is based on stored key ranges, not on key range of layer itself.

Thanks for checking, because this is false :) I think this should be the only place that needs change: https://github.com/neondatabase/neon/blob/58fa4f0eb7f237e5d327566819eb4317b951554f/pageserver/src/tenant/layer_map.rs#L608

Instead of iterating layers, here we should be iterating occupied ranges of all layers.

But yes, I don't see a need for `layer.overlaps`.

pageserver/benches/bench_layer_map.rs
bojanserafimov: > I do not understand why "ctx" in LayerMap is red flag. RequestContext is actually just a way to propagate task specific context. From my point of view it should be everywhere (in each our method) or never (use LS when it is actually needed).

First reason: The layer map is just a data structure. It should be possible to separate it into its own lib crate. Why? When I rewrote the layer map I spent 1 week writing layer map code and 2 months dealing with the lack of separation between the layer map and the caller (no tests, no documented assumptions, too many requirements of a minimum viable implementation, etc.) What we're doing here with `ctx` is equivalent to plumbing `ctx` into the `HashMap` implementation because we want to read a delta file inside `DeltaLayer::hash`. It's bloated. It's not modular.

Second reason: The need to plumb `ctx` for this PR reveals that the layer map needed some kind of access to some kind of expensive operation worth tracking. When someone reads this code, their guess would be that it either:
a) it reads the holes from the header of the delta layer
b) it reads the entire delta layer to construct the holes
c) it downloads layers (pls no!)

A paranoid reader will assume it's (c) and will have to read a lot more code just to make sure. A careless reader will assume it's (a) and move on. In reality it's (b), so nobody is right. Reading code should not be an adventure. The code should make it obvious what's going on and what can be improved. The reader's attention should be on the important bits. Given how complicated the layer map already is, I'd try to remove ctx completely so that there would be no doubt, but if we have to hack something I'd at least be very explicit with comments and TODOs explaining how it can be done better.

Overall, I agree that if we have `ctx` we should plumb it in all methods proactively. But there are exceptions:
1. We should plumb it in all pageserver code, not all libraries. The layer map is not a library, but it really needs to be (see my reasoning above)
2. In theory, more context is always better. But passing the nuclear launch codes into a `get_timezone` function will raise red flags. We're now passing error handling logic into a function that should never attempt to do any of those things that need the error handling.

It seems like an avoidable problem given that we actually compute the holes explicitly whenever we create a delta layer and add it to the layer map. Why not keep or pass that information?

https://github.com/neondatabase/neon/blob/56af67b5931569525b80256b4a9549d05e2fbc9a/pageserver/src/tenant/timeline.rs#L2422-L2429 They're already computed, but we forget this (compile time) information.

hlinnaka: > But passing the nuclear launch codes into a get_timezone function will raise red flags.

:laughing:

FWIW, I agree 100% on your reasoning.

hlinnaka: This shows that it was the right decision to pass `ctx` to all the functions explicitly, instead of using e.g. a task-local variable. It's annoying that we need to pass `ctx` to all the functions - but it's not *literally* all the functions, as this example shows.

knizhnik: Sorry, I am really confused.
- All reviewers complain that propagation of context complicates PR and make review more difficult (frankly speaking I do not completely agree - it is quite easy to recognize that change is caused by adding context parameter and it is not necessary to spend a lot of time trying to understand and verify all this similar changes).
- There was suggestion to separate propagation of context to separate commit (or PR?). I have no idea of how it can be implemented.
- @bojanserafimov  wrote that passing context to layer map is "red flag" but agree that "if we have ctx we should plumb it in all methods proactively". Certainly I am not suggesting to add context to `get_timezone`. Certainly it should affect pageserver code. What is the precise criteria? I am not sure. May be if method is calling some other pageserver methods, then doesn;t matter whether them need context or not, it should propagate context.
- @hlinnaka wrote "I agree 100% on your reasoning" (about red flag) and then "this shows that it was the right decision to pass ctx to all the functions explicitly".  So, if it was right decision, then why so many complaints and concerns? I just have to propagate context to all methods where it is required. I think that if we have to propagate context explicitly, then it is better to do it once for all (or most) pageserver functions rather then do it in lazy way - on demand. But I do not insist on it and didn't try to implement such approach in this PR.

May be it is better to discuss propagation of contexts in separate thread?
From my point of view TLS is still much better approach to associate some context with task.
If you wan to make context read-only: it can be easily enforced by Context interface itself (use public getter functions).
Access to TLS requires minimal overhead: there is register pointing to thread-local area, so it just one more indirection.
If we need to somehow transform context to pass to child, it can be easily done with TLS also.
And the argument that "it is always explicit" is IMHO better to add to pros rather  than cons.The nature of TLS is that it task specific data and so it much better fit for association some context with task than explicit passing some parameter which  relation to task is obscure.

knizhnik: > It seems like an avoidable problem given that we actually compute the holes explicitly whenever we create a delta layer and add it to the layer map. Why not keep or pass that information?

Hmmm.
In this peace of code we are just inserting new delta layer in layer map.
Information about holes is not yet calculated. It is calculated only in one place - in delta layer by traversing disk B-Tree. It is done when layer is "loaded". When new delta layer is created by frozing in-memory layer, then it is marked as "unloaded" and so holes will be constructed on first access to this layer. In principle it is possible to calculate holes for in-memory layer and then pass it to the created delta layer. But why? What are the benefits? And drawback is code duplication.

But in any case, it is possible to pass holes information when new delta layer is created. But this information is 
not stored anywhere. So when we access layer first time after pageserver restart, we in any case have to extract from layer content.

So either I do not understand your proposal, either it can not be implemented.

bojanserafimov: > Information about holes is not yet calculated.

Line 2242 that I linked to, says `let holes = new_delta.get_holes(ctx)?; `. Is this not calculating holes?

> It is calculated only in one place - in delta layer by traversing disk B-Tree.

Let's just **not** traverse the layer B-Tree while holding the layer map write lock. That's all I'm asking. If it can't be done now because it's difficult, add a TODO and issue for later.

Let's move on in this PR, I'll take out the context from the layer map later if necessary.

knizhnik: > Line 2242 that I linked to, says let holes = new_delta.get_holes(ctx)?; . Is this not calculating holes?

Yes, it does. But it is stored in layer's inner. No need to pass it somewhere. Or may be I do not understand you.

> Let's just not traverse the layer B-Tree while holding the layer map write lock. T

It is not done under layer map lock. As I mentioned above it really done in = new_delta.get_holes(ctx)?
where no layer lock is hold.

> Let's address the "reading 256MB under layer map write lock" problem, an

No such problem. And moreover, traversing B-Tree doesn't mean reading all 256Mb. Size of layer index is about 40Mb

bojanserafimov: > It is not done under layer map lock. 

Then why do we need `ctx` under write lock? (Rethorical question, hopefully makes it clear where the code readability problem is)

Ok it's not an **actual** issue, just a readability issue. If it took 10 messages of back and forth to reach this conclusion, it deserves at least a comment in the code. I don't think it's at all obvious.

bojanserafimov: Seems like: "`insert_historic` would usually read the layer to compute holes, but in case of delta insertion that information would already be cached."

Is that the explanation? Or something else?

Even if that's the explanation, it's very fragile to line reordering refactors, and needs an extra comment..

pageserver/src/tenant/layer_map.rs
```diff
@@ -272,50 +294,59 @@ where
...
     ///
     /// Helper function for BatchedUpdates::remove_historic
     ///
-    pub fn remove_historic_noflush(&mut self, layer: Arc<L>) {
-        self.historic
-            .remove(historic_layer_coverage::LayerKey::from(&*layer));
+    pub fn remove_historic_noflush(&mut self, layer: Arc<L>, ctx: &RequestContext) -> Result<()> {
+        let lr = layer.get_lsn_range();
+        for kr in layer.get_occupied_ranges(ctx)? {
+            self.historic.remove(historic_layer_coverage::LayerKey {
```
bojanserafimov, line 329: We assume that no two layers have at least one identical occupied range. If they did, then removing one layer would also remove a part of the other layer.

The assumption holds today and I see no reason it would stop holding. You can even make a stronger statement that "if two layers have identical LSN range, they are disjoint". But IMO the assumption is worth documenting.

pageserver/src/tenant/layer_map.rs
```diff
@@ -705,7 +794,7 @@ where
 
     /// Return all L0 delta layers
     pub fn get_level0_deltas(&self) -> Result<Vec<Arc<L>>> {
-        Ok(self.l0_delta_layers.clone())
+        Ok(self.l0_delta_layers.iter().map(|r| r.0.clone()).collect())
```
bojanserafimov, line 797: Nitpick:
```suggestion
        Ok(self.l0_delta_layers.keys().cloned().collect())
```

bojanserafimov (commented):
Layer map changes look good to me (assuming current open conversations are resolved)

pageserver/benches/bench_layer_map.rs
knizhnik: > Ok it's not an actual issue, just a readability issue. If it took 10 messages of back and forth to reach this conclusion, it deserves at least a comment in the code. I don't think it's at all obvious.

Sorry, looks like it is just some misunderstanding: I didn't understand your concerns and you didn't understand my explanations. As far as almost all reviewer complain about a lot of changes related with propagation of RequestContext and first of all concentrated on this issue I do not notice your main concern - that it can happen that we we perform some expensive operation (extracting holes) under layer map write lock. Actually it is not true, but it happens unintentionally - actually I didn't realize this problem. I will add comment explaining it. But I want to  notice that this problem isn't somehow related with propagation of request contexts. It is present event ven if `DeltaLyaer:::load()` method doesn't require request context  and I do not need to propagate it everywhere.

pageserver/src/tenant/layer_map.rs
knizhnik: > We assume that no two layers have at least one identical occupied range. 

I think you mean not range, but box - including LSN coordinates. Because definitely two layers may have the same key ranges  (i.e. all L0 layers).

> The assumption holds today and I see no reason it would stop holding. 

Actually I am not sure that this assumption is true. Originally the problems with overlapped layer was encountered more than year ago. I don't remember scenarios in which such layers can be generated, but @hlinnaka  has changed layer map to correctly handle this situation. So I believe that current contract is the following:if two layers overlap, then their overlaped segments contains the same content: it is not possible that some wal records corresponding to the particular page and LSN range are stored in one layer and some other wal records - in another layer. @hlinnaka , can you confirm it?

I just want to notice that adding information about holes (or their contrast - occupied ranges) doesn't change something in this assumptions: if we can have two layers with the same key/lsn range as well as we can have two layers with the same occupied key range. So if we need to more precisely document this assumption, it is not directly related with this PR.

pageserver/src/tenant/layer_map.rs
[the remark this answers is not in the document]
koivunej: This was solved with #3558 and #3573.

pageserver/src/tenant/remote_timeline_client/index.rs
koivunej: This looks good, resolving.

pageserver/src/tenant/layer_map.rs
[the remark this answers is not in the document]
knizhnik: I have already integrated this changes in this PR

pageserver/src/tenant/layer_map.rs
```diff
@@ -332,28 +386,67 @@ where
...
+                    Replacement::NotFound => {
+                        // TODO: the downloaded file should probably be removed, otherwise
+                        // it will be added to the layermap on next load? we should
+                        // probably restart any get_reconstruct_data search as well.
+                        //
+                        // See: https://github.com/neondatabase/neon/issues/3533
+                        anyhow::bail!("replacing downloaded layer into layermap failed because layer was not found");
+                    }
+                    Replacement::RemovalBuffered => {
+                        unreachable!("current implementation does not remove anything")
```
koivunej, line 412: This cannot be pulled into the `replace_historic_noflush`, because this was true only for the one callsite.

koivunej (commented):
A quick comment while trying to understand the on-demand calculation of holes while not actually holding any locks. Are there intentional changes for `vendor/postgres-v1{4,5}` as well? I cannot see how these changes could require postgres chjanges.

knizhnik:
> A quick comment while trying to understand the on-demand calculation of holes while not actually holding any locks.

Layers are read-only. So why do we need any locks here?

>  Are there intentional changes for `vendor/postgres-v1{4,5}` as well? I cannot see how these changes could require postgres chjanges.

Sorry, them are completely unrelated. 
I will remove them. Just ignore them for now.

PR #3349, opened 2023-01-15
Update auth.rs

danieltprice:
Update SNI error message. Users now specify the endpoint ID when making a connection to Neon. This should be reflected in the error message.

### DIFF
--- a/proxy/src/auth.rs
+++ b/proxy/src/auth.rs
@@ -42,9 +42,9 @@ pub enum AuthErrorImpl {
     MalformedPassword(&'static str),
 
     #[error(
-        "Project ID is not specified. \
+        "Endpoint ID is not specified. \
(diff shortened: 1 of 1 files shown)


funbringer (changes requested):
Sorry, but this won't cut it. The message mentions `project` because we expect to see `project`, not `endpoint`. This requires changes in the parser, which in its turn requires a decision on whether we should keep supporting the current naming scheme.

Maybe this also affects console UI (legacy urls?).

cc @vadim2404 @kelvich

funbringer (commented):
Ah, after re-reading your change I see that the actual name hasn't changed. Sorry for the confusion. However, the question remains: maybe we should allow `endpoint=...`? We could still preserve `project=` for compatibility.

funbringer (approved):

PR #3351, opened 2023-01-16
Compile unit extension

vadim2404:
There was a Hubspot request about this extension


lubennikovaav (approved):
Approved.

I still don't fully understand, which extensions we are supporting. But I am not going to gatekeep this one. It looks useful, safe and well-maintained.

BTW, why isn't it marked as `trusted` in the original repo?

vadim2404:
> Approved.
> 
> I still don't fully understand, which extensions we are supporting. But I am not going to gatekeep this one. It looks useful, safe and well-maintained.
> 
> BTW, why isn't it marked as `trusted` in the original repo?

I see that no one adds this by default (except some extensions in contrib). Even postgis didn’t do this

Issue #3339, opened 2023-01-15
Demote to info level "sender is dropped while join handle is still alive"


shanyp:
## Logs, links
```
2023-01-11T09:25:33.351344Z  WARN wal_connection_manager{tenant=a90adcfa693b4e13e42ee129e8d5b67a timeline=b4a04357d0201963d0e84f9036f84894}: sender is dropped while join handle is still alive
```

PR #3358, opened 2023-01-16
TaskHandle: tone down `sender is dropped while join handle is still alive`

problame:
Rationale: see comments added as part of this commit.

fixes https://github.com/neondatabase/neon/issues/3339

### DIFF
--- a/pageserver/src/walreceiver.rs
+++ b/pageserver/src/walreceiver.rs
@@ -116,6 +116,10 @@ impl<E: Clone> TaskHandle<E> {
         let join_handle = WALRECEIVER_RUNTIME.spawn(async move {
             events_sender.send(TaskStateUpdate::Started).ok();
             task(events_sender, cancellation_clone).await
+            // events_sender is dropped at some point during the .await above.
+            // But the task is still running on WALRECEIVER_RUNTIME.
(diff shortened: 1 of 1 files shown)


pageserver/src/walreceiver.rs
```diff
@@ -132,7 +136,23 @@ impl<E: Clone> TaskHandle<E> {
                 TaskEvent::End(match self.join_handle.as_mut() {
                     Some(jh) => {
                         if !jh.is_finished() {
-                            warn!("sender is dropped while join handle is still alive");
+                            // Barring any implementation errors in this module, we can
+                            // only arrive here while the task that executes the future
+                            // passed to `Self::spawn()` is still execution. Cf the comment
```
LizardWizzard, line 141: Cf? Is it meant to be check or I miss some abbreviation?

LizardWizzard (approved):

lubennikovaav (approved):

pageserver/src/walreceiver.rs
problame: https://en.wikipedia.org/wiki/Cf.

LizardWizzard: Thanks!

PR #3364, opened 2023-01-17
Enable consumption metrics on prod

lubennikovaav:
This is the final PR to enable metric collection.
After this is merged, I'll prepare a storage release.

The question is : do we want to enable it in all regions at once?


antonyc:
Let's please start from us-west*, then us-east* regions.

PR #3374, opened 2023-01-18
Allow to change pageserver logs dynamically

SomeoneToIgnore:
Part of https://github.com/neondatabase/neon/issues/3218 that allows to tweak pageserver logs dynamically.

For that, https://github.com/neondatabase/tracing/pull/1 is used, see comments in the `libs/utils/Cargo.toml` file.

### DIFF
--- a/libs/utils/Cargo.toml
+++ b/libs/utils/Cargo.toml
@@ -19,8 +19,6 @@ serde_json.workspace = true
 thiserror.workspace = true
 tokio.workspace = true
 tokio-rustls.workspace = true
-tracing.workspace = true
-tracing-subscriber = { workspace = true, features = ["json"] }
 nix.workspace = true
 signal-hook.workspace = true
 rand.workspace = true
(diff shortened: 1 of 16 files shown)


test_runner/regress/test_pageserver_api.py
```diff
@@ -185,3 +190,146 @@ def test_pageserver_http_api_client_auth_enabled(neon_env_builder: NeonEnvBuilde
...
+
+
+def pageserver_custom_scopes() -> List[PageserverLogScope]:
+    return [CustomLogScope("pageserver=debug", True), CustomLogScope("hyper=debug", True)]
+
+
+@pytest.mark.parametrize(
+    "pageserver_log_scope", pageserver_predefined_scopes() + pageserver_custom_scopes()
+)
+def test_pageserver_logs_toggle_overrides_env_var(
```
shanyp, line 209: worth having two timelines here and assert that we don't have the non debug level timeline logs while enabling logs for the second timeline

libs/utils/Cargo.toml
```diff
@@ -31,10 +29,36 @@ serde_with.workspace = true
...
+#
+# Tracing allows the later to change during runtime, and generally cannot statically understand what's behind `%timeline_id` format call.
+# Due to that, env filters like `error,[{tenant=98d670ab7bee6f0051494306a1ab888f}]=debug` are split into two categories.
+# Static, `error` (another example could be `timeline_detail=error`) one that filters based on static event values,
+# and dynamic, `[{tenant=98d670ab7bee6f0051494306a1ab888f}]=debug`, that would require tracing to resolve the filter on every span
+# that might fit to the filter: some span with `tenant` field name and a dynamic value to resolve.
+#
+# Current tracing optimizes the span resolution, and filters out all `[{tenant=98d670ab7bee6f0051494306a1ab888f}]=debug` events because
+# there's a static filter with stricter log level: the global `error` one.
+# Patched version forces tracing to consider such dynamic filters as potential exceptions and check more potentially matching events
```
shanyp, line 54: what is the perf impact ?

test_runner/regress/test_pageserver_api.py
SomeoneToIgnore: Added.

libs/utils/Cargo.toml
koivunej: We could try using the label `run-benchmarks` pre-merge?

pageserver/src/http/routes.rs
```diff
@@ -889,13 +936,17 @@ pub fn make_router(
...
+                .context("Failed to initialize router state")?,
         ))
         .get("/v1/status", status_handler)
         .put(
             "/v1/failpoints",
             testing_api!("manage failpoints", failpoints_handler),
         )
+        .get("/v1/log_filter", get_log_filter_handler)
+        .put("/v1/log_filter", set_log_filter_handler)
+        .post("/v1/reset_log_filter", reset_log_filter_handler)
```
shanyp, line 949: could we use .delete("/v1/log_filter") to maintain the same path for log filtering related?

libs/utils/Cargo.toml
koivunej: It just occured to me that there probably isn't any codepaths exercised in the current benchmarks which would be have been changed, because there's fixed set of filters.

One "OLTP" benchmark would be:

```
pgbench -i -s 10  "hostname=127.0.0.1 port=55432 user=cloud_admin dbname=postgres"
pgbench -S -c 10 -T 100 -P 1  "hostname=127.0.0.1 port=55432 user=cloud_admin dbname=postgres"
pgbench -T 100 -c 10 -P 10  "hostname=127.0.0.1 port=55432 user=cloud_admin dbname=postgres"
pgbench -S -c 10 -T 100 -P 1  "hostname=127.0.0.1 port=55432 user=cloud_admin dbname=postgres"
```

(Used as "OLTP" with walredo work.)

Running this with `neon_local` on a clean `.neon` directory, then repeating it but with:
- another tenant (which doesn't participate in pgbench)
  - could have multiple timelines (branched at initdb_lsn it doesn't matter)
  - don't start pg for these

Then add for example more verbose logging for one timeline, confirm it's not logging much (for example, don't enable gc_loop to log everything), then silence other timeline.

This way the added filter paths would be exercised and we might see small degradation. But as long as we don't have hundreds of filters with matches I don't think there should be any effect...

LizardWizzard:
As discussed moving to draft

PR #3380, opened 2023-01-18
[compute_ctl] Make role deletion spec processing idempotent

ololobus:
Previously we were trying to re-assign owned objects of the already deleted role, which were causing a crash loop in the case when compute was restarted with the same spec right after role deletion. To avoid such cases, check that role is still present before calling `reassign_owned_objects`.

Resolves neondatabase/cloud#3553

### DIFF
--- a/compute_tools/src/compute.rs
+++ b/compute_tools/src/compute.rs
@@ -252,7 +252,7 @@ impl ComputeNode {
         // If connection fails,
         // it may be the old node with `zenith_admin` superuser.
         //
-        // In this case we need to connect with old `zenith_admin`name
+        // In this case we need to connect with old `zenith_admin` name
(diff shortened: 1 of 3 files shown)


lubennikovaav (approved):
the fix looks good to me

What is the spec you refer to in `compute was restarted with the same spec ` ?

ololobus:
> the fix looks good to me

Thanks for looking on that. There are some CI failures, which look unrelated, and e2e passes

> What is the spec you refer to in compute was restarted with the same spec ?

When we delete / rename roles / dbs, we generate a specially formatted spec with some delta operations. So compute processes it, deletes role, for example. But if for some reason k8s restarts compute, it starts with the same spec and tries to follow the same startup sequence, but some objects could be already deleted. I'll rephrase it

PR #3381, opened 2023-01-18
Compile postgres with ICU support

vadim2404:
I saw requests about this

### DIFF
--- a/Dockerfile.compute-node
+++ b/Dockerfile.compute-node
@@ -10,7 +10,8 @@ ARG TAG=pinned
 FROM debian:bullseye-slim AS build-deps
 RUN apt update &&  \
     apt install -y git autoconf automake libtool build-essential bison flex libreadline-dev \
-    zlib1g-dev libxml2-dev libcurl4-openssl-dev libossp-uuid-dev wget pkg-config libssl-dev
+    zlib1g-dev libxml2-dev libcurl4-openssl-dev libossp-uuid-dev wget pkg-config libssl-dev \
+    libicu-dev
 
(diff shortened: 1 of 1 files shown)


kelvich (approved):
Looks good, thank you!

kelvich:
Yes, it will work, but IIUC won't be default unless explicitly specified. Would be good to make it default too. But that may be a different PR.

vadim2404:
At this moment, after `initdb` `en_US.utf-8` will be used for newly created db (because of env var).

I want to test the whole cycle on staging. Most probably, it requires steps you mentioned:)

koivunej:
> At this moment, after initdb en_US.utf-8 will be used for newly created db (because of env var).

If I understand this comment correctly, this will change how initdb's work. Could this introduce some compatibility issues with timelines created before this PR, given the non-reproducable nature of initdb_lsns #2592?

vadim2404:
> > At this moment, after initdb en_US.utf-8 will be used for newly created db (because of env var).
> 
> If I understand this comment correctly, this will change how initdb's work. Could this introduce some compatibility issues with timelines created before this PR, given the non-reproducable nature of initdb_lsns #2592?

Therefore I want to merge this PR after the next release to test these cases. 

Even at this moment, our databases have UTF-8

```
neondb=> select datname, pg_encoding_to_char(encoding) from pg_database;
  datname  | pg_encoding_to_char
-----------+---------------------
 postgres  | UTF8
 template1 | UTF8
 template0 | UTF8
 neondb    | UTF8
```

PR #3383, opened 2023-01-19
tenant::mgr: explicit tracking of initializing & shutting-down states

problame:
This patch wrap the tenants hashmap into an enum that represents the tenant manager's three major states:
- Initializing
- Open for business
- Shutting down. See the enum doc comments for details.

In response, all the users of `TENANTS` are now forced to distinguish those states.
The only major change is in `run_if_no_tenant_in_memory`, which, before this patch, was used by the /attach and /load endpoints. This patch rewrites that method under the name `tenant_map_insert`, replacing the anyhow::Result with a std Result and a dedicated error type.
Introducing this error types allows using `tenant_map_insert` in `tenant_create`, thereby unifying all code paths that create tenants objects to use `tenant_map_insert`.

This is beneficial because we can now systematically prevent tenants from being created, attached, or `/load`ed during pageserver shutdown. The management API remains available, but the endpoints that create new tenants will fail with an error.
More work would need to be done to properly distinguish these errors through HTTP status codes such as 503.

### DIFF
--- a/pageserver/src/consumption_metrics.rs
+++ b/pageserver/src/consumption_metrics.rs
@@ -115,7 +115,7 @@ pub async fn collect_metrics_iteration(
     );
 
     // get list of tenants
-    let tenants = mgr::list_tenants().await;
+    let tenants = mgr::list_tenants().await?;
 
     // iterate through list of Active tenants and collect metrics
     for (tenant_id, tenant_state) in tenants {
@@ -276,7 +276,13 @@ pub async fn calculate_synthetic_size_worker(
             },
         _ = ticker.tick() => {
 
-                let tenants = mgr::list_tenants().await;
+                let tenants = match mgr::list_tenants().await {
+                    Ok(tenants) => tenants,
+                    Err(e) => {
+                        warn!("cannot get tenant list: {e:#}");
+                        continue;
+                    }
+                };
                 // iterate through list of Active tenants and collect metrics
                 for (tenant_id, tenant_state) in tenants {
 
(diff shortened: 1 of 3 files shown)


pageserver/src/tenant/mgr.rs
```diff
@@ -166,21 +197,44 @@ pub fn schedule_local_tenant_processing(
...
+                return;
+            }
+            TenantsMap::Open(tenants) => {
+                let tenants = tenants.clone();
+                *m = TenantsMap::ShuttingDown(tenants.clone());
+                tenants
+            }
+            TenantsMap::ShuttingDown(_) => {
+                error!("already shutting down, this function isn't supposed to be called more than once");
+                return;
```
koivunej, line 224:
```suggestion
        match &mut *m {
            TenantsMap::Initializing => {
                *m = TenantsMap::ShuttingDown(HashMap::default());
                info!("tenants map is empty");
                return;
            }
            TenantsMap::Open(tenants) => {
                let tenants = tenants.clone();
                *m = TenantsMap::ShuttingDown(std::mem::take(tenants));
                tenants
            }
            TenantsMap::ShuttingDown(_) => {
                error!("already shutting down, this function isn't supposed to be called more than once");
                return;
```

pageserver/src/consumption_metrics.rs
```diff
@@ -110,7 +110,12 @@ pub async fn collect_metrics_iteration(
     );
 
     // get list of tenants
-    let tenants = mgr::list_tenants().await;
+    let tenants = match mgr::list_tenants().await {
+        Ok(tenants) => tenants,
+        Err(e) => {
+            anyhow::bail!(e)
+        }
+    };
```
koivunej, line 118:
```suggestion
    let tenants = match mgr::list_tenants().await?;
```

Cannot at least see why this wouldn't work.

koivunej (approved):
I think this looks good.

problame:
Force-push:
- rebased to latest `main`
- addressed @koivunej 's review comments (thanks!)

Issue #3385, opened 2023-01-19
commit `Use actual temporary dir for pageserver unit tests` causes test failures


problame:
Since this commit, when running a sufficient number of tests, plus some bad luck in eviction order, tests fail.

One symptom is this type error message in the test output:

```
2023-01-18T19:41:09.456084Z ERROR writeback of buffer EphemeralPage { file_id: 143, blkno: 1492 } failed: failed to write back to ephemeral file at /tmp/.tmpBRiJZx/tenants/a873a4ca1c37bb76d12a5bc8779f7369/timelines/11223344556677881122334455667788/ephemeral-143 error: No such file or directory (os error 2)
```

The root cause is,presumably, that the `TenantHarness`'s `TempDir` is dropped, and hence the directory deleted from the filesystem, while the ephemeral file ID is still referenced from the page cache.

I'll post more root cause analysis later.

problame:
Root cause analysis in PR #3388

problame:
(cleaning up my stale tickets.)

We ripped out PS page cache's writeback mode a long time ago. 
So, this problem can no longer happen.

PR #3386, opened 2023-01-19
Revert "Use actual temporary dir for pageserver unit tests"

problame:
This reverts commit 826e89b9ce43ce2c4d046b2c5d6376c3de8dbbac.

The problem with that commit was that it deletes the TempDir while there are still EphemeralFile instances open.

At first I thought this could be fixed by simply adding

  Handle::current().block_on(task_mgr::shutdown(None, Some(tenant_id), None))

to TenantHarness::drop, but it turned out to be insufficient.

So, reverting the commit until we find a proper solution.

refs https://github.com/neondatabase/neon/issues/3385

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 /pg_install
 /target
+/tmp_check
+/tmp_check_cli
 __pycache__/
 test_output/
 .vscode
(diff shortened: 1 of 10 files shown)


koivunej (approved):

lubennikovaav (approved):
thank you for fixing it!

Issue #2848, opened 2022-11-17
Pagserver metrics can grow very large


lassizci:
Pageserver `/metrics` endpoint was recently seen producing over 200 Megabyte response. With rather commonly used 60 second scraping interval we are talking about 300 Gigabytes of metrics data per day, which does sound like a lot, although what's lot and how much shoud it be instead might be hard to define.

There's an earlier issue https://github.com/neondatabase/neon/issues/2271 regarding metrics that had been left behind, but the effect wasn't enough for longer term.

See @petuhovskiy 's comments in the PR: https://github.com/neondatabase/neon/pull/2328#discussion_r961601592, https://github.com/neondatabase/neon/pull/2328#pullrequestreview-1094760288

Also findings from @MMeent :
> The worst ones seem to be:
pageserver_storage_operations_seconds_*: 655368 series in buckets, plus
15604 for sum and count each
I suspect this one may be leaking, considering the 15k entries that is much higher than other timeline-distinguished metrics
pageserver_wait_lsn_seconds_*: 109452 series in buckets, plus 2606 in sum and count each
pageserver_getpage_reconstruct_seconds_bucket: also 109452 series in buckets, with 2606 series in sum and count each

> pageserver_io_operations_seconds_*: 79768 series in bucket, plus 9971 in sum and count each
This one may also be leaking; but I'm not 100% sure about that.

> storage_operations_seconds, wait_lsn_seconds and getpage_reconstruct_seconds all have 42 buckets; thats a lot.

kelvich:
cc @stepashka @lassizci @funbringer 

We did some amount of brainstorming on how to "unpublish" Prometheus metrics:

Depending on the service we can have a bit different policies:

* proxy: stop announcing Prometheus metrics when no active connections
* pageserver: stop announcing Prometheus metrics based on the tenant/timeline state, more or less like @lubennikovaav does for usage metering
* safekeeper: stop announcing Prometheus metrics when no connections from compute and pageserver

We also can drop metric with some TTL, e.g. 1 min. 

@stepashka any problems with that approach? With that we would need to have different interpretation of missing data: missing traffic means zero traffic, missing sizes means that it is the last existing value

problame:
Last week we raised the scrape size limit to 600 MiB:
https://github.com/neondatabase/aws/pull/241/files

Reason was that a pageserver got _lots_ of small tenants.

As far as I know, everything has been going fine since then.

So is this really a pressing issue? I think no.
[...]

PR #3409, opened 2023-01-23
Drop per timeline `pageserver_storage_operations_seconds` histograms and create a global one

lassizci:
Related to: https://github.com/neondatabase/neon/issues/2848

`pageserver_storage_operations_seconds` is the most expensive metric we have, as there are a lot of tenants/timelines and the histogram had 42 buckets. These are quite sparse too, so instead of having a histogram per timeline, create a new histogram `pageserver_storage_operations_seconds_global` without tenant and timeline dimensions and replace `pageserver_storage_operations_seconds` with sum and counter.

The `_total` suffix seems to be enforced by the [crate](https://docs.rs/prometheus/latest/prometheus/index.html) so can't keep the old naming. This could be worked around by keeping the histogram, but with just one bucket (perhaps very large or very small value).


test_runner/fixtures/metrics.py
```diff
@@ -61,9 +61,8 @@ def parse_metrics(text: str, name: str = "") -> Metrics:
     "pageserver_smgr_query_seconds_bucket",
     "pageserver_smgr_query_seconds_count",
     "pageserver_smgr_query_seconds_sum",
-    "pageserver_storage_operations_seconds_bucket",
-    "pageserver_storage_operations_seconds_count",
-    "pageserver_storage_operations_seconds_sum",
+    "pageserver_storage_operations_seconds_count_total",
+    "pageserver_storage_operations_seconds_sum_total",
```
problame, line 65: Need `pageserver_storage_operations_seconds_global_{bucket,count,sum}` here as well.

pageserver/src/tenant.rs
```diff
@@ -1231,11 +1234,22 @@ impl Tenant {
...
+                .with_label_values(&["gc"])
+                .observe(duration);
+            STORAGE_TIME_SUM_PER_TIMELINE
                 .with_label_values(&["gc", &self.tenant_id.to_string(), &timeline_str])
-                .start_timer();
-            self.gc_iteration_internal(target_timeline_id, horizon, pitr)
-                .await
+                .inc_by(duration);
+            STORAGE_TIME_COUNT_PER_TIMELINE
+                .with_label_values(&["gc", &self.tenant_id.to_string(), &timeline_str])
```
problame, line 1249: Can we get rid of these `with_label_values()` calls? If the GC loop outlives the `TimelineMetrics`, this will recreate the metrics even though the timeline is deleted.

test_runner/fixtures/metrics.py
lassizci: Thanks, good point, need to test those too. They aren't per tenant anymore, but I'll craft a separate one for those.

pageserver/src/tenant.rs
lassizci: Good question, didn't even think about the possibility. Did you have something already in mind regarding that? Not sure yet how would I approach that.

koivunej: > If the GC loop outlives the `TimelineMetrics`

Unsure if this can happen. Do we have a test case for this? refresh_gc_info would have to pick up the timeline, which would then have to be detached before `timeline.gc().await` is called.. I can see the detach easily happening in between, but the same `Arc<Timeline>` would keep the TimelineMetrics alive, surely?

problame: > Do we have a test case for this? 

Not really, but for me,  `tenant_metrics.py` uncovered missing metric cleanup in the past.

lassizci: > Not really, but for me, tenant_metrics.py uncovered missing metric cleanup in the past.
Do you have any details on how could I reproduce that? I mean we didn't have any kind of cleanup in the past.

> I can see the detach easily happening in between, but the same Arc<Timeline> would keep the TimelineMetrics alive, surely?

That sounds like making sense to me.

Not sure if it would make any difference regarding the behavior, but the original point about calling `with_label_values()` - or perhaps rather initializing the metric here - perhaps seems a bit odd anyway, considering this is the only operation that doesn't go through `TimelineMetrics`. I think we could handle the metrics in `gc_iteration_internal`. That would effectively mean though, that we would drop the metric for garbage collecting all the timelines (`timeline_id="-"`). Unsure how useful is that. But at the same time it could start to be a bit of scope creep in context of the purpose of this PR.

What do you think?

lassizci: So would look like https://github.com/neondatabase/neon/pull/3409/commits/2f1f8c4ebb55458be31a212124ee6e02307cd013, just that would probably make sense to unnest the gc call in that case.

problame: Yeah, that's better.

Regarding the `{timeline_id="-"}` case:

if`cfg(not(testing))`, is `target_timeline_id` ever `Some()`? If it's always `None` in production, are the per_timeline times actually interesting to us?

Or, conversely, we could have a separate histogram that just groups by `"gc", tenant_id`

lassizci: Thank you for pointing this out! I had completely missed the fact that there are no timeline specific ones in production. Probably because all the rest are per timeline (wonder how important that is). Keeping that in mind, I wonder if it would then be a bit confusing to make the call via timeline metrics.

lassizci: >  are the per_timeline times actually interesting to us?

Another thing I previously didn't really think about is that we seem to garbage collect the timelines sequentially: https://github.com/neondatabase/neon/blob/bf63f129ae1b1822dfb2ec689e483f39c772c9a5/pageserver/src/tenant.rs#L1921-L1929. The timer is probably dependent on how many timelines there are, so it's actually a bit unclear to me what are we after with this metric.

That being said, I would think in production use case we actually get a bit more information with the new "summary", as we can `increment(gc_time) / increment(gc_time_count)`. So alleviating the possibly changing timeline count.

So back to the question, I didn't find any stored metrics where there would be `gc` statistics per timeline_id, so that at least supports just getting rid of it.

pageserver/src/metrics.rs
```diff
@@ -375,18 +394,73 @@ pub static WAL_REDO_RECORD_COUNTER: Lazy<IntCounter> = Lazy::new(|| {
...
+        self.metrics.timeline_count.inc();
+        self.metrics.global_histogram.observe(duration);
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct StorageTimeMetrics {
+    timeline_sum: Counter,
+    timeline_count: IntCounter,
+    global_histogram: Arc<Histogram>,
```
koivunej, line 422:
```suggestion
    global_histogram: Histogram,
```

Similar to other counters from `prometheus` this is internally Arc'd.

pageserver/src/metrics.rs
```diff
@@ -495,7 +565,9 @@ impl Drop for TimelineMetrics {
 }
 
 pub fn remove_tenant_metrics(tenant_id: &TenantId) {
-    let _ = STORAGE_TIME.remove_label_values(&["gc", &tenant_id.to_string(), "-"]);
+    let _ = STORAGE_TIME_SUM_PER_TIMELINE.remove_label_values(&["gc", &tenant_id.to_string(), "-"]);
+    let _ =
+        STORAGE_TIME_COUNT_PER_TIMELINE.remove_label_values(&["gc", &tenant_id.to_string(), "-"]);
```
koivunej, line 570:
```suggestion
    let tid = tenant_id.to_string();
    let _ = STORAGE_TIME_SUM_PER_TIMELINE.remove_label_values(&["gc", &tid, "-"]);
    let _ =
        STORAGE_TIME_COUNT_PER_TIMELINE.remove_label_values(&["gc", &tid, "-"]);
```

koivunej: Since the name now changed, why include the timeline_id="-" at all? Does it give some metric storage backend specific benefit?

pageserver/src/tenant/timeline.rs
```diff
@@ -129,7 +129,7 @@ pub struct Timeline {
     ancestor_lsn: Lsn,
 
     // Metrics
-    metrics: TimelineMetrics,
+    pub metrics: TimelineMetrics,
```
koivunej, line 132:
```suggestion
    pub(super) metrics: TimelineMetrics,
```

Using field access from the `gc_iteration_internal` requires the `pub(super)` on the timeline, but no `pub` usage from elsewhere.

koivunej:
I'll see if this can be rebased.

test_runner/regress/test_tenants.py
```diff
@@ -161,6 +162,15 @@ def test_metrics_normal_work(neon_env_builder: NeonEnvBuilder):
...
         )
 
+    # Test (a subset of) pageserver global metrics
+    for metric in PAGESERVER_GLOBAL_METRICS:
+        ps_samples = ps_metrics.query_all(metric, {})
+        assert len(ps_samples) > 0
+        for sample in ps_samples:
+            labels_str = ",".join([f'{key}="{value}"' for key, value in sample.labels.items()])
+            line = "{0}{{{1}}} {2}".format(sample.name, labels_str, sample.value)
+            log.info(line)
```
koivunej, line 172: What kind of testing idea was there, this only logs?

pageserver/src/tenant.rs
koivunej: The current state makes more sense to me.

pageserver/src/tenant/timeline.rs
koivunej: pub(super)

pageserver/src/metrics.rs
koivunej: Inclined to just remove these.

pageserver/src/metrics.rs
koivunej: Removed.

pageserver/src/metrics.rs
koivunej: Aah nope, oki so we still have some metrics which are per timeline, but not just gc.

test_runner/regress/test_tenants.py
koivunej: This does log these three components (histo + sum + count) with only ["operation"], no tenant or timeline. This seems wrong or does prometheus summarize values with more labels them to the one label item?

koivunej: Oki now I understand, histogram has only the operation, count and sum have tenant and timeline. I'll try to add this as a comment.

pageserver/src/tenant.rs
koivunej: Moved this to timeline not to have tenant touch timelines privates. Resolving.

pageserver/src/metrics.rs
koivunej: Nope, this had me misunderstanding this so many times, removed both of these lines because there will never be such values because timeline_id is never `-`. It may have previously been.

koivunej:
@problame, @SomeoneToIgnore I think this is ready to be peeked again. Quick delta if you peeked this earlier:
- gc is now recorded from timeline, because the metric is still (op, tenant, timeline) labeled except for the global one
- comments added
- confusing lines removed
    - see resolved monologues
- record only success cases
    - unsure if @lassizci meant this to change, but I think it makes sense; errors are probably noops anyway which would just create the impression that most of the time we are really fast
    - went through all call sites of `start_timer

I will next audit that those per timeline metrics are in fact removed, wasn't expecting a test pass.

koivunej:
Yeah I think the regression test pass is valid, looked through:
- test_metrics_normal_work
- test_pageserver_metrics_removed_after_detach

problame (approved):
This LGTM.

IIUC we can still reconstruct old metric

```
let _ = STORAGE_TIME.remove_label_values(&["gc", &tenant_id.to_string(), "-"]);
```

in Grafana by using `sum by (operation, tenant_id)`.

The cherry on top would be to replace the `&'static str` that identify the op by an enum.

```
StorageTimeMetrics::new("init logical size", &tenant_id, &timeline_id);
```

Oh and now that I look at it closely, do we want to take the opportunity to replace the spaces with `_` ?

koivunej:
Thanks @problame ...

> The cherry on top would be to replace the &'static str that identify the op by an enum.

Now that you mentioned it I thought we had one already. Can hack it in of course.

> Oh and now that I look at it closely, do we want to take the opportunity to replace the spaces with `_`?

Does grafana not support utf-8 or is this some prometheus naming thing? But yeah, I guess fine for me. All dashboards using these should break.

Issue #3394, opened 2023-01-20
Avoid layer map rebuilds during gc


bojanserafimov:
And test verify that current GC tests actually GC some layers (didn't seem so)

More context + proposed solution:

_Originally posted by @bojanserafimov in https://github.com/neondatabase/neon/pull/2998#discussion_r1081994400_

bojanserafimov:
Tbh the problem is not solved, just severely deprioritized after the increase in gc interval. It's now a p999999 issue or something like that. If people prefer I can keep the issue open, but there's a todo in the code, which is IMO enough for an non-urgent issue nobody is working on.

PR #3417, opened 2023-01-24
Avoid layer map rebuilds

bojanserafimov:
Resolves https://github.com/neondatabase/neon/issues/3394

I wonder if people have comments before I implement the details.

### DIFF
--- a/pageserver/src/tenant/layer_map.rs
+++ b/pageserver/src/tenant/layer_map.rs
@@ -306,6 +306,25 @@ where
         self.historic.rebuild();
     }
 
+    /// Allow (but not require) the layer map to clean up any memory from layers
+    /// with lsn.end <= lsn. This is useful for GC, in order to avoid an expensive
+    /// flush_updates call.
+    pub fn retain_since(&mut self, lsn: Lsn) {
+        // TODO drop entries from HistoricLayerCoverage::historic before this lsn
+        todo!()
+    }
+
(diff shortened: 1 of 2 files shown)


knizhnik (commented):
If some part of relation is not chaged for a long time, then there will be no updates for this key range and layer may remain for arbitrary long time. So it means that `retain_since` will not advane LSN and old layers will never be deleted from it. Unless `remove_historic_lazy` will do it sometimes. But when?  With PiTR=7 days (or even less ) there sees to e completely no sense to run GC frequently. It can be done for example once per day. But is there any sense in this case to postpone layer deletion from layer map even more? Who and when will do it? Spawn one more background task just to preform lazy deletion? Not sure that it is good idea. Why it can not be done by GC even if rebuild of layer map may take a lot of time?

The problem seems to be not time of rebuild itself, but that it is done under exclusive lock preventing any other access to layer map. How lazy delete can help here?

bojanserafimov:
> With PiTR=7 days (or even less ) there sees to e completely no sense to run GC frequently. It can be done for example once per day.

Good point. Let's do it. It's a quick fix that lowers the priority of this issue by a lot.

> If some part of relation is not chaged for a long time, then there will be no updates for this key range and layer may remain for arbitrary long time.

True. Let's go with the other approach (avoiding write lock). If gc runs every second i feel bad taking up 100ms, but if it runs daily then 100ms is negligible.

PR #3423, opened 2023-01-24
fix some anyhow::Context::context calls that should use with_context(format!(...))

problame:
Noticed this while combing through some production logs.


hlinnaka (approved):
I wish there was a clippy lint for this..

PR #3424, opened 2023-01-24
Extract neon deploy jobs into separate workflows

SergeyMelnikov:
Extract deploy jobs from build_and_test.yml to deploy-dev and deploy-prod workflows.
Add trigger to run this workflows after Neon is build and tested on main and release branches.

This will allow us to redeploy/rollback/patch config without full rebuild.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -1,4 +1,4 @@
-name: Test and Deploy
+name: Build and Test
 
 on:
   push:


problame:
> Unfortunately the only way to test this is merge it to main

https://stackoverflow.com/a/67840292

.github/workflows/redeploy.yml
```diff
@@ -0,0 +1,57 @@
...
+      dockerTag:
+        description: 'Docker tag to deploy'
+        required: true
+        type: string
+      branch:
+        description: 'Branch or commit used for configs'
+        required: true
+        type: string
+        default: 'main'
+      environment:
```
problame, line 15: I find the naming of this input confusing, because `environment` as a term is already used by GitHub.
I.e., we have configured GitHub Environments
* dev-eu-west-1
* dev-us-east-2
* prod-us-west-1
* prod-us-east-2
* prod-us-west-2
* prod-eu-central-1
* prod-ap-southeast-1

The `environment` input variable in this PR is one level higher.
How about calling it `environment_kind`.

The naming rule is then:
```
jobs:
 jobname:
    environment:
      # The GH Actions Environment naming scheme is {dev,prod}-$region
      name: ${{ inputs.environment_kind }}-${{matrix.target_region}}

kelvich (approved):

PR #3425, opened 2023-01-24
Tone down walreceiver connection timeout errors

SomeoneToIgnore:
Closes https://github.com/neondatabase/neon/issues/3342

### DIFF
--- a/pageserver/src/walreceiver/walreceiver_connection.rs
+++ b/pageserver/src/walreceiver/walreceiver_connection.rs
@@ -77,9 +77,13 @@ pub async fn handle_walreceiver_connection(
                 info!("DB connection stream finished: {expected_error}");
                 return Ok(());
             }
-            Err(elapsed) => anyhow::bail!(
-                "Timed out while waiting {elapsed} for walreceiver connection to open"
-            ),
+            Err(elapsed) => {
+                // Timing out to connect to a safekeeper node could happen long time, due to
+                // many reasons that pageserver cannot control.
+                // Do not produce an error, but make it visible, that timeouts happen by logging the `event.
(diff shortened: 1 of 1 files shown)


pageserver/src/walreceiver/walreceiver_connection.rs
```diff
@@ -77,9 +77,13 @@ pub async fn handle_walreceiver_connection(
...
                 return Ok(());
             }
-            Err(elapsed) => anyhow::bail!(
-                "Timed out while waiting {elapsed} for walreceiver connection to open"
-            ),
+            Err(elapsed) => {
+                // Timing out to connect to a safekeeper node could happen long time, due to
+                // many reasons that pageserver cannot control.
+                // Do not produce an error, but make it visible, that timeouts happen by logging the `event.
+                info!("Timed out while waiting {elapsed} for walreceiver connection to open");
```
koivunej, line 84: Had to test, what does display on elapsed return: `deadline has elapsed`. Is this intended vs. using `{connect_timeout}` here?

koivunej (approved):
Just small logging related nit, otherwise looking good.

pageserver/src/walreceiver/walreceiver_connection.rs
funbringer: Maybe there should be a metric for this? It's nice to have metrics for bad situations which might spiral out of control.

SomeoneToIgnore: Thank you for spotting, timeout value is indeed what's supposed to be printed, changed.

I agree that here it should be a metric, so we can always see whether we're connected to something and when the last such event had happened.

I planned to use metrics as part of https://github.com/neondatabase/neon/issues/2106 and keep these PRs small for now.

Issue #3341, opened 2023-01-15
Basebackup query handler shouldn't spam the log during upgrade


shanyp:
## Expected result

1. Don't log the stack trace here.
2. If possible, include the current tenant state in the error message, i.e., ... to become Active (current state: Loading)
## Logs, links
- 
```
  2023-01-11T09:37:06.443461Z ERROR query handler for 'basebackup c9f57fb56b3f1055e03e60c903ee9d9c 9ffeb4c14183bdb39217a3abb891bb23 0/1F43208' failed: Timeout waiting for tenant c9f57fb56b3f1055e03e60c903ee9d9c to become Active
Stack backtrace:
   0: anyhow::__private::format_err
             at /home/nonroot/.cargo/registry/src/github.com-1ecc6299db9ec823/anyhow-1.0.66/src/lib.rs:669:13
   1: pageserver::page_service::get_active_tenant_with_timeout::{{closure}}
             at /home/nonroot/pageserver/src/page_service.rs:1013:19
      <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/core/src/future/mod.rs:91:19
   2: pageserver::page_service::get_active_timeline_with_timeout::{{closure}}
             at /home/nonroot/pageserver/src/page_service.rs:1022:46
      <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/core/src/future/mod.rs:91:19
   3: pageserver::page_service::PageServerHandler::handle_basebackup_request::{{closure}}::{{closure}}
             at /home/nonroot/pageserver/src/page_service.rs:639:80
      <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/core/src/future/mod.rs:91:19
   4: <tracing::instrument::Instrumented<T> as core::future::future::Future>::poll
             at /home/nonroot/.cargo/registry/src/github.com-1ecc6299db9ec823/tracing-0.1.37/src/instrument.rs:272:9
   5: pageserver::page_service::PageServerHandler::handle_basebackup_request::{{closure}}
             at /home/nonroot/pageserver/src/page_service.rs:628:5
      <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
[...]

PR #3426, opened 2023-01-24
Tone down page_service timeouts

SomeoneToIgnore:
Closes https://github.com/neondatabase/neon/issues/3341

### DIFF
--- a/pageserver/src/page_service.rs
+++ b/pageserver/src/page_service.rs
@@ -29,7 +29,6 @@ use std::str;
 use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
-use tokio::time::error::Elapsed;
 use tracing::*;
 use utils::id::ConnectionId;
 use utils::postgres_backend_async::QueryError;
@@ -995,11 +994,11 @@ impl postgres_backend_async::Handler for PageServerHandler {
 #[derive(thiserror::Error, Debug)]
 enum GetActiveTenantError {
     #[error(
-        "Timed out waiting {wait_time} for tenant active state. Latest state: {latest_state:?}"
+        "Timed out waiting {wait_time:?} for tenant active state. Latest state: {latest_state:?}"
     )]
     WaitForActiveTimeout {
         latest_state: TenantState,
-        wait_time: Elapsed,
+        wait_time: Duration,
     },
     #[error(transparent)]
     Other(#[from] anyhow::Error),
@@ -1025,18 +1024,19 @@ async fn get_active_tenant_with_timeout(
     tenant_id: TenantId,
 ) -> Result<Arc<Tenant>, GetActiveTenantError> {
     let tenant = mgr::get_tenant(tenant_id, false).await?;
-    match tokio::time::timeout(Duration::from_secs(30), tenant.wait_to_become_active()).await {
+    let wait_time = Duration::from_secs(30);
+    match tokio::time::timeout(wait_time, tenant.wait_to_become_active()).await {
         Ok(Ok(())) => Ok(tenant),
         // no .context(), the error message is good enough and some tests depend on it
         Ok(Err(wait_error)) => Err(GetActiveTenantError::Other(wait_error)),
-        Err(elapsed) => {
+        Err(_) => {
             let latest_state = tenant.current_state();
             if latest_state == TenantState::Active {
                 Ok(tenant)
             } else {
                 Err(GetActiveTenantError::WaitForActiveTimeout {
                     latest_state,
-                    wait_time: elapsed,
+                    wait_time,
                 })
             }
         }
--- a/pageserver/src/page_service.rs
+++ b/pageserver/src/page_service.rs
@@ -13,6 +13,7 @@ use anyhow::Context;
 use bytes::Buf;
 use bytes::Bytes;
 use futures::{Stream, StreamExt};
+use pageserver_api::models::TenantState;
 use pageserver_api::models::{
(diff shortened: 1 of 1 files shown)


pageserver/src/page_service.rs
```diff
@@ -990,27 +992,63 @@ impl postgres_backend_async::Handler for PageServerHandler {
...
+        // no .context(), the error message is good enough and some tests depend on it
+        Ok(Err(wait_error)) => Err(GetActiveTenantError::Other(wait_error)),
+        Err(elapsed) => {
+            let latest_state = tenant.current_state();
+            if latest_state == TenantState::Active {
+                Ok(tenant)
+            } else {
+                Err(GetActiveTenantError::WaitForActiveTimeout {
+                    latest_state,
+                    wait_time: elapsed,
```
koivunej, line 1039: Cannot suggest change but noting a similar use of `tokio::time::error::Elapsed` as the waited time as in #3425.

I suspect tokio::time::timeout does not provide the elapsed time because that'd add N bytes to something which most likely is never needed as most timeouts never fire.

SomeoneToIgnore: Thank you, updated the code to have a proper timeout value in the error message.

pageserver/src/page_service.rs
```diff
@@ -990,27 +991,64 @@ impl postgres_backend_async::Handler for PageServerHandler {
...
+    #[error(transparent)]
+    Other(#[from] anyhow::Error),
+}
+
+impl From<GetActiveTenantError> for QueryError {
+    fn from(e: GetActiveTenantError) -> Self {
+        match e {
+            GetActiveTenantError::WaitForActiveTimeout { .. } => QueryError::Disconnected(
+                ConnectionError::Socket(io::Error::new(io::ErrorKind::TimedOut, e.to_string())),
+            ),
```
koivunej, line 1012: Is it this conversion which prohibits us stacktrace collection and logging..?

SomeoneToIgnore: Yes, since `io::Error` does not have any backtraces in it now.

koivunej: Yeah, just wondering if there should be a comment. For me it always feels wrong to create new `std::io::Error`'s but perhaps this one is good.

SomeoneToIgnore: I'd say, `std::io::Error`'s should be created judiciously, it's not wrong to create them per se with their current design.
One particular pitfall is https://github.com/neondatabase/neon/pull/3254 , when there's 0 path in the 

> Directory not empty (os error 39)

error from `std::io::Error` bubbled up.

Not quite sure what to comment here, since it's basically the type level changes that I did here: it's either `Other(anyhow::Error)` or a certain type of an error, and I've used a more specific type here.
But would be glad to insert anything proposed.

pageserver/src/page_service.rs
```diff
@@ -990,27 +991,64 @@ impl postgres_backend_async::Handler for PageServerHandler {
...
 /// Shorthand for getting a reference to a Timeline of an Active tenant.
-async fn get_active_timeline_with_timeout(
+async fn get_active_tenant_timeline(
     tenant_id: TenantId,
     timeline_id: TimelineId,
-) -> anyhow::Result<Arc<Timeline>> {
-    get_active_tenant_with_timeout(tenant_id)
-        .await
-        .and_then(|tenant| tenant.get_timeline(timeline_id, true))
+) -> Result<Arc<Timeline>, GetActiveTenantError> {
```
shanyp, line 1050: TimelineRef in the future :-)

pageserver/src/page_service.rs
```diff
@@ -990,27 +991,64 @@ impl postgres_backend_async::Handler for PageServerHandler {
...
+}
+
 /// Get active tenant.
 ///
 /// If the tenant is Loading, waits for it to become Active, for up to 30 s. That
 /// ensures that queries don't fail immediately after pageserver startup, because
 /// all tenants are still loading.
-async fn get_active_tenant_with_timeout(tenant_id: TenantId) -> anyhow::Result<Arc<Tenant>> {
+async fn get_active_tenant_with_timeout(
+    tenant_id: TenantId,
```
shanyp, line 1024: shouldn't the mgr expose this functionality?

shanyp (approved):

koivunej (approved):

pageserver/src/page_service.rs
SomeoneToIgnore: I think, the agreed model currently is to follow with something like `TimelineRef(Arc<Timeline>)` which does not make a difference here, but yes, some wrapper should be used here later.

pageserver/src/page_service.rs
koivunej: Perhaps no comment is needed after all. Resolving.

pageserver/src/page_service.rs
SomeoneToIgnore: `mgr` has similar notion of returning active tenants only, when asked: https://github.com/neondatabase/neon/blob/5223b62a194269bc5d19d325eacc357f48ed9302/pageserver/src/tenant/mgr.rs#L248

Generally, we need the timeouts to wait during client request serving, during attaching a certain tenant with many timelines.
For now, something similar to this is done in `tasks.rs` only, where GC and compaction wait for active tenant to start running.
I don't have an answer why this is done so, really, it looks to me that only one approach could be left, but maybe somebody else could know.

PR #3428, opened 2023-01-24
kill switch to change DownloadBehavior::Error to DownloadBehavior::Warn via config

problame:


koivunej (approved):
Looks that all the config places are handled, and nice if this needs to be checked in only one place.

problame:
>  Looks that all the config places are handled, and nice if this needs to be checked in only one place.

To be fair, I just cargo-culted what the last change that added a field to `PageServerConfig` did. 

But, I did some manual testing to ensure that setting the flag does what it's supposed to do.

Merging this now.

PR #3431, opened 2023-01-24
Propagate context to spawn_blocking tasks

koivunej:
Drafted until I test if it would be worth to add `.in_current_span` to `task_mgr`. That'd probably go against most of the expectations, so most likely this is good. And this needs to be tested as well :)

Adds context to stray info messages:

```
<timestamp> Calculating logical size for timeline Y at 0/1F3A1FA8
```

Cc: #3387

### DIFF
--- a/pageserver/src/tenant/timeline.rs
+++ b/pageserver/src/tenant/timeline.rs
@@ -1457,10 +1457,7 @@ impl Timeline {
         up_to_lsn: Lsn,
         cancel: CancellationToken,
     ) -> Result<u64, CalculateLogicalSizeError> {
-        info!(
-            "Calculating logical size for timeline {} at {}",
-            self.timeline_id, up_to_lsn
-        );
+        info!(%up_to_lsn, "Calculating logical size");
         // These failpoints are used by python tests to ensure that we don't delete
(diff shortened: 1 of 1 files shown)


koivunej:
I checked other spawn_blocking call sites, but interestingly there's also in the logs:

```
timestamp kill_and_wait_impl{pid=1048203}: wait successful exit_status=signal: 9 (SIGKILL)
```

koivunej:
Will be getting this ready afternoon. I will try to keep the changes minimal, but just realized that with proper span continuation, we should have the `get tenant/size` span propagated to any `on-demand downloading`, without needing the request contexts.

Similarly we should end up seeing consumption metrics as the source of the downloads, or the walreceiver, or some other get request.

koivunej:
This is not needed at least partly after #3664. (Now there's no more spawn_blocking.)

koivunej:
The relevant log line is now:

```
ts INFO logical_size_calculation_task{tenant=X timeline=Y}: Calculating logical size for timeline Y at $lsn
```

And I don't think it's too interesting to add span to the walredo killing part. Closing.

PR #3432, opened 2023-01-24
Increase gc period to 1h

bojanserafimov:
See discussion https://github.com/neondatabase/neon/pull/3417#issuecomment-1402269432

With 7 days pitr interval, whether we gc every 100 seconds or 1 hour it doesn't make much difference. But with 1 hour gc it means taking the layer map lock for 100ms every hour is not an urgent issue to fix https://github.com/neondatabase/neon/issues/3394

### DIFF
--- a/pageserver/src/tenant/config.rs
+++ b/pageserver/src/tenant/config.rs
@@ -28,7 +28,7 @@ pub mod defaults {
     pub const DEFAULT_COMPACTION_THRESHOLD: usize = 10;
 
     pub const DEFAULT_GC_HORIZON: u64 = 64 * 1024 * 1024;
-    pub const DEFAULT_GC_PERIOD: &str = "100 s";
+    pub const DEFAULT_GC_PERIOD: &str = "1 hr";
     pub const DEFAULT_IMAGE_CREATION_THRESHOLD: usize = 3;
     pub const DEFAULT_PITR_INTERVAL: &str = "7 days";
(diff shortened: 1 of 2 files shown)


LizardWizzard (commented):
What do you ting about changing the configs on the environments instead of changing the default? For most of the tests it makes sense to have a shorter period. Maybe better to do it other way around, have a 1 hr default, and a lower value for tests

bojanserafimov:
@LizardWizzard Yes, can you point me to where the real configs are?

But IMO the default should also be 1h. Tests need a way smaller value to be effective anyway, and they already set it manually. GC tests usually need a smaller layer size too.

shanyp:
We had 100ms when we had PITR set to 30 days..
Should this value by default be a calculated value derived from  PITR ?

configs on disk are located : /storage/page_server/data/tenants/tenant_directory/config

knizhnik (approved):

bojanserafimov:
> Should this value by default be a calculated value derived from PITR ?

Do you have a formula in mind? I couldn't find a formula that we'd change less frequently than the constant itself.

shanyp:
>> Do you have a formula in mind? I couldn't find a formula that we'd change less frequently than the constant itself.
TBH no, but if you don't have one, this seems that this needs to be controlled from the control plane (same as we want for PITR) as once we will settle for some formula we will need to roll it out .

Issue #3438, opened 2023-01-25
sentry should use git commit as version instead of cargo metadata


koivunej:
We don't use cargo metadata. Right now we are hitting the limits of sentry for unknown reason, but it's quite difficult to separate which are interesting issues and which are duplicates.

For pageserver, the version should be the same as defined [here](https://github.com/neondatabase/neon/blob/aabca55d7e7e020a30f6afd286c850af88063f28/pageserver/src/bin/pageserver.rs#L34), so it should be passed to sentry instead of the default cargo metadata based [`release_name!`](https://docs.rs/sentry/latest/sentry/macro.release_name.html) for the init: https://github.com/neondatabase/neon/blob/aabca55d7e7e020a30f6afd286c850af88063f28/pageserver/src/bin/pageserver.rs#L88

PR #3440, opened 2023-01-25
Fix Sentry Version

vadim2404:
Use `GIT_VERSION` to display it in Sentry instead of package version

Closes: #3438

### DIFF
--- a/pageserver/src/bin/pageserver.rs
+++ b/pageserver/src/bin/pageserver.rs
@@ -26,7 +26,7 @@ use utils::{
     logging,
     postgres_backend::AuthType,
     project_git_version,
-    sentry_init::{init_sentry, release_name},
+    sentry_init::init_sentry,
     signals::{self, Signal},
     tcp_listener,
 };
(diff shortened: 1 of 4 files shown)


koivunej (approved):
Previous implementation was just `CARGO_PKG_NAME@CARGO_PKG_VERSION` like `pageserver@0.1.0` which would imply that sentry parses the release string, but the field comment on ClientOptions doesn't mention that: https://docs.rs/sentry/latest/sentry/struct.ClientOptions.html#structfield.release

I think we'll see how this works... I don't know how to create a panic on pageserver in staging for testing how will sentry understand this new version string. We could also just forget about it or we could add a http api which would just `drop(catch_unwind(|| panic!("testing sentry"))); Ok(())` or however one returned a 204 response...

Also noting that we have overrun our sentry plan right now so we couldn't test this right away regardless. So perhaps punt the testing part.

vadim2404:
```2023-01-25T10:07:48.520651Z  INFO version: git:2a54941f03d6c934396acee82c8e6bc13ebe42ea-modified failpoints: true, features: []```

It's how the version will be sent to Sentry

koivunej:
Yeah, I was just wondering where will sentry know that this message comes from the `pageserver` binary? Because before this PR, it could had parsed the version string to know that. Does each pageserver have a different project in sentry, and so the SENTRY_DSN pointing to the project is to determine this?

vadim2404:
yes, there are DSNs per each project

PR #3441, opened 2023-01-25
Leave one Dockerfile for compute node

vadim2404:
I saw no differences between these files except the folder


SomeoneToIgnore (approved):

SomeoneToIgnore:
Interesting, that only v14 postgres-client is installed in the base image, that's used to build these: https://github.com/neondatabase/build/blob/0e25033eb3ddbd555e68f6a5ebb00abaaeff17d2/rust/Dockerfile#L60

lubennikovaav (changes requested):
Dockerfile ARGs didn't work with Kaniko build-args. That was the reason to have two almost identical files.
See this [discussion](https://github.com/neondatabase/neon/pull/2337#issuecomment-1235506580).
 
Was it fixed in the latest version of kaniko?

zoete:
> Dockerfile ARGs didn't work with Kaniko build-args. That was the reason to have two almost identical files. See this [discussion](https://github.com/neondatabase/neon/pull/2337#issuecomment-1235506580).
> 
> Was it fixed in the latest version of kaniko?
We're about to downgrade to 1.7 - Behaviour didn't change when I tested latest.

vadim2404:
we already provide `--build-arg GIT_VERSION=${{ github.sha }}` 

and, as you can see, all images were built

vadim2404:
I found what was wrong with @lubennikovaav's PR.

https://github.com/neondatabase/neon/pull/2337/commits/6e1c179433e198bcce739d2a0f7b0f738664af58

`ARG` must be defined in the corresponding image (when we build Postgres), but not on top of the Dockerfile.

lubennikovaav (approved):
Cool.
If it works, no objections from me

PR #3442, opened 2023-01-25
Expect timeline being stopped during the detach smoke test

SomeoneToIgnore:
Found this error recently: https://neon-github-public-dev.s3.amazonaws.com/reports/main/release/4005062867/index.html#categories/d2116d7e3b88302f27b3d646396b385b/18590f7063e91b53/?attachment=69e899c74f1cbfc5

I could not reproduce it locally, since always received `gc target timeline does not exist` instead, so that test is quite lucky.


koivunej (approved):
I think this is better than to have this be a flakyness reason.

Issue #3062, opened 2022-12-12
Tenant config file update isn't durable which may result in empty/corrupted state


shanyp:
## Steps to reproduce
Create and run a server instance
Update a tenant configuration
Crash the server while it first truncates the tenant configuration file or updates the file

## Expected result
The tenant config file should have the previous configuration

## Actual result
the tenant config file is empty/corrupted

## Logs, links

Suggested fix In persist_tenant_config function:

```[tasklist]
### Tasks
- [ ] create the config file in a config_tmp file
- [ ] flush it
- [ ] rename it to config
- [ ] if all succeeds return OK() 
```

koivunej:
@shanyp renamed this to be about tenant config instead of page server config file, because we don't have AFAIK a way to update pageserver config file via the api.

I did consider fixing this while #3446 but it seemed too large at first, however it's quite trivial. In addition to steps above under:

> In persist_tenant_config function:

I think we'll additionally need:

- require taking a tenant specific, hopefully async mutex in http
- tempfile cleanup as part of tenant load
    - could be left as well, was trying to anticipate review comments

LizardWizzard:
> because we don't have AFAIK a way to update pageserver config file via the api.

Yep, thats correct

> require taking a tenant specific, hopefully async mutex in http

Hmm, I wonder if locking inside fs is enough. E g if two mkdir calls will be racing only one of them should succeed

shanyp:
> tempfile cleanup as part of tenant load
> * could be left as well, was trying to anticipate review comments
Why not leave it in tmpfs and it will be deleted eventually ?

jcsp:
`persist_tenant_config_at` uses `crashsafe_overwrite` in the latest code, so I think this is no longer an issue.

PR #3446, opened 2023-01-25
Allow creating config for attached tenant

koivunej:
This might be a bit controversial change but I thought to write the code takes about as much as writing a proper issue.

Currently `attach` doesn't write a tenant config, because we don't back it up in the first place. The current implementation of `Tenant::persist_tenant_config` does not allow changing tenant's configuration through the http api which will fail because the file wasn't created on attach and `OpenOptions::truncate(true).write(true).create_new(false)` is used.

I think this patch allows for least controversial middle ground which *enables* changing tenant configuration even for attached tenants (not just created tenants).

I also removed the conditional fsync on the directory because it's now a hazard because the file might be created when `!first_save`.

See also:
- #1555

Does not address:
- #3062

Found while trying to reproduce #3387.

### DIFF
--- a/pageserver/src/tenant.rs
+++ b/pageserver/src/tenant.rs
@@ -1854,19 +1854,7 @@ impl Tenant {
         // before this was done conditionally on first_save, but these management actions are rare
         // enough to just fsync it always.
 
-        File::open(target_config_parent)
-            .context("Failed to open config parent")
-            .and_then(|tenant_dir| {
-                tenant_dir
-                    .sync_all()
-                    .context("Failed to fsync config parent")
-            })
-            .with_context(|| {
-                format!(
-                    "Failed to fsync on first save for config {}",
-                    target_config_path.display()
-                )
-            })?;
+        crashsafe::fsync(target_config_parent)?;
 
         Ok(())
     }
--- a/pageserver/src/tenant.rs
+++ b/pageserver/src/tenant.rs
@@ -1823,7 +1823,10 @@ impl Tenant {
             OpenOptions::new()
                 .truncate(true) // This needed for overwriting with small config files
                 .write(true)
-                .create_new(first_save),
+                .create_new(first_save)
+                // when initializing a new tenant, first_save will be true, thus this flag will be
+                // ignored (per rust std docs). later, don't care.
+                .create(true),
         )?;
 
         target_config_file
--- a/pageserver/src/tenant.rs
+++ b/pageserver/src/tenant.rs
@@ -1844,7 +1844,7 @@ impl Tenant {
             .and_then(|_| {
                 target_config_file
                     .sync_all()
-                    .context("Faile to fsync config file")
+                    .context("Failed to fsync config file")
             })
             .with_context(|| {
                 format!(
--- a/pageserver/src/tenant.rs
+++ b/pageserver/src/tenant.rs
@@ -1843,26 +1843,26 @@ impl Tenant {
                 )
             })?;
 
-        // fsync the parent directory to ensure the directory entry is durable
-        if first_save {
-            target_config_path
-                .parent()
-                .context("Config file does not have a parent")
-                .and_then(|target_config_parent| {
-                    File::open(target_config_parent).context("Failed to open config parent")
-                })
-                .and_then(|tenant_dir| {
-                    tenant_dir
-                        .sync_all()
-                        .context("Failed to fsync config parent")
-                })
-                .with_context(|| {
-                    format!(
-                        "Failed to fsync on first save for config {}",
-                        target_config_path.display()
-                    )
-                })?;
-        }
+        // fsync the parent directory to ensure the directory entry is durable.
+        // before this was done conditionally on first_save, but these management actions are rare
+        // enough to just fsync it always.
+        target_config_path
+            .parent()
+            .context("Config file does not have a parent")
+            .and_then(|target_config_parent| {
+                File::open(target_config_parent).context("Failed to open config parent")
+            })
+            .and_then(|tenant_dir| {
+                tenant_dir
+                    .sync_all()
+                    .context("Failed to fsync config parent")
+            })
+            .with_context(|| {
+                format!(
+                    "Failed to fsync on first save for config {}",
+                    target_config_path.display()
+                )
+            })?;
 
         Ok(())
     }
--- a/pageserver/src/tenant.rs
+++ b/pageserver/src/tenant.rs
@@ -1805,6 +1805,13 @@ impl Tenant {
         first_save: bool,
     ) -> anyhow::Result<()> {
         let _enter = info_span!("saving tenantconf").entered();
+        let target_config_parent = target_config_path.parent().with_context(|| {
+            format!(
+                "Config path does not have a parent: {}",
+                target_config_path.display()
+            )
+        })?;
+
         info!("persisting tenantconf to {}", target_config_path.display());
 
         let mut conf_content = r#"# This file contains a specific per-tenant's config.
@@ -1846,12 +1853,9 @@ impl Tenant {
         // fsync the parent directory to ensure the directory entry is durable.
         // before this was done conditionally on first_save, but these management actions are rare
         // enough to just fsync it always.
-        target_config_path
-            .parent()
-            .context("Config file does not have a parent")
-            .and_then(|target_config_parent| {
-                File::open(target_config_parent).context("Failed to open config parent")
-            })
+
+        File::open(target_config_parent)
+            .context("Failed to open config parent")
             .and_then(|tenant_dir| {
                 tenant_dir
                     .sync_all()
--- a/pageserver/src/tenant.rs
+++ b/pageserver/src/tenant.rs
@@ -1840,18 +1840,9 @@ impl Tenant {
 
         target_config_file
             .write(conf_content.as_bytes())
-            .context("Failed to write toml bytes into file")
-            .and_then(|_| {
-                target_config_file
-                    .sync_all()
-                    .context("Failed to fsync config file")
-            })
-            .with_context(|| {
-                format!(
-                    "Failed to write config file into path '{}'",
-                    target_config_path.display()
-                )
-            })?;
+            .context("write toml bytes into file")
+            .and_then(|_| target_config_file.sync_all().context("fsync config file"))
+            .context("write config file")?;
 
         // fsync the parent directory to ensure the directory entry is durable.
         // before this was done conditionally on creating_tenant, but these management actions are rare
--- a/pageserver/src/tenant.rs
+++ b/pageserver/src/tenant.rs
@@ -1807,7 +1807,7 @@ impl Tenant {
     ) -> anyhow::Result<()> {
         let _enter = info_span!("saving tenantconf").entered();
 
-        // imitiate try-block with a closure
+        // imitate a try-block with a closure
         let do_persist = |target_config_path: &Path| -> anyhow::Result<()> {
             let target_config_parent = target_config_path.parent().with_context(|| {
                 format!(
--- a/pageserver/src/tenant.rs
+++ b/pageserver/src/tenant.rs
@@ -1802,7 +1802,7 @@ impl Tenant {
     pub(super) fn persist_tenant_config(
         target_config_path: &Path,
         tenant_conf: TenantConfOpt,
-        first_save: bool,
+        creating_tenant: bool,
     ) -> anyhow::Result<()> {
         let _enter = info_span!("saving tenantconf").entered();
         let target_config_parent = target_config_path.parent().with_context(|| {
@@ -1829,7 +1829,7 @@ impl Tenant {
             OpenOptions::new()
                 .truncate(true) // This needed for overwriting with small config files
                 .write(true)
-                .create_new(first_save)
+                .create_new(creating_tenant)
                 // when creating a new tenant, first_save will be true and `.create(true)` will be
                 // ignored (per rust std docs).
                 //
@@ -1854,7 +1854,7 @@ impl Tenant {
             })?;
 
         // fsync the parent directory to ensure the directory entry is durable.
-        // before this was done conditionally on first_save, but these management actions are rare
+        // before this was done conditionally on creating_tenant, but these management actions are rare
         // enough to just fsync it always.
 
         crashsafe::fsync(target_config_parent)?;
--- a/pageserver/src/tenant.rs
+++ b/pageserver/src/tenant.rs
@@ -1830,8 +1830,11 @@ impl Tenant {
                 .truncate(true) // This needed for overwriting with small config files
                 .write(true)
                 .create_new(first_save)
-                // when initializing a new tenant, first_save will be true, thus this flag will be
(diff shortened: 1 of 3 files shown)


koivunej:
Unrelated regress(release):
- FAILED test_runner/regress/test_tenant_size.py::test_get_tenant_size_with_multiple_branches

pageserver/src/tenant.rs
```diff
@@ -1841,26 +1843,26 @@ impl Tenant {
...
+        // enough to just fsync it always.
+        target_config_path
+            .parent()
+            .context("Config file does not have a parent")
+            .and_then(|target_config_parent| {
+                File::open(target_config_parent).context("Failed to open config parent")
+            })
+            .and_then(|tenant_dir| {
+                tenant_dir
+                    .sync_all()
```
SomeoneToIgnore, line 1857: Since we're touching this, we could use `utils::crashsafe.rs` module here, to have simlar logs for errors.

Also, anyhow context rewording could be used:
`Failed to open config parent` -> `open config parent`, etc.
https://neondb.slack.com/archives/C0277TKAJCA/p1673453329770429

pageserver/src/tenant.rs
```diff
@@ -1823,7 +1822,10 @@ impl Tenant {
             OpenOptions::new()
                 .truncate(true) // This needed for overwriting with small config files
                 .write(true)
-                .create_new(first_save),
+                .create_new(first_save)
+                // when initializing a new tenant, first_save will be true, thus this flag will be
+                // ignored (per rust std docs). later, don't care.
```
SomeoneToIgnore, line 1827: NIT: The comment is somewhat confusing: first, we ignore the parameter, then we don't care? What's the difference? 🙂

test_runner/regress/test_tenant_conf.py
```diff
@@ -158,3 +166,53 @@ def test_tenant_config(neon_env_builder: NeonEnvBuilder):
...
+        env.remote_storage.root / "tenants" / str(tenant_id) / "timelines" / str(timeline_id)
+    )
+    found_any = False
+    for path in Path.iterdir(remote_dir):
+        if not path.is_file():
+            continue
+        found_any = True
+        break
+
+    assert found_any, "should had have files under timeline directory"
```
SomeoneToIgnore, line 197: What do we check, that there's an attach possible afterwards?
We could be more rigorous and check that all data is actually uploaded: https://github.com/neondatabase/neon/blob/0a095894035b4a1ebf660ad04fadbe77d5185475/test_runner/fixtures/neon_fixtures.py#L3260

```suggestion
    assert found_any, "should had have files under timeline remote storage directory"
```

test_runner/regress/test_tenant_conf.py
```diff
@@ -158,3 +166,53 @@ def test_tenant_config(neon_env_builder: NeonEnvBuilder):
...
+
+    # now prepare for attaching it once more
```
SomeoneToIgnore, line 199: Could we instead do `wait_for_upload` and then `detach` HTTP call instead? 
Or is there some certain scenario reproduced in this test? I think, then, some better detailed comments might be placed around the dances with the FS above.

SomeoneToIgnore (commented):
I find the config-related changes always somewhat confusing, since the only real source of truth and use cases now is the console (control plane later?).

Without their specs/demands/stories, all these changes are guessing.
The PR is not bad per se, but I cannot say if it makes any sense due to the above.

pageserver/src/tenant.rs
LizardWizzard: +1 for confusion. I think it refers to combination of `.create_new` + `.create` and the way its handled on stdlib side

pageserver/src/tenant.rs
```diff
@@ -1841,26 +1843,26 @@ impl Tenant {
...
-                })
-                .with_context(|| {
-                    format!(
-                        "Failed to fsync on first save for config {}",
-                        target_config_path.display()
-                    )
-                })?;
-        }
+        // fsync the parent directory to ensure the directory entry is durable.
+        // before this was done conditionally on first_save, but these management actions are rare
```
LizardWizzard, line 1847: NIT: We have the first_save at hand anyway, not sure if this is really needed. If some script changes config in bulk then we can have many extra fsyncs out of that (we will have many fsyncs anyway because config file itself needs an fsync)

Feel free to keep as is. Maybe a comment for such concerns. We can change this at any time if we need to

LizardWizzard (approved):
Thanks! I've noticed that attach doesnt create config files, but didnt realise that config changes can be broken because of that. As an alternative we can always persist default config with attach. And as I see it attach can have tenant knobs as parameters so they're added to the config during the attach itself

> I find the config-related changes always somewhat confusing, since the only real source of truth and use cases now is the console (control plane later?).

> Without their specs/demands/stories, all these changes are guessing.
The PR is not bad per se, but I cannot say if it makes any sense due to the above.

I think this is not really relevant here. The PR fixes a bug which is observed when you call attach and then try to change config with usual config endpoint. I can confirm that this patch fixes the issue (verified locally)

pageserver/src/tenant.rs
koivunej: > I think it refers to combination of `.create_new` + `.create`

Yes, correct. I'll try to word this better.

test_runner/regress/test_tenant_conf.py
koivunej: Detach would probably be better, would allow this to use `neon_simple_env`.

pageserver/src/tenant.rs
koivunej: Note that the `first_save` will be `false` for attached tenants because it's not the initialization time. Perhaps this should be refactored more to get rid of it or renamed as `at_create_tenant` or removed altogether.

We could of course do an exists check before, but less exists checks the better not to add TOCTOU issues. It would be racy right away because we can have two HTTP requests at the same time.

If this were to use in future the mutex (for the race), tempfile (#3062), rename as target, fsync(parent_dir) then it would still require the fsync for the rename.

test_runner/regress/test_tenant_conf.py
koivunej: `wait_for_upload` was used there already.

pageserver/src/tenant.rs
koivunej: Using crashsafe::fsync with the parent.

Moved the parent exists as a precondition, even though we should surely fail without it when writing.

Copied the tenant creation context (outside this fn) as context when calling from http.

pageserver/src/tenant.rs
koivunej: New comment pushed.

koivunej:
@SomeoneToIgnore and @LizardWizzard please peek again for the fixes, tried to make isolated commits.

Workflow failure is not related.

pageserver/src/tenant.rs
koivunej: Renamed `first_save` as `creating_tenant`.

pageserver/src/tenant/mgr.rs
```diff
@@ -245,7 +245,14 @@ pub async fn update_tenant_config(
     get_tenant(tenant_id, true)
         .await?
         .update_tenant_config(tenant_conf);
-    Tenant::persist_tenant_config(&conf.tenant_config_path(tenant_id), tenant_conf, false)?;
+    let tenant_config_path = conf.tenant_config_path(tenant_id);
+    Tenant::persist_tenant_config(&tenant_config_path, tenant_conf, false).with_context(|| {
+        format!(
+            "Failed to write tenant {} config to {}",
```
SomeoneToIgnore, line 251: NIT: note about the context message.

SomeoneToIgnore (approved):

pageserver/src/tenant/mgr.rs
koivunej: Note this is the same as in tenant creation, and is a top level item... What should this be?

SomeoneToIgnore: Oh.

We could just remove the "failed to" prefix or leave as is, if you prefer similarity here.

koivunej: Aah ok. Well, I guess I prefer the similarity, perhaps it's easier picked up on the weekly  refactoring :)

koivunej: On a second thought, why not fix them both.

koivunej: Please see: [435d12d](https://github.com/neondatabase/neon/pull/3446/commits/435d12d912173b4451faf5b22638b4d0e112efc2) and [4f63e93](https://github.com/neondatabase/neon/pull/3446/commits/4f63e930b5e101bf5a4e98a85eccf82d6b0e6ad8)

SomeoneToIgnore (approved):

pageserver/src/tenant/mgr.rs
SomeoneToIgnore: Thank you.

pageserver/src/tenant.rs
LizardWizzard: > Note that the first_save will be false for attached tenants because it's not the initialization time. 

Ah, I got it wrong, I was thinking about:

>  As an alternative we can always persist default config with attach.

So you're right.

pageserver/src/tenant.rs
```diff
@@ -1800,69 +1800,70 @@ impl Tenant {
...
-        target_config_file
-            .write(conf_content.as_bytes())
-            .context("Failed to write toml bytes into file")
-            .and_then(|_| {
-                target_config_file
-                    .sync_all()
-                    .context("Faile to fsync config file")
-            })
-            .with_context(|| {
+        // imitiate try-block with a closure
```
LizardWizzard, line 1810: typo? -> `imitate`

test_runner/regress/test_tenant_conf.py
```diff
@@ -158,3 +166,46 @@ def test_tenant_config(neon_env_builder: NeonEnvBuilder):
...
+    wait_until(
+        number_of_iterations=5,
+        interval=1,
+        func=lambda: assert_tenant_status(http_client, tenant_id, "Active"),
+    )
+
+    env.neon_cli.config_tenant(tenant_id, {"gc_horizon": "1000000"})
+    contents_first = config_path.read_text("utf-8")
+    env.neon_cli.config_tenant(tenant_id, {"gc_horizon": "0"})
+    contents_later = config_path.read_text("utf-8")
```
LizardWizzard, line 207: Were there any problems without explicit utf-8? I think this is the first time in the codebase when encoding is explicitly passed to read_text

LizardWizzard (approved):
Looks good to me, thanks! Two minor comments

test_runner/regress/test_tenant_conf.py
koivunej: I guess I just by default don't trust python environments to be sane and added these on instinct. I'll clear them out.

pageserver/src/tenant.rs
koivunej: Ah, TIL. will fix.

test_runner/regress/test_tenant_conf.py
koivunej: [de63feb](https://github.com/neondatabase/neon/pull/3446/commits/de63febae15e031776badf4c790344112039ab8f)

pageserver/src/tenant.rs
koivunej: [9c0dd23](https://github.com/neondatabase/neon/pull/3446/commits/9c0dd23cc74ce17349144d70f3dabe5b0ae57dac)

pageserver/src/tenant.rs
koivunej: Approvals so I assume both are happy with the de-obfuscation. Apologies for needing the extra roundtrip for these, I cannot seem to shake this habit off.

LizardWizzard: Thats better to have an extra roundtrip to ensure everyone is on the same page, so thats on me (I forgot  to resolve the discussion :))

pageserver/src/tenant/mgr.rs
```diff
@@ -242,10 +242,11 @@ pub async fn update_tenant_config(
...
     info!("configuring tenant {tenant_id}");
-    get_tenant(tenant_id, true)
-        .await?
-        .update_tenant_config(tenant_conf);
-    Tenant::persist_tenant_config(&conf.tenant_config_path(tenant_id), tenant_conf, false)?;
+    let tenant = get_tenant(tenant_id, true).await?;
+
+    tenant.update_tenant_config(tenant_conf);
+    let tenant_config_path = conf.tenant_config_path(tenant_id);
+    Tenant::persist_tenant_config(&tenant.tenant_id(), &tenant_config_path, tenant_conf, false)?;
```
problame, line 249: If we're changing the in-memory config and then fail to persist it, we report an error to the client but still run with the new config. Is this intentional?

koivunej: That part was not explicitly changed on this PR but yeah it seems quite suboptimal. The `update_tenant_config` does a merge all of the given values ... So actually the value we end up writing to the disk is not the *active* tenant config. Raah, I'll open an issue.

koivunej: Modifying the added test in this PR to instead of changing `gc_horizon` twice to in different http `PUT` requests:
1. set `gc_horizon`
2. set `pitr_interval`

Will write to disk as last version:

```toml
# This file contains a specific per-tenant's config.
#  It is read in case of pageserver restart.

[tenant_config]
pitr_interval = "0s"
```

But adding logging to `Tenant::update_tenant_config` shows that the tenant config is first updated to:

```
2023-01-27T15:27:39.514582Z  INFO tenant_config{tenant=db6238f481f8459ce03705f1fa96ab42}: updated config TenantConfOpt {
    checkpoint_distance: None,
    checkpoint_timeout: None,
    compaction_target_size: None,
    compaction_period: None,
    compaction_threshold: None,
    gc_horizon: Some(
        1000000,
    ),
    gc_period: None,
    image_creation_threshold: None,
    pitr_interval: None,
    walreceiver_connect_timeout: None,
    lagging_wal_timeout: None,
    max_lsn_wal_lag: None,
    trace_read_requests: None,
}
```

Then by the second request:

```
2023-01-27T15:27:39.521233Z  INFO tenant_config{tenant=db6238f481f8459ce03705f1fa96ab42}: updated config TenantConfOpt {
    checkpoint_distance: None,
    checkpoint_timeout: None,
    compaction_target_size: None,
    compaction_period: None,
    compaction_threshold: None,
    gc_horizon: Some(
        1000000,
    ),
    gc_period: None,
    image_creation_threshold: None,
    pitr_interval: Some(
        0ns,
    ),
    walreceiver_connect_timeout: None,
    lagging_wal_timeout: None,
    max_lsn_wal_lag: None,
    trace_read_requests: None,
}
```

Creating a new issue, including your comment.

koivunej: Created two new issues, linked below.

PR #3447, opened 2023-01-25
Fix bug while creating unit extension

vadim2404:
after executing

```sql
CREATE EXTENSION unit;
``` 

I saw such error

```
ERROR: could not open file "/usr/local/pgsql/share/extension/unit_prefixes.data" for reading: No such file or directory (SQLSTATE 58P01)
```

### DIFF
--- a/Dockerfile.compute-node
+++ b/Dockerfile.compute-node
@@ -147,6 +147,8 @@ RUN wget https://github.com/df7cb/postgresql-unit/archive/refs/tags/7.7.tar.gz &
     cd postgresql-unit-7.7 && \
     make -j $(getconf _NPROCESSORS_ONLN) PG_CONFIG=/usr/local/pgsql/bin/pg_config && \
     make -j $(getconf _NPROCESSORS_ONLN) install PG_CONFIG=/usr/local/pgsql/bin/pg_config && \
+    # remove pgsql from path because we will move the extension to another location while building the final layer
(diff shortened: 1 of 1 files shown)


Dockerfile.compute-node
```diff
@@ -147,6 +147,8 @@ RUN wget https://github.com/df7cb/postgresql-unit/archive/refs/tags/7.7.tar.gz &
     cd postgresql-unit-7.7 && \
     make -j $(getconf _NPROCESSORS_ONLN) PG_CONFIG=/usr/local/pgsql/bin/pg_config && \
     make -j $(getconf _NPROCESSORS_ONLN) install PG_CONFIG=/usr/local/pgsql/bin/pg_config && \
+    # remove pgsql from path because we will move the extension to another location while building the final layer
```
lubennikovaav, line 150:
```suggestion
    # unit extension's "create extension" script relies on absolute install path to fill some reference tables.
    # We move the extension from '/usr/local/pgsql/' to '/usr/local/'  after it is build. So we need to adjust the path.
    # This one-liner removes pgsql/ part of the path.
    # NOTE: Other extensions that rely on MODULEDIR variable after building phase will need the same fix.
```

lubennikovaav: This fix looks very hacky to me, but I can't suggest any better solution(

Let's at least update the comment that explains what's going on.

lubennikovaav (approved):

PR #3448, opened 2023-01-25
Add metrics for tenants state 

shanyp:
## Describe your changes
Added a metric that allow to monitor tenants state 
https://github.com/neondatabase/neon/issues/3161

- [X] I have added an e2e test for it.

### DIFF
--- a/pageserver/src/metrics.rs
+++ b/pageserver/src/metrics.rs
@@ -112,6 +112,9 @@ static CURRENT_LOGICAL_SIZE: Lazy<UIntGaugeVec> = Lazy::new(|| {
     .expect("failed to define current logical size metric")
 });
 
+// Metrics collected on tenant statss.
+const TENANT_STATE_OPTIONS: &[&str] = &["loading", "attaching", "active", "stopping", "broken"];
+
 pub static TENANT_STATE_METRIC: Lazy<UIntGaugeVec> = Lazy::new(|| {
     register_uint_gauge_vec!(
         "pageserver_tenant_states_count",
@@ -505,6 +508,9 @@ impl Drop for TimelineMetrics {
 
 pub fn remove_tenant_metrics(tenant_id: &TenantId) {
     let _ = STORAGE_TIME.remove_label_values(&["gc", &tenant_id.to_string(), "-"]);
+    for state in TENANT_STATE_OPTIONS {
+        let _ = TENANT_STATE_METRIC.remove_label_values(&[&tenant_id.to_string(), state]);
+    }
 }
 
 use futures::Future;
--- a/pageserver/src/metrics.rs
+++ b/pageserver/src/metrics.rs
@@ -112,6 +112,15 @@ static CURRENT_LOGICAL_SIZE: Lazy<UIntGaugeVec> = Lazy::new(|| {
     .expect("failed to define current logical size metric")
 });
 
+pub static TENANT_STATE_METRIC: Lazy<UIntGaugeVec> = Lazy::new(|| {
+    register_uint_gauge_vec!(
+        "pageserver_tenant_states_count",
+        "Count of tenants per state",
+        &["tenant_id", "state"]
+    )
+    .expect("Failed to register pageserver_tenant_states_count metric")
+});
+
 // Metrics for cloud upload. These metrics reflect data uploaded to cloud storage,
 // or in testing they estimate how much we would upload if we did.
 static NUM_PERSISTENT_FILES_CREATED: Lazy<IntCounterVec> = Lazy::new(|| {
--- a/pageserver/src/metrics.rs
+++ b/pageserver/src/metrics.rs
@@ -5,6 +5,7 @@ use metrics::{
     IntCounter, IntCounterVec, IntGauge, IntGaugeVec, UIntGauge, UIntGaugeVec,
 };
 use once_cell::sync::Lazy;
+use pageserver_api::models::state;
 use utils::id::{TenantId, TimelineId};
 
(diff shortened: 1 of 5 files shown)


pageserver/src/tenant.rs
```diff
@@ -1717,7 +1718,32 @@ impl Tenant {
...
+                        TENANT_STATE_METRIC
+                            .with_label_values(&[&tenant_id.to_string(), new_state.as_str()])
+                            .inc();
+                    }
+                    Err(_sender_dropped_error) => {
+                        info!("Tenant dropped the state updates sender, quitting waiting for tenant state change");
+                        return;
+                    }
+                }
+            }
```
koivunej, line 1744:
```suggestion
            let tid = tenant_id.to_string();
            TENANT_STATE_METRIC
                .with_label_values(&[&tid, current_state.as_str()])
                .inc();
            loop {
                match rx.changed().await {
                    Ok(()) => {
                        let new_state = *rx.borrow();
                        TENANT_STATE_METRIC
                            .with_label_values(&[&tid, current_state.as_str()])
                            .dec();
                        TENANT_STATE_METRIC
                            .with_label_values(&[&tid, new_state.as_str()])
                            .inc();
                    }
                    Err(_sender_dropped_error) => {
                        info!("Tenant dropped the state updates sender, quitting waiting for tenant state change");
                        return;
                    }
                }
            }
```

This will avoid unnecessary allocations.

koivunej (commented):
Rust-wise this looks good. Does the created metric allow efficient querying? Should we be worried about removing the metric, say some time after the tenant has been dropped if it hasn't been re-attached? In relation to #3409 and other metric cleaning up tasks.

If we shutdown then there's of course nothing to worry, and the spawned task doesn't really hold up any shutdown either.

shanyp:
> Rust-wise this looks good. Does the created metric allow efficient querying? Should we be worried about removing the metric, say some time after the tenant has been dropped if it hasn't been re-attached? In relation to #3409 and other metric cleaning up tasks.
> 
> If we shutdown then there's of course nothing to worry, and the spawned task doesn't really hold up any shutdown either.

This is a gauge that will have for each tenant 1 state, with no histogram so this should be efficient
When a tenant is re-attached it will go through the loading phase

pageserver/src/tenant.rs
```diff
@@ -1717,7 +1718,33 @@ impl Tenant {
         tenant_id: TenantId,
         remote_storage: Option<GenericRemoteStorage>,
     ) -> Tenant {
-        let (state, _) = watch::channel(state);
+        let (state, mut rx) = watch::channel(state);
+
+        tokio::spawn(async move {
```
LizardWizzard, line 1723: NIT: Consider using task mgr's version of spawn. Also the fact that we dont store join handle is a little bit concerning but should be fine.

pageserver/src/metrics.rs
```diff
@@ -496,6 +508,9 @@ impl Drop for TimelineMetrics {
 
 pub fn remove_tenant_metrics(tenant_id: &TenantId) {
     let _ = STORAGE_TIME.remove_label_values(&["gc", &tenant_id.to_string(), "-"]);
+    for state in TENANT_STATE_OPTIONS {
```
LizardWizzard, line 511: lets call to_string once in the beginning:

```rust
let tenant_id = tenant_id.to_string();
let _ = STORAGE_TIME.remove_label_values(&["gc", &tenant_id, "-"]);
for state in TENANT_STATE_OPTIONS {
    let _ = TENANT_STATE_METRIC.remove_label_values(&[&tenant_id, state]);
}
```

pageserver/src/metrics.rs
```diff
@@ -112,6 +112,18 @@ static CURRENT_LOGICAL_SIZE: Lazy<UIntGaugeVec> = Lazy::new(|| {
     .expect("failed to define current logical size metric")
 });
 
+// Metrics collected on tenant statss.
+const TENANT_STATE_OPTIONS: &[&str] = &["loading", "attaching", "active", "stopping", "broken"];
```
LizardWizzard, line 116: NIT: we can avoid possible duplication with `TenantState::as_str`.

Nice way to group constants in a module:
```rust
pub mod state {
    const LOADING: &str = "loading";
    // ...
}
```

So they can be referred to as `state::LOADING`.

Naming depends on where this module is placed.

LizardWizzard (approved):
Looks good to me. Left some minor comments. 

Was thinking about alternative approach to change metric in the `set_state` itself instead of separate task. But there doesnt seem to be a real reason to prefer one approach over the other so let it be as it is :)

pageserver/src/tenant.rs
koivunej: Was thinking the same, but then again, using task_mgr here doesn't actually add anything except LOC, since there's only one connection point, which will naturally take care of itself via drop (the watch channel).

pageserver/src/metrics.rs
shanyp: done

pageserver/src/tenant.rs
problame: No, it adds something valuable, namely, that we know that after `task_mgr::shutdown(None, Some(tenant_id), None)`, this task will be gone.

Which is necessary to avoid the following race condition, not covered by this PR:

```
T1: tenant state is changed to Stopping
T1: uninstalls metrics via remove_tenant_metrics
This task: creates new metrics again due to `with_label_values`
T1: drops tenant
This task: exits
```

Result is that now, even though the tenant is gone, there's a stale metric present.

PR #3451, opened 2023-01-26
Add layer_map_analyzer tool

knizhnik:
See #3348

### DIFF
--- a/pageserver/src/bin/layer_map_analyzer.rs
+++ b/pageserver/src/bin/layer_map_analyzer.rs
@@ -153,6 +153,7 @@ fn main() -> Result<()> {
             let mut layers = Vec::new();
             let mut n_deltas = 0usize;
             let mut n_excess_layers = 0usize;
+			let mut n_holes = 0usize;
             for layer in fs::read_dir(timeline.path())? {
                 let layer = layer?;
                 if let Some(mut layer_file) =
@@ -169,21 +170,33 @@ fn main() -> Result<()> {
 
             for i in 0..layers.len() {
                 if !layers[i].is_delta {
+					let mut n_deltas_since_last_image = 0usize;
                     for j in 0..i {
-                        if layers[j].is_delta && layers[j].skips(&layers[i].key_range) {
-                            n_excess_layers += 1;
-                            break;
+						if range_overlaps(&layers[i].key_range, &layers[j].key_range) {
+							if layers[j].is_delta {
+								if layers[j].skips(&layers[i].key_range) {
+									n_holes += 1;
+								} else {
+									n_deltas_since_last_image += 1;
+								}
+							} else {
+								break;
+							}
                         }
                     }
+					if n_deltas_since_last_image < 3 {
+                        n_excess_layers += 1;
+					}
                 }
             }
             println!(
-                "Tenant {} timeline {} delta layers {} image layers {} excess layers {}",
+                "Tenant {} timeline {} delta layers {} image layers {} excess layers {} holes {}",
                 tenant.file_name().into_string().unwrap(),
                 timeline.file_name().into_string().unwrap(),
                 n_deltas,
                 layers.len() - n_deltas,
-                n_excess_layers
+                n_excess_layers,
+				n_holes
             );
             total_delta_layers += n_deltas;
             total_image_layers += layers.len() - n_deltas;
--- a/pageserver/src/bin/layer_map_analyzer.rs
+++ b/pageserver/src/bin/layer_map_analyzer.rs
@@ -171,22 +171,24 @@ fn main() -> Result<()> {
             for i in 0..layers.len() {
                 if !layers[i].is_delta {
                     let mut n_deltas_since_last_image = 0usize;
+					let mut n_skipped = 0usize;
+					let img_key_range = &layers[i].key_range;
                     for j in (0..i).rev() {
-                        if range_overlaps(&layers[i].key_range, &layers[j].key_range) {
+                        if range_overlaps(img_key_range, &layers[j].key_range) {
                             if layers[j].is_delta {
-                                if layers[j].skips(&layers[i].key_range) {
-                                    n_holes += 1;
-                                } else {
-                                    n_deltas_since_last_image += 1;
+                                n_deltas_since_last_image += 1;
+                                if layers[j].skips(img_key_range) {
+                                    n_skipped += 1;
                                 }
                             } else {
                                 break;
                             }
                         }
                     }
-                    if n_deltas_since_last_image < 3 {
+                    if n_deltas_since_last_image >= 3 && n_deltas_since_last_image - n_skipped < 3 {
                         n_excess_layers += 1;
                     }
+					n_holes += n_skipped;
                 }
             }
             println!(
--- a/pageserver/src/bin/layer_map_analyzer.rs
+++ b/pageserver/src/bin/layer_map_analyzer.rs
@@ -2,7 +2,7 @@ use anyhow::Result;
 use std::cmp::Ordering;
 use std::collections::BinaryHeap;
 use std::ops::Range;
-use std::{env, fs, path::PathBuf, str, str::FromStr};
+use std::{env, fs, path::Path, path::PathBuf, str, str::FromStr};
 
 use pageserver::page_cache::PAGE_SZ;
 use pageserver::repository::{Key, KEY_SIZE};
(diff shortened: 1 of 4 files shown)


koivunej:
@knizhnik did you forget to add the new binary, diffs are all about making items accessible?

pageserver/src/bin/layer_map_analyzer.rs
```diff
@@ -0,0 +1,212 @@
...
+    let holes = Vec::new();
+    Some(LayerFile {
+        key_range,
+        lsn_range,
+        is_delta,
+        holes,
+    })
+}
+
+fn get_holes(path: &Path, max_holes: usize) -> Result<Vec<Hole>> {
```
bojanserafimov, line 91: nit: let's document the behavior: "Finds the `max_holes` largest holes, ignoring any that are smaller than `MIN_HOLE_LENGTH`"

pageserver/src/bin/layer_map_analyzer.rs
```diff
@@ -0,0 +1,212 @@
+use anyhow::Result;
```
bojanserafimov, line 1: Top-level docstring would be useful. Suggestion:

"
Tool for extracting content-dependent metadata about layers. Useful for scanning real project layer files and evaluating the effectiveness of different heuristics on them.

Currently it only analyzes holes, which are regions within the layer range that the layer contains no updates for. In the future it might do more analysis (maybe key quantiles?) but it should never return sensitive data.
"

Also maybe `layer_metadata_analyzer` is another alternative name. The current name makes it sound like the input should be a layer map. Your choice.

pageserver/src/bin/layer_map_analyzer.rs
```diff
@@ -0,0 +1,212 @@
...
+        return Ok(());
+    }
+    let storage_path = PathBuf::from_str(&args[1])?;
+    let max_holes = if args.len() > 2 {
+        args[2].parse::<usize>().unwrap()
+    } else {
+        DEFAULT_MAX_HOLES
+    };
+    pageserver::virtual_file::init(10);
+    pageserver::page_cache::init(100);
```
bojanserafimov, line 138: why are these needed? Maybe a comment would help

pageserver/src/bin/layer_map_analyzer.rs
```diff
@@ -0,0 +1,212 @@
...
+        }
+        for timeline in fs::read_dir(tenant.path().join("timelines"))? {
+            let timeline = timeline?;
+            if !timeline.file_type()?.is_dir() {
+                continue;
+            }
+            let mut layers = Vec::new();
+            let mut n_deltas = 0usize;
+            let mut n_excess_layers = 0usize;
+            let mut n_holes = 0usize;
```
bojanserafimov, line 155: Let's declare these before the next for loop because that's when we start computing them. Mutable variables inside a long scope are confusing.

pageserver/src/bin/layer_map_analyzer.rs
```diff
@@ -0,0 +1,212 @@
...
+        let tenant = tenant?;
+        if !tenant.file_type()?.is_dir() {
+            continue;
+        }
+        for timeline in fs::read_dir(tenant.path().join("timelines"))? {
+            let timeline = timeline?;
+            if !timeline.file_type()?.is_dir() {
+                continue;
+            }
+            let mut layers = Vec::new();
```
bojanserafimov, line 152:
```suggestion
            // Collect sorted vec of layers and count deltas
            let mut layers = Vec::new();
```

pageserver/src/bin/layer_map_analyzer.rs
```diff
@@ -0,0 +1,212 @@
...
+                    if layer_file.is_delta {
+                        layer_file.holes = get_holes(&layer.path(), max_holes)?;
+                        n_deltas += 1;
+                    }
+                    layers.push(layer_file);
+                }
+            }
+            layers.sort_by_key(|layer| layer.lsn_range.end);
+
+            for i in 0..layers.len() {
```
bojanserafimov, line 170:
```suggestion
            // Count the number of holes and number of excess layers
            for i in 0..layers.len() {
```

Maybe also define what an excess layer is

pageserver/src/bin/layer_map_analyzer.rs
```diff
@@ -0,0 +1,212 @@
...
+                    let img_key_range = &layers[i].key_range;
+                    for j in (0..i).rev() {
+                        if range_overlaps(img_key_range, &layers[j].key_range) {
+                            if layers[j].is_delta {
+                                n_deltas_since_last_image += 1;
+                                if layers[j].skips(img_key_range) {
+                                    n_skipped += 1;
+                                }
+                            } else {
+                                break;
```
bojanserafimov, line 183: This isn't technically correct, since this image might not cover the full key range. Maybe it's a good approximation but let's put a `// HACK` comment to make it clear.

There's another approximation: We should also check if this image is excess or not. If it's excess we can't break because the image layer wouldn't have existed if we used the new heuristic. **Your call whether it's worth fixing or just documenting.**

And there's yet another: The new layer map doesn't count total deltas, but the max stack of overlapping deltas.

bojanserafimov (approved):
Good stuff!

At some point I'll try integrating this into the visualizer too, so we can see the occupied ranges.

pageserver/src/bin/layer_map_analyzer.rs
bojanserafimov: But probably long term we'd want the real layer_map+reimaging+compaction code to work only with metadata so that we can unit test it and use it here without rewriting it. It's not a small task though.

pageserver/src/bin/layer_map_analyzer.rs
```diff
@@ -0,0 +1,212 @@
...
+use pageserver::tenant::storage_layer::delta_layer::{Summary, DELTA_KEY_SIZE};
+use pageserver::tenant::storage_layer::range_overlaps;
+use pageserver::virtual_file::VirtualFile;
+
+use utils::{bin_ser::BeSer, lsn::Lsn};
+
+const MIN_HOLE_LENGTH: i128 = (128 * 1024 * 1024 / PAGE_SZ) as i128;
+const DEFAULT_MAX_HOLES: usize = 10;
+
+/// Wrapper for key range to provide reverse ordering by range length for BinaryHeap
```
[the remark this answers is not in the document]
knizhnik: Sorry, but `Reverse` can not e used here because I am sorting ranges by length and it requires custom comparator in any case.

pageserver/src/bin/layer_map_analyzer.rs
```diff
@@ -0,0 +1,212 @@
...
+    let is_delta = if lsns.len() == 1 {
+        lsns.push(lsns[0]);
+        false
+    } else {
+        true
+    };
+
+    let key_range = Key::from_hex(keys[0]).unwrap()..Key::from_hex(keys[1]).unwrap();
+    let lsn_range = Lsn::from_hex(lsns[0]).unwrap()..Lsn::from_hex(lsns[1]).unwrap();
+    let holes = Vec::new();
```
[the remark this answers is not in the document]
knizhnik: Sorry, but this code was cut&paste of your draw_timeline_dir.rs
Concerning initialization of `holes` - it will be filled later for delta layers.

pageserver/src/bin/layer_map_analyzer.rs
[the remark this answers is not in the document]
knizhnik: This function was added to Layers trait in #3348

knizhnik: Done

pageserver/src/bin/layer_map_analyzer.rs
knizhnik: 1. Thank you for writing this description for me. Added.
2. Right now holes is not part of layer metadata. It will be if #3348 will be merged.
3.

pageserver/src/bin/layer_map_analyzer.rs
knizhnik: Them are needed to access layer persistent B-Tree.
Comment is added.

pageserver/src/bin/layer_map_analyzer.rs
knizhnik: Done

pageserver/src/bin/layer_map_analyzer.rs
knizhnik: Done

pageserver/src/bin/layer_map_analyzer.rs
knizhnik: Done

pageserver/src/bin/layer_map_analyzer.rs
knizhnik: Well, right now image layer is always dense, despite to the fact that it doesn't contain all possible key values in the specified range: there are may be no keys in the storage belonging to the image layer range but not present in the image layer. 

Comments added.

PR #3454, opened 2023-01-26
Decommission link proxy from old account

SergeyMelnikov:


kelvich:
will the old-style (without region) connection urls still work?

kelvich (approved):

kelvich:
Ah, it is only about link proxy

PR #3456, opened 2023-01-26
move walreceiver module under timeline

problame:
Walreceiver is a per-timeline abstraction. Move it there to reflect the hierarchy of abstractions and task_mgr tasks.
The code that sets up the global storage_broker client is not timeline-scoped. So, break it out into a separate module.

The motivation for this change is to prepare the code base for replacing the task_mgr global task registry with a more ownership-oriented approach to manage task lifetimes.

I removed TaskStateUpdate::Init because, after doing the changes, rustc warned that it was never constructed.
A quick search through the commit history shows that this has always been true since

    commit fb68d01449edb4be9a0d064d69a442dd3688783e
    Author: Dmitry Rodionov <dmitry@neon.tech>
    Date:   Mon Sep 26 23:57:02 2022 +0300

        Preserve task result in TaskHandle by keeping join handle around (#2521)

So, the warning is not an indication of some accidental code removal.

### DIFF
--- a/pageserver/src/bin/pageserver.rs
+++ b/pageserver/src/bin/pageserver.rs
@@ -250,7 +250,7 @@ fn start_pageserver(conf: &'static PageServerConf) -> anyhow::Result<()> {
     let signals = signals::install_shutdown_handlers()?;
 
     // Launch broker client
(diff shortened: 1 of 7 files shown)


koivunej (approved):
Good idea on splitting up and moving.

PR #3457, opened 2023-01-26
use separate type for timeline.rs's representation of Timeline state

problame:
This patch introduces crate::tenant::timeline::State which has the same variants as pageserver_api::models::TimelineState.

This will enable future commits to change state handling in modules `tenant` and `timeline` without affecting the presentation in the API.

### DIFF
--- a/pageserver/src/http/routes.rs
+++ b/pageserver/src/http/routes.rs
@@ -8,16 +8,18 @@ use remote_storage::GenericRemoteStorage;
 use tokio_util::sync::CancellationToken;
 use tracing::*;
 
+use pageserver_api::models as api;
+
 use super::models::{
     StatusResponse, TenantConfigRequest, TenantCreateRequest, TenantCreateResponse, TenantInfo,
-    TimelineCreateRequest, TimelineInfo,
+    TimelineCreateRequest,
 };
 use crate::context::{DownloadBehavior, RequestContext};
 use crate::pgdatadir_mapping::LsnForTimestamp;
 use crate::task_mgr::TaskKind;
 use crate::tenant::config::TenantConfOpt;
 use crate::tenant::mgr::TenantMapInsertError;
-use crate::tenant::{PageReconstructError, Timeline};
+use crate::tenant::{self, PageReconstructError, Timeline};
 use crate::{config::PageServerConf, tenant::mgr};
 use utils::{
     auth::JwtAuth,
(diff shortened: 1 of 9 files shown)


pageserver/src/http/routes.rs
```diff
@@ -167,10 +169,15 @@ fn build_timeline_info_common(
         }
     };
     let current_physical_size = Some(timeline.layer_size_sum().approximate_is_ok());
-    let state = timeline.current_state();
+    let state = match timeline.current_state() {
+        tenant::TimelineState::Loading => api::TimelineState::Loading,
+        tenant::TimelineState::Active => api::TimelineState::Active,
+        tenant::TimelineState::Stopping => api::TimelineState::Stopping,
+        tenant::TimelineState::Broken => api::TimelineState::Broken,
+    };
```
koivunej, line 177: If the objective is to make TimelineState (State) a more local type, we could have

```rust
impl From<tenant::timeline::State> for api::TimelineState {
  fn from(s: tenant::timeline::State) -> Self {
      match s {
        tenant::TimelineState::Loading => api::TimelineState::Loading,
        tenant::TimelineState::Active => api::TimelineState::Active,
        tenant::TimelineState::Stopping => api::TimelineState::Stopping,
        tenant::TimelineState::Broken => api::TimelineState::Broken,
    }
  }
}
```

The above impl could live in pageserver/src/tenant/timeline.rs just fine because I assume this `api` is a dependency of that crate.

And use it here as:

```suggestion
let state = api::TimelineState::from(timeline.current_state());
```

pageserver/src/http/routes.rs
```diff
@@ -8,16 +8,18 @@ use remote_storage::GenericRemoteStorage;
 use tokio_util::sync::CancellationToken;
 use tracing::*;
 
+use pageserver_api::models as api;
```
koivunej, line 11:
```suggestion
use pageserver_api::models;
```

I'd find this less surprising, but I guess it's no biggie with all of the uses we currently need.

koivunej:
I was also thinking about this:

> This will enable future commits to change state handling in modules tenant and timeline without affecting the presentation in the API.

So is the problem that we encode the variants to the openapi yaml?

pageserver/src/http/routes.rs
problame: Generally I'd agree with this suggestion. But:

My WIP changes that I have on top of this PR remove the `tenant::TimelineState::Broken`.
In those WIP changes, if a timeline is Broken, it's a different type.
The `Tenant::timelines` map then has a value of type

```
enum TimelinesMapEntry {
    Broken(BrokenTimeline),
    NotBroken(Arc<Timeline>),
}
```

(Actually, the plans are bigger, but to make this argument here, the above is sufficient).

The status endpoint then looks more like so:
```
   let state = match timelines_map_entry  {
        NotBroken(tl) => match tl.current_state() {
          tenant::TimelineState::Loading => api::TimelineState::Loading,
          tenant::TimelineState::Active => api::TimelineState::Active,
          tenant::TimelineState::Stopping => api::TimelineState::Stopping,
        }
        Broken(_)  => api::TimelineState::Broken,
    };
```

That match then sits inside a 

```
fn api_state(&self: TimelinesMapEntry) => api::TimelineState {
 ...
}
```

IMO preferable over a `From` impl in this case.

problame:
> So is the problem that we encode the variants to the openapi yaml? This change would be the way to go if we'd have this enum and that enum in different repos, in released crates, but we currently don't have that issue.

Nothing outside of pageserver depends on the state variable (thank god!) but a lot of python tests check it.

---

Thanks for the review! Since this patch is so small and is just confusing by itself, I'll let it sit until I have the WIP on top of this ready.

LizardWizzard:
As discussed moving to draft

PR #3460, opened 2023-01-26
Set the last_record_gauge to the value which was persisted metadata

shanyp:
## Describe your changes
Whenever a tenant is detached or the pageserver is restarted the pageserver_last_record_lsn metric is dropped 
This fix resurrects the value from the metadata whenever the tenant is attached again
[3571](https://github.com/neondatabase/cloud/issues/3571)

### DIFF
--- a/test_runner/regress/test_tenant_detach.py
+++ b/test_runner/regress/test_tenant_detach.py
@@ -6,6 +6,7 @@ from threading import Thread
 import asyncpg
 import pytest
 from fixtures.log_helper import log
+from fixtures.metrics import parse_metrics
 from fixtures.neon_fixtures import (
     NeonEnv,
     NeonEnvBuilder,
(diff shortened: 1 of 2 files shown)


koivunej (approved):
This looks good.

SomeoneToIgnore (commented):
LGTM, can we have it tested? 

Afaik, there's a timeline metrics test that should be able to verify such case.

shanyp:
> LGTM, can we have it tested?
> 
> Afaik, there's a timeline metrics test that should be able to verify such case.

@SomeoneToIgnore done

SomeoneToIgnore (approved):

LizardWizzard (approved):

test_runner/regress/test_tenant_detach.py
```diff
@@ -78,9 +79,29 @@ def test_tenant_reattach(
...
 
+    ps_metrics = parse_metrics(pageserver_http.get_metrics(), "pageserver")
+    tenant_metric_filter = {
+        "tenant_id": str(tenant_id),
+        "timeline_id": str(timeline_id),
+    }
+    pageserver_last_record_lsn_before_detach = int(
+        ps_metrics.query_one("pageserver_last_record_lsn", filter=tenant_metric_filter).value
+    )
+
```
koivunej, line 90: I was just let know of the `wait_for_upload` yesterday, why is not needed here? Is it luck?

Because the added assert on line 103 is not always true, or perhaps it could be read from remote_consistent_lsn on line 88 as well.

koivunej: On an unrelated note: the postgres is kept running while doing tenant detach and attach. I am surprised that this works.

On line 63 there could also be:

```patch
- pg = env.postgres.create_start("main", tenant_id)
+ with env.postgres.create_start("main", tenant_id) as pg:
```

Which would have to be repeated, probably with `start` if we don't want to test that postgres will reconnect in this test.

shanyp: wait_for_upload is already in line 72

PR #3470, opened 2023-01-27
Update deploy steps

zoete:
First one isn't optimal, but as it was requested to run the runner as nonroot -> https://github.com/neondatabase/runner/pull/1#discussion_r1069909593 this job will need more significant refactoring. This should unblock the deployment process.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -1178,7 +1178,7 @@ jobs:
           fetch-depth: 0
 
       - name: Configure AWS Credentials
-        uses: aws-actions/configure-aws-credentials@v1
+        uses: aws-actions/configure-aws-credentials@v1-node16
         with:
           role-to-assume: arn:aws:iam::369495373322:role/github-runner
           aws-region: eu-central-1
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -885,7 +885,9 @@ jobs:
 
   deploy-new:
     runs-on: [ self-hosted, gen3, small ]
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -885,7 +885,9 @@ jobs:
 
   deploy-new:
     runs-on: [ self-hosted, gen3, small ]
-    container: 369495373322.dkr.ecr.eu-central-1.amazonaws.com/ansible:pinned
+    container:
+      image: 369495373322.dkr.ecr.eu-central-1.amazonaws.com/ansible:pinned
+      options: --user root --privileged
```
lassizci, line 890: What exactly requires `root`?

zoete: Docker -> /etc/init.d/docker: 96: ulimit: error setting limit (Operation not permitted)

zoete: I switched the user on that image to be "nonroot" else pipeline failures are painful to deal with.

lassizci: > Docker -> /etc/init.d/docker: 96: ulimit: error setting limit (Operation not permitted)

I'm a bit lost. Where do we run /etc/init.d/docker in this job?

zoete: https://github.com/neondatabase/neon/actions/runs/4024470203/jobs/6917288451

in the https://github.com/neondatabase/neon/blob/main/.github/ansible/get_binaries.sh

zoete:
> Well, I guess we have to do that since the pipeline is stuck, but should get rid of running these as root.

100%

Issue #3439, opened 2023-01-25
Add a facility to evict layer files manually


SomeoneToIgnore:
Part of https://github.com/neondatabase/neon/issues/2476

The implementation will use HTTP management API to do that.

We need to:

* be able to return layers via the HTTP API

     - There's currently no layer concept in the HTTP API at all, need to add one and have some unique layer ID there (local FS layer file name is good enough)
API.

     - Layer map dumping mechanism has to be called via that HTTP API and the result converted to the new API object

* be able to evict layers via the HTTP API

Using the IDs produced from the new API handle described above, we can remove the local FS file and replace it with the "Remote" variant of this layer to not to loose it and be able to redownload it again, if needed

Given that the previous API step is implemented, the HTTP API <-> layer map access interface is already there and it's relatively trivial to add the reverse operation code-wise, but there's one potential caveat: we need to hold a lock during layer replacement, and do layer map rebuilds during that. 
It might be slow, according to https://github.com/neondatabase/neon/issues/3394

Most probably, it's not an issue, but should be considered during the impl.

* synchronize with the upload_queue and forbid evicting the layers that are not uploaded

Something similar may happen with GC/compaction that deletes the layer that's not yet uploaded: we can remove that along the way too.
https://github.com/neondatabase/neon/blob/aabca55d7e7e020a30f6afd286c850af88063f28/pageserver/src/tenant/timeline.rs#L2592-L2598

SomeoneToIgnore:
Additional notes:

* we should care about atomic removals: we can be dropped any time, and should 
    * assert that there's no local file in memory when actually removing the file
    * do file removal after the layers in memory are adjusted to have the remote layer

shanyp:
>we can consider soft-deletion and rename the layers files into some layer_name.deleted instead of directly removing them

If we do , there should be a scavenger of some type to delete those upon page server restart

PR #3473, opened 2023-01-27
Avoid `dyn PersistentLayer` interface usage in LayerMap.

SomeoneToIgnore:
Follow-up of https://github.com/neondatabase/neon/pull/3263

This PR proposes more type-based checking of Delta, Image and Remote layers via enums and generics.

* Traits are now are used without `dyn` in almost every place except `impl Deref` for a few enums with different kinds of layers.

* `InMemoryLayer` handling is unchanged.

* `LayerMap`, instead of collections of arbitrary interfaces as 

```rust
historic: BufferedHistoricLayerCoverage<Arc<L>>,
...
l0_delta_layers: Vec<Arc<L>>,
```

became

```rust
historic: BufferedHistoricLayerCoverage<HistoricLayer<D, I>>,
...
l0_delta_layers: Vec<LocalOrRemote<D>>,
```

all related code handling is adjusted

* Layer traits got shrinked and renamed

## Motivation for change.

Old `LayerMap` relied on traits to provide the functionality it has:

```rust 
impl<L> LayerMap<L>
where
    L: ?Sized + Layer,
```

and did not distinguish layers between each other.
Similarly, old `Timeline` did the same, by also placing stricter type boundaries on the layers:

```rust
pub struct Timeline {
    ...
    pub layers: RwLock<LayerMap<dyn PersistentLayer>>,
    ...
```
Due to `pub trait PersistentLayer: Layer`, that means that both traits have to be implemented for every kind of layer behind it.

Inside the `LayerMap`, there are 4.5 kinds of layers:

1. open layer (InMemory kind)
2. frozen layers (InMemory kind)
3. image layers (PersistentLayer kind)
4. delta layers (PersistentLayer kind)
4.5 remote delta or image layers, not present on the local disk

* Generics in layer map cover only the latest 3 categories of layers, but not all the methods in the traits are applicable for every one.
One good example is `impl PersistentLayer for RemoteLayer` and the whole `PersistentLayer` per se, that becomes too much of a god trait with methods like 

```rust
/// Iterate through all keys stored in the layer. Returns key, lsn and value size
/// It is used only for compaction and so is currently implemented only for DeltaLayer
fn key_iter(&self) -> Result<LayerKeyIter<'_>> {
    panic!("Not implemented")
}

fn downcast_remote_layer(self: Arc<Self>) -> Option<std::sync::Arc<RemoteLayer>> {
    None
}

fn is_remote_layer(&self) -> bool {
    false
}
```

Similar mock implementations are used for a specific benchmarking type that's used in the `LayerMap` separately.

* All `Arc<dyn PersistentLayer>` layers inside the `LayerMap` are not distinguishable from each other on the type level, hence delta, image and remote layers could be placed into the wrong places or slip unnoticed, since they all duck type.

* Last but not least, `dyn` support in Rust is generally lacking compared to the generics.
As an example, current nightly async traits feature does not support `dyn` objects.

## Comments on the new version

Due to stricter typisation, big part of the original traits is either not needed at all, or could be moved into certain struct/enum impl instead. Still, they are providing a public, generic API, there's multiple impls on `LayerMap` and `HistoricLayer` that base on that, not on specific generic values for delta and image types.

More specific impls on `DeltaLayer` and `ImageLayer` structs can use whatever exposed fields/methods without extra traits, thus providing "less generic" features: for instance, the methods that compaction uses are now only implemented in `impl LocalOrRemote<DeltaLayer>` section.
No `todo!` or `unimplemented!` trait methods are left.

`InMemoryLayer` could be generified similar way now, by adding another generic to the `LayerMap`.

Further improvements are possible, for instance, `TraversalId` could be changed from `String` to a `Copy` type and no 

```rust
Box::new({
    let layer = layer.clone();
    move || layer.traversal_id()
}),
```

would be needed.

### DIFF
--- a/pageserver/src/tenant/storage_layer/inmemory_layer.rs
+++ b/pageserver/src/tenant/storage_layer/inmemory_layer.rs
@@ -27,7 +27,7 @@ use std::fmt::Write as _;
 use std::ops::Range;
 use std::sync::RwLock;
 
-use super::{DeltaLayer, DeltaLayerWriter, Layer};
+use super::{DeltaLayer, DeltaLayerWriter, LayerContent, LayerRange};
 
 thread_local! {
     /// A buffer for serializing object during [`InMemoryLayer::put_value`].
@@ -38,7 +38,7 @@ thread_local! {
 pub struct InMemoryLayer {
     conf: &'static PageServerConf,
     tenant_id: TenantId,
-    timeline_id: TimelineId,
+    pub timeline_id: TimelineId,
 
     ///
     /// This layer contains all the changes from 'start_lsn'. The
@@ -75,13 +75,7 @@ impl InMemoryLayerInner {
     }
 }
 
-impl InMemoryLayer {
-    pub fn get_timeline_id(&self) -> TimelineId {
-        self.timeline_id
-    }
-}
-
-impl Layer for InMemoryLayer {
+impl LayerRange for InMemoryLayer {
     fn get_key_range(&self) -> Range<Key> {
         Key::MIN..Key::MAX
     }
@@ -108,7 +102,9 @@ impl Layer for InMemoryLayer {
         let end_lsn = inner.end_lsn.unwrap_or(Lsn(u64::MAX));
         format!("inmem-{:016X}-{:016X}", self.start_lsn.0, end_lsn.0)
     }
+}
 
+impl LayerContent for InMemoryLayer {
     /// debugging function to print out the contents of the layer
     fn dump(&self, verbose: bool) -> Result<()> {
         let inner = self.inner.read().unwrap();
--- a/pageserver/src/tenant/storage_layer/inmemory_layer.rs
+++ b/pageserver/src/tenant/storage_layer/inmemory_layer.rs
@@ -27,7 +27,7 @@ use std::fmt::Write as _;
 use std::ops::Range;
 use std::sync::RwLock;
 
-use super::{DeltaLayer, DeltaLayerWriter, LayerContent, LayerRange};
+use super::{DeltaLayer, DeltaLayerWriter, LayerContent};
 
 thread_local! {
     /// A buffer for serializing object during [`InMemoryLayer::put_value`].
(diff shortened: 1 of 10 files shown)


SomeoneToIgnore:
For further demonstration, consider first part of the https://github.com/neondatabase/neon/issues/3439 implemented via both approaches.

We add a new HTTP handle to list metadata on all timeline layers.

Old `LayerMap`: https://github.com/neondatabase/neon/compare/kb/evict-on-demand?expand=1
New `LayerMap`: https://github.com/neondatabase/neon/compare/kb/more-dispatch...kb/evict-on-demand-less-dispatched?expand=1

Both PRs contain identical `models.rs` and `routes.rs` changes, but there are a few differences related to how layers can be accessed via either of the APIs.

* old `LayerMap` cannot get the `layer_map_info` method inside, it only can be inside the `Timeline` unless we add more impls in it for `PersistentLayer`
* old approach required us to add another method in the `PersistentLayer`, that would return the info needed by the endpoint.

`InMemoryLayer` implements the same trait `Layer` trait, so adding the method there would be worse, because different layers have different info types (e.g. `end_lsn` is not present in delta layers, but is present in `info` layers).

`PersistentLayer` might not be most suitable place to add this method, but luckily in the old verison, we do `impl PersistentLayer for RemoteLayer` and that allows us to share the code.

* new approach has a fluency to put `layer_map_info` in either `timeline` or `layer_map`, adds methods to two different kind of layers inside, InMemory and Historic Layers.

pageserver/src/tenant/storage_layer/inmemory_layer.rs
```diff
@@ -339,4 +306,22 @@ impl InMemoryLayer {
...
     }
+
+    pub fn get_current_lsn_range(&self) -> Range<Lsn> {
+        let inner = self.inner.read().unwrap();
+
+        let end_lsn = if let Some(end_lsn) = inner.end_lsn {
+            end_lsn
+        } else {
+            Lsn(u64::MAX)
+        };
```
koivunej, line 317:
```suggestion
        let end_lsn = inner.end_lsn.unwrap_or(Lsn(u64::MAX));
```

To align with the new line 324.

pageserver/src/tenant/storage_layer/inmemory_layer.rs
```diff
@@ -339,4 +306,22 @@ impl InMemoryLayer {
...
+        self.start_lsn..end_lsn
+    }
+
+    pub fn short_id(&self) -> String {
+        let inner = self.inner.read().unwrap();
+
+        let end_lsn = inner.end_lsn.unwrap_or(Lsn(u64::MAX));
+        format!("inmem-{:016X}-{:016X}", self.start_lsn.0, end_lsn.0)
+    }
```
koivunej, line 326: This range concept is high risk for subtle off-by-ones, would prefer to use the single fn above to get it.

```suggestion
    pub fn short_id(&self) -> String {
        let range = self.get_current_lsn_range();

        format!("inmem-{:016X}-{:016X}", range.start.0, range.end.0)
    }
```

SomeoneToIgnore:
This again needs a large rebase, I yield.

PR #3475, opened 2023-01-27
Tracing panic hook

### LINKED ISSUE #3468: Log panics via `tracing`
On pageserver at least, assuming elsewhere as well, we rely on [`std::panicking::default_hook`](https://doc.rust-lang.org/stable/src/std/panicking.rs.html#235) panic hook for printing the panic to stderr. This is not great, because the stderr writes can get interleaved by the stdout writes, and we end up with botched log messages in grafana where there is a bit of the stacktrace in different messages.

What we should do instead is to install a panic hook *after* configuring logging *before* configuring sentry (which also installs a panic hook, which we don't want to override).

This should solve our interleaved messages in grafana issue, as well as "missing span at panic" issue.

This should be considered in other binaries as well, but perhaps easier to test with one.

Additionally it would be nice to have a panic endpoint on http which we could use to ensure both tracing and sentry work on staging and prod.

Relevant:
- [`tracing` has a panic hook example](https://github.com/tokio-rs/tracing/blob/ee8fd2876d5c3459ca4cce38ace29ea222a2a5f5/examples/examples/panic_hook.rs)
- [the std impl, but in release 1.66.1](https://github.com/rust-lang/rust/blob/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/std/src/panicking.rs#L235-L288)

koivunej:
Fixes #3468.

This does change how the panics look, and most importantly, make sure they are not interleaved with other messages. Adds a `GET /v1/panic` endpoint for panic testing (useful for sentry dedup and this hook testing).

The panics are now logged within a new error level span called `panic` which separates it from other error level events. The panic info is unpacked similar to in tracing example with fields:
- thread=mgmt request worker
- location="pageserver/src/http/routes.rs:898:9"
  - instead of:
      - panic.file="pageserver/src/http/routes.rs"
      - panic.line=898
      - panic.column=9

Much interest was not sparked, I am thinking of deciding the below as:
- no frame filtering until std supports it (same for head, tail)
- no getting rid of trailing `\n` (splitting should not care -- lets try this on staging)
- location formatting changed (both are above)
- try first on pageserver on staging because we can test this (added http route)

Decision points:
- [x] should we filter the first frames up to `rust_begin_unwind`?
- [x] should we filter the last frames up with same heuristics as anyhow? (I haven't checked what those are)
- [x] should we get rid of the trailing `\n` on the message? (re: message splitting for promtail)
- [x] should we change the location formatting?
- [x] move directly to utils under logging?
- [x] apply to all binaries directly or first test with `pageserver`?

### DIFF
--- a/pageserver/src/bin/pageserver.rs
+++ b/pageserver/src/bin/pageserver.rs
@@ -517,14 +517,26 @@ fn tracing_panic_hook(info: &std::panic::PanicInfo) {
     let thread = thread.name().unwrap_or("<unnamed>");
     let backtrace = std::backtrace::Backtrace::capture();
 
+    struct PrettyLocation<'a, 'b>(&'a std::panic::Location<'b>);
+
+    impl std::fmt::Display for PrettyLocation<'_, '_> {
+        fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+            write!(f, "{}:{}:{}", self.0.file(), self.0.line(), self.0.column())
+        }
+    }
+
     let _entered = if let Some(location) = location {
-            tracing::error_span!("panic", %thread, panic.file = location.file(), panic.line = location.line(), panic.column = location.column())
-        } else {
-            // very unlikely to hit here, but the guarantees of std could change
-            tracing::error_span!("panic", %thread)
-        }.entered();
+        tracing::error_span!("panic", %thread, location = %PrettyLocation(location))
+    } else {
+        // very unlikely to hit here, but the guarantees of std could change
+        tracing::error_span!("panic", %thread)
+    }
+    .entered();
 
     if backtrace.status() == std::backtrace::BacktraceStatus::Captured {
+        // this has an annoying extra '\n' in the end which anyhow doesn't do, but we cannot really
+        // get rid of it as we cannot get in between of std::fmt::Formatter<'_>; we could format to
+        // string, maybe even to a TLS one but tracing already does that.
         tracing::error!("{msg}\n\nStack backtrace:\n{backtrace}");
     } else {
         tracing::error!("{msg}");
--- a/pageserver/src/bin/pageserver.rs
+++ b/pageserver/src/bin/pageserver.rs
@@ -88,6 +88,13 @@ fn main() -> anyhow::Result<()> {
         }
     };
 
+    // Initialize logging, which must be initialized before the custom panic hook is installed.
+    logging::init(conf.log_format)?;
+
+    // disable the default rust panic hook by using `set_hook`. sentry will install it's own on top
+    // of this, always processing the panic before we log it.
+    std::panic::set_hook(Box::new(tracing_panic_hook));
+
     // initialize sentry if SENTRY_DSN is provided
     let _sentry_guard = init_sentry(
         Some(GIT_VERSION.into()),
@@ -210,9 +217,6 @@ fn start_pageserver(
     launch_ts: &'static LaunchTimestamp,
     conf: &'static PageServerConf,
 ) -> anyhow::Result<()> {
-    // Initialize logging
-    logging::init(conf.log_format)?;
-
     // Print version and launch timestamp to the log,
     // and expose them as prometheus metrics.
     // A changed version string indicates changed software.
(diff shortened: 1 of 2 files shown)


koivunej:
Old panic looked like (via `GET /v1/panic`) with prev and next line:

```
2023-01-27T16:28:14.440275Z  INFO GET /v1/always_panic 404 Not Found
thread 'mgmt request worker' panicked at 'unconditional panic for testing panic hook integration', pageserver/src/http/routes.rs:898:9
stack backtrace:
   0: rust_begin_unwind
--snip--
  64: tokio::runtime::blocking::pool::Spawner::spawn_thread::{{closure}}
             at /home/joonas/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.24.2/src/runtime/blocking/pool.rs:469:13
2023-01-27T16:28:26.059654Z  INFO GET /v1/panic 204 No Content
```

New panic looks like (with anyhow backtrace above):

```
             at ./misc/../sysdeps/unix/sysv/linux/x86_64/clone3.S:81
2023-01-27T16:56:48.866015Z ERROR panic{thread=mgmt request worker panic.file="pageserver/src/http/routes.rs" panic.line=898 panic.column=9}: unconditional panic for testing panic hook integration

Stack backtrace:
   0: pageserver::tracing_panic_hook
             at /home/joonas/src/neon/neon-main/pageserver/src/bin/pageserver.rs:503:21
   1: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/alloc/src/boxed.rs:2001:9
   2: std::panicking::rust_panic_with_hook
             at /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/std/src/panicking.rs:692:13
   3: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/std/src/panicking.rs:577:13
   4: std::sys_common::backtrace::__rust_end_short_backtrace
             at /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/std/src/sys_common/backtrace.rs:137:18
   5: rust_begin_unwind
--snip--
  82: clone3
             at ./misc/../sysdeps/unix/sysv/linux/x86_64/clone3.S:81

2023-01-27T16:56:48.970587Z  INFO GET /v1/panic 204 No Content
```

Above also shows the new frames.

So there's an extra newline *after*. I wonder if that's a problem for message splitting, we could try to make it go away by depending directly on `backtrace-rs` instead of `std` (like `anyhow`) but I would prefer not doing that.

Could @lassizci comment on if this will create an issue for the message splitting pipeline, or point me to the implementation?

lassizci:
> Could @lassizci comment on if this will create an issue for the message splitting pipeline, or point me to the implementation?

@koivunej Shouldn't affect, the matcher seems to be matching a timestamp and then collecting the lines until the next timestamp.

koivunej:
Postponing this.

pageserver/src/bin/pageserver.rs
```diff
@@ -495,6 +499,50 @@ fn cli() -> Command {
...
+        Some(s) => *s,
+        None => match info.payload().downcast_ref::<String>() {
+            Some(s) => &s[..],
+            None => "Box<dyn Any>",
+        },
+    };
+
+    let thread = std::thread::current();
+    let thread = thread.name().unwrap_or("<unnamed>");
+    let backtrace = std::backtrace::Backtrace::capture();
```
problame, line 518:
```suggestion
    # Use force_capture: we want the stack trace in the logs, regardless of
    # whether RUST_BACKTRACE is set or not.
    let backtrace = std::backtrace::Backtrace::force_capture();
```

pageserver/src/http/routes.rs
```diff
@@ -1029,6 +1029,16 @@ async fn active_timeline_of_active_tenant(
         .map_err(ApiError::NotFound)
 }
 
+async fn always_panic_handler(_: Request<Body>) -> Result<Response<Body>, ApiError> {
+    // catch_unwind ensures that tokio nor hyper are distracted by our panic, but panic hooks
+    // (at least sentry) will report the issue. if tested across fleet, this should be a good test
+    // case for panic deduplication as well per version.
```
problame, line 1035:
```suggestion
    // Deliberately cause a panic to exercise the panic hook registered via std::panic::set_hook().
    // For pageserver, the relevant panic hook is `tracing_panic_hook` , and the `sentry` crate's wrapper around it.
    // Use catch_unwind to ensure that tokio nor hyper are distracted by our panic.
```

pageserver/src/http/routes.rs
```diff
@@ -1029,6 +1029,16 @@ async fn active_timeline_of_active_tenant(
...
+    let _ = std::panic::catch_unwind(|| {
+        panic!("unconditional panic for testing panic hook integration")
```
problame, line 1037: How about repeating a caller-provided string in the panic message?
Caller can still use a constant, but that way, they can pass in a nonce and check the logs / sentry for that exact nonce.

problame (approved):
Generally LGTM, minor nits and one improvement suggestion.

pageserver/src/bin/pageserver.rs
koivunej: We have that setting already set in deployments, what if you are manual testing and want to hit something often? That'd be altering the existing behaviour, so don't like it.

pageserver/src/http/routes.rs
koivunej: I'll add an optional message query parameter.

pageserver/src/http/routes.rs
koivunej: Is this just stylistic rewrite or is there some changes I cannot see? unsure if std's panic hook needs to be mentioned, it's the only panic hook in std..

pageserver/src/http/routes.rs
koivunej: I'll just dump the whole url decoded query actually.

pageserver/src/http/routes.rs
koivunej: Hmm ok, now I can see.

pageserver/src/http/routes.rs
koivunej: [9d8e796](https://github.com/neondatabase/neon/pull/3475/commits/9d8e796e866ed4e6f5b9f2f4ab98f00b6ec0c47d)

pageserver/src/bin/pageserver.rs
koivunej: Resolving.

PR #3486, opened 2023-01-30
Evict layers on demand

SomeoneToIgnore:
Closes https://github.com/neondatabase/neon/issues/3439

Adds a set of commands to manipulate the layer map:
* dump the layer map contents
* evict the layer form the layer map (remove the local file, put the remote layer instead in the layer map)
* download the layer (operation, reversing the eviction)

The commands will change later, when the statistics is added on top, so the swagger schema is not adjusted.

The commands might have issues with big amount of layers: no pagination is done for the dump command, eviction and download commands look for the layer to evict/download by iterating all layers sequentially and comparing the layer names.
For now, that seems to be tolerable ("big" number of layers is ~2_000) and further experiments are needed.

### DIFF
--- a/libs/pageserver_api/src/models.rs
+++ b/libs/pageserver_api/src/models.rs
@@ -255,6 +255,8 @@ pub enum InMemoryLayerInfo {
 pub enum HistoricLayerInfo {
     Delta {
         layer_file_name: String,
+        layer_file_size: Option<u64>,
+
         #[serde_as(as = "DisplayFromStr")]
         lsn_start: Lsn,
         #[serde_as(as = "DisplayFromStr")]
@@ -263,6 +265,8 @@ pub enum HistoricLayerInfo {
     },
     Image {
         layer_file_name: String,
+        layer_file_size: Option<u64>,
+
         #[serde_as(as = "DisplayFromStr")]
         lsn_start: Lsn,
         remote: bool,
--- a/libs/pageserver_api/src/models.rs
+++ b/libs/pageserver_api/src/models.rs
@@ -227,6 +227,48 @@ pub struct TimelineInfo {
     pub state: TimelineState,
 }
 
+#[derive(Debug, Clone, Serialize)]
+pub struct LayerMapInfo {
+    pub in_memory_layers: Vec<InMemoryLayerInfo>,
+    pub historic_layers: Vec<HistoricLayerInfo>,
+}
+
+#[serde_as]
+#[derive(Debug, Clone, Serialize)]
+#[serde(tag = "kind")]
+pub enum InMemoryLayerInfo {
+    Open {
+        #[serde_as(as = "DisplayFromStr")]
+        lsn_start: Lsn,
+    },
+    Frozen {
+        #[serde_as(as = "DisplayFromStr")]
+        lsn_start: Lsn,
+        #[serde_as(as = "DisplayFromStr")]
+        lsn_end: Lsn,
+    },
+}
+
+#[serde_as]
+#[derive(Debug, Clone, Serialize)]
+#[serde(tag = "kind")]
+pub enum HistoricLayerInfo {
+    Delta {
+        layer_file_name: String,
+        #[serde_as(as = "DisplayFromStr")]
+        lsn_start: Lsn,
+        #[serde_as(as = "DisplayFromStr")]
+        lsn_end: Lsn,
+        remote: bool,
+    },
+    Image {
+        layer_file_name: String,
+        #[serde_as(as = "DisplayFromStr")]
+        lsn_start: Lsn,
+        remote: bool,
+    },
+}
+
 #[derive(Debug, Serialize, Deserialize)]
 pub struct DownloadRemoteLayersTaskSpawnRequest {
     pub max_concurrent_downloads: NonZeroUsize,
--- a/pageserver/src/http/routes.rs
+++ b/pageserver/src/http/routes.rs
@@ -7,7 +7,7 @@ use pageserver_api::models::DownloadRemoteLayersTaskSpawnRequest;
 use remote_storage::GenericRemoteStorage;
 use tokio_util::sync::CancellationToken;
 use tracing::*;
-use utils::http::request::{must_get_query_param, parse_query_param};
+use utils::http::request::{get_request_param, must_get_query_param, parse_query_param};
 
 use super::models::{
     StatusResponse, TenantConfigRequest, TenantCreateRequest, TenantCreateResponse, TenantInfo,
@@ -317,10 +317,7 @@ async fn get_lsn_by_timestamp_handler(request: Request<Body>) -> Result<Response
     let timestamp_pg = postgres_ffi::to_pg_timestamp(timestamp);
 
     let ctx = RequestContext::new(TaskKind::MgmtRequest, DownloadBehavior::Download);
-    let timeline = mgr::get_tenant(tenant_id, true)
-        .await
-        .and_then(|tenant| tenant.get_timeline(timeline_id, true))
-        .map_err(ApiError::NotFound)?;
+    let timeline = active_timeline_of_active_tenant(tenant_id, timeline_id).await?;
     let result = timeline
         .find_lsn_for_timestamp(timestamp_pg, &ctx)
         .await
@@ -526,6 +523,62 @@ async fn tenant_size_handler(request: Request<Body>) -> Result<Response<Body>, A
             inputs,
         },
     )
+}
+
+async fn layer_map_info_handler(request: Request<Body>) -> Result<Response<Body>, ApiError> {
+    let tenant_id: TenantId = parse_request_param(&request, "tenant_id")?;
+    let timeline_id: TimelineId = parse_request_param(&request, "timeline_id")?;
+    check_permission(&request, Some(tenant_id))?;
+
+    let timeline = active_timeline_of_active_tenant(tenant_id, timeline_id).await?;
+    let layer_map_info = timeline.layer_map_info();
+
+    json_response(StatusCode::OK, layer_map_info)
+}
+
+async fn layer_download_handler(request: Request<Body>) -> Result<Response<Body>, ApiError> {
+    let tenant_id: TenantId = parse_request_param(&request, "tenant_id")?;
+    check_permission(&request, Some(tenant_id))?;
+    let timeline_id: TimelineId = parse_request_param(&request, "timeline_id")?;
+    let layer_file_name = get_request_param(&request, "layer_file_name")?;
+    check_permission(&request, Some(tenant_id))?;
+
+    let timeline = active_timeline_of_active_tenant(tenant_id, timeline_id).await?;
+    let downloaded = timeline
+        .download_layer(layer_file_name)
+        .await
+        .map_err(ApiError::InternalServerError)?;
+
+    match downloaded {
+        Some(true) => json_response(StatusCode::OK, ()),
+        Some(false) => json_response(StatusCode::NOT_MODIFIED, ()),
+        None => json_response(
+            StatusCode::BAD_REQUEST,
+            format!("Layer {tenant_id}/{timeline_id}/{layer_file_name} not found"),
+        ),
+    }
+}
+
+async fn evict_timeline_layer_handler(request: Request<Body>) -> Result<Response<Body>, ApiError> {
+    let tenant_id: TenantId = parse_request_param(&request, "tenant_id")?;
+    check_permission(&request, Some(tenant_id))?;
+    let timeline_id: TimelineId = parse_request_param(&request, "timeline_id")?;
+    let layer_file_name = get_request_param(&request, "layer_file_name")?;
+
+    let timeline = active_timeline_of_active_tenant(tenant_id, timeline_id).await?;
+    let evicted = timeline
+        .evict_layer(layer_file_name)
+        .await
+        .map_err(ApiError::InternalServerError)?;
+
+    match evicted {
+        Some(true) => json_response(StatusCode::OK, ()),
+        Some(false) => json_response(StatusCode::NOT_MODIFIED, ()),
+        None => json_response(
+            StatusCode::BAD_REQUEST,
+            format!("Layer {tenant_id}/{timeline_id}/{layer_file_name} not found"),
+        ),
+    }
 }
 
 // Helper function to standardize the error messages we produce on bad durations
@@ -804,12 +857,7 @@ async fn timeline_checkpoint_handler(request: Request<Body>) -> Result<Response<
     check_permission(&request, Some(tenant_id))?;
 
     let ctx = RequestContext::new(TaskKind::MgmtRequest, DownloadBehavior::Download);
-    let tenant = mgr::get_tenant(tenant_id, true)
-        .await
-        .map_err(ApiError::NotFound)?;
-    let timeline = tenant
-        .get_timeline(timeline_id, true)
-        .map_err(ApiError::NotFound)?;
+    let timeline = active_timeline_of_active_tenant(tenant_id, timeline_id).await?;
     timeline
         .freeze_and_flush()
         .await
@@ -830,12 +878,7 @@ async fn timeline_download_remote_layers_handler_post(
     let body: DownloadRemoteLayersTaskSpawnRequest = json_request(&mut request).await?;
     check_permission(&request, Some(tenant_id))?;
 
-    let tenant = mgr::get_tenant(tenant_id, true)
-        .await
-        .map_err(ApiError::NotFound)?;
-    let timeline = tenant
-        .get_timeline(timeline_id, true)
-        .map_err(ApiError::NotFound)?;
+    let timeline = active_timeline_of_active_tenant(tenant_id, timeline_id).await?;
     match timeline.spawn_download_all_remote_layers(body).await {
         Ok(st) => json_response(StatusCode::ACCEPTED, st),
         Err(st) => json_response(StatusCode::CONFLICT, st),
@@ -846,20 +889,27 @@ async fn timeline_download_remote_layers_handler_get(
     request: Request<Body>,
 ) -> Result<Response<Body>, ApiError> {
     let tenant_id: TenantId = parse_request_param(&request, "tenant_id")?;
-    let timeline_id: TimelineId = parse_request_param(&request, "timeline_id")?;
     check_permission(&request, Some(tenant_id))?;
+    let timeline_id: TimelineId = parse_request_param(&request, "timeline_id")?;
 
-    let tenant = mgr::get_tenant(tenant_id, true)
-        .await
-        .map_err(ApiError::NotFound)?;
-    let timeline = tenant
-        .get_timeline(timeline_id, true)
-        .map_err(ApiError::NotFound)?;
+    let timeline = active_timeline_of_active_tenant(tenant_id, timeline_id).await?;
     let info = timeline
         .get_download_all_remote_layers_task_info()
         .context("task never started since last pageserver process start")
         .map_err(ApiError::NotFound)?;
     json_response(StatusCode::OK, info)
+}
+
+async fn active_timeline_of_active_tenant(
+    tenant_id: TenantId,
+    timeline_id: TimelineId,
+) -> Result<Arc<Timeline>, ApiError> {
+    let tenant = mgr::get_tenant(tenant_id, true)
+        .await
+        .map_err(ApiError::NotFound)?;
+    tenant
+        .get_timeline(timeline_id, true)
+        .map_err(ApiError::NotFound)
 }
 
 async fn handler_404(_: Request<Body>) -> Result<Response<Body>, ApiError> {
@@ -957,6 +1007,18 @@ pub fn make_router(
         .delete(
             "/v1/tenant/:tenant_id/timeline/:timeline_id",
             timeline_delete_handler,
+        )
+        .get(
+            "/v1/tenant/:tenant_id/timeline/:timeline_id/layer",
+            layer_map_info_handler,
+        )
+        .get(
+            "/v1/tenant/:tenant_id/timeline/:timeline_id/layer/:layer_file_name",
+            layer_download_handler,
+        )
+        .delete(
+            "/v1/tenant/:tenant_id/timeline/:timeline_id/layer/:layer_file_name",
+            evict_timeline_layer_handler,
         )
         .any(handler_404))
 }
--- a/pageserver/src/tenant/timeline.rs
+++ b/pageserver/src/tenant/timeline.rs
@@ -30,7 +30,8 @@ use crate::broker_client::is_broker_client_initialized;
 use crate::context::{DownloadBehavior, RequestContext};
 use crate::tenant::remote_timeline_client::{self, index::LayerFileMetadata};
 use crate::tenant::storage_layer::{
-    DeltaFileName, DeltaLayerWriter, ImageFileName, ImageLayerWriter, InMemoryLayer, RemoteLayer,
+    DeltaFileName, DeltaLayerWriter, ImageFileName, ImageLayerWriter, InMemoryLayer, LayerFileName,
+    RemoteLayer,
 };
 use crate::tenant::{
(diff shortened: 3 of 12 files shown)


pageserver/src/tenant/timeline.rs
```diff
@@ -818,6 +818,94 @@ impl Timeline {
...
+                &layer_metadata,
+            ),
+            #[cfg(test)]
+            LayerFileName::Test(_) => unreachable!(),
+        });
+
+        let mut layers = self.layers.write().unwrap();
+        let mut updates = layers.batch_update();
+        local_layer.delete()?;
+        updates.remove_historic(local_layer);
```
koivunej, line 902: Checked the impl again.

For level 0 deltas we still have an `Arc::ptr_eq` check which I suspect might be troublesome (but could not prove) -- that'd be easily fixed via `Arc::as_ptr(&left) as *const () == Arc::as_ptr(&right) as *const ()` which the std lib might eventually use as well (no link right now).

This would be one risky place of poisoning the lock, requiring `/ignore` + `/load`, but I think we cannot test this out.

Also, looking at the `insert_historic` this makes me wonder if we could fair with just `insert_historic` + `flush`, but perhaps that'd be bit of a misuse.

problame (changes requested):
The dump endpoint look fine to me :+1:  

Download endpoint as well (we already have the "download all remote layers" endpoint, but I guess this one is good to have as well).

Eviction endpoint needs more work.

Can we factor out the changes for the dump endpoint so that we can merge it before tomorrow's release?

libs/pageserver_api/src/models.rs
```diff
@@ -227,6 +230,39 @@ pub struct TimelineInfo {
...
+    Open { lsn_start: Lsn },
+    Frozen { lsn_start: Lsn, lsn_end: Lsn },
+}
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize)]
+#[serde(tag = "kind")]
+pub enum HistoricLayerInfo {
+    Delta {
+        layer_file_name: String,
+        key_start: String,
```
problame, line 251: Why is this a `String`? We have a `impl fmt::Display for Key`

problame (changes requested):

libs/pageserver_api/src/models.rs
```diff
@@ -227,6 +230,39 @@ pub struct TimelineInfo {
...
+}
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Copy, Clone, Serialize, Deserialize)]
+#[serde(tag = "kind")]
+pub enum InMemoryLayerInfo {
+    Open { lsn_start: Lsn },
+    Frozen { lsn_start: Lsn, lsn_end: Lsn },
+}
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize)]
```
problame, line 246: The `Ord` is annoying here because we want to include the layer access stats in this struct.
I suppose we need it for sorted output. Can we just not sort the output and let the client deal with this?

problame: Similarly Eq and PartialEq

problame: Will prepare a patch

pageserver/src/tenant/timeline.rs
```diff
@@ -818,6 +818,94 @@ impl Timeline {
...
+                self.tenant_id,
+                self.timeline_id,
+                &delta_name,
+                &layer_metadata,
+            ),
+            #[cfg(test)]
+            LayerFileName::Test(_) => unreachable!(),
+        });
+
+        let mut layers = self.layers.write().unwrap();
```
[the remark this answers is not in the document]
SomeoneToIgnore: Thank you, added.

libs/pageserver_api/src/models.rs
SomeoneToIgnore: I did not want to drag the whole `Key` with impls into `models` and change imports everywhere.

But sure, moved and fixed all places.

SomeoneToIgnore: If I add `#[serde_as(as = "DisplayFromStr")]` on top of the `key_start: Key,` and similar fields, I still don't get the `Display`-based serialization, interesting.

SomeoneToIgnore: TIL, `#[serde_as]` has to be the topmost annotation placed (or, at least above `#[derive(Serialize,Deserialize)]`, otherwise silently won't (de)serialize in the proper format.

Fixed that, need to add tests.

libs/pageserver_api/src/models.rs
problame: https://github.com/neondatabase/neon/pull/3486#discussion_r1091712440

SomeoneToIgnore:
My branch builds correctly locally, but here, CI fails with odd errors.
I will rebase and push.

pageserver/src/tenant/timeline.rs
```diff
@@ -3086,20 +3195,8 @@ impl Timeline {
             // while iterating it. BTreeMap::retain() would be another option)
             let mut layer_names_to_delete = Vec::with_capacity(layers_to_remove.len());
             for doomed_layer in layers_to_remove {
-                if let Some(path) = doomed_layer.local_path() {
-                    self.metrics
-                        .resident_physical_size_gauge
-                        .sub(path.metadata()?.len());
-                }
                 layer_names_to_delete.push(doomed_layer.filename());
-                doomed_layer.delete()?; // FIXME: schedule succeeded deletions before returning?
```
problame, line 3095: This FIXME is still valid, please keep it, and add it to ther other call site at compaction

pageserver/src/tenant/timeline.rs
```diff
@@ -1621,6 +1701,41 @@ impl Timeline {
...
+                return Some(historic_layer);
+            }
+        }
+
+        None
+    }
+
+    /// Removes the layer from local FS (if present) and from memory.
+    /// Remote storage is not affected by this operation.
+    fn delete_historic_layer(
```
problame, line 1718: Can we take the `layer_removal_cs` guard as an unused argument here, in order to hint to the caller that they need to hold it when they call this function?

SomeoneToIgnore: Nice idea, added.

pageserver/src/tenant/timeline.rs
SomeoneToIgnore: Returned.

test_runner/regress/test_layer_eviction.py
```diff
@@ -0,0 +1,128 @@
...
+            local_layer.name == returned_layer.layer_file_name
+            for local_layer in initial_local_layers
+        ), f"Did not find returned layer {returned_layer} in local layers {initial_local_layers}"
+
+    # redownload all evicted layers and ensure the initial state is restored
+    for local_layer in initial_local_layers:
+        client.download_layer(
+            tenant_id=tenant_id, timeline_id=timeline_id, layer_name=local_layer.name
+        )
+    time.sleep(1)  # ensure all layer downloads are finished
```
problame, line 115: This is going to be flaky eventually.
The `download_all_remote_layers` API exists specifically for this purpose.
There's even Python bindings that do the polling for you, see `client.timeline_download_remote_layers()`

SomeoneToIgnore: I think `timeline_download_remote_layers` is not needed since will always errors leading to no downloads, but whatever, used that.

pageserver/src/tenant/timeline.rs
```diff
@@ -832,6 +834,86 @@ impl Timeline {
...
+            LayerFileName::Delta(delta_name) => RemoteLayer::new_delta(
+                self.tenant_id,
+                self.timeline_id,
+                &delta_name,
+                &layer_metadata,
+            ),
+            #[cfg(test)]
+            LayerFileName::Test(_) => unreachable!(),
+        });
+
```
koivunej, line 905: failpoint idea discussed with @problame

pageserver/src/tenant/timeline.rs
```diff
@@ -832,6 +834,86 @@ impl Timeline {
...
+        }
+
+        let Some(remote_layer) = layer.downcast_remote_layer() else {return  Ok(Some(false)) };
+
+        self.download_remote_layer(remote_layer).await?;
+        Ok(Some(true))
+    }
+
+    pub async fn evict_layer(&self, layer_file_name: &str) -> anyhow::Result<Option<bool>> {
+        let Some(local_layer) = self.find_layer(layer_file_name) else { return Ok(None) };
```
koivunej, line 872: Idea: wrap the layermap as "PersistentLayerMap", move "loading" and "evict" etc. there?

libs/pageserver_api/src/models.rs
```diff
@@ -470,6 +524,175 @@ impl PagestreamBeMessage {
     }
 }
 
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq, Ord, PartialOrd, Serialize, Deserialize)]
+/// Key used in the Repository kv-store.
+///
+/// The Repository treats this as an opaque struct, but see the code in pgdatadir_mapping.rs
+/// for what we actually store in these fields.
```
koivunej, line 531:
```suggestion

/// Key used in the Repository kv-store.
///
/// The Repository treats this as an opaque struct, but see the code in pgdatadir_mapping.rs
/// for what we actually store in these fields.
#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq, Ord, PartialOrd, Serialize, Deserialize)]
```

libs/pageserver_api/src/models.rs
```diff
@@ -227,6 +231,56 @@ pub struct TimelineInfo {
...
+}
+
+#[serde_as]
+#[derive(Debug, Clone, Serialize)]
+#[serde(tag = "kind")]
+pub enum HistoricLayerInfo {
+    Delta {
+        layer_file_name: String,
+        #[serde_as(as = "DisplayFromStr")]
+        key_start: Key,
```
koivunej, line 263: Unsure if key needs to be here, we do have the `layer_file_name` already.

pageserver/src/tenant/timeline.rs
```diff
@@ -832,6 +834,86 @@ impl Timeline {
...
+        }
+    }
+
+    pub async fn download_layer(&self, layer_file_name: &str) -> anyhow::Result<Option<bool>> {
+        let Some(layer) = self.find_layer(layer_file_name) else { return Ok(None) };
+        if self.remote_client.is_none() {
+            return Ok(Some(false));
+        }
+
+        let Some(remote_layer) = layer.downcast_remote_layer() else {return  Ok(Some(false)) };
```
koivunej, line 865:
```suggestion
        let Some(remote_layer) = layer.downcast_remote_layer() else { return Ok(Some(false)) };
```

Interesting `rustfmt` doesn't handle this.

libs/pageserver_api/src/models.rs
problame: Yeah, unless there's s a particular need to have the keys repeated, I don't see why we need them in `LayerMapInfo`.
If we don't need them, we can move back the `Key` struct into the `pageserver` crate.

pageserver/src/tenant/timeline.rs
SomeoneToIgnore: Well, my 3rd stab on that idea is https://github.com/neondatabase/neon/pull/3473 too sad I cannot make it work.

pageserver/src/http/routes.rs
```diff
@@ -554,6 +555,79 @@ async fn tenant_size_handler(request: Request<Body>) -> Result<Response<Body>, A
...
+    let layer_map_info = timeline.layer_map_info();
+
+    json_response(StatusCode::OK, layer_map_info)
+}
+
+async fn layer_download_handler(request: Request<Body>) -> Result<Response<Body>, ApiError> {
+    let tenant_id: TenantId = parse_request_param(&request, "tenant_id")?;
+    let timeline_id: TimelineId = parse_request_param(&request, "timeline_id")?;
+    let layer_file_name = get_request_param(&request, "layer_file_name")?;
+    check_permission(&request, Some(tenant_id))?;
```
shanyp, line 579: shouldn't check_permission be right after tenant id parsing ? if there is no permission there is no value in parsing the rest

pageserver/src/http/routes.rs
```diff
@@ -554,6 +555,79 @@ async fn tenant_size_handler(request: Request<Body>) -> Result<Response<Body>, A
...
+
+    let tenant = mgr::get_tenant(tenant_id, true)
+        .await
+        .map_err(ApiError::NotFound)?;
+    let timeline = tenant
+        .get_timeline(timeline_id, true)
+        .map_err(ApiError::NotFound)?;
```
shanyp, line 614: as this repeats several times, did we consider get_timeline in the mgr ? a

libs/pageserver_api/src/models.rs
SomeoneToIgnore: We can go with `LayerFileName` alone and remove the rest of the fields.
Yet, it has to be changed, since has `Range` in its fields.

pageserver/src/tenant/timeline.rs
```diff
@@ -832,6 +834,86 @@ impl Timeline {
     pub fn subscribe_for_state_updates(&self) -> watch::Receiver<TimelineState> {
         self.state.subscribe()
     }
+
+    pub fn layer_map_info(&self) -> LayerMapInfo {
+        let mut in_memory_layers = Vec::new();
+        let layer_map = self.layers.read().unwrap();
```
shanyp, line 840: shouldn't the vec initialized with capacity ?

pageserver/src/tenant/timeline.rs
```diff
@@ -832,6 +834,86 @@ impl Timeline {
...
+
+        let Some(remote_layer) = layer.downcast_remote_layer() else {return  Ok(Some(false)) };
+
+        self.download_remote_layer(remote_layer).await?;
+        Ok(Some(true))
+    }
+
+    pub async fn evict_layer(&self, layer_file_name: &str) -> anyhow::Result<Option<bool>> {
+        let Some(local_layer) = self.find_layer(layer_file_name) else { return Ok(None) };
+        let Some(remote_client) = &self.remote_client else { return Ok(Some(false)) };
```
shanyp, line 873: should move after the check for is not remote layer

pageserver/src/tenant/timeline.rs
```diff
@@ -1621,6 +1703,43 @@ impl Timeline {
...
+        }
+
+        None
+    }
+
+    /// Removes the layer from local FS (if present) and from memory.
+    /// Remote storage is not affected by this operation.
+    fn delete_historic_layer(
+        &self,
+        // we cannot remote layers otherwise, since gc and compaction will race
```
shanyp, line 1722: did you mean: "we cannot delete remote layers "

test_runner/regress/test_layer_eviction.py
```diff
@@ -0,0 +1,133 @@
...
+        ), f"All layers should be evicted and not present locally, but got {returned_layer}"
+        assert any(
+            local_layer.name == returned_layer.layer_file_name
+            for local_layer in initial_local_layers
+        ), f"Did not find returned layer {returned_layer} in local layers {initial_local_layers}"
+
+    # redownload all evicted layers and ensure the initial state is restored
+    for local_layer in initial_local_layers:
+        client.download_layer(
+            tenant_id=tenant_id, timeline_id=timeline_id, layer_name=local_layer.name
```
shanyp, line 111: can it be done through SQL that will cause on demand download ?

SomeoneToIgnore: It could, but I think that's a different test, here we test the HTTP API.

pageserver/src/tenant/timeline.rs
SomeoneToIgnore: Yep, `remove` layers, fixed, thank you.

pageserver/src/tenant/timeline.rs
SomeoneToIgnore: Did also similar for downloads.

pageserver/src/tenant/timeline.rs
SomeoneToIgnore: It could, but the number of inmem layers should not be big and critical for this.
Done.

pageserver/src/tenant/timeline.rs
SomeoneToIgnore: Yes, `rustfmt` does not handle anything related to the feature, and `clippy` already has the lints to enforce this, odd 🙂

libs/pageserver_api/src/models.rs
SomeoneToIgnore: Moved `LayerFileName`, had to leave `Key` for that too, but now the API is better.

pageserver/src/tenant/timeline.rs
SomeoneToIgnore: For simple HTTP test, feels excessive.
I think https://github.com/neondatabase/neon/pull/3486#discussion_r1092993355 would work well with this idea.

test_runner/regress/test_layer_eviction.py
SomeoneToIgnore: We can add some later, with failpoints, as proposed in https://github.com/neondatabase/neon/pull/3486#discussion_r1092870621

pageserver/src/http/routes.rs
SomeoneToIgnore: I don't think the manager should know about `ApiError`, but extracted this into a function in `routes.rs`.

pageserver/src/repository.rs
```diff
@@ -1,177 +1,9 @@
...
-use anyhow::{bail, Result};
-use byteorder::{ByteOrder, BE};
 use bytes::Bytes;
 use serde::{Deserialize, Serialize};
-use std::fmt;
-use std::ops::{AddAssign, Range};
+use std::ops::AddAssign;
 use std::time::Duration;
 
-#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq, Ord, PartialOrd, Serialize, Deserialize)]
```
koivunej, line 10: Should this be moved back or not?

shanyp (approved):

pageserver/src/repository.rs
SomeoneToIgnore: I cannot, if I continue with the "file names in `models` module" approach: `DeltaLayer` has a `Range<Key>` inside of it, so `Key` has to be in the `models` module too.

libs/pageserver_api/src/models/filename.rs
```diff
@@ -242,12 +223,8 @@ impl serde::Serialize for LayerFileName {
...
     {
         match self {
-            LayerFileName::Image(fname) => serializer.serialize_str(&format!("{}", fname)),
-            LayerFileName::Delta(fname) => serializer.serialize_str(&format!("{}", fname)),
-            #[cfg(test)]
-            LayerFileName::Test(t) => {
-                serializer.serialize_str(&format!("{LAYER_FILE_NAME_TEST_PREFIX}{t}"))
-            }
+            LayerFileName::Image(fname) => serializer.serialize_str(&format!("{fname}")),
+            LayerFileName::Delta(fname) => serializer.serialize_str(&format!("{fname}")),
```
koivunej, line 227:
```suggestion
            LayerFileName::Image(fname) | LayerFileName::Delta(fname) => serializer.collect_str(&fname),
```

But I guess the scope of this PR is already quite much.

koivunej (commented):
I'm not too happy about moving keys, layerfilename to utils for this feature but we don't really have a repository either, perhaps this is not a big deal.

Since there's ample python typing support, should there also be openapi description? I never understood for who is that file maintained -- if control plane autogenerates something out of that, perhaps it's better to leave it out so far.

pageserver/src/repository.rs
koivunej: Yeah I wasn't expecting the LayerFileName to move either, but since it has now been done... Like I commented, probably doesn't matter so much where these are in the monorepo after all. Sadly the comments describing key are now a more far away but that's a problematic structure anyway. (Checked that those are referenced as `pgdatadir_mapping.rs` which is still accurate.)

At least we are not serializing Key's anymore which I think is more important -- I don't see value in that breakdown into fields.

Alternatives would had included:
- be generic over `LayerName: Serialize`
- `String`

SomeoneToIgnore:
> Since there's ample python typing support, should there also be openapi description? I never understood for who is that file maintained -- if control plane autogenerates something out of that, perhaps it's better to leave it out so far.

I've tried to state it in the PR description, and after recent discussions, seems that neither of the API is final, so adding them to the openapi schema seems preliminary?
I agree that it should be done soon, but now
* dumping layer map command will get more stat types and might change a few times
* evicting and downloading commands might be removed in their regular form

So, not sure about this really. Do you think I still should describe these and potentially leak them into the go codegen?

libs/pageserver_api/src/models/filename.rs
SomeoneToIgnore: I don't think they could be united that easily, since one is `DeltaFileName` and another is `ImageFileName`.
So, I'll leave it as is, thank you.

koivunej: Ah yeah didn't realize they were different types as well :) But yeah the point was `collect_str` which collects any Display and doesn't require to_string here.

problame:
I'm strongly against documenting any of these APIs in the openapi YML. They're for debugging only, the control plane should never rely on such implementation details.

problame (commented):
> I'm strongly against documenting any of these APIs in the openapi YML. They're for debugging only, the control plane should never rely on such implementation details.

In that spirit, I'd prefer if we 

- move `Key` back to where it was
- use `String` instead of `LayerFileName` in the api response for layer file names, and undo the move of `libs/pageserver_api/src/models/filename.rs`

Makes for a smaller PR overall and keeps the core logic of pageserver closer together (in terms of location of .rs files in the filesystem tree). `libs/pageserver_api` is just not the place where I'd expect to look for core types of pageserver.

SomeoneToIgnore:
I've decided to rebase and squash into one commit, since the number of separate commits became quite big for intended, small, set of changes.

I've removed `key_start: String` and `key_end`, but left `lsn_start: Lsn` in the API responses for layers. Layer file names are `String`, no entities are moved anymore, as it was initially.
The PR seems to contain the rest of the feedback in it, but please note if I missed something.

pageserver/src/tenant/timeline.rs
```diff
@@ -1290,7 +1370,7 @@ impl Timeline {
                     updates.insert_historic(remote_layer);
                 }
                 #[cfg(test)]
-                LayerFileName::Test(_) => unreachable!(),
+                LayerFileName::Test(_) => unimplemented!(),
```
problame, line 1374: I think this is actually unreachable, the tests never exercise this code path.

problame (approved):
Cleaned up some import churn (pushed to this branch) and added one more NIT-level comment. LGTM

pageserver/src/tenant/timeline.rs
SomeoneToIgnore: I plan to remove the entire variant after this PR lands.
I've actually done that in this PR after all moves, but reverted that.

problame: Yeah, saw that, but let's do that in a separate PR

SomeoneToIgnore: Sure, that's what I wanted.
My point here is that let's just leave it as is and it'll be gone in the next PR hopefully.

SomeoneToIgnore: Since had to push more code, changed this back to `unreachable!`

problame (approved):

pageserver/src/repository.rs
SomeoneToIgnore: Now it's a `String`.

PR #3495, opened 2023-01-31
Fix create release PR

vadim2404:


shanyp (approved):

bayandin (approved):
👍 in general, but I just wanted to highlight a couple of questions:

- Should we name this by the date of the last commit instead of the current date?
- What if we have opened release PR from the previous week?
- What if we don't have any commits in a week (no changes in code that we deploy)?

vadim2404:
> * Should we name this by the date of the last commit instead of the current date?
I don't want to spend a lot of effort on this automation. I see no differences in PR name
> * What if we have opened release PR from the previous week?
I don't want to face this situation. But if we face, we can just close it
> * What if we don't have any commits in a week (no changes in code that we deploy)?
Empty PR that we will close

PR #3502, opened 2023-01-31
add parse_query_param(); use Cow<> where possible; move param parsing code to utils::http::request

problame:
Predecessor: https://github.com/neondatabase/neon/pull/3503
Successor: https://github.com/neondatabase/neon/pull/3502

### DIFF
--- a/libs/utils/src/http/request.rs
+++ b/libs/utils/src/http/request.rs
@@ -1,4 +1,5 @@
-use std::str::FromStr;
+use core::fmt;
+use std::{borrow::Cow, str::FromStr};
 
 use super::error::ApiError;
 use anyhow::anyhow;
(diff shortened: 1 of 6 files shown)


pageserver/src/http/routes.rs
```diff
@@ -270,36 +271,11 @@ async fn timeline_list_handler(request: Request<Body>) -> Result<Response<Body>,
...
-    )
-}
-
 async fn timeline_detail_handler(request: Request<Body>) -> Result<Response<Body>, ApiError> {
     let tenant_id: TenantId = parse_request_param(&request, "tenant_id")?;
     let timeline_id: TimelineId = parse_request_param(&request, "timeline_id")?;
-    let include_non_incremental_logical_size =
-        query_param_present(&request, "include-non-incremental-logical-size");
+    let include_non_incremental_logical_size: Option<bool> =
+        parse_query_param(&request, "include-non-incremental-logical-size")?;
```
koivunej, line 278: Do we need to change semantics here (existence => bool)?

Though, this is not documented, so hopefully there are no users. Searching did not yield any uses.

koivunej (approved):

koivunej:
Breaks regression tests:

```
 FAILED test_runner/regress/test_pageserver_api.py::test_pageserver_http_api_client
FAILED test_runner/regress/test_pageserver_api.py::test_pageserver_http_api_client_auth_enabled
FAILED test_runner/regress/test_pageserver_api.py::test_pageserver_http_get_wal_receiver_not_found
FAILED test_runner/regress/test_tenants_with_remote_storage.py::test_tenant_redownloads_truncated_file_on_startup[local_fs]
FAILED test_runner/regress/test_timeline_size.py::test_timeline_size - fixtur...
FAILED test_runner/regress/test_timeline_size.py::test_timeline_size_createdropdb
FAILED test_runner/regress/test_timeline_size.py::test_timeline_initial_logical_size_calculation_cancellation[timeline_delete]
FAILED test_runner/regress/test_timeline_size.py::test_timeline_initial_logical_size_calculation_cancellation[tenant_detach]
FAILED test_runner/regress/test_timeline_size.py::test_timeline_size_quota - ...
```

libs/utils/src/http/request.rs
```diff
@@ -29,6 +30,52 @@ pub fn parse_request_param<T: FromStr>(
...
+        Some(q) => q,
+        None => return Ok(None),
+    };
+    let mut values = url::form_urlencoded::parse(query.as_bytes()).filter_map(|(k, v)| {
+        if k == param_name {
+            Some(v)
+        } else {
+            None
+        }
+    });
```
koivunej, line 47:
```suggestion
    }).fuse();
```

Since the `value1` might be None, it the second `.next` could do whatever.

koivunej: Or alternatively below:

```rust
let Some(value1) = values.next() else { return Ok(None) };
```

test_runner/fixtures/neon_fixtures.py
```diff
@@ -1232,9 +1232,9 @@ def timeline_list(
 
         params = {}
         if include_non_incremental_logical_size:
-            params["include-non-incremental-logical-size"] = "yes"
+            params["include-non-incremental-logical-size"] = "true"
         if include_timeline_dir_layer_file_size_sum:
-            params["include-timeline-dir-layer-file-size-sum"] = "yes"
+            params["include-timeline-dir-layer-file-size-sum"] = "true"
```
koivunej, line 1237: Another pair of these under `timeline_detail`.

libs/utils/src/http/request.rs
problame: > Since the value1 might be None, it the second .next could do whatever.

Good catch, thx!

pageserver/src/http/routes.rs
problame: Yeah, it's only used by the python tests.
I think properly parsing the arg is better, because it makes it easier for callers to use variables.

```
curl "http://...?include-non-incremental-logical-size=$include_size"
```

problame:
Deleted the target branch because it got cherry-picked-merged, now this PR is force-closed by GitHub :thinking: 

Will force-push to this PR once the underlying #3486  has been merged.

problame:
That didn't work, opened up https://github.com/neondatabase/neon/pull/3510 with the same patch, but against `main`.

PR #3503, opened 2023-01-31
replace btreemap with vec

problame:
see https://github.com/neondatabase/neon/pull/3486#discussion_r1091712440


SomeoneToIgnore:
Merged into the target branch, thank you.

PR #3505, opened 2023-01-31
Use the same runners/container for old prod deployments as for new prod

SergeyMelnikov:
docker doesn't work on the new runners yet: https://github.com/neondatabase/neon/actions/runs/4054496885/jobs/6976395557


zoete (approved):
Will add this to ticket for refactoring

PR #3510, opened 2023-02-01
add parse_query_param(); use Cow<> where possible; move param parsing code to utils::http::request

problame:
This was originally PR https://github.com/neondatabase/neon/pull/3502 which targeted a different branch.


koivunej (changes requested):
Commit style should be git, could be changed at squash time.

koivunej (approved):
With different commit message this is looking good (shorter first line) -- didn't realize I could turn off the auto-rebase.

PR #3514, opened 2023-02-01
Kb/better build stats

SomeoneToIgnore:


SomeoneToIgnore:
Do I need to solve the retention issue here? The files are small, so if the bucket is able to clean up its old contents already, it would be great.

.github/workflows/neon_extra_builds.yml
```diff
@@ -122,7 +125,31 @@ jobs:
...
+        id: upload-stats
+        env:
+          BUCKET: neon-github-public-dev
+          SHA: ${{ github.event.pull_request.head.sha || github.sha }}
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_DEV }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_KEY_DEV }}
+        run: |
+          REPORT_URL=https://${BUCKET}.s3.amazonaws.com/build-stats/${SHA}/${GITHUB_RUN_ID}/cargo-timing.html
+          aws s3 cp --recursive --only-show-errors ./target/cargo-timings/ "s3://${BUCKET}/build-stats/${SHA}/${GITHUB_RUN_ID}/"
+          echo "report-url=${REPORT_URL}" >> $GITHUB_OUTPUT
```
shanyp, line 137: can you set a TTL for files to be one week?

SomeoneToIgnore: https://aws.amazon.com/blogs/aws/amazon-s3-object-expiration/ says that in the management console, I can set up those per prefix.

So, I guess I can. Will do that after the merge.

.github/workflows/neon_extra_builds.yml
```diff
@@ -109,11 +117,6 @@ jobs:
           submodules: true
           fetch-depth: 1
 
-      - name: Install Ubuntu postgres dependencies
-        run: |
-          sudo apt update
-          sudo apt install build-essential libreadline-dev zlib1g-dev flex bison libseccomp-dev libssl-dev protobuf-compiler
-
```
koivunej, line 116: Are these already installed on gen3 or why remove these...?

koivunej (approved):
I think this looks all right, just one question on removed deps which is probably what I assume.

.github/workflows/neon_extra_builds.yml
SomeoneToIgnore: Before this PR, the "clean build" was run on GH Actions' runner, `ubuntu-latest`.
But now we need to write to our S3 bucket and that requires creds and tools installed, easiest way is to use something with all this, such as our runner, now gen3.

And yes, that has all tools installed, so no need to install them again.
I also have to disable cachepot now due to this runner switch.

.github/workflows/neon_extra_builds.yml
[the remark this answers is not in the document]
SomeoneToIgnore: Not sure I follow the question, but let me explain what happens here: 
* we have a public S3 bucket that is able to display its contents, anyone can go to `https://${BUCKET}.s3.amazonaws.com` , add some relative path in S3 and get the file returned.

Allure manipulates (and exposes) files under `reports/` subpath, I'm here adding another one, `build_stats/` and put there the HTMLs `cargo --timings` produces.

Now, I have to expose that html url (REPORT_URL) as a GH commit status (same way Allure reports do that), but that uses `script: |` for that and I'm not sure it's mixable with plain bash commands here, so instead of doing that directly, I write the URL into a shared var , `report-url`, that's exposed in the task output.

Then access this var and do `createCommitStatus` on its value, to display the report file URL on a commit.

SomeoneToIgnore: Set the retention for 60 days, just in case somebody wants to draw charts based on that data.
Check out the [corresponding bucket settings](https://s3.console.aws.amazon.com/s3/management/neon-github-public-dev/lifecycle/view?region=eu-central-1&id=Rust+build+stats+cleanup) if you want to verify me.

Both report files seem identical, so I've changed the code to copy only the `cargo-timing.html` one.
That leaves us with 408.2 KiB per build run (`main` + manual invocations via tags).
Feels ok to leave these for two months.

```
work/neon/neon kb/evict-on-demand:origin/kb/evict-on-demand ❯ aws s3 ls --recursive --human-readable --summarize --profile main s3://neon-github-public-dev/build-stats/
2023-02-01 17:26:52  408.2 KiB build-stats/03ef6ec3b798c00a71f79f45c48d1cd388f4b9d5/4065912911/cargo-timing-20230201T152518Z.html
2023-02-01 17:26:52  408.2 KiB build-stats/03ef6ec3b798c00a71f79f45c48d1cd388f4b9d5/4065912911/cargo-timing.html
2023-02-01 17:37:07  403.0 KiB build-stats/9bb4c80aa102b72585bdeb141de41dd613ff088e/4066006107/cargo-timing-20230201T153532Z.html
2023-02-01 17:37:07  403.0 KiB build-stats/9bb4c80aa102b72585bdeb141de41dd613ff088e/4066006107/cargo-timing.html

Total Objects: 4
   Total Size: 1.6 MiB
```

shanyp (approved):

PR #3516, opened 2023-02-01
carry stats across eviction & on-demand download

problame:
Heikki and Kirill asked for this in the review.

### DIFF
--- a/pageserver/src/tenant/storage_layer.rs
+++ b/pageserver/src/tenant/storage_layer.rs
@@ -111,7 +111,8 @@ struct LayerAccessStatsInner {
 pub enum LayerResidenceStatus {
     Resident {
         timestamp: SystemTime,
-        creating: bool,
+        /// If `true`, then this resident status marks the birth time of the layer.
+        created: bool,
     },
     Evicted {
         timestamp: SystemTime,
@@ -146,28 +147,42 @@ impl LayerAccessStatFullDetails {
 }
 
 impl LayerResidenceStatus {
+    /// Residence status for a layer file that only exists on the remote.
     pub fn evicted() -> Self {
         LayerResidenceStatus::Evicted {
             timestamp: SystemTime::now(),
         }
     }
 
-    pub fn resident(creating: bool) -> Self {
+    /// Residence status for a layer file that exists locally.
+    /// It may also exist on the remote, we don't care here.
+    /// NB: use this for existing layer files, e.g., during timeline load.
+    /// For newly written layer files, use [`created`].
+    pub fn resident() -> Self {
         LayerResidenceStatus::Resident {
             timestamp: SystemTime::now(),
-            creating,
+            created: false,
+        }
+    }
+
+    /// Residence status for a local layer file that we just wrote.
+    /// Example: a layer file created by compaction.
+    /// Private, because callers are supposed to use [`LayerAccessStats::new_for_new_layer_file`].
+    fn created() -> Self {
+        LayerResidenceStatus::Resident {
+            timestamp: SystemTime::now(),
+            created: true,
         }
     }
 
     fn to_api_model(&self) -> pageserver_api::models::LayerResidenceStatus {
         match self {
-            LayerResidenceStatus::Resident {
-                timestamp,
-                creating,
-            } => pageserver_api::models::LayerResidenceStatus::Resident {
-                timestamp_millis_since_epoch: system_time_to_millis_since_epoch(timestamp),
-                creating: *creating,
-            },
+            LayerResidenceStatus::Resident { timestamp, created } => {
+                pageserver_api::models::LayerResidenceStatus::Resident {
+                    timestamp_millis_since_epoch: system_time_to_millis_since_epoch(timestamp),
+                    created: *created,
+                }
+            }
             LayerResidenceStatus::Evicted { timestamp } => {
                 pageserver_api::models::LayerResidenceStatus::Evicted {
                     timestamp_millis_since_epoch: system_time_to_millis_since_epoch(timestamp),
@@ -205,7 +220,7 @@ impl LayerAccessStats {
 
     pub(crate) fn new_for_new_layer_file() -> Self {
         let new = LayerAccessStats(Mutex::new(LayerAccessStatsInner::default()));
-        new.record_residence_change(LayerResidenceStatus::resident(true));
+        new.record_residence_change(LayerResidenceStatus::created());
         new
     }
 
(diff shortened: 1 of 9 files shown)


pageserver/src/tenant/storage_layer.rs
```diff
@@ -138,12 +186,56 @@ impl Default for LayerAccessStatsInner {
...
+            last_residence_changes: VecDeque::with_capacity(
+                LayerAccessStats::LAST_RESIDENCE_CHANGES_MAX_LEN,
+            ),
         }
     }
 }
 
 impl LayerAccessStats {
     const LAST_ACCESSES_MAX_LEN: usize = 16;
+    const LAST_RESIDENCE_CHANGES_MAX_LEN: usize = 16;
```
SomeoneToIgnore, line 198: Optional: If you plan to keep these elements const, you can go down to arrays, there's https://doc.rust-lang.org/std/primitive.slice.html#method.rotate_right to help with the buffer.

SomeoneToIgnore (commented):
It's not bad, but I wonder if we could structure it differently?

We may keep `EnumMap<LayerAccessKind` with more info (not only count, but latest time accessed, etc.) and keep only one `VecDeque<LayerEvent>` with all accesses, downloads, evictions and the rest, maybe growing the allowed size for the vec.
`LayerEvent` could contain in one enum all possible kinds of events we want to track.

pageserver/src/tenant/storage_layer.rs
```diff
@@ -88,21 +88,34 @@ pub enum ValueReconstructResult {
...
     task_kind_flag: EnumSet<TaskKind>,
     last_accesses: VecDeque<LayerAccessStatFullDetails>,
+    last_residence_changes: VecDeque<LayerResidenceStatus>,
+}
+
+#[derive(Debug, Clone)]
+pub enum LayerResidenceStatus {
+    Resident {
+        timestamp: SystemTime,
+        creating: bool,
```
koivunej, line 114:
```suggestion
        /// Layer was created during this session.
        created: bool,
```

The `ing` seem like something ongoing, but our file creations should be atomic, layer doesn't really exist before we are done creating and writing a new file.

pageserver/src/tenant/storage_layer.rs
```diff
@@ -119,16 +138,45 @@ impl LayerAccessStatFullDetails {
...
+            timestamp: SystemTime::now(),
+        }
+    }
+
+    pub fn resident(creating: bool) -> Self {
+        LayerResidenceStatus::Resident {
+            timestamp: SystemTime::now(),
+            creating,
+        }
+    }
```
koivunej, line 160:
```suggestion
    /// Residence status for a new layer file created during this session.
    pub fn created() -> Self {
        LayerResidenceStatus::Resident {
            timestamp: SystemTime::now(),
            created: true,
        }
    }
    
    /// Residence status for a layer file found already on the disk.
    pub fn existing() -> Self {
        LayerResidenceStatus::Resident { timestamp: SystemTime::now(), created: false }
    }
```

koivunej (changes requested):
Hate to nitpick on naming but I find it confus**ing**, had some suggestions. Implementation-wise this seems quite clearcut.

Related: Carrying the `new_` prefix for names doesn't really carry information since those have the type `Fn() -> Type` so we know they are factories.

pageserver/src/tenant/storage_layer.rs
koivunej: Granted, we don't really have a word for the time period between `(load|attach) ---> (pageserver shutdown|detach)` but figured it could be a _tenant session_? :)

pageserver/src/tenant/storage_layer.rs
problame: I think the correct abstraction would be https://docs.rs/heapless/latest/heapless/struct.HistoryBuffer.html

Not sure I want to drag that library in, though :)

pageserver/src/tenant/timeline.rs
```diff
@@ -1190,6 +1197,7 @@ impl Timeline {
                     self.tenant_id,
                     &deltafilename,
                     file_size,
+                    LayerAccessStats::new_for_loading_layer(LayerResidenceStatus::resident(false)),
```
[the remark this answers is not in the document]
problame: Should be fixed with the separate constructor methods that you suggested.

pageserver/src/tenant/storage_layer.rs
koivunej: TIL: HistoryBuffer, even though I try to peek every now and then what's in heapless.

pageserver/src/tenant/storage_layer.rs
```diff
@@ -111,6 +125,12 @@ pub enum LayerAccessStatsReset {
     AllStats,
 }
 
+fn system_time_to_millis_since_epoch(ts: &SystemTime) -> u128 {
+    ts.duration_since(UNIX_EPOCH)
+        .expect("better to die in this unlikely case than report false stats")
+        .as_millis()
+}
+
```
koivunej, line 133:
```suggestion
fn system_time_to_millis_since_epoch(ts: &SystemTime) -> u64 {
    ts.duration_since(UNIX_EPOCH)
        .expect("better to die in this unlikely case than report false stats")
        .as_millis()
        .try_into()
        .expect("64 bits is enough for few more years")
}

```

problame:
>  It's not bad, but I wonder if we could structure it differently?
> We may keep EnumMap<LayerAccessKind with more info (not only count, but latest time accessed, etc.) and keep only one VecDeque<LayerEvent> with all accesses, downloads, evictions and the rest, maybe growing the allowed size for the vec.
> LayerEvent could contain in one enum all possible kinds of events we want to track.

I think we want to keep the history of last accesses vs residence changes separate. Residence changes will happen much less frequently. We don't want to lose that info just because of a few layer accesses.

pageserver/src/tenant/storage_layer.rs
problame: Pushed a change that switches to HistoryBuffer to this branch. @SomeoneToIgnore  what do you think?

SomeoneToIgnore: A bit sad that there's no release for that, but if you like `HistoryBuffer` more, why not.

koivunej (approved):

koivunej:
I don't think the `heapless` is necessary but fine with this as is.

PR #3519, opened 2023-02-01
well-defined marker file contents to support storing TenantConfOpt in the future

problame:
This is for https://github.com/neondatabase/neon/pull/3466#discussion_r1093413931

WDYT?

TODOs:

- [ ] adjust migration code

### DIFF
--- a/pageserver/src/tenant.rs
+++ b/pageserver/src/tenant.rs
@@ -19,6 +19,7 @@ use futures::TryFutureExt;
 use pageserver_api::models::TimelineState;
 use remote_storage::DownloadError;
 use remote_storage::GenericRemoteStorage;
+use scopeguard::ScopeGuard;
 use tokio::sync::watch;
 use tokio::task::JoinSet;
 use tracing::*;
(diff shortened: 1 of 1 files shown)


LizardWizzard (commented):
The approach sounds ok to me, but IMO it would be good to think about larger picture with all the mark files we have now and in which direction we would like to evolve it. Settle on format, json is fine, but is there a better option? Why not to go with something binary or toml? Tenant config is in toml, so it is confusing to me that attach marker which contains the same config needs to be in different format? If we're in a hurry to do something that works I'm on board with this PR. Another option on how to proceed without doing this right now would be to rename attach marker file when we change format. This way we can distinguish between empty file of v0 (which is fine) and half written v1 (which is not fine).

It may sound picky, but I think we may have another round of discussion when control plane will be integrating with this api and there may be some changes coming from this direction (unlikely, but who knows). Also IMO we need to be mindful to not pile things on top of current already big modules like tenant.rs or timeline.rs. WDYT?

Also I'm thinking about the case when we persisted the attach file, but crashed without replying back to client. In this case client will be required to retry and realize that attach succeeded despite failed request.

cc @SomeoneToIgnore WDYT?

problame:
Thanks for the review, good write-up of your thoughts!

> Another option on how to proceed without doing this right now would be to rename attach marker file when we change format. This way we can distinguish between empty file of v0 (which is fine) and half written v1 (which is not fine).

Sounds good.
So, we'll leave this PR unmerged, and merge the underlying #3466 as is?

LizardWizzard:
> So, we'll leave this PR unmerged, and merge the underlying https://github.com/neondatabase/neon/pull/3466 as is?

If there are no objections, yes

problame:
Idea abandoned

PR #3520, opened 2023-02-01
Send cached metrics every nth iteration

lubennikovaav:
@antonyc , can you clarify, if we need new config option for or just const is enough?
and what should be the default?

### DIFF
--- a/pageserver/src/consumption_metrics.rs
+++ b/pageserver/src/consumption_metrics.rs
@@ -93,7 +93,7 @@ pub async fn collect_metrics(
                     iter_num = 0;
                 }
 
-                collect_metrics_iteration(&client, &mut cached_metrics, metric_collection_endpoint, node_id, &ctx).await;
+                collect_metrics_iteration(&client, &mut cached_metrics, metric_collection_endpoint, node_id, &ctx, send_cached).await;
             }
         }
     }
--- a/pageserver/src/consumption_metrics.rs
+++ b/pageserver/src/consumption_metrics.rs
@@ -46,12 +46,14 @@ pub struct PageserverConsumptionMetricsKey {
 pub async fn collect_metrics(
     metric_collection_endpoint: &Url,
     metric_collection_interval: Duration,
+    cached_metric_collection_interval: Duration,
     synthetic_size_calculation_interval: Duration,
     node_id: NodeId,
     ctx: RequestContext,
 ) -> anyhow::Result<()> {
     let mut ticker = tokio::time::interval(metric_collection_interval);
-
+    let send_cached_every_iter =
+        cached_metric_collection_interval.as_secs() / metric_collection_interval.as_secs();
     info!("starting collect_metrics");
 
     // spin up background worker that caclulates tenant sizes
@@ -75,6 +77,7 @@ pub async fn collect_metrics(
     // define client here to reuse it for all requests
     let client = reqwest::Client::new();
     let mut cached_metrics: HashMap<PageserverConsumptionMetricsKey, u64> = HashMap::new();
+    let mut iter_num: u64 = 0;
 
     loop {
         tokio::select! {
@@ -83,6 +86,13 @@ pub async fn collect_metrics(
                 return Ok(());
             },
             _ = ticker.tick() => {
+                // send cached metrics every send_cached_every_iter iterations
+                iter_num += 1;
+                let send_cached = iter_num >= send_cached_every_iter;
+                if send_cached {
+                    iter_num = 0;
+                }
+
                 collect_metrics_iteration(&client, &mut cached_metrics, metric_collection_endpoint, node_id, &ctx).await;
             }
         }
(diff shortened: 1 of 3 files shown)


libs/consumption_metrics/src/lib.rs
```diff
@@ -42,6 +42,10 @@ pub fn idempotency_key(node_id: String) -> String {
 
 pub const CHUNK_SIZE: usize = 1000;
 
+// How often to send cached metrics.
+// Set to 10 to send cached metrics on every 10th iteration.
+pub const SEND_CACHED_ITER_STEP: u32 = 10;
```
lubennikovaav, line 47: better naming ideas are welcome

pageserver/src/consumption_metrics.rs
```diff
@@ -83,10 +86,18 @@ pub async fn collect_metrics(
                 return Ok(());
             },
             _ = ticker.tick() => {
-                if let Err(err) = collect_metrics_iteration(&client, &mut cached_metrics, metric_collection_endpoint, node_id, &ctx).await
+                // send cached metrics every SEND_CACHED_ITER_STEP iterations
+                iter_num += 1;
+                let send_cached = iter_num % SEND_CACHED_ITER_STEP == 0;
```
koivunej, line 91: This could also be:
```suggestion
                let send_cached = iter_num >= SEND_CACHED_ITER_STEP;
```

Since you reset it (though would move the reset closer to this one from lines 97...100). Whenever I see a `%` used for this I need to think extra-carefully does this only allow on 0 or end or both. I think the original would send on 10th, reset to 0, then send on 10th again because of the `iter_num += 1`.

Is this meant to clear out even on error..? It might be safer that way.

libs/consumption_metrics/src/lib.rs
koivunej:
```suggestion
pub const SEND_CACHED_EVERY_STEPS: u32 = 10;
```

antonyc:
I would put it into the config and set the timeout to 1 hour.

pageserver/src/consumption_metrics.rs
lubennikovaav: Thank you. I used this approach in new version

lubennikovaav:
@antonyc , 
> I would put it into the config and set the timeout to 1 hour.

Done. New config param is `cached_metric_collection_interval`

Do I need to add similar logic for proxy metric too?

antonyc:
> Do I need to add similar logic for proxy metric too?

Thanks! No, for proxy metrics it's not needed, they are not absolute, not needed to be recalculated hourly.

pageserver/src/consumption_metrics.rs
```diff
@@ -222,11 +233,13 @@ pub async fn collect_metrics_iteration(
         ));
     }
 
-    // Filter metrics
-    current_metrics.retain(|(curr_key, curr_val)| match cached_metrics.get(curr_key) {
-        Some(val) => val != curr_val,
-        None => true,
-    });
+    // Filter metrics, unless we want to send all metrics, including cached ones.
```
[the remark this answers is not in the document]
lubennikovaav: added in my last commit

pageserver/src/consumption_metrics.rs
```diff
@@ -83,7 +86,14 @@ pub async fn collect_metrics(
...
             _ = ticker.tick() => {
-                collect_metrics_iteration(&client, &mut cached_metrics, metric_collection_endpoint, node_id, &ctx).await;
+                // send cached metrics every send_cached_every_iter iterations
+                iter_num += 1;
+                let send_cached = iter_num >= send_cached_every_iter;
+                if send_cached {
+                    iter_num = 0;
+                }
+
+                collect_metrics_iteration(&client, &mut cached_metrics, metric_collection_endpoint, node_id, &ctx, send_cached).await;
```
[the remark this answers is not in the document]
lubennikovaav: Fixed. Only set the timer before calling the iteration function to not depend on its duration.

pageserver/src/consumption_metrics.rs
```diff
@@ -86,12 +84,13 @@ pub async fn collect_metrics(
...
-                iter_num += 1;
-                let send_cached = iter_num >= send_cached_every_iter;
-                if send_cached {
-                    iter_num = 0;
-                }
+
+                // send cached metrics every cached_metric_collection_interval
+                let send_cached = prev_iteration_time
+                .map(|x| x.elapsed() >= cached_metric_collection_interval)
+                .unwrap_or(false);
```
koivunej, line 91: Small formatting nit, but okay to merge as is.

koivunej (approved):

pageserver/src/consumption_metrics.rs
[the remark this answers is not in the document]
koivunej: Looking good.

PR #3522, opened 2023-02-02
WIP safekeeper walsender: read-write from single task.

arssher:
- Use postgres_backend_async throughout safekeeper.
- Use Framed in postgres_backend_async, it allows polling interface and takes some logic.
- Do read-write from single task in walsender.

The latter turned out to be more complicated than I initially expected due to 1) borrow checking and 2) anon Future types. 1) required SendRc<Refcell<...>> construct just to satisfy the checker; 2) is currently done via boxing futures, which is a pointless heap allocation in active path.

I'll probably try to workaround 2) with transmute, but it made me wonder whether socket split, like it was done previously, would be better. It is also messy though:
- we need to manage two tasks, properly join them and should on exit/error should join pgbackend back to leave it in valid state; pgbackend itself must swell a bit to provide splitted interface.
- issues with tls
- tokio::io::split has pointless mutex inside

fixing walreceiver and proxy is not done yet


arssher:
@funbringer this is a draft, as mentioned, but I'm interested in comments around `send_wal.rs` polling machinery: basically ReplicationHandler implementation. See also commit message.

CC @petuhovskiy

arssher:
And here is the version without boxes:
https://github.com/neondatabase/neon/blob/asher/sk-walsender-framed-pg-read-write-single-task-no-boxes/safekeeper/src/send_wal.rs

arssher:
> And here is the version without boxes:

Haven't tested it yet, will try. But considering how unappealing it looks, I realized that I can attempt sitting on both chairs: split the stream (and postgres backend), but keep parts in the same task, polling both. No issues with TLS, no need to synchronize tasks, and using async fns shouldn't require squats. Will try.

arssher:
> Haven't tested it yet, will try. 

Tested with pg_receivewal, it works. Nice!

arssher:
The next version is at
https://github.com/neondatabase/neon/tree/asher/sk-async-pg-backend

I added split to postgres_backend_async, switching manual framing, and started adapting receive_wal.rs accordingly. I wanted to have proper error propagation, and in the process realized that it would be more natural and easier there to have two threads one of which fully does network IO, and the other one writes to disk -- not as currently, where one thread reads data from network, and another writes to disk *and* replies back. So will follow this approach.

arssher:
closing in favor of
https://github.com/neondatabase/neon/pull/3576

PR #3523, opened 2023-02-02
Moving the template file location 

shanyp:
see https://github.com/appsmithorg/appsmith/issus/826#issuecomment-703093426 for details


shanyp:
> Not sure how that will change the behavior, but let's see.

me too, but I can't experiment without PRs for it :-(

PR #3524, opened 2023-02-02
Use regular layer names instead of a special test ones

SomeoneToIgnore:
We do not need special enum variant for testing the file names, neither its special handling across the code.
Current tests are able to create regular layers with normal layer names, as the PR shows.


koivunej (approved):
Looks simple enough, and I think you've done this many times by now :)

Issue #3200, opened 2022-12-23
Epic: Take into account the AZ while connecting PS to SK


vadim2404:
## Motivation
To reduce latency and expenditures on cross-AZ traffic, we need to consider the AZ while choosing the SK to which we need to connect. 

## DoD
The cross-AZ traffic has decreased.

## Implementation ideas
At first glance, we need to connect to SK the max LSN. But in our current situation, 2 of the 3 SKs will be with the same LSN. So, at the very beginning, I would suggest extending such criteria: when LSNs are equal, compare the SK AZ and PS SK and prefer to connect to SK in the same AZ.

## Tasks
- [x] start measuring cross-AZ traffic
- [ ] connect to closest SK in terms of the network when LSN difference is absent 

## Other related tasks and Epics
- neondatabase/cloud#2975

vadim2404:
@neondatabase/safekeepers, guys, can you provide ETA for this task?

petuhovskiy:
@vadim2404 I'm going to work on this, hope to add code for measuring cross-AZ traffic this week. If there will be no change in plans, it should reach staging before the end of next week.

vadim2404:
Thanks for keeping me updated!

petuhovskiy:
PR with metrics has reached prod, here is the query to look at traffic metrics 24 hours after deploy: https://neonprod.grafana.net/goto/K6NX5IfVz?orgId=1

Results look expected so far:
- 8% of safekeeper traffic is inter-AZ
- 16% of safekeeper traffic is cross-AZ
- 75% of safekeeper traffic has unknown source (but I'm almost sure it comes from computes)

We can also measure compute traffic, for that we only need to get access to AZ identifier inside a k8s pod and pass it in the connection string to safekeeper. That should fix unknown label on the 75% metrics but not urgent and can be done later.

16% of cross-AZ corresponds directly to PS connecting to SK from another AZ. That means that PR with advanced connection logic should move 16% of safekeeper traffic to inter-AZ category.

I also looked at AWS costs, at the `-DataTransfer-Regional-Bytes` which should be exactly cross-AZ traffic. I calculated we can only optimize 4% of our cross-AZ traffic costs with this epic, which is really not much compared to other EC2 costs.

vadim2404:
why only 4%? Do you have a metric about PS <-> SK traffic? We need to optimize only this particular part.

PR #3526, opened 2023-02-02
RFC about read-only replica

lubennikovaav:
RFC for https://github.com/neondatabase/neon/issues/769

### DIFF
--- a/docs/rfcs/read-only-replica.md
+++ b/docs/rfcs/read-only-replica.md
@@ -1,0 +1,60 @@
+# Read-only replicas
+
+We want to be able to spin-up read-only compute nodes.
+
+## Scope of this feature
+
+- We want to be able to spin-up multiple read-only replicas.
+- We should be able to spin-up RO node, if primary is inactive.
+
+- We don't want "snapshot" read-only computes at fixed LSN. Users can use branches for that.
+- We don't need replica promotion. For failower we rely on quick primary node restarts.
+
+- Implement it for v15 first.
+
+## Design
+
+### Replica creation
+
+1. At the moment of endpoint creation, user should be able to specify that they want to create a read-only endpoint.
+This setting will be passed via compute spec to `compute_ctl`. The replica will be created in the same region as primary (and all project's branches).
+
+Q: Should we allow to specify region for replica? It will affect latency and costs.
+
+2. `compute_ctl` will spin up a read-only replica.
+
+3. replica then connects to safekeepers and starts streaming WAL from them
+It would be good to connect to the safekeeper in the same AZ to reduce network latency.
+
+### Configuration & UI
+
+We take the same approach as with primary nodes and don't expose any configuration knobs to users.
+It may be useful to add a dashboard to show replica lag.
+
+### Replication
+
+- Replica receives WAL stream from safekeeper to keep up with primary.
+- Replica is connected to pageserver to receive page images.
+
+There are two cases of WAL apply:
+
+- page is already in buffer cache.
+In this case, replica should update page's `LastWriteLSN` and apply the WAL record to the page.
+
+NOTE: We must either hold buffer lock (pin) or update `LastWriteLSN` before applying WAL record. Otherwise concurrent process may request a stale page from pageserver.
+
+- page is not in buffer cache
+In this case, replica should update `LastWriteLSN` and ignore the WAL record.
+
+NOTE: We must be careful about atomicity of operations that modify multiple pages. See buffer lock coupling/crabbing ( https://github.com/neondatabase/neon/issues/383) for details. Most likely everything will work fine, but we need to test it properly.
+
+
+To request page, replica should use current WAL applyLSN to make sure that we don't get pages from the future.
+
+
+## Safekeepers
+
+- safekeepers should be able to stream WAL to multiple replicas and handle their feedback.
+
+- lagging replica can make safekeepers to keep a lot of WAL.
+If they can offload WAL to S3 and download it back on demand, it will be fine.


docs/rfcs/read-only-replica.md
```diff
@@ -0,0 +1,60 @@
...
+### Replica creation
+
+1. At the moment of endpoint creation, user should be able to specify that they want to create a read-only endpoint.
+This setting will be passed via compute spec to `compute_ctl`. The replica will be created in the same region as primary (and all project's branches).
+
+Q: Should we allow to specify region for replica? It will affect latency and costs.
+
+2. `compute_ctl` will spin up a read-only replica.
+
+3. replica then connects to safekeepers and starts streaming WAL from them
```
vadim2404, line 26: And to the page server in the same AZ.

There is the issue neondatabase/neon#3200 to connect ps to the sk in the same az

docs/rfcs/read-only-replica.md
```diff
@@ -0,0 +1,60 @@
...
+It may be useful to add a dashboard to show replica lag.
+
+### Replication
+
+- Replica receives WAL stream from safekeeper to keep up with primary.
+- Replica is connected to pageserver to receive page images.
+
+There are two cases of WAL apply:
+
+- page is already in buffer cache.
```
vadim2404, line 41: @knizhnik, will it affect file cache implementation?

docs/rfcs/read-only-replica.md
```diff
@@ -0,0 +1,60 @@
...
+
+
+To request page, replica should use current WAL applyLSN to make sure that we don't get pages from the future.
+
+
+## Safekeepers
+
+- safekeepers should be able to stream WAL to multiple replicas and handle their feedback.
+
+- lagging replica can make safekeepers to keep a lot of WAL.
```
vadim2404, line 59: Our replica instances will be terminated after 5 minutes of inactivity. What will happen with WAL on SK in case, when replica was terminated, but primary instance continues generating WAL?

docs/rfcs/read-only-replica.md
knizhnik: Certainly.
Most likely we should lookup target page not only in shared buffers, but also in local file cache.
And if we do not perform such lookup, then we should invalidate correspondent entry in local file cache.

docs/rfcs/read-only-replica.md
knizhnik: There are two reasons for using read-only replicas:
1. Fast switch-over in case of master fault (unlike spawning new compute node, replica is expected to have prewarmed cache)
2. Load balancing of red-only queries. 

If there is no workload (i.e. no requests within 5 minutes), then there is no sense in keeping replica. So we should remove (detach) replica fro master in this case before compute termination.

vadim2404: We have a matrix of states.

| # | Primary | Replica |
|--|---------|--------|
| 1 |  ❌ | ❌ |
| 2 | ✅ | ❌ |
| 3 | ❌ | ✅ |
| 4 | ✅ | ✅ | 

And all these states are possible. Imagine the small online shop, where primary accepts only write requests and all reads are sent to the replica.

1. There is no load of write and read queries (at night)
2. CRON job changes sth, but still, we don't have users at this moment
3. Google bot scans all pages 
4. Users actively use the application

And we should handle all such terminations (voluntary and/or involuntary).

docs/rfcs/read-only-replica.md
knizhnik: Add branch https://github.com/neondatabase/neon/pull/new/read_only_replica
with implementation of redo filter for read-only replica which performs redo only for available pages.
I believe that it is correct (explaining why in comments) and doesn't require any changes in Postgres core.

docs/rfcs/read-only-replica.md
```diff
@@ -0,0 +1,60 @@
...
+It would be good to connect to the safekeeper in the same AZ to reduce network latency.
+
+### Configuration & UI
+
+We take the same approach as with primary nodes and don't expose any configuration knobs to users.
+It may be useful to add a dashboard to show replica lag.
+
+### Replication
+
+- Replica receives WAL stream from safekeeper to keep up with primary.
```
SomeoneToIgnore, line 36: Also below,

>  - safekeepers should be able to stream WAL to multiple replicas and handle their feedback.

Similar things is [done](https://github.com/neondatabase/neon/tree/main/pageserver/src/tenant/timeline) by the pageserver now already: there's a lagging node detection and other heuristics.

A few related concerns:

* safekeepers share their state via broker, that uses `grpc` protocol. 

If I'm getting this right, the new functionality is supposed to appear in compute, so is it even feasible (possible?) to message over grpc from compute? 

* Is it so, that the current architecture still has pageserver as the only node connecting to safekeepers for data?
While pageserver has [some](https://github.com/neondatabase/neon/issues/2106) issues, it still aims at "get data form SKs ASAP and return a page for it" response handling.
Recent WAL (that's presumably understood under "to keep up with primary" activity) is stored in pageserver's memory and ready to serve page requests as soon as it's received from SK.

Since current proposal keeps the RO computes to pageserver, what are the benefits of making replicas to stream latest WAL from SKs directly?

vadim2404: > Since current proposal keeps the RO computes to pageserver, what are the benefits of making replicas to stream latest WAL from SKs directly?

To clear/update internal caches (shared buffers and file cache). Read-only replica != proxy to pageserver

docs/rfcs/read-only-replica.md
```diff
@@ -0,0 +1,60 @@
...
+
+We want to be able to spin-up read-only compute nodes.
+
+## Scope of this feature
+
+- We want to be able to spin-up multiple read-only replicas.
+- We should be able to spin-up RO node, if primary is inactive.
+
+- We don't want "snapshot" read-only computes at fixed LSN. Users can use branches for that.
+- We don't need replica promotion. For failower we rely on quick primary node restarts.
```
knizhnik, line 11: The primary goal of using replicas in Postgres is HA, not load balancing.
Yes, Neon allows to quickly restart compute. But with cold cache.
And performance of Neon greatly depends on local caches: if data set fits in compute node then it is comparable with standalone Postgres, if not, then at random-access workload it is  ~10x times slower.

So if we assume that in most cases working data set fit in compute cache, then keeping warm replica seems to be good solution for fast switch-over without cache warming. But it will not be possible if there no read-only workload at replica and we apply WAL records at replica only to available pages.

So I am not sure whether we need replicas for fast HA, but at least I think that we should mention this option and explain related problems.

docs/rfcs/read-only-replica.md
```diff
@@ -0,0 +1,60 @@
...
+
+- Implement it for v15 first.
+
+## Design
+
+### Replica creation
+
+1. At the moment of endpoint creation, user should be able to specify that they want to create a read-only endpoint.
+This setting will be passed via compute spec to `compute_ctl`. The replica will be created in the same region as primary (and all project's branches).
```
knizhnik, line 20: As far as I understand (at least I heard this requirement from somebody) one of the motivations of read-only replicas is to provide fast read-only access to the data from different region. In this case replica should be create in different region. And most likely we also needs pageserver in this region. So we also need to have primary/standby pageservers....

knizhnik (approved):
I still do not feel that we have a consensus concerning using of replicas.
There are at least four different use cases:
1. Load balancing of read-only workload within the same region. It assumes that either client application is designed in this way that it can separate update and read-only queries. Either therei is some intermediate proxy (like `pgbouncer-rr`) which can do it transparently to application.
2. Separate OLTP from OLAP. It is mostly imilar with 1), but in this case there are different applications which working with master and replicas.
3. Provide fast access to remote data(from other region). In this case replica needs to be install in different region than master. And we either need local copy of pageserver, either assume that replica is able to hold most of the data.
4. HA. Fast switch-over without cache warming. In case of Neon at can be achieved only if we have read-only workload on replicas which can keep cache warm.

I am not sure if we want to support all of them. But at least we should mention them and explain related challenges.

docs/rfcs/read-only-replica.md
pfwang80s: > Replica receives WAL stream from safekeeper to keep up with primary.

To keep PG's SSI, it seems that when RW commits a transaction, RW has to wait until WAL is received and applied to all of replicas?

knizhnik: Just waiting for applying WAL at all replicas is not enough to provide SSI. Some other mechanism (like commit serial numbers - CSN, or logical time) are need to enforce DSSI.

Standard postgres replication mechanism support async and sync mode. In the last case commit is blocked until all sync replicas confirm receiving the WAL. But not applying the WAL. So it is still possible with sync replication the changes visible at master are not yet visible at replica.

ololobus:
@lubennikovaav let's close this one? I think we merged this one instead https://github.com/neondatabase/neon/pull/8546 Or do we need something from this PR?

PR #3527, opened 2023-02-02
introduce LaunchTimestamp to identify process restarts

problame:
This patch adds a LaunchId type, libmetric_ Prometheus metric, and response header for the pageserver's management API.

The idea is to generate a unique string upon process launch, and expose that string in places where an outside observer needs to identify a process restart.

The motivation for this is that we plan to scrape the pageserver's /v1/tenant/:tenant_id/timeline/:timeline_id/layer
HTTP endpoint over time. It will soon expose access metrics (#3496) which reset upon process restart. We will use the pageserver's launch ID to identify a restart between two scrape points.

However, there are other use cases for the launch ID. For example, we could use the Prometheus metric to annotate Grafana plots.

Or, we could add it to the `tracing` span to enable filtering logs by process.

### DIFF
--- a/libs/utils/src/http/endpoint.rs
+++ b/libs/utils/src/http/endpoint.rs
@@ -1,7 +1,8 @@
 use crate::auth::{Claims, JwtAuth};
 use crate::http::error;
-use anyhow::anyhow;
-use hyper::header::AUTHORIZATION;
+use anyhow::{anyhow, Context};
+use hyper::header::{HeaderName, AUTHORIZATION};
+use hyper::http::HeaderValue;
 use hyper::{header::CONTENT_TYPE, Body, Request, Response, Server};
 use metrics::{register_int_counter, Encoder, IntCounter, TextEncoder};
 use once_cell::sync::Lazy;
@@ -13,6 +14,7 @@ use tracing::info;
 
 use std::future::Future;
 use std::net::TcpListener;
+use std::str::FromStr;
 
 use super::error::ApiError;
 
@@ -141,6 +143,38 @@ pub fn auth_middleware<B: hyper::body::HttpBody + Send + Sync + 'static>(
         }
         Ok(req)
     })
+}
+
+pub fn add_response_header_middleware<B>(
+    header: &str,
+    value: &str,
+) -> anyhow::Result<Middleware<B, ApiError>>
+where
+    B: hyper::body::HttpBody + Send + Sync + 'static,
+{
+    let name =
+        HeaderName::from_str(header).with_context(|| format!("invalid header name: {header}"))?;
+    let value =
+        HeaderValue::from_str(value).with_context(|| format!("invalid header value: {value}"))?;
+    Ok(Middleware::post_with_info(
+        move |mut response, request_info| {
+            let name = name.clone();
+            let value = value.clone();
+            async move {
+                let headers = response.headers_mut();
+                if headers.contains_key(&name) {
+                    tracing::warn!(
+                        "{} response already contains header {:?}",
+                        request_info.uri(),
+                        &name,
+                    );
+                } else {
+                    headers.insert(name, value);
+                }
+                Ok(response)
+            }
+        },
+    ))
 }
 
 pub fn check_permission_with(
--- a/pageserver/src/bin/pageserver.rs
+++ b/pageserver/src/bin/pageserver.rs
@@ -7,6 +7,7 @@ use std::{env, ops::ControlFlow, path::Path, str::FromStr};
 use anyhow::{anyhow, Context};
 use clap::{Arg, ArgAction, Command};
 use fail::FailScenario;
+use metrics::launch_timestamp::{set_launch_timestamp_metric, LaunchTimestamp};
 use remote_storage::GenericRemoteStorage;
 use tracing::*;
 
@@ -52,6 +53,8 @@ fn version() -> String {
 }
 
 fn main() -> anyhow::Result<()> {
+    let launch_ts = Box::leak(Box::new(LaunchTimestamp::generate()));
+
     let arg_matches = cli().get_matches();
 
     if arg_matches.get_flag("enabled-features") {
@@ -108,7 +111,7 @@ fn main() -> anyhow::Result<()> {
     virtual_file::init(conf.max_file_descriptors);
     page_cache::init(conf.page_cache_size);
 
-    start_pageserver(conf).context("Failed to start pageserver")?;
+    start_pageserver(launch_ts, conf).context("Failed to start pageserver")?;
 
     scenario.teardown();
     Ok(())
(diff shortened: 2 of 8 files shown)


libs/metrics/src/launch_id.rs
```diff
@@ -0,0 +1,40 @@
...
+            .try_into()
+            .expect("many years to come until this doesn't fit into u63");
+        let rand: u64 = rand::random();
+        let alphabet = base32::Alphabet::RFC4648 { padding: false };
+        Self(format!(
+            "{}-{}-{}",
+            prefix,
+            base32::encode(alphabet, &startup_time_millis_since_epoch.to_le_bytes()[..]),
+            base32::encode(alphabet, &rand.to_le_bytes()[..]),
+        ))
```
koivunej, line 25: Is this not some format for which we have an existing crate?

koivunej: https://github.com/mmacedoeu/rulid.rs

But do we really need all this vs. a timestamp, which could be a counter value (instead of a label)? Prometheus probably cannot handle 128-bit values.

libs/metrics/src/launch_id.rs
```diff
@@ -0,0 +1,40 @@
...
+pub struct LaunchId(String);
+
+impl LaunchId {
+    pub fn generate(prefix: &str) -> Self {
+        let startup_time_millis_since_epoch: u64 = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap()
+            .as_millis()
+            .try_into()
+            .expect("many years to come until this doesn't fit into u63");
```
koivunej, line 17: Note we are not enforcing this to be u63 but u64, but I am unsure why would that be done?

libs/utils/src/http/endpoint.rs
```diff
@@ -143,6 +143,31 @@ pub fn auth_middleware<B: hyper::body::HttpBody + Send + Sync + 'static>(
...
+        if headers.contains_key(header) {
+            tracing::warn!(
+                "{} response already contains header {}",
+                request_info.uri(),
+                header,
+            );
+        } else {
+            headers.insert(
+                header,
+                value.parse().expect("any UTF-8 string should work here"),
```
koivunej, line 164: If you'd parse the value outside of this `async move {}` in the body, the panic would be accessible to the `main` during start, instead of on first request.

If I remember correctly hyper would also prefer an already converted HeaderName, which could be converted in the main body.

libs/metrics/src/launch_id.rs
koivunej: Hehe just checked it's 5y old.

koivunej: https://docs.rs/ulid/latest/ulid/ is more up to date :)

koivunej: ulid-rs looks all right.

koivunej (changes requested):
If we really need this ulid-alikeness, I'd go with `ulid` crate instead of bespoke ulid-alike, which will also live as u128 instead of a string.

If we only want metrics, I think we could do with the 64-bit epoch millis, and not wrap that as a string, it might be useful for keeping other in-session calculations later on -- for a year or so with u32 :)

problame:
ulid looks nice, will switch this PR to that. It should address most of your review comments as well.

> If we only want metrics, I think we could do with the 64-bit epoch millis, and not wrap that as a string, it might be useful for keeping other in-session calculations later on -- for a year or so with u32 :)

Not really sure what you mean by this.

I think with the switch to `ulid`, I'd just keep `LaunchId` a newtypre around `ulid::Ulid`.

koivunej:
>> If we only want metrics, I think we could do with the 64-bit epoch millis, and not wrap that as a string, it might be useful for keeping other in-session calculations later on -- for a year or so with u32 :)
>
> Not really sure what you mean by this.

As in have a metric which just sets a counter/gauge to millis since epoch at startup. That cannot be mixed with any other value because how our ... who ever scrapes it enhances the metrics with additional labels. So we don't need any randomness. We could add a millisecond sleep before "generating" this 64-bit millis value, which would guarantee we wont see duplicates.

This is to say, we don't need any randomness, because other parts provide a replacement for that.

With the latter part:

> it might be useful for keeping other in-session calculations later on -- for a year or so with u32

I meant that if we were to need relative timestamps within pageserver, we could in some cases represent them with u32 since our own epoch (pageserver startup), which could help in certain situations like trying to keep message payloads smaller. Well, perhaps that'd still be too small for millis, but over a year with 0.01s.

problame:
> As in have a metric which just sets a counter/gauge to millis since epoch at startup. That cannot be mixed with any other value because how our ... who ever scrapes it enhances the metrics with additional labels. So we don't need any randomness. We could add a millisecond sleep before "generating" this 64-bit millis value, which would guarantee we wont see duplicates.
> This is to say, we don't need any randomness, because other parts provide a replacement for that.

Ok, now I get it.
Sounds appealing, especially in combination with the 1ms of sleep to guarantee uniqueness.
Problems I see:
* clock drift
* a human looking at just timestamps has a harder time deciding whether they're the same or different compared to ulids

koivunej (approved):

pageserver/src/http/routes.rs
```diff
@@ -937,6 +939,11 @@ pub fn make_router(
         }))
     }
 
+    router = router.middleware(
+        endpoint::add_response_header_middleware("PAGESERVER_LAUNCH_ID", &launch_id.to_string())
```
koivunej, line 943: Hmm just realized, is this the way to use custom headers?

koivunej: Oki, I was late about 10y and missed RFC 6648. This should be good. Resolving.

libs/utils/src/http/endpoint.rs
koivunej: This seems to be implemented.

koivunej:
Still against for putting this into all-span though :) There has to be a better way.

problame:
As discussed on Slack, changed this PR to just use a timestamp (`LaunchTimestamp`).
Using that in the metric, startup log message, and header.

pageserver/src/bin/pageserver.rs
```diff
@@ -203,13 +206,24 @@ fn initialize_config(
...
+) -> anyhow::Result<()> {
     // Initialize logging
     logging::init(conf.log_format)?;
 
-    // Print version to the log, and expose it as a prometheus metric too.
-    info!("version: {}", version());
+    // Print version and launch timestamp to the log,
+    // and expose them as prometheus metrics.
+    // A changed version string indicates changed software.
+    // A changed launc timestamp indicates a pageserver restart.
```
koivunej, line 219:
```suggestion
    // A changed launch timestamp indicates a pageserver restart.
```

koivunej (approved):
Fine with this as well.

problame:
Final commit message:

```
    introduce LaunchTimestamp to identify process restarts
    
    This patch adds a LaunchTimestamp type to the `metrics` crate,
    along with a `libmetric_` Prometheus metric.
    
    The initial user is pageserver.
    In addition to exposing the Prometheus metric, it also reproduces
    the launch timestamp as a header in the API responses.
    
    The motivation for this is that we plan to scrape the pageserver's
    /v1/tenant/:tenant_id/timeline/:timeline_id/layer
    HTTP endpoint over time. It will soon expose access metrics (#3496)
    which reset upon process restart. We will use the pageserver's launch
    ID to identify a restart between two scrape points.
    
    However, there are other potential uses. For example, we could use
    the Prometheus metric to annotate Grafana plots whenever the launch
    timestamp changes.
```

PR #3528, opened 2023-02-02
Local docker build instructions

bojanserafimov:
People suggested various ways to docker build locally. I chose the one method that requires no AWS permissions or special tools to be installed, and allows local edits of the base rust image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,13 @@
 ### The image itself is mainly used as a container for the binaries and for starting e2e tests with custom parameters.
 ### By default, the binaries inside the image have some mock parameters and can start, but are not intended to be used
 ### inside this image in the real deployments.
+###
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -2,6 +2,13 @@
 ### The image itself is mainly used as a container for the binaries and for starting e2e tests with custom parameters.
 ### By default, the binaries inside the image have some mock parameters and can start, but are not intended to be used
 ### inside this image in the real deployments.
+###
+### Local build instructions:
+### 1. Clone the `neondatabase/build` repo somewhere. Recommended location is `../build`.
```
SomeoneToIgnore, line 7: image

That would be amazing to do indeed, but alas.

bojanserafimov:

PR #3529, opened 2023-02-02
[proxy] Fix project (aka endpoint) init in the password hack handler

funbringer:
The project/endpoint should be set in the original (non-as_ref'd) creds, because we call `wake_compute` not only in `try_password_hack` but also later in the connection retry logic.

This PR also removes the obsolete `as_ref` method and makes the code simpler because we no longer need this complication after a recent refactoring.

Further action points: finally introduce typestate in creds (planned).


LizardWizzard (approved):
LGTM

PR #3536, opened 2023-02-03
std::fmt::Debug for Layer implementations

koivunej:
Follow-up to #3513.

This removes the old blanket `std::fmt::Debug` impl on `dyn Layer` which did not seem to be used from anywhere (no compilation errors after removing).

Adds `std::fmt::Debug` requirement and implementations for `trait Layer` implementors:
- LayerDescriptor (derived)
- RemoteLayer (manual)
- DeltaLayer (manual)
- ImageLayer (manual)

Adds and adjusts some doc comments to be more rustdoc alike.

In follow-up #3544: Use this `std::fmt::Debug` when logging a `LayerMap::replace` failure for which we currently have no tests but speculation (see #3387 comments).

Alternatives for adding the `std::fmt::Debug`:
- add a custom `Layer::get_type_name(&self) -> &'static str`
  - would give no additional information to the place of use except the type name
- add other custom `std::fmt::Debug` alike functionality to the trait
  - `std::fmt::Debug` is the way to debug format objects

### DIFF
--- a/pageserver/src/tenant/storage_layer/delta_layer.rs
+++ b/pageserver/src/tenant/storage_layer/delta_layer.rs
@@ -188,6 +188,17 @@ pub struct DeltaLayer {
     inner: RwLock<DeltaLayerInner>,
 }
 
+impl std::fmt::Debug for DeltaLayer {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("DeltaLayer")
+            .field("key_range", &self.key_range)
+            .field("lsn_range", &self.lsn_range)
+            .field("file_size", &self.file_size)
+            .field("inner", &self.inner)
+            .finish()
+    }
+}
+
 pub struct DeltaLayerInner {
     /// If false, the fields below have not been loaded into memory yet.
     loaded: bool,
@@ -198,6 +209,16 @@ pub struct DeltaLayerInner {
 
     /// Reader object for reading blocks from the file. (None if not loaded yet)
     file: Option<FileBlockReader<VirtualFile>>,
+}
+
+impl std::fmt::Debug for DeltaLayerInner {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("DeltaLayerInner")
+            .field("loaded", &self.loaded)
+            .field("index_start_blk", &self.index_start_blk)
+            .field("index_root_blk", &self.index_root_blk)
+            .finish()
+    }
 }
 
 impl Layer for DeltaLayer {
--- a/pageserver/src/tenant/storage_layer/delta_layer.rs
+++ b/pageserver/src/tenant/storage_layer/delta_layer.rs
@@ -167,14 +167,13 @@ impl DeltaKey {
     }
 }
 
+/// DeltaLayer is the in-memory data structure associated with an on-disk delta
+/// file.
 ///
-/// DeltaLayer is the in-memory data structure associated with an
-/// on-disk delta file.  We keep a DeltaLayer in memory for each
-/// file, in the LayerMap. If a layer is in "loaded" state, we have a
-/// copy of the index in memory, in 'inner'. Otherwise the struct is
-/// just a placeholder for a file that exists on disk, and it needs to
-/// be loaded before using it in queries.
-///
+/// We keep a DeltaLayer in memory for each file, in the LayerMap. If a layer
+/// is in "loaded" state, we have a copy of the index in memory, in 'inner'.
+/// Otherwise the struct is just a placeholder for a file that exists on disk,
+/// and it needs to be loaded before using it in queries.
 pub struct DeltaLayer {
     path_or_conf: PathOrConf,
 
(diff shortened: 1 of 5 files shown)


pageserver/src/tenant/storage_layer/delta_layer.rs
```diff
@@ -188,6 +187,17 @@ pub struct DeltaLayer {
     inner: RwLock<DeltaLayerInner>,
 }
 
+impl std::fmt::Debug for DeltaLayer {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("DeltaLayer")
+            .field("key_range", &self.key_range)
+            .field("lsn_range", &self.lsn_range)
+            .field("file_size", &self.file_size)
+            .field("inner", &self.inner)
```
koivunej, line 196: I did not check, but at least `std::sync::Mutex` prints out the contents of a mutex if try_lock succeeds. Could be it's not safe/possible for `RwLock`, if so, this might not add any new information except for the type.

pageserver/src/tenant/storage_layer/remote_layer.rs
```diff
@@ -39,6 +46,16 @@ pub struct RemoteLayer {
     pub(crate) ongoing_download: Arc<tokio::sync::Semaphore>,
 }
 
+impl std::fmt::Debug for RemoteLayer {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("RemoteLayer")
+            .field("file_name", &self.file_name)
+            .field("layer_metadata", &self.layer_metadata)
+            .field("is_incremental", &self.is_incremental)
+            .finish()
```
problame, line 55: would like to see `is_delta` in here.

problame (approved):
Nice job!

pageserver/src/tenant/storage_layer/remote_layer.rs
koivunej: AFAIK It's in the filename (lsn range) -- I was quite close to axing the extra fields in RemoteLayer but just left them out here. Or is there a case where ... we have a single lsn delta layer? Or some other case I couldn't figure out?

Semi-related: While adding the tests to layermap, noted it considers any (delta or image) covering the whole range as `l0_delta` (so lsn is not checked at all). This most likely follows the inability to create an image for anything at 0 or 0xff..ff but still.

pageserver/src/tenant/storage_layer/delta_layer.rs
koivunej: It does `try_read()` so this is great: https://doc.rust-lang.org/stable/src/std/sync/rwlock.rs.html#482

pageserver/src/tenant/storage_layer/remote_layer.rs
koivunej: `is_delta` is not needed because it is always `matches!(self.filename, LayerFileName::Delta(_))` and `file_name` is already included. Should be removed.

koivunej:
This should probably wait eviction related stats because we might want to include those.

koivunej:
After looking a bit how to use this on the new evict_layer endpoint, I am thinking I'll just separate that to different PR altogether (with the [c8b3139](https://github.com/neondatabase/neon/pull/3536/commits/c8b3139e2bd264fcb2f33a7283c7df8cea4c984d)).

SomeoneToIgnore (approved):
I'm a bit sad that it makes it harder (at least, a bigger patch) to go away from the "`&dyn Trait` as an interface" approach, but otherwise looks good.

koivunej:
@SomeoneToIgnore please elaborate how will this make it more difficult? I think the main point is that we should have `Debug` for Layer impls, surely this is as easy to do with more typed values than not?

SomeoneToIgnore:
Maybe I'm wrong, just imagined that now every place with `Arc<dyn Layer>` could be debugged, and for the later swap, more code changes could be required in either of such place.
But it's still noting big, having `Debug` seems to be a good thing in general.

pageserver/src/tenant/storage_layer/remote_layer.rs
koivunej: Not merging until this is clarified :)

koivunej:
> more code changes could be required in either of such place.

Oki now I understood. I think in future it's just important to use `#[derive(Debug)]` *regardless of the shape*. Here it cannot *yet* be used because we could dump a lot of indices for InMemoryLayer and few of those extra fields, perhaps EphmeralFile would had had quite long `Debug` output as well (the full hashmap of files).

I don't want to drop fields (above discussion on RemoteLayer I'm keeping open) or add newtypes which would suppress any long output (InMemoryLayer's index) to move towards that **on this PR** but having `Debug` impls in general will be a good thing.

pageserver/src/tenant/storage_layer/remote_layer.rs
problame: I think you're right, I was just skimming over the definition of `RemoteLayer` and thought `is_delta` would be relevant information, without checking what it actually means.
Feel free to resolve this conversation.

koivunej: Oki, will type up a tech debt issue of this, resolving and merging.

Issue #3537, opened 2023-02-03
Benchmarks fail with assertion failure


koivunej:
@bayandin thought they started failing on https://github.com/neondatabase/neon/actions/runs/4078437880

Latest failure in https://neon-github-public-dev.s3.amazonaws.com/reports/main/release/4084641341/index.html:

```
test_runner/performance/test_seqscans.py:58: in test_seqscans
    assert int(shared_buffers) < int(table_size)
E   assert 2147483648 < 3629056
E    +  where 2147483648 = int(2147483648)
E    +  and   3629056 = int(3629056)
```

PR #3538, opened 2023-02-03
Revert accidental change

bojanserafimov:
Resolves https://github.com/neondatabase/neon/issues/3537

In https://github.com/neondatabase/neon/commit/ada933eb426cc26fd95e9a03039531eea2522598#diff-aad3374c917096378e2433fd7e6d7e87611c734ea96dc4f4da150d51d45c28cd I used it for debugging and forgot to take it out


koivunej (approved):

knizhnik:
May be we should also fix assert test_seqscan because assumption that shared buffer can not be grater than 2GBis not correct.

PR #3541, opened 2023-02-05
Expose the tenant calculated synthetic size as a Prometheus metric

shanyp:
## Describe your changes
Expose the currently calculated synthetic size as a Prometheus metric
#3509


koivunej (approved):
This should work. I do feel that the temporary workaround to not using channels in consumption metrics collection and using the `cached_synthetic_tenant_size` instead is starting to get out of hand..  

Note I went ahead and added the related issue as linked issue, please remove it if that was not the right thing to do.

Issue #3533, opened 2023-02-03
Proper handling for download failing to replace layer


koivunej:
First discussed in https://github.com/neondatabase/neon/issues/3387#issuecomment-1398311655, made more robust in #3513, but still remains to be done:
- reproducing test case
- proper handling for metrics
- downloaded file on disk

koivunej:
Because of compaction never downloading a remote layer and gc depending on the presence of image layers, I haven't yet figured out a way for this to be reproduced in a test. We certainly have the conditions for it.

If we supported compaction doing on-demand download while not holding `layer_removal_cs` #3589 suggested we could run into a replacement failure with the inserts and removals caused by compaction. I don't think that is valid either, because the two threads would download the *same* instance of RemoteLayer, so they would wait on the same semaphore, and either of those would win and do the replacement, then close the semaphore.

(The "both wait on same semaphore" issue has been overlooked on a number of on-demand related guesses for replacement failures.)

koivunej:
Now suspected failure to replace layer happened in production while migrating a tenant, which caused a retry loop which killed the logging with:

```
2023-02-15T12:39:28.295686Z  INFO synthetic_size_worker:calculate_synthetic_size{tenant_id=X}:gather_size_inputs{tenant_id=X}:download_remote_layer{tenant_id=X timeline_id=Y layer=000000067F00004002000004EB0000000005-030000000000000000000000000000000002__00000000D2B03F89-00000000D2E79FE9}: download of layer has already finished
```

This comes from the semaphore being closed, which will be looped forever after one or more threads discover this RemoteLayer still in LayerMap (commit unrelated): https://github.com/neondatabase/neon/blob/a974602f9f4311d6b1e778d31e5c2ea17aacebf8/pageserver/src/tenant/timeline.rs#L3496

I suspect it is similar to #3387 in that it happened via `refresh_gc_info` because we have a span on the only message (related fix: #3431).

This was on a version before #3558 fixes (we don't have git commits for production right now, I thought closest would be 248563c5).

@LizardWizzard could you link this issue if you write something up about the production issue?

koivunej:
Looking at the scrapes for the tenant and timeline, this problematic layer must have been the last remote layer. This is something not discovered at least on #3387.

koivunej:
Discussed today with @problame: this shouldn't be an issue now that #3613 is merged. If a replacement fails:
- the `get_reconstruct_data`'s will be restarted
[...]

PR #3544, opened 2023-02-06
Use LayerMap::replace in eviction

koivunej:
Follow-up to #3536, to actually use the new `Debug` in replacing the layers.

Turns out the two paths share a lot of handling of `Replacement` but didn't unify the two (need 3).

Related to #3486.

### DIFF
--- a/pageserver/src/tenant/timeline.rs
+++ b/pageserver/src/tenant/timeline.rs
@@ -930,7 +930,7 @@ impl Timeline {
                 true
             }
             Replacement::NotFound => {
-                debug!(evicted=?local_layer, "lost the race to evict layer");
+                debug!(evicted=?local_layer, "layer was no longer in layer map");
                 false
             }
             Replacement::RemovalBuffered => {
--- a/pageserver/src/tenant/timeline.rs
+++ b/pageserver/src/tenant/timeline.rs
@@ -3412,6 +3412,7 @@ impl Timeline {
                                 error!(
                                     expected.ptr = ?Arc::as_ptr(&l),
                                     other.ptr = ?Arc::as_ptr(&other),
+                                    ?other,
                                     "replacing downloaded layer into layermap failed because another layer was found instead of expected"
                                 );
                             }
--- a/pageserver/src/tenant/timeline.rs
+++ b/pageserver/src/tenant/timeline.rs
@@ -871,7 +871,9 @@ impl Timeline {
     ///
     /// Returns:
     /// - `Ok(Some(true))` when the layer was replaced
-    /// - `Ok(Some(false))` when the layer was found, but is not downloaded
+    /// - `Ok(Some(false))` when the layer was found, but no changes were made
+    ///    - evictee was not yet downloaded
+    ///    - layermap replacement failed
     /// - `Ok(None)` when the layer is not found
     pub async fn evict_layer(&self, layer_file_name: &str) -> anyhow::Result<Option<bool>> {
         let Some(local_layer) = self.find_layer(layer_file_name) else { return Ok(None) };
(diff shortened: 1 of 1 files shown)


pageserver/src/tenant/timeline.rs
```diff
@@ -903,13 +912,43 @@ impl Timeline {
         let gc_lock = self.layer_removal_cs.lock().await;
         let mut layers = self.layers.write().unwrap();
         let mut updates = layers.batch_update();
-        self.delete_historic_layer(&gc_lock, local_layer, &mut updates)?;
-        updates.insert_historic(new_remote_layer);
+
+        let replaced = match updates.replace_historic(&local_layer, new_remote_layer)? {
+            super::layer_map::Replacement::Replaced { .. } => {
```
koivunej, line 917: I'll add a `use super..Replacement` for this.

pageserver/src/tenant/timeline.rs
```diff
@@ -903,13 +912,43 @@ impl Timeline {
         let gc_lock = self.layer_removal_cs.lock().await;
         let mut layers = self.layers.write().unwrap();
         let mut updates = layers.batch_update();
-        self.delete_historic_layer(&gc_lock, local_layer, &mut updates)?;
-        updates.insert_historic(new_remote_layer);
+
+        let replaced = match updates.replace_historic(&local_layer, new_remote_layer)? {
```
koivunej, line 916: What do you think, should there be a comment about the `?`? That'd be a precondition failure once more, which I don't think we even be able to have -- basically meaning that right hand side lines 892..910 would be buggy. It's not an assertion because we have taken locks.

knizhnik: Franking speaking I have to completely rewrite this replacement logic in  #3348.
If we are really going to somehow address the problem with storage ballooning caused with excessive image layers generation (#2948) then lets first do something with this PR before doing some refactoring.

koivunej: Oki, did not see this work colliding... I'll need to look into your PR. Please note this is not just refactoring or prettying up, it's about making the replacement across locks precise.

koivunej: @knizhnik Note we've been worried about a race condition in relation to this from #3387, which I was planning to work on next after this PR: #3533.

koivunej:
I think this is ready to be merged, it's a quite minor changing of the new endpoint, don't see any issues arising because of it.

SomeoneToIgnore (approved):

problame (approved):
Small nits only, LGTM

pageserver/src/tenant/timeline.rs
```diff
@@ -903,13 +914,43 @@ impl Timeline {
...
+                }
+
+                if let Some(layer_size) = layer_size {
+                    self.metrics.resident_physical_size_gauge.sub(layer_size);
+                }
+
+                true
+            }
+            Replacement::NotFound => {
+                debug!(evicted=?local_layer, "lost the race to evict layer");
```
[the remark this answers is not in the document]
koivunej: Well, we did search for the layer, and not it didn't get replaced so someone replaced it before us right? But you are right, two evicts doing the same operation at the same time should result in `Unexpected` on the other, winner gets `Replaced`..

So something like:

```suggestion
                debug!(evicted=?local_layer, "layer was no longer in layer map");
```

Or do you have a better idea?

pageserver/src/tenant/timeline.rs
```diff
@@ -903,13 +914,43 @@ impl Timeline {
...
+        let replaced = match updates.replace_historic(&local_layer, new_remote_layer)? {
+            Replacement::Replaced { .. } => {
+                let layer_size = local_layer.file_size();
+
+                if let Err(e) = local_layer.delete() {
+                    error!("failed to remove layer file on evict after replacement: {e:#?}");
+                }
+
+                if let Some(layer_size) = layer_size {
+                    self.metrics.resident_physical_size_gauge.sub(layer_size);
```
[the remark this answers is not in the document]
koivunej: I'd like to just fire events off of these accepted commands (evict layer X), handle them separatedly, I think that'd be only good way to handle orthogonal concerns like metrics and stats. There's a limit how much we can safely push code to a single path, and I think we are already past it just by looking at how many lines there are in `tenant.rs` and `timeline.rs`...

pageserver/src/tenant/timeline.rs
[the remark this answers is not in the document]
problame: That looks good

pageserver/src/tenant/timeline.rs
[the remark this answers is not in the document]
problame: Ok, let's merge it as is, then.

PR #3549, opened 2023-02-06
Create folder for file_cache

vadim2404:
## Describe your changes

This folder is used for file cache

On VMs: it will be mounted to tmpfs
On pods: local disk will be used

neondatabase/cloud#3828

- [x] If it is a core feature, I have added thorough tests.
- [x] Do we need to implement analytics? if so did you add the relevant metrics to the dashboard?
- [x] If this PR requires public announcement, mark it with /release-notes label and add several sentences in this section.

### DIFF
--- a/Dockerfile.compute-node
+++ b/Dockerfile.compute-node
@@ -228,7 +228,9 @@ RUN mkdir /var/db && useradd -m -d /var/db/postgres postgres && \
     mkdir /var/db/postgres/compute && mkdir /var/db/postgres/specs && \
     chown -R postgres:postgres /var/db/postgres && \
     chmod 0750 /var/db/postgres/compute && \
-    echo '/usr/local/lib' >> /etc/ld.so.conf && /sbin/ldconfig
+    echo '/usr/local/lib' >> /etc/ld.so.conf && /sbin/ldconfig && \
+    # create folder for file cache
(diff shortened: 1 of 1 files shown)


cicdteam (approved):

Dockerfile.compute-node
```diff
@@ -251,7 +251,8 @@ RUN apt update &&  \
         libsfcgal1 \
         gdb && \
     rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* && \
-    localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8
+    localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8 && \
+    mkdir -p -m 777 /neonvm/cache
```
sharnoff, line 255: Given that something related to NeonVM is a little unexpected here, I'd like to see a comment here about why it's `/neonvm`. Maybe something along the lines of:

> The control plane creates compute node VMs with a tmpfs mount at `/neonvm/cache`. Manually creating this directory allows a fallback when not running in a VM.

sharnoff (approved):
Manually adding a `/neonvm` path seems fishy, but if it works it works. Ideally we'd have a different path for non-VM compute nodes, but it seems like doing that wouldn't be trivial.

PR #3540, opened 2023-02-03
script to scrape layer access stats

problame:
How to use it on a pageserver, or, a machine that is running pageserver through `neon_local`

1. Create a postgres database from somewhere (not the pageserver you want to scrape). E.g., neon.tech.
2. Prepare a `scraper.env`, based on `scraper.env.example`
3. Perform the following steps

```
 $ source scraper.env
 # load schema into your db
 $ ./pg_install/v14/bin/psql < scraper.db.schema.sql
 # start the scraper and watch log output
 $ poetry run python scraper.py ALL
```

Then watch the data pile up in the database, and do something useful with it in the future (TM).

PR #3552, opened 2023-02-06
automatic layer eviction

problame:
```
automatic layer eviction

This patch adds a per-timeline periodic task that executes an eviction
policy. The eviction policy is configurable per tenant.

Two policies exist:
- NoEviction (the default one)
- LayerAccessThreshold

The LayerAccessThreshold policy examines the last access timestamp per
layer in the layer map and evicts the layer if that last access is
further in the past than a configurable threshold value.
This policy kind is evaluated periodically at a configurable period.
It logs a summary statistic at `info!()` or `warn!()` level, depending
on whether any evictions failed.

This feature has no explicit killswitch since it's off by default.
```

---

Try out with `neon_local` by putting the following into `pageserver.toml`

```
[tenant_config]
eviction_policy = { kind = "LayerAccessThreshold" , period = "2s", threshold = "10s" }
```


problame:
NB: the stats reset feature conflicts with this one: if we continuously reset the stats, then the eviction loop has no last access date to base the decision on. Need to figure out what to do here, reset is quite useful in #3540

pageserver/src/tenant/timeline.rs
```diff
@@ -1135,6 +1158,162 @@ impl Timeline {
...
+            }
+        }
+    }
+
+    async fn eviction_iteration_threshold(self: &Arc<Self>, p: EvictionPolicyLayerAccessThreshold) {
+        let now = SystemTime::now();
+        let remote_client = match self.remote_client.as_ref() {
+            None => {
+                error!("no remote storage configured but eviction requires that");
+                // check again in 1 hour; XXX config watch mechanism
```
koivunej, line 1222: Restart is the only way to reconfigure pageserver.

pageserver/src/tenant/timeline.rs
```diff
@@ -1135,6 +1158,162 @@ impl Timeline {
...
+            candidates
+        };
+        stats.candidates = candidates.len();
+        for l in candidates {
+            // XXX would be good to batch the evictions.
+            // The way it is now, we have to rebuild the layer map after every eviction.
+            // Idea for this:
+            // - a layer should have an 'evicting' state. We set it here.
+            // - If GC / Compaction find such a layer, they skip it or otherwise
+            //   guarantee to not remove it from the layer map until 'evicting' state is over.
```
koivunej, line 1281: Why this instead of the lock which is currently being used?

pageserver/src/tenant/timeline.rs
```diff
@@ -869,10 +872,23 @@ impl Timeline {
 
     pub async fn evict_layer(&self, layer_file_name: &str) -> anyhow::Result<Option<bool>> {
         let Some(local_layer) = self.find_layer(layer_file_name) else { return Ok(None) };
+        let Some(remote_client) = &self.remote_client else { return Ok(Some(false)) };
```
koivunej, line 875: I made this into an error in #3544 ...

pageserver/src/tenant/timeline.rs
problame: Actually I think my performance conerns are unfounded with `LayerMap::replace`, aren't they?

Anyways, batch API seems useful. WDYT: https://github.com/neondatabase/neon/pull/3554 ?

koivunej: Yeah the batching impl looked exactly like I was thinking, forgot to add any positive comments on the PR before you merged it :)

pageserver/src/tenant/timeline.rs
koivunej: With this I meant that perhaps just error out quickly, or `INFO` out quickly (for tests, local usage)

pageserver/src/tenant/timeline.rs
```diff
@@ -869,17 +872,71 @@ impl Timeline {
...
+        let layer_removal_guard = self.layer_removal_cs.lock().await;
+
+        // start the batch update
+        let mut layer_map = self.layers.write().unwrap();
+        let mut batch_updates = layer_map.batch_update();
+
+        let mut results = Vec::with_capacity(layers_to_evict.len());
+
+        for l in layers_to_evict.iter() {
+            let res = if task_mgr::is_shutdown_requested() {
```
koivunej, line 913: While this right now reads a cancellation token, I do feel a bit uneasy with it.

```rust
let expected = layers_to_evict.len();

results.extend(
  layers_to_evict
    .into_iter()
    .take_while(|_| task_mgr::is_shutdown_requested())
    .map(|l| self.evict_layer_batch_impl(&guard, l, &mut batch))
    .map(Some));

// fill in the blanks
results.extend(std::iter::repeat(None).take(expected - results.len()));
```

This would read the `false` from `is_shutdown_requested` only once.

pageserver/src/tenant/timeline.rs
```diff
@@ -1135,6 +1193,160 @@ impl Timeline {
...
+            match result {
+                None => {
+                    stats.skipped_for_shutdown += 1;
+                }
+                Some(Ok(true)) => {
+                    debug!("evicted layer {}", l.short_id());
+                    stats.evicted += 1;
+                }
+                Some(Ok(false)) => {
+                    debug!("layer is not evictable: {}", l.short_id());
```
koivunej, line 1330: Also now there's the `Debug` impls after rebase.

pageserver/src/tenant/timeline.rs
koivunej: Though ... I guess the functionality without a remote client could be to just "maintain the set of would-be-evicted".

koivunej (approved):
I think this is looking good! So the minimum value for unused is the periodic e2e check interval, and ... That's it? What's missing for this to be put very aggressively into staging?

Well, could move all of the eviction stuff into `tenant/timeline/eviction.rs` for example.

pageserver/src/tenant/timeline.rs
problame: Hm, that re-orders the results though.
Nothing relies on it right now, but still.
Let's take this discussion to https://github.com/neondatabase/neon/pull/3557

problame:
The last pushes should address all review comments.

I'll rebase this PR on top of #3557  once that's merged.

problame:
> NB: the stats reset feature conflicts with this one: if we continuously reset the stats, then the eviction loop has no last access date to base the decision on. Need to figure out what to do here, reset is quite useful in https://github.com/neondatabase/neon/pull/3540

Addressed this by duplicating the LayerAccessStatsInner.
See commit
[maintain stats twice, once for api, once for eviction policy](https://github.com/neondatabase/neon/pull/3552/commits/8185760d7573ba5e3a8a54b4360362e181a283da)

> Well, could move all of the eviction stuff into tenant/timeline/eviction.rs for example.

Good idea, did that.

problame:
Need to figure out how to implement tenant config update via API before we can merge this.

control_plane/src/pageserver.rs
```diff
@@ -419,6 +419,11 @@ impl PageServerNode {
                     .map(|x| x.parse::<bool>())
                     .transpose()
                     .context("Failed to parse 'trace_read_requests' as bool")?,
+                eviction_policy: settings
+                    .get("eviction_policy")
+                    .map(|x| serde_json::from_str(x))
+                    .transpose()
+                    .context("Failed to parse 'eviction_policy' json")?,
```
problame, line 426: Ok to require user to pass JSON here?

pageserver/src/config.rs
```diff
@@ -731,6 +731,13 @@ impl PageServerConf {
                 })?);
         }
 
+        if let Some(eviction_policy) = item.get("eviction_policy") {
+            t_conf.eviction_policy = Some(
+                toml_edit::de::from_item(eviction_policy.clone())
```
problame, line 736: The other config keys are deserialized using hand-rolled FromStr's. Don't understand why.
Is this acceptable?
For the alternative hand-rolled variant, see the left-hand-side of https://github.com/neondatabase/neon/pull/3552/commits/15fb8a3824150d9c7c6b0c833f89603a52d2cb0b

pageserver/src/tenant/timeline/eviction_task.rs
```diff
@@ -0,0 +1,198 @@
...
+                    warn!("failed to evict layer {l:?}: {e:?}");
+                    stats.errors += 1;
+                }
+            }
+        }
+        if stats.not_considered_due_to_clock_skew > 0 || stats.errors > 0 || stats.not_evictable > 0
+        {
+            warn!(stats=?stats, "eviction iteration complete");
+        } else {
+            info!(stats=?stats, "eviction iteration complete");
```
problame, line 194: debug, or `info!()` only if any of the numbers is non-zero?

pageserver/src/tenant/timeline.rs
koivunej: Taken, discussed, it did not reorder, but for loop was kept due to this being quite unreadable and semi-misusing map.

pageserver/src/config.rs
koivunej: Note here we are deserializing it as toml, though I cannot find docs for `toml_edit::de::from_item`, nor the fn from their repository. Aah ok: https://docs.rs/toml_edit/0.17.1/toml_edit/de/fn.from_item.html (we are using a bit older version).

I think it makes sense to rely on serde here. I understand in the past this hasn't been done to give more clearer errors, because by default serde errors have not been helpful.

How are the errors looking for this one?

control_plane/src/pageserver.rs
koivunej: So this is json within json, why? Aah, unsure if I've ever used this neon_local functionlity. Does seem weird but what are the alternatives?

libs/pageserver_api/src/models.rs
```diff
@@ -155,6 +155,7 @@ pub struct TenantConfigRequest {
     pub lagging_wal_timeout: Option<String>,
     pub max_lsn_wal_lag: Option<NonZeroU64>,
     pub trace_read_requests: Option<bool>,
+    pub eviction_policy: Option<serde_value::Value>,
```
koivunej, line 158: Why is this a dom object?

Related: serde_json::Value is probably the same as serde_value.

pageserver/src/tenant/storage_layer.rs
```diff
@@ -92,7 +93,23 @@ pub enum ValueReconstructResult {
...
+/// The `for_eviction_policy` is never reset.
+#[derive(Debug, Default, Clone)]
+struct LayerAccessStatsLocked {
+    for_scraping_api: LayerAccessStatsInner,
+    for_eviction_policy: LayerAccessStatsInner,
+}
+
+impl LayerAccessStatsLocked {
+    fn iter_mut(&mut self) -> impl Iterator<Item = &mut LayerAccessStatsInner> {
+        vec![&mut self.for_scraping_api, &mut self.for_eviction_policy].into_iter()
```
koivunej, line 110: This allocation is needless. You can get the same with
```suggestion
        [&mut self.for_scraping_api, &mut self.for_eviction_policy].into_iter()
```

Or should at least, didn't test this. Might need to add a named lifetime, or not.

pageserver/src/tenant/timeline.rs
```diff
@@ -867,30 +869,98 @@ impl Timeline {
...
+            Some(Ok(b)) => Ok(Some(b)),
+            Some(Err(e)) => Err(e),
         }
+    }
 
-        // ensure the current layer is uploaded for sure
-        self.remote_client
-            .as_ref()
-            .ok_or_else(|| anyhow::anyhow!("remote storage not configured; cannot evict"))?
+    /// Try to evict the given `layers_to_evict` by
```
koivunej, line 894:
```suggestion
    /// Evict multiple layers at once, continuing through errors.
    ///
    /// Try to evict the given `layers_to_evict` by
```

Minor rustdocification by giving the summary on the first line.

pageserver/src/tenant/timeline/eviction_task.rs
```diff
@@ -0,0 +1,198 @@
...
+
+        let remote_client = match self.remote_client.as_ref() {
+            None => {
+                error!("no remote storage configured but eviction requires that");
+                // Break lhe eviction task loop: remote storage is not runtime-configurable,
+                // so, it can never become Some().
+                return ControlFlow::Break(());
+            }
+            Some(c) => c,
+        };
```
koivunej, line 108: Did you consider that we could just skip the actual eviction, but list "would evict: {candidates:#?}" in this case?

pageserver/src/tenant/timeline/eviction_task.rs
koivunej: Not too fond of these but perhaps it makes sense in the first attempt to get comfort it completes, so as is with `info!`.

pageserver/src/tenant/storage_layer.rs
```diff
@@ -104,10 +121,10 @@ struct LayerAccessStatsInner {
 }
 
 #[derive(Debug, Clone)]
```
koivunej, line 123: Might be worthwhile to add `#[derive(Copy)]` to these which are always memcpyable.

koivunej (commented):
First round of questions, minor suggestions.

control_plane/src/pageserver.rs
koivunej: Understood the point raised here a bit late and started wondering if neon_local should at *all* parse the payload, except to check that it's JSON.

However using from_str here might not be great; better would be `Box<serde_json::value::RawValue>` then this property would not have to be a string containing json in a json document but just a arbitrary json subtree string. If we do have full control of the object, this could also be `#[serde(borrowed)] name: &'a RawValue`.

pageserver/src/tenant/storage_layer.rs
problame: https://github.com/neondatabase/neon/pull/3574

libs/pageserver_api/src/models.rs
problame: So, the (wrong) reasoning behind this was that a `serde_json::Value` would only deserialize JSON.
Of course that's wrong, as demonstrated in this playground:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b5352d8d94a429e74dc6fe981ef6454b

However, I don't understand why the `serde` crate doesn't provide such a type.

problame:
For some reason GitHub doesn't show a comment box for some of your comments, so, I'll reply here:

> So this is json within json, why? Aah, unsure if I've ever used this neon_local functionlity. Does seem weird but what are the alternatives?

Well, I don't want to spend time on implementing a CLI-friendly representation of the eviction policy variants here. IMO there should be a `serde_cli` crate that standardizes such stuff.

Any objections with going forward with JSON?

control_plane/src/pageserver.rs
problame: Agree in principle on the `RawValue` proposal, but, neon_local uses the `pageserver_api` models directly here, and I think having the proper type in the API model is useful.

So, not doing that in this PR

libs/pageserver_api/src/models.rs
koivunej: Oki but I think the why is that "this type is in pageserver_api and the real type is not" right? IMHO should be a comment, is the plan to expose the one type but not now?

pageserver/src/config.rs
problame: Looked fine. I tried with a wrong `kind`, and it even spit out the list of valid `kind`s.

control_plane/src/pageserver.rs
koivunej: Aah ok this is a darker corner I haven't yet ran into :) Yes agreed.

> Well, I don't want to spend time on implementing a CLI-friendly representation of the eviction policy variants here. IMO there should be a serde_cli crate that standardizes such stuff.

yeah this is a weird ... middle ground. Didn't check, but I'm pretty sure our argv parser *could* be hooked to parse additional parts as serde deserialized values by turning them into `HashMap<&str, &str>` which we already have but perhaps ... yeah. Well, we can change this in future.

This could use a comment for the above reasoning just as well, so the picking up would be easier.

pageserver/src/config.rs
koivunej: Yes for that it should work nicely ... Can't remember for which it didn't work. Resolving!

koivunej (approved):

problame:
Force-push: squash & proper commit message (updated top-level PR description as well)

PR #3554, opened 2023-02-07
batch api for eviction

problame:
ref https://github.com/neondatabase/neon/pull/3552#discussion_r1098268019 

What do you think, @koivunej ?

### DIFF
--- a/pageserver/src/tenant/timeline.rs
+++ b/pageserver/src/tenant/timeline.rs
@@ -873,28 +873,58 @@ impl Timeline {
     pub async fn evict_layer(&self, layer_file_name: &str) -> anyhow::Result<Option<bool>> {
         let Some(local_layer) = self.find_layer(layer_file_name) else { return Ok(None) };
         let Some(remote_client) = &self.remote_client else { return Ok(Some(false)) };
-        self.evict_layer_impl(local_layer, remote_client)
-            .await
-            .map(|res| Some(res))
+        let results = self
+            .evict_layer_batch(remote_client, &[local_layer])
+            .await?;
+        assert_eq!(results.len(), 1);
+        let result: Option<anyhow::Result<bool>> = results.into_iter().next().unwrap();
+        match result {
+            None => anyhow::bail!("task_mgr shutdown requested"),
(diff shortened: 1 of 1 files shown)


pageserver/src/tenant/timeline.rs
```diff
@@ -922,14 +952,12 @@ impl Timeline {
             ),
         });
 
-        let gc_lock = self.layer_removal_cs.lock().await;
         let mut layers = self.layers.write().unwrap();
         let mut updates = layers.batch_update();
-        self.delete_historic_layer(&gc_lock, local_layer, &mut updates)?;
+        self.delete_historic_layer(layer_removal_cs, Arc::clone(local_layer), &mut updates)?;
         updates.insert_historic(new_remote_layer);
```
koivunej, line 958: If you rebase, you should see the `replace` now being used here in a conflict.

pageserver/src/tenant/timeline.rs
```diff
@@ -873,29 +873,59 @@ impl Timeline {
...
+            } else {
+                Some(self.evict_layer_batch_impl(&layer_removal_guard, l).await)
+            };
+            results.push(res);
+        }
+
+        Ok(results)
+    }
+
+    async fn evict_layer_batch_impl(
```
koivunej, line 920: This is no longer async right?

IMHO the `layers` could be kept locked as well, since the other lock is already. Also, only one batch should happen. Or perhaps there's a reason why multiple batches?

problame: > This is no longer async right?

Right. Will remove it for now and keep using the synchronous unlink code.

> IMHO the layers could be kept locked as well, since the other lock is already.Also, only one batch should happen. Or perhaps there's a reason why multiple batches?

Yeah. Both were oversights from my side, sorry about that.

pageserver/src/tenant/timeline.rs
problame: Yeah, right now this branch targets #3552 but I think I'll spin it out once your #3513 is merged.

PR #3557, opened 2023-02-07
refactor: allow for eviction of layers in a batch

problame:
refactor: allow for eviction of layers in a batch

The auto-eviction PR (#3552) operates in two phaes:

  1. find candidate layers
  2. evict them.

For (2), a batch API like the one added in this commit is useful.

Note that this PR requires #3558 to be merged first.
Otherwise, the tests won't pass.

### DIFF
--- a/pageserver/src/tenant/timeline.rs
+++ b/pageserver/src/tenant/timeline.rs
@@ -867,29 +867,97 @@ impl Timeline {
         Ok(Some(true))
     }
 
-    /// Evicts one layer as in replaces a downloaded layer with a remote layer
-    ///
-    /// Returns:
-    /// - `Ok(Some(true))` when the layer was replaced
-    /// - `Ok(Some(false))` when the layer was found, but no changes were made
-    ///    - evictee was not yet downloaded
-    ///    - layermap replacement failed
-    /// - `Ok(None)` when the layer is not found
+    /// Like [`evict_layer_batch`], but for just one layer.
+    /// Additional case `Ok(None)` covers the case where the layer could not be found by its `layer_file_name`.
     pub async fn evict_layer(&self, layer_file_name: &str) -> anyhow::Result<Option<bool>> {
-        use super::layer_map::Replacement;
-
         let Some(local_layer) = self.find_layer(layer_file_name) else { return Ok(None) };
-        if local_layer.is_remote_layer() {
-            return Ok(Some(false));
-        }
-
-        // ensure the current layer is uploaded for sure
-        self.remote_client
+        let remote_client = self
+            .remote_client
             .as_ref()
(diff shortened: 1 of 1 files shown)


pageserver/src/tenant/timeline.rs
```diff
@@ -867,30 +867,98 @@ impl Timeline {
...
+        let mut results = Vec::with_capacity(layers_to_evict.len());
+
+        for l in layers_to_evict.iter() {
+            let res = if cancel.is_cancelled() {
+                None
+            } else {
+                Some(self.evict_layer_batch_impl(&layer_removal_guard, l, &mut batch_updates))
+            };
+            results.push(res);
+        }
```
koivunej, line 939: What do you mean my earlier suggestion reorders?

```rust
let expected = layers_to_evict.len();

results.extend(
  layers_to_evict
    .iter()
    .take_while(|_| task_mgr::is_shutdown_requested())
    .map(|l| self.evict_layer_batch_impl(&guard, l, &mut batch))
    .map(Some));

// fill in the blanks
results.extend(std::iter::repeat(None).take(expected - results.len()));
```

As in I cannot see how it will reorder?

problame: Do you think this `.extend()`business is more readable than the for-loop?
I don't, but I don't have strong feelings about it

koivunej (approved):
This is good as we do this with single rebuild and all, my refactoring suggestion isn't critical, but I would like to why does it not work.

pageserver/src/tenant/timeline.rs
```diff
@@ -867,30 +867,98 @@ impl Timeline {
...
-
         let Some(local_layer) = self.find_layer(layer_file_name) else { return Ok(None) };
-        if local_layer.is_remote_layer() {
-            return Ok(Some(false));
+        let remote_client = self
+            .remote_client
+            .as_ref()
+            .ok_or_else(|| anyhow::anyhow!("remote storage not configured; cannot evict"))?;
+
+        let cancel = CancellationToken::new();
```
koivunej, line 879: hmm if this were to be made into `drop_guard` this would be useful, right now it's never cancelled.

pageserver/src/tenant/timeline.rs
koivunej: Not really but it wouldn't query the cancellation after getting one true, apart from that's quite horrible with the map...

But does it reorder?

problame: No, it doesn't reorder. I don't think I was very attentive when I said that

problame:
Your refactoring suggestion _does_ work, I was wrong about the reordering.

Re drop_guard: we confirmed on Slack that it isn't actually helpful here, because it's all the same future.

PR #3566, opened 2023-02-08
fix f64 counter value -> u64 metric type conversion

lubennikovaav:

### DIFF
--- a/proxy/src/metrics.rs
+++ b/proxy/src/metrics.rs
@@ -82,7 +82,7 @@ pub fn gather_proxy_io_bytes_per_client() -> Vec<(Ids, (u64, DateTime<Utc>))> {
                         .find(|l| l.get_name() == "endpoint_id")
                         .unwrap()
                         .get_value();
-                    let value = ms.get_counter().get_value() as u64;
+                    let value = ms.get_counter().get_value().round() as u64;
 
                     debug!("endpoint_id:val - {}: {}", endpoint_id, value);
                     current_metrics.push((


lubennikovaav:
proxy [failed](https://neonprod.grafana.net/explore?orgId=1&left=%7B%22datasource%22:%22grafanacloud-logs%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22grafanacloud-logs%22%7D,%22editorMode%22:%22code%22,%22expr%22:%22%7Bneon_service%3D%5C%22proxy-scram%5C%22%7D%20%7C%3D%20%60metrics%20endpoint%20refused%20the%20sent%20metrics%60%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22from%22:%221675321166500%22,%22to%22:%221675322382500%22%7D%7D) to send metric values

`18446744073707869856` and `18446744073709536698`

These values are abnormal.
VictoriaMetrics has correct values, so I suppose the problem is somewhere in the metrics.rs code.

I haven't found anything that could cause the problem, except for this line.

Now, that I think about it, this fix doesn't make sense, because `round()`  doesn't change anything about large values [truncation](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0358061fcca8d81de89dc02305783896).

koivunej:
> Now, that I think about it, this fix doesn't make sense, because round() doesn't change anything about large values [truncation](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0358061fcca8d81de89dc02305783896).

Agreed with the conclusion here.

These numbers seem 64 bit, which is beyond the safe range for integer precision meaning they have less than 1 integer precision, limit was in 2^53 or so. It seems that this counter or whatever should had been an u64 counter perhaps?

PR #3569, opened 2023-02-08
[proxy] Fix possible unsoundness in the websocket machinery

funbringer:
This PR replaces the ill-advised `unsafe Sync` impl with a de-facto standard way to solve the underlying problem.

TLDR:
- tokio::task::spawn requires future to be Send
- ∀t. (t : Sync) <=> (&t : Send)
- ∀t. (t : Send + !Sync) => (&t : !Send)

### DIFF
--- a/proxy/src/http/websocket.rs
+++ b/proxy/src/http/websocket.rs
@@ -1,161 +1,134 @@
+use crate::{
+    cancellation::CancelMap, config::ProxyConfig, error::io_error, proxy::handle_ws_client,
+};
 use bytes::{Buf, Bytes};
 use futures::{Sink, Stream, StreamExt};
-use hyper::server::accept;
-use hyper::server::conn::AddrIncoming;
-use hyper::upgrade::Upgraded;
-use hyper::{Body, Request, Response, StatusCode};
-use hyper_tungstenite::{tungstenite, WebSocketStream};
-use hyper_tungstenite::{tungstenite::Message, HyperWebsocket};
+use hyper::{
+    server::{accept, conn::AddrIncoming},
+    upgrade::Upgraded,
+    Body, Request, Response, StatusCode,
+};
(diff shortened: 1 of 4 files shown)


proxy/src/http/websocket.rs
```diff
@@ -1,161 +1,134 @@
...
+            }
+
+            let res = ready!(this.stream.as_mut().get_pin_mut().poll_next(cx));
+            match res.transpose().map_err(io_error)? {
+                Some(message) => match message {
+                    Message::Ping(_) => {}
+                    Message::Pong(_) => {}
+                    Message::Text(text) => {
+                        // We expect to see only binary messages.
+                        warn!("unexpected text: {text}");
```
petuhovskiy, line 113: Anyone can connect to our public websocket server and write arbitrary text data, consider changing logs to `debug!` level.

petuhovskiy (approved):
Tested locally, it works. One day we need to add WS tests with serverless driver, but that's not trivial.

proxy/src/http/websocket.rs
koivunej: Would it make sense to drop the connection for invalid message right away?

funbringer: I also think that we should drop connection in that case.

The idea behind `warn!` instead of `debug!` is that receiving a text playload is an example of abnormal behavior, which could mean that we either didn't implement the protocol correctly (i.e. there are non-trivial messages which we drop without even a warning or explanation) or the user is trying to do something unusual (like sending unexpected no-op payloads). We won't be able to see those `debug!` logs, which makes them pointless. Granted, I wouldn't like to have millions of user-controlled log lines, so maybe this should be a metric instead.

koivunej (approved):
I think this looks good.

LizardWizzard (approved):
LGTM

petuhovskiy (approved):

PR #3573, opened 2023-02-09
fix: one leftover Arc::ptr_eq

koivunej:
@knizhnik noticed that one instance of `Arc::<dyn PersistentLayer>::ptr_eq` was missed in #3558.

Now all `ptr_eq` which remain are in comments.

### DIFF
--- a/pageserver/src/tenant/layer_map.rs
+++ b/pageserver/src/tenant/layer_map.rs
@@ -731,6 +731,9 @@ where
         Ok(())
     }
 
+    /// Similar to `Arc::ptr_eq`, but only compares the object pointers, not vtables.
+    ///
+    /// Returns `true` if the two `Arc` point to the same layer, false otherwise.
     #[inline(always)]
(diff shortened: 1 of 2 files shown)


koivunej:
Remaining are:

```
pageserver/src/tenant/layer_map.rs
301:            // this assertion is related to use of Arc::ptr_eq in Self::compare_arced_layers,
734:    /// Similar to `Arc::ptr_eq`, but only compares the object pointers, not vtables.
744:        // `Arc::ptr_eq` as of writing (at least up to 1.67) uses a comparison where both the

pageserver/src/tenant/layer_map/historic_layer_coverage.rs
436:    /// against some expectation. This allows to use `Arc::ptr_eq` or similar which would be
```

As I've checked a few times now. I never do remember search all in the first place, as I thought I had already dedupped the callsites when introducing `compare_arced_layers`.

shanyp (approved):

PR #3575, opened 2023-02-09
fix: readable Debug for Layers

koivunej:
#3536 added the custom Debug implementations but it using derived Debug on Key lead to too verbose output. Instead of making `Key`'s `Debug` unconditionally or conditionally do the `Display` variant (for table space'd keys), opted to build a newtype to provide `Debug` for `Range<Key>` via `Display` which seemed to work unconditionally.

Also orders Key to have:
1. comment
2. derive
3. `struct Key`

### DIFF
--- a/pageserver/src/repository.rs
+++ b/pageserver/src/repository.rs
@@ -7,11 +7,11 @@ use std::fmt;
 use std::ops::{AddAssign, Range};
 use std::time::Duration;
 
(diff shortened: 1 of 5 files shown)


problame (approved):

koivunej:
I don't think this needs many more reviewers since it fixes an omission I did recently, nor were too many interested in debug, it's only used in eviction and layermap::replace.

PR #3577, opened 2023-02-10
Run compute_ctl in a cgroup in VMs

sharnoff:
Not expected to work without <https://github.com/neondatabase/neonvm/pull/41> and <https://github.com/neondatabase/autoscaling/pull/40>.

Related work: <https://github.com/neondatabase/autoscaling/pull/41>.

---

I know this _mostly_ looks like unrelated changes. The basic idea is: It's easier to manage cgroups and the vm-informant outside of compute_ctl. With a little bit of added functionality to NeonVM, it's possible to implement this pretty simply, while keeping compute_ctl running as user `postgres` (instead of e.g., root, to manage cgroups).

~~This PR is mostly intended to be exploratory - as an alternative to writing a custom vm-builder for this repo. IMO this approach will be simpler and easier to maintain.~~

_edit:_ We decided to go with this approach for now.

---

The first two commits are from the previous approach I used.


Dockerfile.vm-compute-node
```diff
@@ -0,0 +1,31 @@
...
+# Pull VM informant and set up inittab
+FROM neondatabase/vm-informant:$VM_INFORMANT_VERSION as informant
+
+RUN set -e \
+	&& rm -f /etc/inittab \
+	&& touch /etc/inittab
+
+ADD vm-cgconfig.conf /etc/cgconfig.conf
+RUN echo "::sysinit:cgconfigparser -l /etc/cgconfig.conf -s 1664" >> /etc/inittab
+RUN echo "::respawn:su vm-informant -c '/bin/vm-informant --auto-restart --cgroup=neon-postgres"
```
tychoish, line 15: fix the thing

tychoish (approved):

Dockerfile.vm-compute-node
sharnoff: thing fixed :)

Dockerfile.vm-compute-node
```diff
@@ -0,0 +1,31 @@
...
+# Pull VM informant and set up inittab
+FROM neondatabase/vm-informant:$VM_INFORMANT_VERSION as informant
+
+RUN set -e \
+	&& rm -f /etc/inittab \
+	&& touch /etc/inittab
+
+ADD vm-cgconfig.conf /etc/cgconfig.conf
+RUN echo "::sysinit:cgconfigparser -l /etc/cgconfig.conf -s 1664" >> /etc/inittab
+RUN echo "::respawn:su vm-informant -c '/bin/vm-informant --auto-restart --cgroup=neon-postgres" >> /etc/inittab
```
vadim2404, line 15: Do you need two different layers?

sharnoff: Nope. Merged them

sharnoff:
Merging this, will monitor staging to check it works ok.

Issue #3571, opened 2023-02-08
Fix synthetic tenant size calculation in trivial (1 branch) case


lubennikovaav:
## Steps to reproduce

This can be observed on this [dashboard](https://neonprod.grafana.net/d/dcdAph04z/metrics-playground?orgId=1&var-data_source=victoria-metrics-aws-dev&var-timeline_id=f15ae0cf21cce2ba27e4d80c6709a6cd&from=1675274810190&to=1675879610190&viewPanel=9).

NOTE: graph scale may the problem 

## Expected result

Synthetic size is calculated as described in the [documentation](https://github.com/neondatabase/neon/pull/3328/files)

TLDR:
```
In the simple case that the project contains just one branch (main),
and a fixed PITR horizon, the synthetic size is the sum of:

- the logical size of the branch *at the beginning of the PITR
  horizon*, i.e. at the oldest point that you can still recover to, and
- the size of the WAL covering the PITR horizon.

                             WAL
   -----------------------#########>
                          ^
                       snapshot
Legend:
  ##### PITR horizon. This is the region that you can still access
        with Point-in-time query and you can still create branches
        from.
  ----- history that has fallen out of the PITR horizon, and can no
        longer be accessed
```

## Actual result

Synthetic size always returns result that matches "last_lsn"

## Environment

all envs

## Logs, links

`investigate_synthetc_size` branch contains unit test and debug outputs, that illustrate the issue.

cc: @vadim2404 , @koivunej , @hlinnaka

PR #3586, opened 2023-02-12
Add earlier update to the logical size metric

shanyp:
## Describe your changes
Add early as possible update to the time line logical size metric 
#3571

### DIFF
--- a/pageserver/src/pgdatadir_mapping.rs
+++ b/pageserver/src/pgdatadir_mapping.rs
@@ -524,7 +524,9 @@ impl Timeline {
                 total_size += relsize as u64;
             }
         }
-        Ok(total_size * BLCKSZ as u64)
+        let calculated_size = total_size * BLCKSZ as u64;
+        self.metrics.current_logical_size_gauge.set(calculated_size);
+        Ok(calculated_size)
     }
 
     ///
--- a/pageserver/src/tenant/timeline.rs
+++ b/pageserver/src/tenant/timeline.rs
@@ -135,7 +135,7 @@ pub struct Timeline {
     ancestor_timeline: Option<Arc<Timeline>>,
     ancestor_lsn: Lsn,
 
-    metrics: TimelineMetrics,
+    pub metrics: TimelineMetrics,
 
     /// Ensures layers aren't frozen by checkpointer between
     /// [`Timeline::get_layer_for_write`] and layer reads.
@@ -1630,7 +1630,13 @@ impl Timeline {
                     .initial_logical_size
                     .set(calculated_size)
                 {
-                    Ok(()) => (),
+                    Ok(()) => {
+                        self_clone
+                            .metrics
+                            .current_logical_size_gauge
+                            .set(calculated_size);
+                        ()
+                    }
                     Err(existing_size) => {
                         // This shouldn't happen because the semaphore is initialized with 1.
                         // But if it happens, just complain & report success so there are no further retries.
--- a/pageserver/src/tenant/timeline.rs
+++ b/pageserver/src/tenant/timeline.rs
@@ -1635,7 +1635,6 @@ impl Timeline {
                             .metrics
                             .current_logical_size_gauge
                             .set(calculated_size);
-                        ()
                     }
                     Err(existing_size) => {
                         // This shouldn't happen because the semaphore is initialized with 1.


pageserver/src/tenant/timeline.rs
```diff
@@ -1630,7 +1630,12 @@ impl Timeline {
                     .initial_logical_size
                     .set(calculated_size)
                 {
-                    Ok(()) => (),
+                    Ok(()) => {
+                        self_clone
+                            .metrics
+                            .current_logical_size_gauge
+                            .set(calculated_size);
```
lubennikovaav, line 1637: I am afraid this `set` will mess the data if timeline is active.

See
```
 /// Note that we also expose a copy of this value as a prometheus metric,
    /// see `current_logical_size_gauge`. Use the `update_current_logical_size`
    /// to modify this, it will also keep the prometheus metric in sync.
    size_added_after_initial: AtomicI64,
    ```
    
WDYT?

**Review by lubennikovaav — CHANGES_REQUESTED** [review]

File: pageserver/src/tenant/timeline.rs (hunk 1 above)
Comment (lubennikovaav) [reply pageserver/src/tenant/timeline.rs:1637]: BTW, Do we have any tests for logical size?
It would be nice to add some, while we're on it.

Comment (shanyp) [reply pageserver/src/tenant/timeline.rs:1637]: update_current_logical_size takes a delta and not total value

Comment (koivunej) [reply pageserver/src/tenant/timeline.rs:1637]: With this I see only one possible issue; there's a race updating the metric from `Timeline::update_current_logical_size`, so I'd suggest that before setting the `initial_logical_size`, we set the gauge value. This way the two cannot conflict. With this order, we could be receiving wal at the same time as we complete the logical size calculation and we lower the gauge value unintentionally here.

Additionally I think we should take the `size_added_after_initial` into account for the gauge value right away, this way it would not need fixing by receiving wal afterwards.

We should be guarded by the semaphore already so that should be safe, and the `Err` case in this match should be unreachable.

If a new path should be added for the setting the `initial_logical_size` without having acquired the semaphore well all bets are off.

File: pageserver/src/pgdatadir_mapping.rs [hunk 2]
```diff
@@ -524,7 +524,9 @@ impl Timeline {
                 total_size += relsize as u64;
             }
         }
-        Ok(total_size * BLCKSZ as u64)
+        let calculated_size = total_size * BLCKSZ as u64;
+        self.metrics.current_logical_size_gauge.set(calculated_size);
```
koivunej, line 528: This is the wrong place to set the gauge, because this is "calculate non-incremental logical size at any lsn", which is just incidentially called by "calculate initial logical size". It is used by tenant_size (synthetic_size inputs gathering).

koivunej (changes requested):

pageserver/src/tenant/timeline.rs
lubennikovaav: The code here is pretty complex.

`logical_size` = `initial_size` + `size_added_after_initial`

while `initial_size` calculation is in progress (which may take some time, if layers are offloaded to S3), current logical size is `Approximate` and only contains increment part. See `CurrentLogicalSize` for details.

So, let's say we have some timeline 
1. it's initial logical size is 50MB (but we don't know it yet, it is in process of calculation)
2. Meanwhile,  it receives some updates. and has increment 200MB
3.  `current_logical_size_gauge` reports 200
4.  initial size calculation is finished. and we set gauge to `50MB`
5. `current_logical_size_gauge` reports 50 as if we removed something.
While actual logical_size is 250.

I don't think this is a desired behavior.

there is already comment about similar race:
```
        // Also set the value in the prometheus gauge. Note that
        // there is a race condition here: if this is is called by two
        // threads concurrently, the prometheus gauge might be set to
        // one value while current_logical_size is set to the
        // other.
```

please, correct me if I am missing something

koivunej: > there is already comment about similar race:
> 
> ```
>         // Also set the value in the prometheus gauge. Note that
>         // there is a race condition here: if this is is called by two
>         // threads concurrently, the prometheus gauge might be set to
>         // one value while current_logical_size is set to the
>         // other.
> ```
> 
> please, correct me if I am missing something

Well that's a different race, if we'd have two threads ingesting WAL at the same time, but we don't support that, at least it's my understanding; this method is only called from `TimelineWriter` which in turn is only used from `DatadirModification` -- hopefully not possible to have two of those at the same time ... Probably because this is only called from import functionality..

Now that I went through the callers, how exactly does walingest end up incrementing the approx counter?

lubennikovaav: >Now that I went through the callers, how exactly does walingest end up incrementing the approx counter?

check `update_current_logical_size` occurrences in `flush()` and `commit()`

koivunej: > > Now that I went through the callers, how exactly does walingest end up incrementing the approx counter?
> 
> check `update_current_logical_size` occurrences in `flush()` and `commit()`

Yes but `flush` and `commit` are only called from the `import_basebackup_from_tar` and `import_from_postgres_datadir` right?

> So, we do need to update the gauge here.

This I agree, but I would do it before setting the initial size to avoid the race steps you described in https://github.com/neondatabase/neon/pull/3586#discussion_r1104371149.
 
> ` .set(calculated_size);` should be changed to `.set(caclulated_size + size_added_after_initial)` but it's not a simple sum, so instead we need to call `get_current_logical_size()` or its internal `current_size()` somehow.

Yes, I think I agree on this as well and tried to word it on https://github.com/neondatabase/neon/pull/3586#discussion_r1104363709. If done *before* setting the `initial_logical_size` like I suggested in my comment (and above), it would avoid the race you explained.

lubennikovaav: > Yes but flush and commit are only called from the import_basebackup_from_tar and import_from_postgres_datadir right?

I didn't check it that far :)  But you're right.
Does it mean that we never update logical size (not just metric gauge, but the value itself)? Like ever?

koivunej: > Does it mean that we never update logical size (not just metric gauge, but the value itself)? Like ever?

I have no clue on this part. I was hoping you'd know :)

After I started making the fix in a separate PR, I realize I had misunderstood the match in `update_current_logical_size` (it's not about having Exact vs. Approximate but having anything vs. overflowing), so this might be a bit more complicated.

koivunej: Oki now looking at `pageserver/src/walingest.rs` for the first time, it'd appear that all changes happen through a `DatadirModification` which is just dropped when it goes out of scope. `ingest_record` does call `commit` on it though, so this should be all valid.

shanyp:
closing for the alternative path

PR #3587, opened 2023-02-13
[proxy] Restore INFO as the default tracing level

funbringer:
Also move tracing init to its own function.


petuhovskiy (approved):
Looks ok, but it seems that we already have `utils::logging::init` with a similar purpose, why we're not reusing it here?

ref https://github.com/neondatabase/neon/blob/main/libs/utils/src/logging.rs#L26

PR #3588, opened 2023-02-13
docs: add dependency requirements for arch based systems

kashw2:
## Describe your changes

This pull request adds information on building neon on Arch based system such as Artix, Manjaro, Antergos, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ dnf install flex bison readline-devel zlib-devel openssl-devel \
 * On Arch based systems, these packages are needed:
 ```bash
 pacman -S base-devel readline zlib libseccomp openssl clang \
-pkg-config postgresql-libs cmake postgresql protobuf
+postgresql-libs cmake postgresql protobuf
 ```
 
 2. [Install Rust](https://www.rust-lang.org/tools/install)
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ dnf install flex bison readline-devel zlib-devel openssl-devel \
 ```
 * On Arch based systems, these packages are needed:
 ```bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -34,6 +34,11 @@ dnf install flex bison readline-devel zlib-devel openssl-devel \
   libseccomp-devel perl clang cmake postgresql postgresql-contrib protobuf-compiler \
   protobuf-devel
 ```
+* On Arch based systems, these packages are needed:
+```bash
+pacman -S base-devel libtool readline zlib flex bison libseccomp openssl clang \
```
funbringer, line 39: It's a bit more nuanced than `pacman -S`, because you might also need to synchronize the database with a package server (and perform a mandatory upgrade). However, this is not a proper place to teach arch users good manners or nuances, so this should be enough in practice.

README.md
```diff
@@ -34,6 +34,11 @@ dnf install flex bison readline-devel zlib-devel openssl-devel \
...
+pkg-config postgresql-libs cmake postgresql protobuf
```
sharnoff, line 40: Not an expert by any means, but my reading is that `pkg-config` has since been replaced by `pkgconf`, which is already provided by `base-devel`. Is there a reason to include `pkg-config` specifically?

kashw2: I hadn't realised that it'd been replaced by pkgconfig, I can remove that from the command alongside flex, bison and libtool which are also found in base-devel.

sharnoff: unresolving; `pkg-config` is still there ­— otherwise looks good.

kashw2: oops, missed that one when removing the others, fixed in 70e29527ae4d7d78574eb457774d66f9cc6c1436

sharnoff (approved):

PR #3592, opened 2023-02-13
Update current logical size gauge

koivunej:
Alternative to #3586.

Draft while we discuss on the original PR, specifically in thread https://github.com/neondatabase/neon/pull/3586#discussion_r1104424155.

N+1 commits:
- replace our code with latest std methods (1.66)
- stop updating the gauge on approx results
    - this allows us to use OnceCell within as a "latch" on whether to update or not the gauge
- finally update it right after calculating the initial logical size, but before exposing that initial size
- (formatting)
- wrong value in log message if that ever would be logged

TODO:

- atomic orders postponed to #3601
- tests postponed

### DIFF
--- a/pageserver/src/tenant/timeline.rs
+++ b/pageserver/src/tenant/timeline.rs
@@ -1616,6 +1616,19 @@ impl Timeline {
                     }
                     x @ Err(_) => x.context("Failed to calculate logical size")?,
                 };
+
+                // we cannot query current_logical_size.current_size() to know the current
+                // *negative* value, only truncated to u64.
+                let added = self_clone
+                    .current_logical_size
+                    .size_added_after_initial
+                    .load(AtomicOrdering::Relaxed);
+
+                let sum = calculated_size.saturating_add_signed(added);
+
+                // set the gauge value before it can be set in `update_current_logical_size`.
+                self_clone.metrics.current_logical_size_gauge.set(sum);
+
                 match self_clone
                     .current_logical_size
                     .initial_logical_size
--- a/pageserver/src/tenant/timeline.rs
+++ b/pageserver/src/tenant/timeline.rs
@@ -293,8 +293,8 @@ impl LogicalSize {
         match self.initial_logical_size.get() {
             Some(initial_size) => {
                 initial_size.checked_add_signed(size_increment)
-                .with_context(|| format!("Overflow during logical size calculation, initial_size: {initial_size}, size_increment: {size_increment}"))
-                .map(CurrentLogicalSize::Exact)
+                    .with_context(|| format!("Overflow during logical size calculation, initial_size: {initial_size}, size_increment: {size_increment}"))
+                    .map(CurrentLogicalSize::Exact)
             }
             None => {
                 let non_negative_size_increment = u64::try_from(size_increment).unwrap_or(0);
--- a/pageserver/src/tenant/timeline.rs
+++ b/pageserver/src/tenant/timeline.rs
@@ -1805,10 +1805,15 @@ impl Timeline {
         // one value while current_logical_size is set to the
         // other.
         match logical_size.current_size() {
-            Ok(new_current_size) => self
+            Ok(CurrentLogicalSize::Exact(new_current_size)) => self
                 .metrics
                 .current_logical_size_gauge
-                .set(new_current_size.size()),
+                .set(new_current_size),
+            Ok(CurrentLogicalSize::Approximate(_)) => {
+                // don't update the gauge yet, this allows us not to update the gauge back and
+                // forth between the initial size calculation task.
+            }
+            // this is overflow
             Err(e) => error!("Failed to compute current logical size for metrics update: {e:?}"),
         }
     }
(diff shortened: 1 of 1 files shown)


koivunej:
Hmm currently it seems that the `size_added_after_initial` is being written as `SeqCst` and read as `Acquire` and now I added `Relaxed`. Mixing SeqCst and Acquire seems strange... Also I am unclear could there be some other operations we would want this to be ordered with?

lubennikovaav:
>finally update it right after calculating the initial logical size, but before exposing that initial size

Why in this order? I don't get it

koivunej:
> > finally update it right after calculating the initial logical size, but before exposing that initial size
> 
> Why in this order? I don't get it

By itself it does nothing, but with the [0edec19](https://github.com/neondatabase/neon/pull/3592/commits/0edec19ff8cb031686082cc717df8a62f86d1fb8) it does prevent clashing with the gauge update from `update_current_logical_size`.

Also it would appear that this change will prevent us ever setting the gauge size to 0 as well -- which I can now see having happened when `size_added_after_initial` was negative.

koivunej:
> > finally update it right after calculating the initial logical size, but before exposing that initial size
> 
> Why in this order? I don't get it

Better explanation:

With #3586 we have these operations:

1. update the `size_added_after_initial`
2. read `size_added_after_initial`
3. read current_size
4. set current_size
5. update the gauge to approx value
6. update the gauge to exact value

Let's consider two threads:

T1: has just completed calculating the initial size (tokio blocking threadpool)
T2: `update_current_logical_size` after `DataDirModification::commit`

It is possible to schedule these as:

```
bad:

T1: set current_size
T2: update the `size_added_after_initial`
T2: read `current_size`
T2: update the gauge to exact value
T1: update the gauge to exact value without the increment by T2 (bad)

bad:

T1: set current_size
T1: read `size_added_after_initial`
T2: update the `size_added_after_initial`
T2: read `current_size`
T2: update the gauge to exact value
T1: update the gauge to exact value without the increment by T2 (bad) 

happy:

T1: set current_size
T1: read `size_added_after_initial`
T1: update the gauge to exact value (OK)
T2: update `size_added_after_initial`
T2: read `current_size`
T2: update the gauge to exact value (OK)
```

Now that I think of it ... OnceCell provides synchronization only for two set's, but the "get current_size" is never synchronized, it's just atomic, so all interleavings of the two are possible above.

Now after [0edec19](https://github.com/neondatabase/neon/pull/3592/commits/0edec19ff8cb031686082cc717df8a62f86d1fb8) only these are possible:

```
not bad:

T1: read `size_added_after_initial`
T1: update the gauge to exact value
T2: update `size_added_after_initial`
T2: read `current_size`
T2: (do not update the gauge to approx value)
T1: set current_size

good:

T2: update `size_added_after_initial`
T1: read `size_added_after_initial`
T1: update the gauge to exact value
T2: read `current_size`
T2: (do not update the gauge to approx value)
T1: set current_size

eventually ok:

T1: read `size_added_after_initial`
T2: update `size_added_after_initial`
T1: update the gauge to exact value (missing the latest update)
T2: read `current_size`
T2: (do not update the gauge to approx value)
T1: set current_size

eventually ok:

T1: read `size_added_after_initial`
T2: update `size_added_after_initial`
T1: update the gauge to exact value (missing the latest update)
T1: set current_size
T2: read `current_size`
T2: update the gauge to exact value (latest value)
```

Again, above is probably missing some cases, but the end result is that iff `update_current_logical_size` does not ever update the gauge with approx result, it will never be updated to an invalid value.

It is of course unfortunate if we fail to calculate the initial size altogether, but that's also a different story, which needs my other PRs.

pageserver/src/tenant/timeline.rs
```diff
@@ -292,18 +292,9 @@ impl LogicalSize {
...
-                } else {
-                    initial_size.checked_add(absolute_size_increment)
-                }.with_context(|| format!("Overflow during logical size calculation, initial_size: {initial_size}, size_increment: {size_increment}"))
-                .map(CurrentLogicalSize::Exact)
+                initial_size.checked_add_signed(size_increment)
+                    .with_context(|| format!("Overflow during logical size calculation, initial_size: {initial_size}, size_increment: {size_increment}"))
+                    .map(CurrentLogicalSize::Exact)
             }
             None => {
                 let non_negative_size_increment = u64::try_from(size_increment).unwrap_or(0);
```
lubennikovaav, line 300: let's add warn message here to catch possible negative increments
before `unwrap_or(0)`

SomeoneToIgnore (approved):
The code does not seem odd, but looks like it's very simply to make some mistakes here?

I wonder, how feasible it is to create a few unit tests on such things.

koivunej:
> I wonder, how feasible it is to create a few unit tests on such things.

Metrics are yet another unit test hostile part.. Weren't we able to trigger the logical size calculation through http? That could be tested that requesting it eventually sets the gauge to sensible value..

As for the 2 atomics + 1 once_cell ... I've been thinking of loom tests but at the same time I'm feeling out of time already. I think the reasoning for the using once_cell as "latch" is clear, nor do we see any approx results in the metrics either for the problematic tenants.

@SomeoneToIgnore did you have any knowledge on why the orderings (Acquire, SeqCst) are used around these parts: https://github.com/neondatabase/neon/pull/3592#issuecomment-1427915206?

SomeoneToIgnore:
Nope.

lubennikovaav (approved):
The code looks good to me. 
Let's rely on staging for testing, until we have idea for better tests..

PR #3594, opened 2023-02-13
fix: initial logical size timer and caching

koivunej:
initially added sometime ago, introduced in earlier PR of mine #2714. then there was confusion about for which lsn logical size was initially calculated for, and now it appears it is not the initdb_lsn but disk_consistent_lsn.

I don't think this is any correctness change, except for the used histogram, which also means that the other histogram has been seeing all of the use except for tenants which have never been touched after creation.

### DIFF
--- a/pageserver/src/tenant/timeline.rs
+++ b/pageserver/src/tenant/timeline.rs
@@ -1780,7 +1780,7 @@ impl Timeline {
             // need to return something
             Ok(0)
         });
-        let timer = if up_to_lsn == self.initdb_lsn {
+        let timer = if Some(up_to_lsn) == self.current_logical_size.initial_part_end {
             if let Some(size) = self.current_logical_size.initialized_size() {
                 if size != 0 {
                     // non-zero size means that the size has already been calculated by this method
--- a/pageserver/src/tenant/timeline.rs
+++ b/pageserver/src/tenant/timeline.rs
@@ -1780,7 +1780,17 @@ impl Timeline {
             // need to return something
             Ok(0)
         });
-        let timer = if Some(up_to_lsn) == self.current_logical_size.initial_part_end {
+
+        // is this *the* initial logical size calculation from `try_spawn_size_init_task`;
+        // if so, we could optimistically cache the size calculation.
+        let init_logical_size_calculation = match self.current_logical_size.initial_part_end {
+            // this timeline has been loaded from the disk
+            Some(lsn) => lsn == up_to_lsn,
+            // this timeline was just created during this pageserver startup
+            None => up_to_lsn == self.initdb_lsn,
+        };
+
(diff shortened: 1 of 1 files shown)


koivunej:
The fix was realized during #3592 in discussions with @lubennikovaav, decided to move it in separate PR as it would had made the changes in #3592 less clear.

pageserver/src/tenant/timeline.rs
```diff
@@ -1780,7 +1780,7 @@ impl Timeline {
             // need to return something
             Ok(0)
         });
-        let timer = if up_to_lsn == self.initdb_lsn {
+        let timer = if Some(up_to_lsn) == self.current_logical_size.initial_part_end {
```
lubennikovaav, line 1783: as comment says,  `initial_part_end` could be absent for freshly created timelines.
(not sure what exactly this "could be" means)

Am I getting it right, that this change will break  `init_logical_size_histo` for new timelines?

koivunej: Aah I did not consider this.

koivunej: new attempt: [5cfc5f3](https://github.com/neondatabase/neon/pull/3594/commits/5cfc5f398c86f4af4e8a26ba88af7dba659f8989)

SomeoneToIgnore (commented):
Does not look bad to me, but still somewhat hard to understand the context here after a while.

pageserver/src/tenant/timeline.rs
SomeoneToIgnore: "Freshly crated timeline" does not seem hard to create, could some unit tests help detection here?

koivunej:
@SomeoneToIgnore:

> Does not look bad to me, but still somewhat hard to understand the context here after a while.

The context was that when doing this optimization for tenant_size initially, we deemed the `up_to_lsn == initdb_lsn` check correct. Later I realized that no, that is only calculated once for just now created timelines, later restarts use `disk_consistent_lsn` (IIRC) here, so this is more about enabling the proper timer to be used (which might be just as bad => more metrics), but also enabling the caching to be used. I guess with the latest rewriting #2817 the caching could be just taken out. I don't think that this is going to be hit...

At the same time it is a valid thing to cache. Now I am more and more doubtful about the metric, should it be trimmed at the same time or in a follow-up?

pageserver/src/tenant/timeline.rs
koivunej: Good idea on the unit test. I've abused this CurrentLogicalSize enough to warrant some refactoring :)

koivunej:
Needs an unit test.

koivunej:
Superceded by #3874.

PR #3598, opened 2023-02-14
Compaction with on-demand download

### LINKED ISSUE #3591: Compaction with on-demand download is unimplemented
```rust
impl PersistentLayer for RemoteLayer {

    ...

    fn iter(&self, _ctx: &RequestContext) -> Result<LayerIter<'_>> {
        bail!("cannot iterate a remote layer");
    }

    fn key_iter(&self, _ctx: &RequestContext) -> Result<LayerKeyIter<'_>> {
        bail!("cannot iterate a remote layer");
    }

    ...

}
```

Or actual lines: https://github.com/neondatabase/neon/blob/f383b4d5401bd8ddb4e96a683af4d72496967d0a/pageserver/src/tenant/storage_layer/remote_layer.rs#L139-L145

Related:
- #3348 adds hole information for DeltaLayer
    - but added Layer method will work on RemoteLayer just as well
    - risk of creating different compactions repartitioning based on when the layers are downloaded
    - it also switches from using in-order Vec of l0_deltas to a HashSet
        - didn't check if at load time these are actually sorted, but their order is maintained before #3348

Related thoughts:

Currently we don't have facilities to follow a lineage of `{Delta,Image}Layer => RemoteLayer => {Delta,Image}Layer => ...` state transitions, which cause a new layer to appear in the old ones place, as we use the `Arc::as_ptr` as layer identity (before or without ever comparing the layer filename, which we use for logging). So it follows, we cannot really implement a "before compaction on-demand downloading but allowing eviction during compaction".

Discovered while discussing race conditions for #3533.

koivunej:
## Describe your changes

Repeatedly try to download the compaction targeted layers before actual compaction.

Fixes #3591

- [x] If it is a core feature, I have added thorough tests.
    - don't think it's necessary, we have ondemand metrics already
    - enables eviction, not customer facing hopefully

### DIFF
--- a/pageserver/src/tenant/timeline.rs
+++ b/pageserver/src/tenant/timeline.rs
@@ -622,6 +622,7 @@ impl Timeline {
         self.flush_frozen_layers_and_wait().await
     }
 
+    /// Outermost timeline compaction operation; downloads needed layers.
     pub async fn compact(&self, ctx: &RequestContext) -> anyhow::Result<()> {
         const ROUNDS: usize = 2;
 
@@ -639,7 +640,7 @@ impl Timeline {
             // should we error out with the most specific error?
             let last_round = round == ROUNDS - 1;
 
-            let res = self.compact_without_ondemand_downloads(ctx).await;
+            let res = self.compact_inner(ctx).await;
 
             let rls = match res {
                 Ok(()) => return Ok(()),
@@ -689,10 +690,8 @@ impl Timeline {
         unreachable!("retry loop exits")
     }
 
-    async fn compact_without_ondemand_downloads(
-        &self,
-        ctx: &RequestContext,
-    ) -> Result<(), CompactionError> {
+    /// Compaction which might need to be retried after downloading remote layers.
+    async fn compact_inner(&self, ctx: &RequestContext) -> Result<(), CompactionError> {
         //
         // High level strategy for compaction / image creation:
         //
--- a/pageserver/src/tenant/timeline.rs
+++ b/pageserver/src/tenant/timeline.rs
@@ -670,12 +670,21 @@ impl Timeline {
 
             let mut failed = 0;
 
-            while let Some(res) = downloads.next().await {
-                match res {
-                    Ok(()) => {}
-                    Err(e) => {
-                        warn!("Downloading remote layer for compaction failed: {e:#}");
-                        failed += 1;
+            let cancelled = task_mgr::shutdown_watcher();
+            tokio::pin!(cancelled);
+
+            loop {
+                tokio::select! {
+                    _ = &mut cancelled => anyhow::bail!("Cancelled while downloading remote layers"),
+                    res = downloads.next() => {
+                        match res {
+                            Some(Ok(())) => {},
+                            Some(Err(e)) => {
+                                warn!("Downloading remote layer for compaction failed: {e:#}");
+                                failed += 1;
+                            }
+                            None => break,
+                        }
                     }
                 }
             }
--- a/pageserver/src/tenant/timeline.rs
+++ b/pageserver/src/tenant/timeline.rs
@@ -642,6 +642,22 @@ impl Timeline {
 
             let res = self.compact_inner(ctx).await;
 
+            // If `create_image_layers' or `compact_level0` scheduled any
+            // uploads or deletions, but didn't update the index file yet,
+            // do it now.
+            //
+            // This isn't necessary for correctness, the remote state is
+            // consistent without the uploads and deletions, and we would
+            // update the index file on next flush iteration too. But it
+            // could take a while until that happens.
+            //
+            // Additionally, only do this on the terminal round before sleeping.
+            if last_round {
+                if let Some(remote_client) = &self.remote_client {
+                    remote_client.schedule_index_upload_for_file_changes()?;
+                }
+            }
+
             let rls = match res {
                 Ok(()) => return Ok(()),
                 Err(CompactionError::DownloadRequired(rls)) if !last_round => {
@@ -769,18 +785,6 @@ impl Timeline {
                 self.compact_level0(&layer_removal_cs, target_file_size, ctx)
                     .await?;
                 timer.stop_and_record();
-
-                // If `create_image_layers' or `compact_level0` scheduled any
-                // uploads or deletions, but didn't update the index file yet,
-                // do it now.
-                //
-                // This isn't necessary for correctness, the remote state is
-                // consistent without the uploads and deletions, and we would
-                // update the index file on next flush iteration too. But it
-                // could take a while until that happens.
-                if let Some(remote_client) = &self.remote_client {
-                    remote_client.schedule_index_upload_for_file_changes()?;
-                }
             }
             Err(err) => {
                 // no partitioning? This is normal, if the timeline was just created
--- a/pageserver/src/tenant/timeline.rs
+++ b/pageserver/src/tenant/timeline.rs
@@ -644,17 +644,17 @@ impl Timeline {
 
             let rls = match res {
                 Ok(()) => return Ok(()),
-                Err(CompactionError::Phase1(Phase1Error::DownloadRequired(rls))) if !last_round => {
+                Err(CompactionError::DownloadRequired(rls)) if !last_round => {
                     // this can be done at most one time before exiting, waiting
                     rls
                 }
-                Err(CompactionError::Phase1(e @ Phase1Error::DownloadRequired(_))) => {
+                Err(e @ CompactionError::DownloadRequired(_)) => {
                     // it is possible that this keeps happening, because we are battling
                     // some other process which evicts the layers we download; we should
                     // retry quite soon anyways.
                     return Err(e.into());
                 }
-                Err(CompactionError::Phase1(Phase1Error::Other(e)) | CompactionError::Other(e)) => {
+                Err(CompactionError::Other(e)) => {
                     return Err(e);
                 }
             };
@@ -2773,18 +2773,12 @@ struct CompactLevel0Phase1Result {
     deltas_to_compact: Vec<Arc<dyn PersistentLayer>>,
 }
 
-#[derive(Debug, thiserror::Error)]
-enum Phase1Error {
-    #[error("level 0 compaction requires downloading of {} layer files", Vec::len(.0))]
-    DownloadRequired(Vec<Arc<RemoteLayer>>),
-    #[error(transparent)]
-    Other(#[from] anyhow::Error),
-}
-
+/// Top-level failure to compact.
 #[derive(Debug, thiserror::Error)]
 enum CompactionError {
-    #[error(transparent)]
-    Phase1(#[from] Phase1Error),
+    #[error("level 0 compaction requires downloading of {} layer files", Vec::len(.0))]
+    DownloadRequired(Vec<Arc<RemoteLayer>>),
+    /// Compaction cannot be done right now; page reconstruction and so on.
     #[error(transparent)]
     Other(#[from] anyhow::Error),
 }
@@ -2794,7 +2788,7 @@ impl Timeline {
         &self,
         target_file_size: u64,
         ctx: &RequestContext,
-    ) -> Result<CompactLevel0Phase1Result, Phase1Error> {
+    ) -> Result<CompactLevel0Phase1Result, CompactionError> {
         let layers = self.layers.read().unwrap();
         let mut level0_deltas = layers.get_level0_deltas()?;
         drop(layers);
@@ -2863,7 +2857,7 @@ impl Timeline {
             //     deltas_to_compact.len(),
             //     level0_deltas.len(),
             //     remotes.len());
-            return Err(Phase1Error::DownloadRequired(remotes));
+            return Err(CompactionError::DownloadRequired(remotes));
         }
 
         info!(
@@ -3047,9 +3041,9 @@ impl Timeline {
             }
 
             fail_point!("delta-layer-writer-fail-before-finish", |_| {
-                return Err(Phase1Error::Other(anyhow::anyhow!(
-                    "failpoint delta-layer-writer-fail-before-finish"
-                )));
+                return Err(
+                    anyhow::anyhow!("failpoint delta-layer-writer-fail-before-finish").into(),
+                );
             });
 
             writer.as_mut().unwrap().put_value(key, lsn, value)?;
--- a/pageserver/src/tenant/timeline.rs
+++ b/pageserver/src/tenant/timeline.rs
@@ -2873,7 +2873,8 @@ impl Timeline {
 
         if !remotes.is_empty() {
             // caller is holding the lock to layer_removal_cs, and we don't want to download while
-            // holding that; in future download_remote_layer might take it as well.
+            // holding that; in future download_remote_layer might take it as well. this is
+            // regardless of earlier image creation downloading on-demand, while holding the lock.
             return Err(CompactionError::DownloadRequired(remotes));
         }
 
--- a/pageserver/src/tenant/timeline.rs
+++ b/pageserver/src/tenant/timeline.rs
@@ -2784,8 +2784,14 @@ enum CompactionError {
 }
 
 impl Timeline {
+    /// Level0 files first phase of compaction, explained in the [`compact_inner`] comment.
+    ///
+    /// This method takes the `_layer_removal_cs` guard to highlight it required downloads are
+    /// returned as an error. If the `layer_removal_cs` boundary is changed not to be taken in the
+    /// start of level0 files compaction, the on-demand download should be revisited as well.
     async fn compact_level0_phase1(
         &self,
+        _layer_removal_cs: &tokio::sync::MutexGuard<'_, ()>,
         target_file_size: u64,
         ctx: &RequestContext,
     ) -> Result<CompactLevel0Phase1Result, CompactionError> {
@@ -3081,7 +3087,9 @@ impl Timeline {
         let CompactLevel0Phase1Result {
             new_layers,
             deltas_to_compact,
-        } = self.compact_level0_phase1(target_file_size, ctx).await?;
+        } = self
+            .compact_level0_phase1(layer_removal_cs, target_file_size, ctx)
+            .await?;
 
         if new_layers.is_empty() && deltas_to_compact.is_empty() {
             // nothing to do
--- a/pageserver/src/tenant/timeline.rs
+++ b/pageserver/src/tenant/timeline.rs
@@ -2850,13 +2850,6 @@ impl Timeline {
         if !remotes.is_empty() {
             // caller is holding the lock to layer_removal_cs, and we don't want to download while
             // holding that; in future download_remote_layer might take it as well.
-            //
-            // unsure if this is required, probably not:
-            // info!(
-            //     "Level0 compaction in LSN range {lsn_range:?} for {} layers ({} deltas in total) requires downloading {} layers",
-            //     deltas_to_compact.len(),
-            //     level0_deltas.len(),
-            //     remotes.len());
             return Err(CompactionError::DownloadRequired(remotes));
         }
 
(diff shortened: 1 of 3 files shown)


pageserver/src/tenant/timeline.rs
```diff
@@ -2945,7 +3058,7 @@ impl Timeline {
 
             // Fsync all the layer files and directory using multiple threads to
             // minimize latency.
-            par_fsync::par_fsync(&layer_paths)?;
+            par_fsync::par_fsync(&layer_paths).context("failed to fsync all new layers")?;
```
problame, line 3061: omit the `failed to`

pageserver/src/tenant/timeline.rs
```diff
@@ -2995,7 +3108,12 @@ impl Timeline {
         for l in new_layers {
             let new_delta_path = l.path();
 
-            let metadata = new_delta_path.metadata()?;
+            let metadata = new_delta_path.metadata().with_context(|| {
+                format!(
+                    "failed to read file metadata for new created layer {}",
```
problame, line 3113: same here, omit the `failed to`

problame (commented):
To keep this PR smaller, can you move the unchanged bulk of the code into an outer function?

I.e.:

- current `main`'s `async fn compact` becomes `async fn compact_impl -> Result<..., CompactionError>`
- your loop in a new `async fn compact`

In terms of the general approach: I think download & retry will work reliably with threshold-based eviction iff
1. the retry happens faster than the eviction (fair assumption) and
2. the `compact_level0_phase1` makes the same selection of layers that it wants to download in the second round.

What guarantees (2)? 

Regarding (1), I think it would be helpful if compaction logged a message like `we downloaded all the layers successfully, but they got evicted again; likely eviction threshold is too aggressive or the compaction code is too slow`. Just to make it obvious in the code & to operators what's going on.

koivunej:
> To keep this PR smaller, can you move the unchanged bulk of the code into an outer function?

Yeah ... Thought of this, but there's already a chain of N functions like this :D But you are right, perhaps N+1 with the smaller diff is better.

> What guarantees (2)?

I don't think it needs to be guaranteed. We have the 2s retry for compaction, as long as we don't evict anything within the timescale of seconds (currently aiming for days) -- as long as this doesn't happen every time, progress will still be made.

I think implementing this would require `fn previous_version(&self) -> Weak<L>` or so to track the state changes (and that chain would be unboundedly long) for the message you suggested:

> we downloaded all the layers successfully, but they got evicted again; likely eviction threshold is too aggressive or the compaction code is too slow

Otherwise we cannot really be sure if something else gets rid of the layers we downloaded, or if gc took them away in between. The gc might be a valid possibility with some sort of data layouts. Or not, I am a bit fuzzy on the image layer requirement and if it can only be created by compaction.

pageserver/src/tenant/timeline.rs
koivunej: [06bf49c](https://github.com/neondatabase/neon/pull/3598/commits/06bf49cb9eafe38f2e95782cf7dcbc4dd28c52e1)

pageserver/src/tenant/timeline.rs
koivunej: [06bf49c](https://github.com/neondatabase/neon/pull/3598/commits/06bf49cb9eafe38f2e95782cf7dcbc4dd28c52e1)

pageserver/src/tenant/timeline.rs
```diff
@@ -631,6 +633,66 @@ impl Timeline {
...
+            }
+
+            if failed != 0 {
+                anyhow::bail!("{failed} out of {total} layers failed to download, retrying later");
+            }
+
+            // if everything downloaded fine, lets try again
+        }
+
+        unreachable!("retry loop exists")
```
koivunej, line 689:
```suggestion
        unreachable!("retry loop exits")
```

koivunej: Done in af9034f2

pageserver/src/tenant/timeline.rs
```diff
@@ -2747,16 +2843,42 @@ impl Timeline {
...
+        if !remotes.is_empty() {
+            // caller is holding the lock to layer_removal_cs, and we don't want to download while
+            // holding that; in future download_remote_layer might take it as well.
+            //
+            // unsure if this is required, probably not:
+            // info!(
+            //     "Level0 compaction in LSN range {lsn_range:?} for {} layers ({} deltas in total) requires downloading {} layers",
+            //     deltas_to_compact.len(),
+            //     level0_deltas.len(),
+            //     remotes.len());
```
koivunej, line 2866: Didn't miss this logging at all. Thinking of removing.

test_runner/regress/test_ondemand_download.py
```diff
@@ -449,3 +453,168 @@ def get_resident_physical_size():
...
+    tenant_id, timeline_id = env.neon_cli.create_tenant(conf=stringify(conf))
+    env.initial_tenant = tenant_id
+    pageserver_http = env.pageserver.http_client()
+
+    with env.postgres.create_start("main") as pg:
+        # no particular reason to create the layers like this, but we are sure
+        # not to hit the image_creation_threshold here.
+        with pg.cursor() as cur:
+            cur.execute("create table a (id bigserial primary key, some_value bigint not null)")
+            cur.execute("select pg_relation_size('a')")
```
koivunej, line 563: Oops

pageserver/src/tenant/timeline.rs
```diff
@@ -631,6 +633,66 @@ impl Timeline {
             return Ok(());
         }
 
+        // retry two times to allow first round to find layers which need to be downloaded, then
+        // download them, then retry compaction
+        for round in 0..ROUNDS {
+            // should we error out with the most specific error?
+            let last_round = round == ROUNDS - 1;
+
+            let res = self.compact_without_ondemand_downloads(ctx).await;
```
koivunej, line 642: Should a subcontext be created which would ERROR on ... downloads.

No, and but the name needs to be changed, because it can on-demand download for image creation, shown by the added test :)

test_runner/regress/test_ondemand_download.py
```diff
@@ -449,3 +453,168 @@ def get_resident_physical_size():
...
+    # to be less flaky
+    env.neon_cli.config_tenant(tenant_id, {"image_creation_threshold": "1"})
+
+    pageserver_http.timeline_compact(tenant_id, timeline_id)
+    layers = pageserver_http.layer_map_info(tenant_id, timeline_id)
+    kinds_after = defaultdict(int)
+    for layer in layers.historic_layers:
+        kinds_after[layer.kind] += 1
+
+    assert dict(kinds_after) == {"Delta": 4, "Image": 1}
```
koivunej, line 616: Does python have flexible dict order for equality comparisons...?

test_runner/regress/test_ondemand_download.py
```diff
@@ -449,3 +453,168 @@ def get_resident_physical_size():
...
+        # loop in case flushing is still in progress
+        layers = pageserver_http.layer_map_info(tenant_id, timeline_id)
+        if not layers.in_memory_layers:
+            break
+        time.sleep(0.2)
+
+    layers = pageserver_http.layer_map_info(tenant_id, timeline_id)
+    assert not layers.in_memory_layers, "no inmemory layers expected after post-commit checkpoint"
+
+    kinds_before = defaultdict(int)
```
koivunej, line 596: mypy does not like this.

koivunej:
Marking this ready for review but I'll continue on it tomorrow morning.

pageserver/src/tenant/timeline.rs
```diff
@@ -631,6 +633,66 @@ impl Timeline {
...
+            let total = rls.len();
+
+            let mut downloads = rls
+                .into_iter()
+                .map(|rl| self.download_remote_layer(rl))
+                .collect::<futures::stream::FuturesUnordered<_>>();
+
+            let mut failed = 0;
+
+            while let Some(res) = downloads.next().await {
```
problame, line 672: - Add TODO that this `.await` should be sensitive to `ctx` cancellation once that is supported.
- In the meantime, make it sensitive to `task_mgr` cancellation (I think even compaction via mgmt API runs in a task_mgr task nowadays, should work fine)

pageserver/src/tenant/timeline.rs
```diff
@@ -631,6 +633,66 @@ impl Timeline {
...
+                Ok(()) => return Ok(()),
+                Err(CompactionError::Phase1(Phase1Error::DownloadRequired(rls))) if !last_round => {
+                    // this can be done at most one time before exiting, waiting
+                    rls
+                }
+                Err(CompactionError::Phase1(e @ Phase1Error::DownloadRequired(_))) => {
+                    // it is possible that this keeps happening, because we are battling
+                    // some other process which evicts the layers we download; we should
+                    // retry quite soon anyways.
+                    return Err(e.into());
```
problame, line 654: would still like to see that comment in the error message and ultimately a `WARN` or `ERROR` level log message

problame (commented):
Much better diff with the `compact_without_ondemand_downloads`.

---

Another point I noticed: the second round will run the image layer generation again. Do we want that to happen?
Before you jump to address this here: maybe this is the time to split up the compaction task into two tasks? compact_level0, and image layer generation. I think This has been suggested multiple times in the past, cc @knizhnik

pageserver/src/tenant/timeline.rs
```diff
@@ -2693,25 +2767,47 @@ impl Timeline {
...
 }
+
 #[derive(Default)]
 struct CompactLevel0Phase1Result {
     new_layers: Vec<DeltaLayer>,
     deltas_to_compact: Vec<Arc<dyn PersistentLayer>>,
 }
 
+#[derive(Debug, thiserror::Error)]
+enum Phase1Error {
```
koivunej, line 2778: I don't think the phase1error carries it's weight, combining this into the CompactionError.

test_runner/regress/test_ondemand_download.py
koivunej: With some manual testing it would appear yes.

pageserver/src/tenant/timeline.rs
koivunej: Good point, this will be output as `{:?}` so we currently have needlessly the whole stacktrace at least from the compaction bg task.

Is that not enough? Would like to get rid of the stacktrace though.

test_runner/regress/test_ondemand_download.py
koivunej: Removed: 78a1a41e

pageserver/src/tenant/timeline.rs
koivunej: Removed: fe7b6ec6

pageserver/src/tenant/timeline.rs
koivunej: Only one remains: 1560f0ed

pageserver/src/tenant/timeline.rs
koivunej: Good point. Implemented racing against cancellation in 297b1a70.

Do you want an extra todo as well? Note that it doesn't seem like there are any other cancellation checks in compaction...

pageserver/src/tenant/timeline.rs
koivunej: Renamed in eb95d489 as `compact_inner`.

pageserver/src/tenant/timeline.rs
```diff
@@ -2566,21 +2636,25 @@ impl Timeline {
...
-                    debug!(
-                        "key range {}-{}, has {} deltas on this timeline in LSN range {}..{}",
-                        img_range.start, img_range.end, num_deltas, img_lsn, lsn
-                    );
+                    max_deltas = max_deltas.max(num_deltas);
                     if num_deltas >= threshold {
+                        debug!(
+                            "key range {}-{}, has {} deltas on this timeline in LSN range {}..{}",
+                            img_range.start, img_range.end, num_deltas, img_lsn, lsn
+                        );
```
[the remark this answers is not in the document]
koivunej: Seems fine, but keeping this open until merge in case someone wants to object.

koivunej:
> Another point I noticed: the second round will run the image layer generation again.

A good observation, ignored this myself even though I did witness it with the changed logging.

> Do we want that to happen? Before you jump to address this here: maybe this is the time to split up the compaction task into two tasks? compact_level0, and image layer generation. I think This has been suggested multiple times in the past, cc @knizhnik

Well, the implementation is based on the assumption that it's not critical when compaction is ran as it is just a background task while I don't agree with this being a sensible implementation (racing against other operations) I think right now it is sort of idempotent because of the `Timeline::time_for_new_image_layer`, similarly `repartition` should not repartition unless checkpoint_distance/10 has been advanced.

Now I can see we might leave the new image layer uploaded because of this esp. on repeated download required... Will look a bit more on this, what else could I had missed. Asking for the index part to be uploaded is easy to do upon exiting with error or returning ok from the new `Timeline::compact(...)`.

Unsure what would the benefits be of splitting to *two tasks* however. Then we would have even more competing tasks which can only be ran sequentially. Perhaps the retry should only cover the "non-remote layer tolerating" part.

pageserver/src/tenant/timeline.rs
koivunej: Aah but because it's a thiserror::Error, the added variant will not have a stacktrace, only `anyhow::anyhow!` will have it, which means "failed to download some layers etc" will have them just as well. I should add for those specific error types to avoid having the stacktrace.

koivunej: [0790957](https://github.com/neondatabase/neon/pull/3598/commits/0790957c3ab8c4467d9b6cfabb931b8246939d68)

koivunej: It seems it might capture the stacktrace even on into coversion?

koivunej: Yes it does: https://github.com/dtolnay/anyhow/blob/546342c3bd5418734ea0313331f847fa9517fbeb/src/backtrace.rs#L48-L53

Ok so only way to avoid printing the stacktrace is to turn all compaction reporting `{:#}` or expose the CompactionError..

koivunej: Well, I'll just drop the last commit (0790957), and we'll just need to drop the feature from anyhow some other time.

But to the original point; if we have repeated downloads issue like you were asking:
> would still like to see that comment in the error message and ultimately a WARN or ERROR level log message

It would be logged by https://github.com/neondatabase/neon/blob/3569c1bacdaea7b357f927cef32cd657e82af941/pageserver/src/tenant/tasks.rs#L80

The message will look something like:

```
ts WARN Compaction failed, retrying in 2s: Repeated downloading of layers for compaction required

Stack backtrace:
   0: <T as core::convert::Into<U>>::into
             at /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/core/src/convert/mod.rs:726:9
      pageserver::tenant::timeline::Timeline::compact::{{closure}}
             at /home/joonas/src/neon/neon-main/pageserver/src/tenant/timeline.rs:672:32
             at /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/core/src/future/mod.rs:91:19
   1: <tracing::instrument::Instrumented<T> as core::future::future::Future>::poll
```

So no logging and returning an error is required.

koivunej: Oki perhaps I missed what you wanted, guessing this with the next push:

```
ts INFO span compact requires download of 000000000000000000000000000000000000-FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF__000000000169ACF0-000000000169AD69
ts INFO span download complete: <unfortunately full path to layer file>
ts INFO span Rebuilt layer map. Did 2 insertions to process a batch of 1 updates.
ts INFO span compact requires download of 000000000000000000000000000000000000-FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF__000000000169AD69-00000000024DACB1
ts ERROR span Compaction failed, retrying in 2s: Compaction requires downloading multiple times (last was 1 layers), possibly battling against eviction

Stack backtrace:
```

problame: However, if you don't think it's worth the effort, feel free to mark this as resolved.

koivunej: Now [b59d12f](https://github.com/neondatabase/neon/pull/3598/commits/b59d12fc1bfee07627ad3fb3fd8d7a5221edbd1f) mentiones battling eviction, but doesn't include the amount of loops.

koivunej:
Skipping reporting of test_ondemand_download_timetravel[real_s3] failure and just re-ran. Message was, looks typical:

```
Exception: timed out while waiting for remote_consistent_lsn to reach 0/43038D0, was 0/4303898
```

problame (approved):

pageserver/src/tenant/timeline.rs
koivunej: Discussed that it's fine, had to rebase though so: [dcd322c](https://github.com/neondatabase/neon/pull/3598/commits/dcd322cb895b2b05957b4476f1ebee843a20ac7b) is fine.

koivunej:
Waiting to make sure @knizhnik does not object to this per above comments.

pageserver/src/tenant/timeline.rs
```diff
@@ -669,7 +736,7 @@ impl Timeline {
         // Is the timeline being deleted?
         let state = *self.state.borrow();
         if state == TimelineState::Stopping {
-            anyhow::bail!("timeline is Stopping");
+            return Err(anyhow::anyhow!("timeline is Stopping").into());
```
shanyp, line 739: shouldn't the lock be taken after this check ?

koivunej: The lock guards against concurrent layer removals (which for example eviction does) ... I don't think the order here matters; for this checked state the state change does not take the same lock.

I guess this does seem a bit out of place check for me, but it's not wrong.

pageserver/src/tenant/timeline.rs
koivunej: Extra todo assumed not necessary, resolving.

pageserver/src/tenant/timeline.rs
koivunej: Chatted, this was a generic point about preconditions before taking any locks. I am inclined not to change this on this PR, current change is only for the introduced error type. Resolving.

pageserver/src/tenant/timeline.rs
```diff
@@ -2747,16 +2856,35 @@ impl Timeline {
...
+                l.clone()
+                    .downcast_remote_layer()
+                    .expect("just checked it is remote layer")
+            })
+            .collect::<Vec<_>>();
+
+        if !remotes.is_empty() {
+            // caller is holding the lock to layer_removal_cs, and we don't want to download while
+            // holding that; in future download_remote_layer might take it as well.
+            return Err(CompactionError::DownloadRequired(remotes));
```
koivunej, line 2873: Just noticed that we are in fact holding the lock while image creation downloads the layers. Should it be noted here, or alternative approach be taken still (download here)?

koivunej: Noted in [1af4b79](https://github.com/neondatabase/neon/pull/3598/commits/1af4b79d5895e237e6faf8fbc839ce6ea1b5555a)

koivunej:
If the ci passes on latest commits, I will merge it right then.

I moved the index part uploading out of the retried `compact_inner`. This should have the restore worst-case of 1 upload instead of 2 (after first image creation, then after compaction).

koivunej:
The added test fails on merge commit for the first time (non-locally):

https://neon-github-public-dev.s3.amazonaws.com/reports/main/debug/4194510385/index.html#suites/b97efae3a617afb71cb8142f5afa5224/ab71fb4c33a6daaa/

Added a wait timeout for this which took the amount of failures to zero locally and didn't think about this after but it seems that something new might come up after the last checkpoint, had not noticed. Perhaps the last checkpoint should be done only after pg is stopped, or just ignore the inmemory layer.

PR #3605, opened 2023-02-14
[proxy] Fix: don't cache user & dbname in node info cache

funbringer:
```
    Upstream proxy erroneously stores user & dbname in compute node info
    cache entries, thus causing "funny" connection problems if such an entry
    is reused while connecting to e.g. a different DB on the same compute node.
    
    This PR fixes the problem but doesn't eliminate the root cause (complexity) just yet.
    I'll revisit this code and make it more type-safe in the upcoming PR.
```

Discussion: https://neondb.slack.com/archives/C02S1G47EQ3/p1675804192491309

### DIFF
--- a/proxy/src/auth/backend/link.rs
+++ b/proxy/src/auth/backend/link.rs
@@ -78,6 +78,8 @@ pub(super) async fn handle_user(
 
     client.write_message_noflush(&Be::NoticeResponse("Connecting to database."))?;
 
+    // This config should be self-contained, because we won't
+    // take username or dbname from client's startup message.
     let mut config = compute::ConnCfg::new();
     config
         .host(&db_info.host)
--- a/proxy/src/auth/credentials.rs
+++ b/proxy/src/auth/credentials.rs
@@ -32,7 +32,6 @@ impl UserFacingError for ClientCredsParseError {}
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct ClientCredentials<'a> {
     pub user: &'a str,
-    pub dbname: &'a str,
     // TODO: this is a severe misnomer! We should think of a new name ASAP.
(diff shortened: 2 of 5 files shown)


koivunej (approved):
After some chatting approving this because we are hurrying up. Would prefer to see a regression test using proxy as a whole to demonstrate this particular bug because from the changed lines it is not obvious how these fix it.

kelvich:
Looks good as a quick fix. But speaking more generally in that structure:

```
pub struct Config {
    pub(crate) user: Option<String>,
    pub(crate) password: Option<Vec<u8>>,
    pub(crate) auth_keys: Option<Box<AuthKeys>>,
    pub(crate) dbname: Option<String>,
    pub(crate) options: Option<String>,
    pub(crate) application_name: Option<String>,
    pub(crate) ssl_mode: SslMode,
    pub(crate) host: Vec<Host>,
    pub(crate) port: Vec<u16>,
    pub(crate) connect_timeout: Option<Duration>,
    pub(crate) keepalives: bool,
    pub(crate) keepalive_config: KeepaliveConfig,
    pub(crate) target_session_attrs: TargetSessionAttrs,
    pub(crate) channel_binding: ChannelBinding,
    pub(crate) replication_mode: Option<ReplicationMode>,
}
```

probably we don't need to cache anything besides `host` / `port`?

funbringer:
> Looks good as a quick fix. But speaking more generally in that structure:
> ...

Precisely, but it takes more time.

funbringer:
> After some chatting approving this because we are hurrying up. Would prefer to see a regression test using proxy as a whole to demonstrate this particular bug because from the changed lines it is not obvious how these fix it.

I'll implement a proper mocked console api for local proxy tests. This will close the gap.

PR #3610, opened 2023-02-14
compute_ctl: support for fetching spec from control plane

tychoish:
This code won't run yet, but allows the compute to request the spec
from the control plane rather than require that it be passed in as an argument.

### DIFF
--- a/compute_tools/src/bin/compute_ctl.rs
+++ b/compute_tools/src/bin/compute_ctl.rs
@@ -66,6 +66,9 @@ fn main() -> Result<()> {
     let spec = matches.get_one::<String>("spec");
     let spec_path = matches.get_one::<String>("spec-path");
 
+    let compute_id = matches.get_one::<String>("compute-id");
+    let control_plane_uri = matches.get_one::<String>("control-plane-uri");
+
     // Try to use just 'postgres' if no path is provided
     let pgbin = matches.get_one::<String>("pgbin").unwrap();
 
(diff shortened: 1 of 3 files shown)


compute_tools/src/bin/compute_ctl.rs
```diff
@@ -78,6 +81,16 @@ fn main() -> Result<()> {
                 let path = Path::new(sp);
                 let file = File::open(path)?;
                 serde_json::from_reader(file)?
+            } else if let Some(id) = compute_id {
+                if let Some(cp_base) = control_plane_uri {
+                    let cp_uri = cp_base.to_owned() + "/management/api/v1/" + id + "/spec";
```
vadim2404, line 86: `format!` to build the URI. And it will make obsolete usage of `to_owned`

compute_tools/src/bin/compute_ctl.rs
```diff
@@ -78,6 +81,16 @@ fn main() -> Result<()> {
...
+                    reqwest::blocking::get(cp_uri)?
```
vadim2404, line 87: And what are you doing with this JSON afterward?

compute_tools/src/bin/compute_ctl.rs
```diff
@@ -78,6 +81,16 @@ fn main() -> Result<()> {
...
+                        // TODO: fetch a jwt token from the environment
+                        // .header("Authorization", "JWT TOKEN")
```
ololobus, line 89: Maybe just implement it right away instead of putting a TODO? Seems to be pretty trivial

compute_tools/src/bin/compute_ctl.rs
```diff
@@ -78,6 +81,16 @@ fn main() -> Result<()> {
...
+                        .json()?
+                } else {
+                    panic!("must specify --control-plane-uri with --compute-id");
+                }
             } else {
                 panic!("cluster spec should be provided via --spec or --spec-path argument");
```
ololobus, line 95: NIT: control flow is a bit flawed now. If you provide just `control_plane_uri` without `compute_id` it will error out with `cluster spec should be provided via --spec or --spec-path argument`, which doesn't give you any hint on what you did wrong

ololobus: Also, we can change `cluster spec` to the `compute spec` in this message, while on it

ololobus (approved):

PR #3612, opened 2023-02-15
[proxy] Don't forward empty `options` to compute nodes

funbringer:
Clients may specify endpoint/project name via `options=project=...`, so we should not only remove `project=` from `options` but also drop `options` entirely, because connection pools don't support it.

Discussion: https://neondb.slack.com/archives/C033A2WE6BZ/p1676464382670119

### DIFF
--- a/proxy/src/compute.rs
+++ b/proxy/src/compute.rs
@@ -77,14 +77,9 @@ impl ConnCfg {
             self.dbname(dbname);
         }
 
-        if let Some(options) = params.options_raw() {
-            // We must drop all proxy-specific parameters.
-            #[allow(unstable_name_collisions)]
-            let options: String = options
-                .filter(|opt| !opt.starts_with("project="))
-                .intersperse(" ") // TODO: use impl from std once it's stabilized
-                .collect();
-
+        // Don't add `options` if they were only used for specifying a project.
+        // Connection pools don't support `options`, because they affect backend startup.
+        if let Some(options) = filtered_options(params) {
             self.options(&options);
         }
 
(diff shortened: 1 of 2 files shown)


Daniel-ef:
Could you remove me from the CODE_OWNERS? :) I'm not sure it makes sense to keep me there.

proxy/src/compute.rs
```diff
@@ -225,3 +220,46 @@ impl ConnCfg {
...
+
+        // It's likely that clients will only use options to specify endpoint/project.
+        let params = StartupMessageParams::new([("options", "project=foo")]);
+        assert_eq!(filtered_options(&params), None);
+
+        // Same, because unescaped whitespaces are no-op.
+        let params = StartupMessageParams::new([("options", " project=foo ")]);
+        assert_eq!(filtered_options(&params).as_deref(), None);
+
+        let params = StartupMessageParams::new([("options", r"\  project=foo \ ")]);
```
ololobus, line 259: I wonder is it libpq / client lib responsible for converting from url-safe encoding like `?options=project%3Dep-raspy-unit-343115%20something` to this `\` escaping?

ololobus (approved):
New logic looks good to me

ololobus:
> Could you remove me from the CODE_OWNERS? :) I'm not sure it makes sense to keep me there.

I think it's rather question to @vadim2404. Current GH teams are a bit messy, indeed

ololobus:
But I think it's generally good that people from console / control-plane are aware of new proxy changes. It doesn't necessary mean that you need to thoroughly review Rust code (I cannot do that for sure)

proxy/src/compute.rs
funbringer: Correct!

Examples:
* https://github.com/postgres/postgres/blob/5352ca22e0012d48055453ca9992a9515d811291/src/interfaces/libpq/fe-connect.c#L5961
* https://github.com/lib/pq/blob/922c00e176fb3960d912dc2c7f67ea2cf18d27b0/url.go#L32
* https://github.com/sfackler/rust-postgres/blob/4bae134f501db5c08b8dd43a44474e515bbb39d3/tokio-postgres/src/config.rs#L921

